PR #1, opened 2020-10-24
initial spec proposal

julienledem:
The goal is to specify the initial model (just plain lineage) that will be extended with various facets.
It does not intend to restrict to HTTP. Those same PUT calls without output can be translated to any async protocol


mars-lan:
Not sure if OpenAPI is the right place to start if the idea is to use an async protocol like Kafka. Something like Avro or porotobuf might be more appropriate in that case.

julienledem:
@mars-lan: I've opened an issue to discuss that: https://github.com/OpenLineage/OpenLineage/issues/2. This spec doesn't prevent having an Avro serialization as well.

julienledem:
FYI, you can generate avro schemas form OpenAPI as well.
Here is the corresponding output on the spec in this PR:
```
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "Dataset",
  "fields": [
    {
      "name": "type",
      "type": "string",
      "doc": "The type of the dataset."
    },
    {
      "name": "updatedAt",
      "type": ["null", "string"],
      "doc": "An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the date/time the dataset was updated."
    },
    {
      "name": "physicalDescriptor",
      "type": ["null", "string"],
      "doc": "The URL pointing to the dataset"
    },
    {
      "name": "source",
      "type": ["null", "string"],
      "doc": "The URL pointing to the source."
    },
    {
      "name": "description",
      "type": ["null", "string"],
      "doc": "The description of the table."
    },
    {
      "name": "runId",
      "type": ["null", "string"],
      "doc": "The ID associated with the run modifying the table."
    }
  ]
}
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "Job",
  "fields": [
    {
      "name": "type",
      "type": ["null", "string"],
      "doc": "The type of the job."
    },
    {
      "name": "updatedAt",
      "type": ["null", "string"],
      "doc": "An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the date/time the job was updated."
    },
    {
      "name": "inputs",
      "type": ["null", {
        "type": "array",
        "items": "model.JobInputs"
      }],
      "doc": "The set of **input** datasets."
    },
    {
      "name": "outputs",
      "type": ["null", {
        "type": "array",
        "items": "model.JobOutputs"
      }],
      "doc": "The set of **output** datasets."
    },
    {
      "name": "location",
      "type": ["null", "string"],
      "doc": "The URL of the job source code or artifact."
    },
    {
      "name": "context",
      "type": ["null", ],
      "doc": "A key/value pair that must be of type `string`. A context can be used for getting additional details about the job."
    },
    {
      "name": "description",
      "type": ["null", "string"],
      "doc": "The description of the job."
    }
  ]
}
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "JobInputs",
  "fields": [
    {
      "name": "namespace",
      "type": ["null", "string"],
      "doc": "The namespace of the **input** dataset."
    },
    {
      "name": "name",
      "type": ["null", "string"],
      "doc": "The name of the **input** dataset."
    }
  ]
}
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "JobOutputs",
  "fields": [
    {
      "name": "namespace",
      "type": ["null", "string"],
      "doc": "The namespace of the **output** dataset."
    },
    {
      "name": "name",
      "type": ["null", "string"],
      "doc": "The name of the **output** dataset."
    }
  ]
}
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "Namespace",
  "fields": [
    {
      "name": "name",
      "type": ["null", "string"],
      "doc": "The name of the namespace."
    },
    {
      "name": "ownerName",
      "type": ["null", "string"],
      "doc": "The owner of the namespace."
    },
    {
      "name": "description",
      "type": ["null", "string"],
      "doc": "The description of the namespace."
    }
  ]
}
{
  "namespace": "model",
  "type": "record",
  "doc": "",
  "name": "Run",
  "fields": [
    {
      "name": "nominalStartTime",
      "type": ["null", "string"],
      "doc": "An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the nominal start time of the run."
    },
    {
      "name": "nominalEndTime",
      "type": ["null", "string"],
      "doc": "An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the nominal end time of the run."
    },
    {
      "name": "args",
      "type": ["null", ],
      "doc": "The arguments of the run."
    }
  ]
}
```

julienledem:
And here is the corresponding protos:
```
/*
  OpenLineage

  OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.

  The version of the OpenAPI document: 0.0.1

  Generated by OpenAPI Generator: https://openapi-generator.tech
*/

syntax = "proto3";

package ;


message Dataset {

  // The type of the dataset.
  string type = 1;

  // An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the date/time the dataset was updated.
  string updatedAt = 2;

  // The URL pointing to the dataset
  string physicalDescriptor = 3;

  // The URL pointing to the source.
  string source = 4;

  // The description of the table.
  string description = 5;

  // The ID associated with the run modifying the table.
  string runId = 6;

}
/*
  OpenLineage

  OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.

  The version of the OpenAPI document: 0.0.1

  Generated by OpenAPI Generator: https://openapi-generator.tech
*/

syntax = "proto3";

package ;

import public "models/job_inputs.proto";
import public "models/job_outputs.proto";
import public "models/todo_object_mapping.proto";

message Job {

  // The type of the job.
  string type = 1;

  // An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the date/time the job was updated.
  string updatedAt = 2;

  // The set of **input** datasets.
  repeated JobInputs inputs = 3;

  // The set of **output** datasets.
  repeated JobOutputs outputs = 4;

  // The URL of the job source code or artifact.
  string location = 5;

  // A key/value pair that must be of type `string`. A context can be used for getting additional details about the job.
  TODO_OBJECT_MAPPING context = 6;

  // The description of the job.
  string description = 7;

}
/*
  OpenLineage

  OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.

  The version of the OpenAPI document: 0.0.1

  Generated by OpenAPI Generator: https://openapi-generator.tech
*/

syntax = "proto3";

package ;


message JobInputs {

  // The namespace of the **input** dataset.
  string namespace = 1;

  // The name of the **input** dataset.
  string name = 2;

}
/*
  OpenLineage

  OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.

  The version of the OpenAPI document: 0.0.1

  Generated by OpenAPI Generator: https://openapi-generator.tech
*/

syntax = "proto3";

package ;


message JobOutputs {

  // The namespace of the **output** dataset.
  string namespace = 1;

  // The name of the **output** dataset.
  string name = 2;

}
/*
  OpenLineage

  OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.

  The version of the OpenAPI document: 0.0.1

  Generated by OpenAPI Generator: https://openapi-generator.tech
*/

syntax = "proto3";

package ;


message Namespace {

  // The name of the namespace.
  string name = 1;

  // The owner of the namespace.
  string ownerName = 2;

  // The description of the namespace.
  string description = 3;

}
/*
  OpenLineage

  OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.

  The version of the OpenAPI document: 0.0.1

  Generated by OpenAPI Generator: https://openapi-generator.tech
*/

syntax = "proto3";

package ;

import public "models/todo_object_mapping.proto";

message Run {

  // An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the nominal start time of the run.
  string nominalStartTime = 1;

  // An [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) timestamp representing the nominal end time of the run.
  string nominalEndTime = 2;

  // The arguments of the run.
  TODO_OBJECT_MAPPING args = 3;

}
```

julienledem:
I added a simple generator to make it a bit more concrete.
As an example, this is what a model class looks like:
```
/**
 * Namespace
 */
@JsonPropertyOrder({
  "name",
  "ownerName",
  "description"
})
@JsonTypeName("Namespace")

public class Namespace {

  private String name = null;

  private String ownerName = null;

  private String description = null;


  public Namespace name(String name) {
    this.name = name;
    return this;
  }

   /**
   * The name of the namespace.
   * @return name
  **/
  @javax.annotation.Nullable
  @ApiModelProperty(example = "null", value = "The name of the namespace.")
  @JsonProperty("name")
  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
  public String getName() {
    return name;
  }

  public void setName(String name) {
    this.name = name;
  }


  public Namespace ownerName(String ownerName) {
    this.ownerName = ownerName;
    return this;
  }

   /**
   * The owner of the namespace.
   * @return ownerName
  **/
  @javax.annotation.Nullable
  @ApiModelProperty(example = "null", value = "The owner of the namespace.")
  @JsonProperty("ownerName")
  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
  public String getOwnerName() {
    return ownerName;
  }

  public void setOwnerName(String ownerName) {
    this.ownerName = ownerName;
  }


  public Namespace description(String description) {
    this.description = description;
    return this;
  }

   /**
   * The description of the namespace.
   * @return description
  **/
  @javax.annotation.Nullable
  @ApiModelProperty(example = "null", value = "The description of the namespace.")
  @JsonProperty("description")
  @JsonInclude(value = JsonInclude.Include.USE_DEFAULTS)
  public String getDescription() {
    return description;
  }

  public void setDescription(String description) {
    this.description = description;
  }


  @Override
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    Namespace Namespace = (Namespace) o;
    return Objects.equals(this.name, Namespace.name) &&
        Objects.equals(this.ownerName, Namespace.ownerName) &&
        Objects.equals(this.description, Namespace.description);
  }

  @Override
  public int hashCode() {
    return Objects.hash(name, ownerName, description);
  }

  @Override
  public String toString() {
    StringBuilder sb = new StringBuilder();
    sb.append("class Namespace {\n");
    sb.append("    name: ").append(toIndentedString(name)).append("\n");
    sb.append("    ownerName: ").append(toIndentedString(ownerName)).append("\n");
    sb.append("    description: ").append(toIndentedString(description)).append("\n");
    sb.append("}");
    return sb.toString();
  }

  /**
   * Convert the given object to string with each line indented by 4 spaces
   * (except the first line).
   */
  private String toIndentedString(Object o) {
    if (o == null) {
      return "null";
    }
    return o.toString().replace("\n", "\n    ");
  }

}
```
An what the api looks like:
```
public class OpenLineageApi {

  private final Backend backend;

  public OpenLineageApi(Backend backend) {
    this.backend = backend;
  }

  /**
   * Create a run
   * Creates a new run object for a job.
   * @param namespaceName The name of the namespace. (required)
   * @param jobName The name of the job. (required)
   * @param id The ID of the run. (required)
   * @param run  (optional)
   * @throws ApiException if fails to make API call
   */
  public void createRun(String namespaceName, String jobName, java.util.UUID id, Run run) {
    
    // verify the required parameter 'namespaceName' is set
    if (namespaceName == null) {
      throw new IllegalArgumentException("Missing the required parameter 'namespaceName' when calling createRun");
    }
    
    // verify the required parameter 'jobName' is set
    if (jobName == null) {
      throw new IllegalArgumentException("Missing the required parameter 'jobName' when calling createRun");
    }
    
    // verify the required parameter 'id' is set
    if (id == null) {
      throw new IllegalArgumentException("Missing the required parameter 'id' when calling createRun");
    }
    
    backend.call(
      "POST", 
      "/namespaces/{namespaceName}/jobs/{jobName}/runs/{runId}",
      new Parameters().add("namespaceName", namespaceName).add("jobName", jobName).add("id", id),
      new Parameters(), 
      run);
  }
...
```

julienledem:
Then, one can implement multiple Backends to write to the protocol they like (HTTP, Kafka,...) we can also automatically generate corresponding avro and protos if that's what people want to do.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,212 @@
...
+      type: object
+      properties:
+        id:
+          description: the unique identifier for that run
+          type: string
+          format: uuid
+        transition:
+          description: transitions the run state
+          type: string #TODO: add enum
+          example: 'START|COMPLETE|ABORT|FAIL'
```
henneberger, line 38: It may be nice to record multiple state transitions in a single event since the relationship of `runs` -> `run state` is 1-n. However it changes the relationship of an event with the dataset.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,212 @@
...
+components:
+  schemas:
+    RunStateUpdate:
+      type: object
+      properties:
+        id:
+          description: the unique identifier for that run
+          type: string
+          format: uuid
+        transition:
```
henneberger, line 35: It may be nice to record multiple state transitions in a single event since the relationship of `runs` -> `run state` is 1-n. However it changes the relationship of an event with the dataset.

julienledem: I was thinking we could have an endpoint where we send an array of RunStateUpdate. The idea being that the job and dataset are described as observed at the time of the transition. If nothing has changed or we didn't observe it we can have a RunStateUpdate event that contains only the transition information.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+      responses:
+        '200':
+          description: OK
+
+components:
+  schemas:
+    RunStateUpdates:
+      type: array
+      items:
+        $ref: '#/components/schemas/RunStateUpdate'
```
henneberger, line 31: I can't really think of an instance where I want to describe a dataset that doesn't encompass all available information so this doesn't make sense to me.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+
+    RunStateUpdate:
+      type: object
+      properties:
+        runId:
+          description: the unique identifier for that run
+          type: string
+          format: uuid
+        jobId:
+          $ref: '#/components/schemas/JobId'
```
henneberger, line 41: The jobid can be rolled into the job vertex and marked as required to achieve the same effect, since all the other job attributes are optional. If optional job fields are present at write, it can be treated as an upsert to a standalone job entity.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+
+    ParentRunFacet:
+      description: the id of the parent run and job, iff this run was spawn from an other run (for example, the Dag run scheduling its tasks)
+      type: object
+      properties:
+        runId:
+          type: string
+          format: uuid
+        jobId:
+          $ref: '#/components/schemas/JobId'
```
henneberger, line 199: This should be able to take an entire job entity rather than just the id components. This will allow it to be completely defined in a single event.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+      properties:
+        runId:
+          description: the unique identifier for that run
+          type: string
+          format: uuid
+        jobId:
+          $ref: '#/components/schemas/JobId'
+        transition:
+          description: transitions the run state
+          type: string #TODO: add enum
```
henneberger, line 44: Defining this an an enum may be limiting since systems have a wide array of transition states

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+          items:
+            $ref: '#/components/schemas/Dataset'
+        outputs:
+          description: The set of **output** datasets.
+          type: array
+          items:
+            $ref: '#/components/schemas/Dataset'
+        facets:
+          description: The run facets.
+          type: object
```
henneberger, line 65: Should this be an array?

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+          type: string
+        facets:
+          description: The job facets.
+          type: object
+          properties:
+            sourceCodeLocation:
+              $ref: '#/components/schemas/SourceCodeLocationJobFacet'
+            sql:
+              $ref: '#/components/schemas/SQLJobFacet'
+          additionalProperties:
```
henneberger, line 94: This type of semi-structured data is really tough to model in strongly typed languages (e.g. graphql) since its tough to have arbitrary keys. How would you feel about facets having the following def. The downside is that we'd need to introduce some tooling but it'd make it facets more future proof.
```
    Job:
      type: object
      properties:
        description:
          description: The description of the job.
          type: string
        facets:
          description: The job facets.
          type: array
          items:
            $ref: '#/components/schemas/Facet'

    Facet:
      type: object
      properties:
        origin:
          description: The URI of the facet property
          type: string
          example: http://openlineage.com/facets/schema/postgres
        content:
          description: The content of the facet
          type: object
      required:
        - origin
```

henneberger: I'm drawing inspiration from https://github.com/graphql/graphql-spec/issues/635

henneberger: Heres an overview of some approaches to input union types: https://github.com/graphql/graphql-spec/blob/master/rfcs/InputUnion.md

spec/OpenLineage.yml
julienledem: I agree with this statement. I moved the jobId here because the runId can be linked to only one jobId. They are both  required definitions for the RunStateUpdate and they must be constant for all StateUpdates of the same run. Nesting JobId in `job` gives the false impression that more than one job can be linked to the run.
I'm open to other suggestions as well.

spec/OpenLineage.yml
julienledem: The intent here is to allow batching of those updates in a single event, independently of other factors.

spec/OpenLineage.yml
julienledem: This is a great point. We do need standard transitions and a mechanism to extend them with custom ones.
Possibly this could be achieved with prefixes (similar to adding facets)

spec/OpenLineage.yml
julienledem: It could. Since we're using JSON and it is required for the facet name to be unique in the list, I took the approach to model it as:
```
facets: {
  "{facet_name}": { .... facet properties ...}
  ...
}
```
The array equivalent would look like this:
```
facets: [
   { name: "{facet_name}", ... facet properties ...},
   ...
]
```
I have a preference for the object approach but I'm happy to hear other opinions.

spec/OpenLineage.yml
julienledem: Yes, I agree. I am somewhat intending that in the definition of CustomFacet bellow with the origin and version fields that could actually be replaced by an URI to the definition.
It's akin to Map<String, CustomFacet> with CustomeFacet having some required fields.
The goal is to allow sending facets that are not part of the spec to strike a balance between standard facets that might be lossy and allowing adding all the details specific to a given system.

spec/OpenLineage.yml
julienledem: I am intending that the Run in question has started before that one and so we could have 2 RunStateUpdates at the top level. One for each job run.
```
POST /events [
  { runID: "1", jobId: { ns: "foo", name: "parent"} ... },
  { runID: "2", jobId: { ns: "foo", name: "child"}, parent: {runID: "1", jobId: { ns: "foo", name: "parent"}} ... }
]
```

spec/OpenLineage.yml
julienledem: This also simplified the notion of parent run and job. 
I would consider the runid to be the composite (jobId, runID)

julienledem (commented):
Thank you for your comments. It feels like we're getting closer. 
a few more iterations and it will be ready.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
+openapi: 3.0.2
+info:
+  title: OpenLineage
+  version: 0.0.1
+  description: OpenLineage is an open source **lineage and metadata collection API** for the data ecosystem.
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+paths:
+  /events:
```
wslulciuc, line 10: We might want to contextualize the endpoint a bit more. Meaning, `events` is rather generic and can mean all types of events (not just specific to open lineage). I'd also expect to the payload structured as an array of different event types. That said, I understand the endpoints overall intent: _capture event **X** at a given time **T** with metadata observed_.

I think some alternative endpoints can be:

-  `/metadata`: Yeah, still "generic", but now conveys that the payload will contain objects / properties specific to metadata collection
-  `/lineage`: Collect metadata specific to data lineage, mapping the objects / properties in the payload to the domain of open lineage.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,250 @@
...
+          description: OK
+
+components:
+  schemas:
+    RunStateUpdates:
+      type: array
+      items:
+        $ref: '#/components/schemas/RunStateUpdate'
+
+    RunStateUpdate:
```
wslulciuc, line 33: Is the `RunStateUpdate` object the only event that will be handled by `/events`?

spec/OpenLineage.yml
julienledem: `/lineage` makes sense to me, I'll update accordingly.

spec/OpenLineage.yml
julienledem: yes

spec/OpenLineage.md
```diff
@@ -0,0 +1,110 @@
...
+    }
+  ],
+  "outputs": [
+    {
+      "namespace": "my-datasource-namespace",
+      "name": "instance.schema.output_table",
+    }
+  ],
+  "producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client"
+}
```
wslulciuc, line 50: I would expect the `job` to have `inputs` / `outputs` as fields, rather than be top-level fields associated with the run state change. That is:

```
job: {
  inputs: {...}
  outputs: {...}
}
```

But, if the goal is to associated the `inputs` / `outputs` at the run-level, then should `run` include them?

```
run: {
  inputs: {...}
  outputs: {...}
}
```

spec/OpenLineage.yml
```diff
@@ -0,0 +1,268 @@
...
+
+components:
+  schemas:
+    RunStateUpdate:
+      type: object
+      properties:
+        transition:
+          description: the current transition of the run state. It is required to issue 1 START event and 1 of [ COMPLETE, ABORT, FAIL ] event per run. Additional events with no transition can be added to the same run. For example to send additional metadata after the run is complete
+          type: string
+          enum: [START, COMPLETE, ABORT, FAIL]
```
henneberger, line 34: How about a NONE transition and have transition be required, or not required with a default value of NONE. Since transition would be a queryable field (e.g. all events that are COMPLETE), it'd save us from having a placeholder field and doing an additional transformation on the server.

spec/OpenLineage.yml
```diff
@@ -0,0 +1,268 @@
...
+            required:
+              - name
+              - type
+
+    DatasourceDatasetFacet:
+      type: object
+      properties:
+        name:
+          type: string
+        url:
```
dorzey, line 266: Should the property also be called `uri`? I am assuming that a Datasource for a Dataset could be something like a Kafka topic; or at least something that wouldn't have a URL.

dorzey (commented):
This is super exciting; let me know how I can help.

spec/OpenLineage.yml
julienledem: that is correct. Thanks for the feedback. 
Feel free to reach out on the mailing list and slack or comment on the PRs. You can also open github issues for specific things you'd want to contribute.

henneberger (approved):

jcampbell:
I really like the direction of this proposal, and can definitely see sufficient flexibility via the facet customizations to support adding metadata in the right places. I'd recommend merging and moving this forward as is. The main high-level challenge I see comes from reconciling the namespaces of jobs and datasets, specifically for cases where datasets have a concept of "batches" inside them where there are logically related units (e.g. "this month's data is approximately the same volume as last month's").

spec/OpenLineage.yml
```diff
@@ -0,0 +1,268 @@
...
+            $ref: '#/components/schemas/CustomFacet'
+      required:
+       - namespace
+       - name
+
+
+# Custom facets
+    CustomFacet:
+      type: object
+      properties:
```
drewbanin, line 142: Maybe we could consider creating a `data` property, rather than leveraging `additionalProperties` at the root-level of a `CustomFacet`. In my experience, mixing reserved fields and arbitrary user-supplied fields can be really messy - it can make it challenging to add new reserved fields into the mix in the future, as they might conflict with field names that are already utilized in practice.

drewbanin (commented):
I dropped one comment in here, but would not consider it a blocker to getting the initial proposal for this spec merged. I really like the way things are shaking out here, and am keen to dig further into some of the custom facets that might make this spec extensible for use with some of the common databases we see folks using in an analytics engineering capacity.

wslulciuc (approved):
The initial draft of the spec looks great! 💯 🥇

spec/OpenLineage.md
```diff
@@ -0,0 +1,110 @@
...
+- Dataset
+
+A facet is an atomic piece of metadata identified by its name. This means that emiting a new facet whith the same name for the same entity replaces the previous facet instance for that entity entirely). It is defined as a JSON object that can be either part of the spec or custom facets defined in a different project. 
+
+Custom facets must use a distinct prefix named after the project defining them to avoid colision with standard facets defined in the [OpenLineage.yml](OpenLineage.yml) OpenAPI spec. 
+They have a schemaURL field pointing to the corresponding version of the facet schema (as a [$ref URL location](https://swagger.io/docs/specification/using-ref/) ).
+
+Example: https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/spec/OpenLineage.yml#MyCustomJobFacet
+
+Custom facets can be promoted to the standard by including them in the spec.
```
rdblue, line 90: What isn't very clear to me is how facet information is added by providers. For example, when Iceberg commits to a table, we get lots of information like what partitions were changed, what the operation was (append, overwrite, etc.), how long the commit took, number of retries, etc. That information isn't exposed to Spark, which has a simple API that is generic across sources. How would Iceberg attach the facet information?

I can think of a couple options:
1. Add facets to Spark's source API
2. Send more lineage events directly from Iceberg
3. Add calls in Iceberg to pass facet information to some listener

I don't think that option 1 is a good idea because it would add complication to the API and would need to be done in every compute engine.

Option 2 is also not a great option because it isn't clear exactly what the "run" is in this case. I think that a run would probably be at the level of a SQL statement here, and further divisions seem like hacky ways to work around the problem. It isn't clear what the Iceberg commit "run" would use as its "job", for example.

Option 3 sounds more reasonable to me, but is really more a question of what the JVM API looks like.

spec/OpenLineage.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Standard Facets
+
+#### Run Facets
+
+- **nominalTime**: Captures the time this run is scheduled for. This is a typical usage for time based scheduled job. The job has a nominal schedule time that will be different from the actual time it is running at.
+
+- **parent**: Captures the parent job and Run when the run was spawn from a parent run.
```
rdblue, line 98: I would like to see a longer description of how to use the "parent" relationship that calls out that runs will very likely have parents and explains how the parent should be set. This is going to be used quite a bit because there is probably a scheduler run that kicks off a container run in which an application runs (Spark) with queries (SQL) in that application.

I would probably also make this a required part of a run, but that is probably debatable.

spec/OpenLineage.md
```diff
@@ -0,0 +1,110 @@
...
+
+Facets are pieces of metadata that can be attached to the core entities:
+- Run
+- Job
+- Dataset
+
+A facet is an atomic piece of metadata identified by its name. This means that emiting a new facet whith the same name for the same entity replaces the previous facet instance for that entity entirely). It is defined as a JSON object that can be either part of the spec or custom facets defined in a different project. 
+
+Custom facets must use a distinct prefix named after the project defining them to avoid colision with standard facets defined in the [OpenLineage.yml](OpenLineage.yml) OpenAPI spec. 
+They have a schemaURL field pointing to the corresponding version of the facet schema (as a [$ref URL location](https://swagger.io/docs/specification/using-ref/) ).
```
rdblue, line 86: I would really like to see some guidance here on how to version facets and what facet consumers can rely on.

spec/OpenLineage.yml
julienledem: that's a great point.
Another mechanism could be to have a special prefix for those reserved field that can not be used in the user defined schemas. That would alleviate the extra nesting of having a `data` field to nest those. I'll submit an update along those lines.

spec/OpenLineage.md
julienledem: That's a good point, I have another PR with a code generation draft as the beginning of a java client.
https://github.com/OpenLineage/OpenLineage/pull/3
We should explore what such an API looks like.

spec/OpenLineage.md
julienledem: I'm updating this description to mention those examples.

julienledem: There's always a root run that doesn't have a parent. But we could make it part of the root RunStateUpdate as just `parent`

spec/OpenLineage.md
julienledem: Thanks. Updating

PR #14, opened 2020-12-28
Conform formatting of runId, fix required fields.

henneberger:


spec/OpenLineage.yml
```diff
@@ -68,6 +68,7 @@ components:
         runId:
           description: The id of the run, unique relative to the job
           type: string
+          format: uuid
```
julienledem, line 71: Do we want runIds to be UUIDs?

henneberger: updated the PR so runid in the parent facet is the same as this runid

julienledem (approved):

PR #17, opened 2021-01-14
adding Json-Schema for OpenLineage

### LINKED ISSUE #15: [PROPOSAL] Move the OpenLineage spec to JonSchema (from OpenAPI)
**Purpose:**
OpenAPI and JsonSchema are mostly the same model with a goal of OpenAPI to be a strict superset of JSONSchema in the near future. 
The goal is to make OpenLineage independent from the underlying protocol. OpenAPI assumes a REST endpoint which is not our intent here. This would still make possible to define an OpenLineage HTTP endpoint while decoupling the model and make it more obvious that it can be used with Kafka for example.

**Proposed implementation**
Convert the OpenAPI spec to a json schema.

julienledem:
resolves #15


spec/OpenLineage.json
```diff
@@ -0,0 +1,425 @@
...
+        },
+        {
+          "type": "object",
+          "properties": {
+            "run": {
+              "type": "object",
+              "properties": {
+                "runId": {
+                  "type": "string",
+                  "format": "uuid"
```
henneberger, line 239: type is string, not uuid

spec/OpenLineage.json
```diff
@@ -0,0 +1,425 @@
...
+        {
+          "type": "object",
+          "properties": {
+            "type": {
+              "type": "string",
+              "example": "git"
+            },
+            "url": {
+              "type": "string",
+              "format": "uri"
```
henneberger, line 309: It is a bit confusing to me if this should be a url or uri.

spec/OpenLineage.json
```diff
@@ -0,0 +1,425 @@
...
+      "type": "object"
+    },
+    "SourceCodeLocationJobFacet": {
+      "allOf": [
+        {
+          "$ref": "#/definitions/BaseFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
```
henneberger, line 302: All of these properties are required

spec/OpenLineage.json
```diff
@@ -0,0 +1,425 @@
...
+            }
+          },
+          "required": [
+            "description"
+          ]
+        }
+      ],
+      "type": "object"
+    },
+    "SourceCodeLocationJobFacet": {
```
henneberger, line 295: should these canned facets also extend additional properties for future backwards compatibility for changing facet versions?

spec/OpenLineage.json
julienledem: thanks

spec/OpenLineage.json
julienledem: that's a good point. Is it required?
Should all nested object types do the same?

spec/OpenLineage.json
julienledem: thanks

spec/OpenLineage.json
julienledem: a URL is a URI, so this is correct? https://tools.ietf.org/html/rfc3986#section-1.1.3
built-in formats have uri only. https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers

spec/OpenLineage.json
```diff
@@ -0,0 +1,426 @@
+{
+  "$schema": "http://json-schema.org/schema#",
```
wslulciuc, line 2: Should we add the [specific draft](https://json-schema.org/specification-links.html#published-drafts) of the standard we are using? For example: `http://json-schema.org/draft-07/schema#`

wslulciuc (approved):
Left a minor comment, but otherwise this looks great! And looks like there are plenty of [libraries](http://json-schema.org/implementations.html) for working with JSON Schema 🥇 💯

spec/OpenLineage.json
julienledem: good point

PR #18, opened 2021-01-28
OpenLineage java model Generator 

julienledem:
This a generator of the OpenLineage model as java classes with Jackson serialization


client/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  public static TypeName getTypeName(Type type) {
+    return type.accept(new TypeVisitor<TypeName>(){
+
+      @Override
+      public TypeName visit(PrimitiveType primitiveType) {
+        if (primitiveType.getName().equals("string")) {
+          return ClassName.get(String.class);
+        }
```
henneberger, line 159: Add date-time with a validating class? I personally have mixed feelings about it. It may be nice to be able to programmatically resolve scalar types & additional facets via a compile-time type annotation.

henneberger (approved):

client/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore Gradle project-specific cache directory
+.gradle
+
+# Ignore Gradle build output directory
+build
```
wslulciuc, line 5: Should we also add `.classpath`, `.idea/`, `.settings`, and other idea specific dir / files to the list?

client/generator/bin/.gitignore
```diff
@@ -0,0 +1,3 @@
+/functionalTest/
+/main/
+/test/
```
wslulciuc, line 3: _Minor_: `bind/` can also be ignored

client/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
```diff
@@ -0,0 +1,174 @@
...
+import io.openlineage.client.TypeResolver.Type;
+import io.openlineage.client.TypeResolver.TypeVisitor;
+
+/**
+ * Generates a JavaClass with all the types as inner classes
+ */
+public class JavaPoetGenerator {
+
+  private static final String PACKAGE = "io.openlineage.client";
+  private static final String CONTAINER_CLASS_NAME = "OpenLineage";
```
wslulciuc, line 47: _Minor_: Would we want to name the class `OpenLineageClient`?

wslulciuc: I see, this is generating the `OpenLineage` model with the nested internal class. Is the idea to also generate `OpenLineageClient`?

client/settings.gradle
```diff
@@ -0,0 +1,11 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * The settings file is used to specify which projects to include in your build.
+ *
+ * Detailed information about configuring a multi-project build in Gradle can be found
+ * in the user manual at https://docs.gradle.org/6.7/userguide/multi_project_builds.html
+ */
+
+rootProject.name = 'openlineage-client'
```
wslulciuc, line 10: _Minor_: I think appending the language type to the client name is a nice way to communicate support? `openlineage-java`, `openlineage-python`, `openlineage-go`, etc.

client/src/test/java/io/openlineage/client/OpenLineageTest.java
```diff
@@ -0,0 +1,44 @@
...
+    JobFacets jobFacets = new JobFacets(null, null, null);
+    Job job = new Job(namespace, name, jobFacets);
+    List<Dataset> inputs = Arrays.asList();
+    List<Dataset> outputs = Arrays.asList();
+    OpenLineage.RunStateUpdate runStateUpdate = new OpenLineage.RunStateUpdate("START", "123", run, job, inputs, outputs, producer );
+
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.configure(SerializationFeature.INDENT_OUTPUT, true);
+    String json = mapper.writeValueAsString(runStateUpdate);
+    System.out.println(json);
```
wslulciuc, line 40: _Minor_: May want to remove the `println` here :)

client/src/test/java/io/openlineage/client/OpenLineageTest.java
```diff
@@ -0,0 +1,44 @@
...
+import io.openlineage.client.OpenLineage.Job;
+import io.openlineage.client.OpenLineage.JobFacets;
+import io.openlineage.client.OpenLineage.Run;
+import io.openlineage.client.OpenLineage.RunFacets;
+import io.openlineage.client.OpenLineage.RunStateUpdate;
+
+public class OpenLineageTest {
+
+  @Test
+  public void jsonSerialization() throws JsonProcessingException {
```
wslulciuc, line 23: _Minor_: I wondering if we want to load the expected json from a file, then check that against the `OpenLineage.RunStateUpdate` object? or maybe define the json string in-line:

```
String expectedJson = "{...";
```

wslulciuc (approved):
Left some comments, but certainly not blockers on getting this merged 👍 💯

client/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
julienledem: I think the Client would not be generated and be a class in the repo. This is the first step.

client/settings.gradle
julienledem: since each language has their own artifact repository, I think it is easier to look for the -client artifact in each repo. maven would only have the java client etc.

client/src/test/java/io/openlineage/client/OpenLineageTest.java
julienledem: ooops

client/src/test/java/io/openlineage/client/OpenLineageTest.java
julienledem: agreed, this needs to be improved.

client/.gitignore
julienledem: sure

client/generator/bin/.gitignore
julienledem: will cleanup

client/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
julienledem: I suggest we add the support for the `format` as a follow up.

PR #21, opened 2021-02-12
fix member visibility; fix javadoc; generate from URL

julienledem:

### DIFF
--- a/client/generator/src/main/java/io/openlineage/client/Generator.java
+++ b/client/generator/src/main/java/io/openlineage/client/Generator.java
@@ -2,7 +2,9 @@ package io.openlineage.client;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.PrintWriter;
+import java.net.URL;
 
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonMappingException;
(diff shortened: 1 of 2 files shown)


wslulciuc (approved):
LGTM 👍

client/generator/src/main/java/io/openlineage/client/Generator.java
```diff
@@ -12,13 +14,13 @@
 public class Generator {
 
   public static void main(String[] args) throws JsonParseException, JsonMappingException, IOException {
-    String baseURL = "https://raw.githubusercontent.com/OpenLineage/OpenLineage/jsonschema/spec/OpenLineage.json";
-    File input = new File("../spec/OpenLineage.json");
+    String baseURL = "https://raw.githubusercontent.com/OpenLineage/OpenLineage/main/spec/OpenLineage.json";
```
wslulciuc, line 17: _Minor_: Would we always want to generate the model off `main`? or based on a release tag (ex: `0.1.0`)?

wslulciuc (approved):

client/generator/src/main/java/io/openlineage/client/Generator.java
julienledem: We would want to generate from a tag. I use main in the meantime until we tag our first version

PR #23, opened 2021-02-18
add input and output facets

### LINKED ISSUE #20: [PROPOSAL] Add InputFacets and OutputFacets
**Purpose:**
We currently have JobFacets for everything that does not change every run (example: the current version in source control like the git sha). We have RunFacets for everything that changes every Run (example: the schedule time).
We are missing the same notion for input and output datasets.
Currently we only have Dataset facets for things that don't change every run (example: Schema). I propose to add Input and Output facets to capture things that change in datasets every run (ex: number of row produced, etc)

**Proposed implementation**
Add InputFacets and OutputFacets in the core model. Directly in the Input or Output field.
```
"inputs": {
          "description": "The set of **input** datasets.",
          "type": "array",
          "items": {
            "allOf": [
              { "$ref": "#/definitions/Dataset" },
              {
                "type": "object",
                "properties": {
                  "inputFacets": {
                    ...
                  }
                }
              }
            ]
          }
        }
```

julienledem:
This PR resolves #20 

### DIFF
--- a/spec/OpenLineage.md
+++ b/spec/OpenLineage.md
@@ -113,4 +113,8 @@ Custom facets can be promoted to the standard by including them in the spec.
 
 #### Input Dataset Facets
 
+*Note: for now there no input facets defined*
+
 #### Output Dataset Facets
+
+*Note: for now there no output facets defined*
--- a/spec/OpenLineage.md
+++ b/spec/OpenLineage.md
@@ -2,8 +2,9 @@
 
 ## Specification
 
-The specification for OpenLineage is formalized as an OpenAPI spec: [OpenLineage.yml](OpenLineage.yml)
-published at: https://openlineage.github.io/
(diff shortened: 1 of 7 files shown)


spec/OpenLineage.md
```diff
@@ -111,3 +110,7 @@ Custom facets can be promoted to the standard by including them in the spec.
 - **schema**: Captures the schema of the dataset
 
 - **dataSource**: Captures the Database instance containing this datasets (ex: Database schema. Object store bucket, ...)
+
+#### Input Dataset Facets
```
wslulciuc, line 114: _Minor_: Should we add a comment communicating that no standard inputs / outputs? facets are yet defined?

wslulciuc (approved):
LGTM 👍

spec/OpenLineage.md
julienledem: added

PR #27, opened 2021-02-27
quick proof of concept to embed the facet schema in the generated class

julienledem:


julienledem:
this would enable a schema registry for custom facets

client/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
```diff
@@ -228,6 +236,63 @@ private void generateTypes(TypeSpec.Builder containerTypeBuilder) {
...
+        List<String> elements = new ArrayList<String>();
+        for (Field field : properties) {
+          elements.add("\"" + field.getName() + "\": " + field.getType().accept(this));
+        }
+        return String.join(", ", elements);
+      }
+
+      @Override
+      public String visit(ArrayType arrayType) {
+        return "{ \"type\" : \"array\""
```
julienledem, line 289: This would Have to use a proper JsonNode.

PR #31, opened 2021-03-18
Create unique naming spec

julienledem:


spec/Naming.md
```diff
@@ -0,0 +1,122 @@
...
+ * Host
+ * Port
+
+Naming hierarchy:
+ * Database
+ * Schema
+ * Table
+
+Identifier:
+ * Namespace: postgres://{host}:{port} of the service instance.
```
collado-mike, line 25: JDBC connection URIs include the database name. Should the clients be responsible for parsing the connection URI to extract the database name and only report the host/port? The connection URI is going to vary by driver (e.g., oracle, postgres, and sqlite all have different uri formats).

Should we consider leaving the database name as part of the namespace and only including the schema/table in the name? That'll make it easier for clients to capture that information without parsing the different URIs.

spec/Naming.md
```diff
@@ -0,0 +1,122 @@
...
+Naming hierarchy:
+ * bucket name => globally unique
+ * Path 
+
+Identifier :
+ * Namespace: gs://{bucket name}
+   * Scheme = gs
+   * Authority = {bucket name}
+ * Unique name: {path}
+   * URI =   gs://{bucket name}{path}
```
collado-mike, line 85: Quick clarification here- GCS and S3 object URLs always include a slash between the bucket and the key, but the key itself doesn't start with a slash. Will this URI look like `gs://{bucket name}/{path}`? I assume yes, so maybe worth clarifying that here.

spec/Naming.md
```diff
@@ -0,0 +1,122 @@
...
+
+Identifier :
+ * Namespace: hdfs://{namenode host}:{namenode port}
+   * Scheme = hdfs
+   * Authority = {namenode host}:{namenode port}
+ * Unique name: {path}
+   * URI =   hdfs://{namenode host}:{namenode port}{path}
+
+### Hive
+
```
collado-mike, line 112: I assume here, something like

> Naming hierarchy:
>   * Metastore host
>   * Database name
>   * Table name
>
> Identifier :
>  * Namespace: hive://{metastore host}:{metastore port}
>    * Scheme = hive
>    * Authority = {metastore host}:{metastore port}
>  * Unique name: {database name}.{table name}
>    * URI =  hive://{metastore host}:{metastore port}/{database name}.{table name}

spec/Naming.md
julienledem: I was assuming the path starts with `/` in that case.
But we can also have the name, not start with `/` and define the URI as `gs://{bucket name}/{path}`

spec/Naming.md
julienledem: Ah yes, this was a TODO. Item, should I leave it out for now or add this to the spec?

spec/Naming.md
julienledem: That makes sense. 
This spec is trying to stay close to the current Marquez implementation (dataset name = {db}.{schema}.{table})
So it's a bit of a trade off.
I'm assuming we would want to change the other db types as well to have the names consistently in 2 levels?

spec/Naming.md
collado-mike: Seems to me that it would be just as easy to add it in now vs later.

spec/Naming.md
collado-mike: In my mind, there's a tradeoff between consistency, existing conventions, and ease of implementation. Personally, I'd simply prefer not to parse JDBC connection URIs, if I can avoid it (have you seen Oracle connection URIs? 😬). Implementation details do seem to vary among databases. E.g., a Snowflake connection doesn't require the database in the connection URI- https://docs.snowflake.com/en/user-guide/jdbc-configure.html . The spark connector seems to use the same connection URI w/o database in the URI. BigQuery has no connection URI, so implementation is completely flexible. Hive connection URIs don't specify the database. Queries can use fully qualified table names or use the `USE <db>` statement.

I don't know that there has to be 100% consistency between each of these, but if consistency is important, I think it would be easier to move the `database` portion into the namespace than to extract it from the connection URI. If consistency _and_ existing convention is important, we'll need parsers for each flavor of JDBC connection URIs to extract database and form the FQ dataset name.

spec/Naming.md
```diff
@@ -0,0 +1,122 @@
...
+ * Project ID: {project id} => readable: is unique across google cloud
+ * dataset: {dataset name} => is unique within a project
+ * table: {table name} => is unique within a dataset
+
+Identifier :
+ * Namespace: bigquery
+   * Scheme = bigquery
+   * Authority = 
+ * Unique name: {project id}.{dataset name}.{table name}
+   * URI =   bigquery:{database}.{schema}.{table}
```
mobuchowski, line 70: `bigquery://{database}.{schema}.{table}` ?
I know the `//` is optional in URIs, but in practice it's rarely seen

mobuchowski, line 70: Shouldn't `{database}.{schema}.{table}` here be same as `{project id}.{dataset name}.{table name}`?

spec/Naming.md
julienledem: A couple of points:
- We should consistently build those dataset names for all SQL data{base,warehouse}. I'd be happy with standardizing on `{schema}.{table}`. But it should be we should follow the same convention for bigquery and Snowflake for example.  
- We cannot get away with not looking at the connection URL. The namespace must be canonical as well. The fact that some parameters are added to a connection URL or that the database is optional should not lead to different names. Those names need to be canonical, if there's a default db name, then we need to add it explicitly in the namespace here.

spec/Naming.md
julienledem: The others have been researched and there are existing integrations. This is not the case yet for Hive. So I'd rather add it later.

spec/Naming.md
julienledem: you are correct. I'll fix.

spec/Naming.md
```diff
@@ -0,0 +1,120 @@
...
+ * Unique name: {path}
+   * URI =   gs://{bucket name}{path}
+
+#### S3
+Naming hierarchy:
+ * bucket name => globally unique
+ * Path 
+
+Identifier :
+ * Namespace: s3://{bucket name}
```
APMoctezuma, line 93: A quick question here surrounding this namespace. I know the suggestion is to use the s3 path but looking at the Open Lineage documentation it says that Namespaces should only have letters and numbers. Does this mean there will be an additional change that will allow '/' to be included in a namespace? Thank you

julienledem: Yes. This has been updated in Marquez ":" and "/" are now allowed. can you point me to the doc you are looking at, we need to update this?

APMoctezuma: Hey Julien, thank you for getting back to me here, I'm assuming this change will be in 0.14.0? The documentation I was looking at was here in the API reference documentation: https://marquezproject.github.io/marquez/openapi.html#tag/Namespaces/paths/~1namespaces~1{namespace}/put

julienledem: It looks like the documentation was not updated when that change was made in 0.13:
https://github.com/MarquezProject/marquez/commit/4e2cee04e2040bf55d37a2a29c0c981cda016717
We're going to fix it.
It is in 0.13.1 already.

APMoctezuma: Ah ok. Looks like we behind on versions anyways. I'll look into updating to 0.13.1. I believe my colleague @KevinMellott91 contributed to Marquez by making that Java 8 compatible, which was our original issue with updating. Thank you for looking into this Julien!

collado-mike (approved):

julienledem:
Thank you all!

PR #49, opened 2021-05-17
default schema field description should not serialize to empty string

mobuchowski:


julienledem (approved):
thanks!

wslulciuc (approved):
LGTM 👍

PR #52, opened 2021-05-21
ci/python: format RC versions compliant to PEP 440, use tag to version if exists, unpin dependencies

mobuchowski:
`0.0.1-rc.1` -> `0.0.1rc1` (or `0.0.1-rc1`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ checkout_project_root: &checkout_project_root
 only-on-release: &only-on-release
   filters:
     tags:
-      only: /^[0-9]+(\.[0-9]+){2}(rc[0-9]+)?$/
+      only: /^[0-9]+(\.[0-9]+){2}(-?rc[0-9]+)?$/
     branches:
       ignore: /.*/
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,11 +8,11 @@ checkout_project_root: &checkout_project_root
     path: ~/openlineage
 
 
-# Only trigger CI job on release (=X.Y.Z)
+# Only trigger CI job on release (=X.Y.Z) with possible (rcX)
 only-on-release: &only-on-release
   filters:
     tags:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -8,11 +8,11 @@ checkout_project_root: &checkout_project_root
     path: ~/openlineage
 
 
-# Only trigger CI job on release (=X.Y.Z)
+# Only trigger CI job on release (=X.Y.Z) with possible (rcX)
 only-on-release: &only-on-release
   filters:
     tags:
-      only: /^[0-9]+(\.[0-9]+){2}(-rc\.[0-9]+)?$/
+      only: /^[0-9]+(\.[0-9]+){2}(rc[0-9]+)?$/
```
wslulciuc, line 15: I don't think this regex would work for jvm release, for example: `0.1.0-rc1` (which is more common). I think we'll want to support `X.Y.Z-rcN` for jvm specific releases and `X.Y.ZrcN` for python (as per [`PEP 440`](https://www.python.org/dev/peps/pep-0440/#pre-release-spelling)). You can update the regex to support both by adding `-?`:

```
^[0-9]+(\.[0-9]+){2}(-?rc[0-9]+)?$
```

mobuchowski: Makes sense - but maybe have two separate release paths for clients? We don't necessarily want to release both clients together.

mobuchowski: @wslulciuc for now let's go with `-?` version

wslulciuc (approved):
LGTM 👍

.circleci/config.yml
mobuchowski: >My line of thinking is that, when we update the OpenLineage spec, it would generate the client models and trigger a release with each client since they are within the same repo.

That's one case where we want to update all the clients. Second one is when we want to fix bug in particular client implementation, where release of other clients is not needed, and I'm not sure how to reconcile those two.

Issue #29, opened 2021-03-15
[PROPOSAL] Make job run ids unique 


collado-mike:
**Purpose:**
Currently, the `Run` object has a `runId` property that is defined as "_The id of the run, unique relative to the job_" https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L61-L63 . This implies a run id may be reused across jobs so that to link a job run to a parent run, the job name and namespaces are required in addition to the run id (see https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L273-L311 ).

In practice, Marquez generates UUIDs for the job run ids (e.g., see the airflow integration at https://github.com/MarquezProject/marquez/blob/main/integrations/airflow/marquez_airflow/dag.py#L228-L229 ). In the spark integration, there's confusion about whether the passed job run parameters belong to the parent run or to a new job run (see the README at https://github.com/MarquezProject/marquez/blob/main/integrations/spark/README.md and the context at https://github.com/MarquezProject/marquez/blob/main/integrations/spark/src/main/java/marquez/spark/agent/MarquezContext.java#L29-L31 ).

Given the implementation already uses UUIDs for job runs, I think the requirement to pass the additional job information is a bit burdensome and redundant. It also leaves room for confusion in implementation projects, which would be nice to eliminate.

**Proposed implementation**
I propose requiring job run ids be globally unique (at least within a given OpenLineage backend) and remove the `job` parameter from the `ParentRunFacet`. Integrations should pass around these globally unique runIds to link job runs to parents. This is in line with the existing Marquez model (see https://github.com/MarquezProject/marquez/blob/main/clients/python/marquez_client/client.py#L325-L327 ) and will reduce confusion in integration implementations, such as the spark integration project.

julienledem:
I'm in favor of making the runId a UUID

wslulciuc:
+1

wslulciuc:
Fixed #65

PR #65, opened 2021-06-17
Set `Run.runId` format as uuid, according to RFC4122

wslulciuc:
This PR sets the`Run.runId` format as `uuid`, according to [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122). A `uuid` is a valid Json schema resource identifier introduced in `2019-09`, see [section 7.3.5](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3.5).

Note, please see #29 outlining the reasoning for using the `uuid` type for `Run.runId`.

### DIFF
--- a/client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
+++ b/client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
@@ -16,6 +16,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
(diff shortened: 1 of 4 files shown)


julienledem (commented):
This looks good.
Can you also update the spec?
https://github.com/OpenLineage/OpenLineage/blame/f0eb44722f7308dea9a5332258ed2294fb4bf957/spec/OpenLineage.md#L22
https://github.com/OpenLineage/OpenLineage/blame/f0eb44722f7308dea9a5332258ed2294fb4bf957/spec/OpenLineage.md#L33

julienledem (approved):

PR #85, opened 2021-07-07
dbt: parsing manifest

mobuchowski:
This PR adds basic library for parsing dbt artifacts and generating OpenLineage events based on them. This supports only v2 json artifacts, with schema hosted on https://schemas.getdbt.com/. In particular, it does not provide

- support for database-specific facets
- support for databases other than snowflake and bigquert
- support for running dbt in a way that results in event being generated 
- support for sending events

Subsequent PR will add dbt python wrapper that emits events to OL endpoint, as well as more facets and supported databases.

### DIFF
--- a/integration/common/openlineage/common/provider/dbt.py
+++ b/integration/common/openlineage/common/provider/dbt.py
@@ -1,0 +1,265 @@
+import datetime
+import json
+import yaml
+import os
+import uuid
+from typing import List, Tuple, Dict, Optional
+
+import attr
+
+from openlineage.client.facet import DataSourceDatasetFacet, SchemaDatasetFacet, SchemaField, \
+    SqlJobFacet
+from openlineage.common import __version__
+from openlineage.client.run import RunEvent, RunState, Run, Job, Dataset
+from openlineage.common.utils import get_from_nullable_chain
+
+PRODUCER = f"openlineage-dbt/{__version__}"
+
+
+@attr.s
+class DbtRun:
+    started_at: str = attr.ib()
+    completed_at: str = attr.ib()
+    status: str = attr.ib()
+    inputs: List[Dict] = attr.ib()
+    output: Dict = attr.ib()
+    job_name: str = attr.ib()
+    namespace: str = attr.ib()
+    run_id: str = attr.ib(factory=lambda: str(uuid.uuid4()))
+
+
+@attr.s
+class DbtEvents:
+    starts: List[RunEvent] = attr.ib()
+    completes: List[RunEvent] = attr.ib()
+    fails: List[RunEvent] = attr.ib()
+
+
+@attr.s
+class DbtRunResult:
+    start: RunEvent = attr.ib()
+    complete: Optional[RunEvent] = attr.ib(default=None)
+    fail: Optional[RunEvent] = attr.ib(default=None)
+
+
+class DbtArtifactProcessor:
+    def __init__(self, project: str = 'dbt_project.yaml', skip_errors: bool = False):
+        self.dir = os.path.abspath(os.path.dirname(project))
+        self.project = self.load_yaml(project)
+        self.namespace = ""
+        self.skip_errors = skip_errors
+
+    def parse(self, target: Optional[str] = None) -> DbtEvents:
+        """
+            Parse dbt manifest and run_result and produce OpenLineage events.
+        """
+        manifest = self.load_manifest(
+            os.path.join(self.dir, self.project['target-path'], 'manifest.json')
+        )
+        run_result = self.load_run_results(
+            os.path.join(self.dir, self.project['target-path'], 'run_results.json')
+        )
+
+        profile_dir = run_result['args']['profiles_dir']
+
+        profile = self.load_yaml(
+            os.path.join(profile_dir, 'profiles.yml')
+        )[self.project['profile']]
+
+        if target:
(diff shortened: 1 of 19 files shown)


integration/common/tests/dbt/large/result.json
```diff
@@ -0,0 +1,501 @@
...
+				}]
+			}
+		}
+	}]
+}, {
+	"eventType": "COMPLETE",
+	"eventTime": "2021-07-05T11:43:17.218104Z",
+	"run": {
+		"runId": "1a69c0a7-04bb-408b-980e-cbbfb1831ef7",
+		"facets": {}
```
julienledem, line 166: Could we add a dbt run facet that captures the raw dbt metadata for that run?

julienledem (approved):
This looks great!
I'm looking forward to be able to try this

integration/common/tests/dbt/large/result.json
julienledem: in dbt parlance it's the node?

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -0,0 +1,202 @@
...
+
+import attr
+
+from openlineage.client.facet import DataSourceDatasetFacet, SchemaDatasetFacet, SchemaField, \
+    SqlJobFacet
+from openlineage.common import __version__
+from openlineage.client.run import RunEvent, RunState, Run, Job, Dataset
+from openlineage.common.utils import get_from_nullable_chain
+
+PRODUCER = f"openlineage-dbt/{__version__}"
```
wslulciuc, line 15: It looks like the convention for DBT adapters is `dbt-<adapter>` (see [Community Supported](https://docs.getdbt.com/docs/available-adapters#community-supported)). So we'll want to use `dbt-openlineage`

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -0,0 +1,202 @@
...
+class DbtEvents:
+    starts: List[RunEvent] = attr.ib()
+    completes: List[RunEvent] = attr.ib()
+
+
+class DbtArtifactProcessor:
+    def __init__(self, project: str = 'dbt_project.yaml'):
+        self.dir = os.path.abspath(os.path.dirname(project))
+        self.project = self.load_yaml(project)
+        self.namespace = ""
```
wslulciuc, line 40: Is an empty string a reasonable namespace default?

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -0,0 +1,202 @@
...
+        else:
+            profile = profile['outputs'][profile['target']]
+
+        self.set_namespace(profile)
+
+        runs = self.parse_artifacts(manifest, run_result)
+
+        start_events, end_events = [], []
+        for run in runs:
+            start, end = self.to_openlineage_events(run)
```
wslulciuc, line 67: We should handle the `ValueError` thrown by `to_openlineage_events()`. We either want to log the error and continue, or exit out of the loop.

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -0,0 +1,202 @@
...
+                ),
+                producer=PRODUCER,
+                inputs=[self.node_to_dataset(node, facets=True) for node in run.inputs],
+                outputs=[self.node_to_dataset(run.output, facets=True)]
+            )
+        except Exception as e:
+            raise ValueError(e)
+
+    def node_to_dataset(self, node: Dict, facets: bool = False) -> Dataset:
+        if facets:
```
wslulciuc, line 174: I would rename the boolean `facets` to `has_facets` (or similar) to avoid confusing the variable with dict `facets`.

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+
+    def set_namespace(self, profile: Dict):
+        if profile['type'] == 'snowflake':
+            self.namespace = f"snowflake://{profile['account']}"
+        elif profile['type'] == 'bigquery':
+            self.namespace = "bigquery"
+        else:
+            raise NotImplementedError(
+                "Only 'snowflake' and 'bigquery' adapters are supported right now."
```
wslulciuc, line 201: Can you also log the unsupported type in the error message?

wslulciuc (approved):
Some general comments, but otherwise HUGE improvement in the overall implementation 💯

integration/common/openlineage/common/provider/dbt.py
mobuchowski: Yes, but we're not dbt adapter here - I'm using our producer terminology, as `openlineage-airflow`. What do you think?

integration/common/openlineage/common/provider/dbt.py

integration/common/openlineage/common/provider/dbt.py

integration/common/openlineage/common/provider/dbt.py
mobuchowski: The namespace is set later in `extract_namespace` method. We need dbt profile to properly get the namespace, so the value is just a placeholder.

integration/common/openlineage/common/provider/dbt.py
wslulciuc: Sounds good 👍

PR #99, opened 2021-07-21
openlineage#98 fix spark.openlineage.url

OleksandrDvornik:


OleksandrDvornik:
I have inlined default values for ArgumentParser, cause they were different for parse url and detailed config.

collado-mike (approved):

Issue #89, opened 2021-07-12
Update OpenLineage CI to publish to artifactory


collado-mike:
See https://github.com/OpenLineage/OpenLineage/issues/90
Unreleased packages should be deployed to 	
https://datakin.jfrog.io/artifactory/datakin-pypi/ for testing

PR #109, opened 2021-07-27
Update OpenLineage CI to publish to artifactory #89

OleksandrDvornik:


client/java/build.gradle
```diff
@@ -135,10 +135,18 @@ publishing {
...
+                credentials {
+                    username = System.getenv('SONATYPE_USER')
+                    password = System.getenv('SONATYPE_PASSWORD')
+                }
+            } else {
+                url = 'https://datakin.jfrog.io/artifactory/maven-public-libs-snapshot'
+                credentials {
+                    username = System.getenv('ARTIFACTORY_USERNAME')
+                    password = System.getenv('ARTIFACTORY_PASSWORD')
+                }
```
collado-mike, line 149: I think we can manage this switching in the CI, rather than in the gradle build

integration/spark/build.gradle
```diff
@@ -231,10 +231,18 @@ publishing {
...
+                credentials {
+                    username = System.getenv('SONATYPE_USER')
+                    password = System.getenv('SONATYPE_PASSWORD')
+                }
+            } else {
+                url = 'https://datakin.jfrog.io/artifactory/maven-public-libs-snapshot'
+                credentials {
+                    username = System.getenv('ARTIFACTORY_USERNAME')
+                    password = System.getenv('ARTIFACTORY_PASSWORD')
+                }
```
collado-mike, line 245: Same comment as above

collado-mike (commented):
Given we have different release targets in the CI workflow, let's restrict the switching of what credentials/endpoint to use in those targets

client/java/build.gradle
OleksandrDvornik: Yes, we can manage it via context, for https://oss.sonatype.org/service/local/staging/deploy/maven2' we can use exist 'release' context, for https://datakin.jfrog.io/artifactory/maven-public-libs-snapshot we can create a new context like 'snapshot'. Then we also need common variable name like PUBLISH_USERNAME /  PUBLISH_PASSWORD which we will manage via different context. Are you ok with that and can you help to manage that?

collado-mike: Yeah, I think having a common env variable makes sense. You can set the common variables from circle ci- see [this example](https://circleci.com/docs/2.0/env-vars/?utm_medium=SEM&utm_source=gnb&utm_campaign=SEM-gnb-DSA-Eng-ni&utm_content=&utm_term=dynamicSearch-&gclid=CjwKCAjwxo6IBhBKEiwAXSYBs20tejw_8UMAqcLsqXwxinCO8DkENPv08q_G3aW1VmQZlyK-Svq_OhoCnbkQAvD_BwE#example-configuration-of-environment-variables)

OleksandrDvornik: I'm having error when opening circleCI/OpenLineage/Organization settings.
_Something Unexpected Happened
This is certainly annoying, our apologies. Maybe a refresh would help. If that doesn't work, you may not have permissions to view this page or you may have the incorrect URL._
@julienledem  maybe I need some additional permissions for that? Cause it works for Marquez

OleksandrDvornik:
Screenshot from 2021-08-04 11-54-52
I've tested publish job, it works, I see snapshots in repo.
I've changed back config to publish only on main

mobuchowski (approved):

PR #113, opened 2021-07-27
add Job naming convention

julienledem:


mobuchowski:
I'd think about one sentence or two to address potential confusion (that I had!) that jobs namespaces and dataset namespaces are somehow connected. While job namespace should be unique to the orchestrator instance/environment, job running in that namespace can potentially operate on datasets in all dataset namespaces.

spec/Naming.md
```diff
@@ -108,13 +108,52 @@ Identifier :
...
+For example an Airflow DAG contains tasks. An instance of the DAG is finished when all of the tasks are finished. Similarly a Spark job can spawn multiple actions each of them running independently. Additionally, a Spark job can be launched by an Airflow task within a DAG.
+
+Since what we care about is identifying the job as rooted in a recurring schedule, we want to capture that connection and make sure that we treat the same application logic triggered at different schedules as different jobs. For example: if an Airflow DAG runs individual tasks per partition (for example market segments) using the same underlying job logic, they will be tracked as separate jobs.
+
+To capture this, a run event provides a `ParentRun` facet, referring to the parent `Job` and `Run`. This allows tracking a recurring job from the root of the schedule it is running for. 
+
+https://github.com/OpenLineage/OpenLineage/blob/9445948e41b1880a349cb45ea31f2d6f65225ee0/spec/OpenLineage.json#L268-L317
+
+Example:
+{
```
wslulciuc, line 150: _Minor_: You'll want to use a code block here to keep the json indentation.

spec/Naming.md
```diff
@@ -108,13 +108,52 @@ Identifier :
...
+### Parent job run: a nested hierarchy of Jobs
+
+It is often the case that jobs are part of a nested hierarchy.
+For example an Airflow DAG contains tasks. An instance of the DAG is finished when all of the tasks are finished. Similarly a Spark job can spawn multiple actions each of them running independently. Additionally, a Spark job can be launched by an Airflow task within a DAG.
+
+Since what we care about is identifying the job as rooted in a recurring schedule, we want to capture that connection and make sure that we treat the same application logic triggered at different schedules as different jobs. For example: if an Airflow DAG runs individual tasks per partition (for example market segments) using the same underlying job logic, they will be tracked as separate jobs.
+
+To capture this, a run event provides a `ParentRun` facet, referring to the parent `Job` and `Run`. This allows tracking a recurring job from the root of the schedule it is running for. 
+
+https://github.com/OpenLineage/OpenLineage/blob/9445948e41b1880a349cb45ea31f2d6f65225ee0/spec/OpenLineage.json#L268-L317
```
wslulciuc, line 147: _Minor_: This links to the `ParentRunFacet` definition, but you'll also want to use the version that it's `main` (not a commit). I would just add the link in the paragraph above:

> To capture this, a run event provides a [`ParentRun`](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json) facet, referring ...

wslulciuc (approved):
Left some minot comments, otherwise 👍

spec/Naming.md
julienledem: makes sense

spec/Naming.md
julienledem: It's a tradeoff here. I wanted to make sure the line numbers would stay correct by pointing to a fixed commit. main is probably better though.

wslulciuc: Ahh, ok 👍

julienledem:
Per @mobuchowski 's comment, I have updated the text at the top of `Naming.md` and added links from the main spec int the definition of Jobs and Datasets

julienledem:
I have also adressed @wslulciuc 's comments

spec/Naming.md
```diff
@@ -108,13 +108,52 @@ Identifier :
...
+Jobs have a `name` that is unique to them in their `namespace` by construction.
+
+The Namespace is the root of the naming hierarchy. The job name is constructed to identify the job within that namespace.
+
+Example:
+ - Airflow:
+   - Namespace: the namespace is assigned to the airflow instance. Ex: airflow-staging, airflow-prod
+   - Job: each task in a DAG is a job. name: {dag name}.{task name}
+ - Spark:
+   - Namespace: the namespace is provided as a configuration parameter as in airflow. If there's a parent job, we use the same namespace, otherwise it is provided by configuration.
```
collado-mike, line 134: I'd move the line about the parent job namespace to the `Parent job run` section

collado-mike: Also, what happens if a client reports a job with a namespace and a parent run id, but the namespace doesn't match the parent run's namespace?

julienledem: In that case, the consumer of the event takes the namespace into account in the fully qualified path, using the same namespace allows for a shorter (and more readable) one.

julienledem:
And addressed @collado-mike 's comments

PR #114, opened 2021-07-27
add OutputStatistics facet

julienledem:

### DIFF
--- a/spec/OpenLineage.json
+++ b/spec/OpenLineage.json
@@ -153,6 +153,9 @@
               "description": "The output facets for this dataset",
               "type": "object",
               "properties": {
+                "outputStatistics": {
+                  "$ref": "#/definitions/OutputStatisticsOutputDatasetFacet"
+                }
               },
               "additionalProperties": {
                 "$ref": "#/definitions/CustomFacet"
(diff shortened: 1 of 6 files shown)


mobuchowski:
I think if we're adding something to official json schema, we need to mention it in docs too.

spec/OpenLineage.json
```diff
@@ -380,7 +383,29 @@
       ],
       "type": "object"
     },
-
+    "OutputStatisticsOutputDatasetFacet": {
```
mobuchowski, line 386: I think if we're adding facets to json schema, we need to mention them in the docs here too:

https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#standard-facets

julienledem: Yes! Thank you for pointing this out

julienledem: I just added it.

spec/OpenLineage.json
```diff
@@ -380,7 +383,29 @@
...
+        },
+        {
+          "type": "object",
+          "properties": {
+            "rowCount": {
+              "description": "The number of rows written to the dataset",
+              "type": "integer"
+            },
+            "size": {
+              "description": "The size in mb written to the dataset",
```
collado-mike, line 399: Both the Spark and BigQuery integrations publish this in bytes

julienledem: thanks!

collado-mike (approved):

PR #117, opened 2021-07-28
dbt: wrapper script

mobuchowski:

### DIFF
--- a/integration/common/openlineage/common/provider/dbt.py
+++ b/integration/common/openlineage/common/provider/dbt.py
@@ -34,6 +34,9 @@ class DbtEvents:
     completes: List[RunEvent] = attr.ib()
     fails: List[RunEvent] = attr.ib()
 
+    def events(self):
+        return self.starts + self.completes + self.fails
+
 
 @attr.s
 class DbtRunResult:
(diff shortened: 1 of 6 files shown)


integration/common/openlineage/common/provider/dbt.py
```diff
@@ -253,6 +254,17 @@ def node_to_dataset(self, node: Dict, has_facets: bool = False) -> Dataset:
             facets=has_facets
         )
 
+    def extract_fields(self, columns: List[Dict]) -> List[SchemaField]:
+        fields = []
+        for field in columns:
+            type = None
+            if 'data_type' in field and field['data_type'] is None:
+                type = field['data_type']
```
julienledem, line 262: do you mean field['data_type'] is **not** None ?
wouldn't type be always None?

julienledem (approved):
This looks good to me.
I added a comment on the type field logic in schema.
Can you add a README on how to use it?

integration/common/openlineage/common/provider/dbt.py
mobuchowski: Yes. Fixed.

julienledem (approved):

PR #118, opened 2021-07-28
separate facets to allow versioning them independently

julienledem:
This PR includes:
- Splitting facet outside of the core spec into their own file.
  - I wrote a separate automated program to generate each facet file from the original so you don't need to check that they have all the fields.
  - You might to review what has been removed from OpenLineage.json and the general structure of each facet
- Updating code gen to deal with that
  - Code generation merges all the definitions to be compatible with the existing logic.
  - Adding more precise list of BaseFacets (for each entity type). Which in code gen allows for type checking adding the right type of facet in the right place
  - It also adds a default implementation to the BaseFacets to allow deserializing custom facets.


client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
```diff
@@ -230,6 +215,138 @@ private void generateTypes(TypeSpec.Builder containerTypeBuilder) {
...
+      }
+
+      // additionalFields
+      if (type.hasAdditionalProperties()) {
+        addAdditionalProperties(type, classBuilder, constructor);
+      }
+      classBuilder.addMethod(constructor.build());
+      containerTypeBuilder.addType(classBuilder.build());
+    }
+    ///////////////////////////////
```
collado-mike, line 261: Maybe just split the above into its own method?

collado-mike (approved):

client/java/generator/src/main/java/io/openlineage/client/TypeResolver.java
```diff
@@ -31,140 +35,170 @@
...
-            // base interface
-            baseTypes.add(resolvedChildType.getName());
-            parents.add(resolvedChildType.getName());
+
+        @Override
+        public ResolvedType visit(OneOfType oneOfType) {
+          List<Type> types = oneOfType.getTypes();
+          if (types.size() != 1) {
+            throw new UnsupportedOperationException("Only oneOf of size 1 accepted for now: " + types);
           }
```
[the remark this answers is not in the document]
julienledem: That's a good point, I've been moving away from this. This was functionally equivalent to a Ref.

client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
julienledem: done

client/java/generator/src/main/java/io/openlineage/client/Generator.java
```diff
@@ -16,67 +24,155 @@
...
+  public static URL verifySchemaVersion(URL url) throws IOException {
+    InputStream input = url.openStream();
+    ObjectMapper mapper = new ObjectMapper();
+    JsonNode schema = mapper.readValue(input, JsonNode.class);
+    if (schema.has("$id") && schema.get("$id").isTextual()) {
+      URL idURL = new URL(schema.get("$id").asText());
+      try (InputStream openStream = idURL.openStream();) {
+        JsonNode published = mapper.readValue(openStream, JsonNode.class);
+        if (!published.equals(schema)) {
+          throw new RuntimeException("You must increment the version when modifying the schema. The current schema at " + url + " has the $id " + idURL + " but the version at that URL does not match.");
```
wslulciuc, line 86: _Minor:_ We can be more specific and throw an `IllegalArgumentException`

client/java/generator/src/main/java/io/openlineage/client/Generator.java
```diff
@@ -16,67 +24,155 @@
...
   }
 
-  public static void generate(String baseURL, InputStream input, File output) {
-    if (!output.getParentFile().exists()) {
-      if (!output.getParentFile().mkdirs()) {
-        throw new RuntimeException("can't create output " + output.getAbsolutePath());
+  public static void generate(Set<URL> urls, File outputBase) {
+    if (!outputBase.exists()) {
+      if (!outputBase.mkdirs()) {
+        throw new RuntimeException("can't create output " + outputBase.getAbsolutePath());
```
wslulciuc, line 99: _Minor:_ Similar comment as above to throw an `IllegalArgumentException`

client/java/generator/src/main/java/io/openlineage/client/Generator.java
```diff
@@ -16,67 +24,155 @@
...
+        if (path.endsWith(JSON_EXT)) {
+          String containerClassName = path.substring(path.lastIndexOf('/') + 1, path.lastIndexOf('.'));
+          containerToID.put(containerClassName, url);
+        } else {
+          throw new IllegalArgumentException("inputs should end in " + JSON_EXT + " got " + path);
         }
       }
-      TypeResolver typeResolver = new TypeResolver(schema);
+
+      String containerClassName = "OpenLineage";
```
wslulciuc, line 121: _Minor:_ Can these be defined as class constants?

```java
CONTAINER_CLASS_NAME = "OpenLineage";
JAVA_PATH = CONTAINER_CLASS_NAME+ "OpenLineage";
```

client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
```diff
@@ -220,6 +229,11 @@ public String getDescription() {
       return description;
     }
 
+    @Override
+    public String toString() {
+      return "Field{name: " + name + ", type: " + type + "}";
```
wslulciuc, line 234: _Minor:_ We may want to standardize on `toString()` method overrides with lombok's [`ToString`](https://projectlombok.org/features/ToString) annotation. Looks like throughout the generator, a few formats are used.

wslulciuc (approved):
Great to see facets being moved into their own schema definitions to allow for independent versioning from the core spec! 💯 💯 💯

client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
julienledem: I fixed the inconsistencies thanks for pointing it out.
lombok is not a dependency at the moment so holding off on that.

client/java/generator/src/main/java/io/openlineage/client/Generator.java
julienledem: Ok

PR #124, opened 2021-07-30
add version in spec

julienledem:


collado-mike (approved):

wslulciuc:
@julienledem: Based on the versioning for `SchemaVer`, versioning starts from `1`, not `0` as in SemVer:

> We gave some serious thought to recreating SemVer’s unstable `MAJOR` version 0 idea. On balance, this seemed a bad idea for SchemaVer: because inevitably some `MODEL` version 0s will go into production, and then we lose our all-important guarantees about the relationship between schema versions and the historical data.

see: https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/

Based on this, should the version be `1-0-0`?

PR #130, opened 2021-08-03
[spark-client-integration] changes in java client for spark integration

OleksandrDvornik:
Changes required for [PR](https://github.com/OpenLineage/OpenLineage/pull/131)

PR #131, opened 2021-08-03
Integrate spark with java client spark part

OleksandrDvornik:
Depends on [PR](https://github.com/OpenLineage/OpenLineage/pull/130)
Java client should be published to snapshot, before merge this PR.
Test can fail due to dependency and latest changes in openlineage-java


julienledem:
This looks good to me. @collado-mike @wslulciuc ?

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/RddExecutionContext.java
```diff
@@ -190,19 +200,16 @@ protected ZonedDateTime toZonedTime(long time) {
     return ZonedDateTime.ofInstant(i, ZoneOffset.UTC);
```
collado-mike, line 200: I think we should stop reporting the timestamps in UTC- the processing jobs are often running in locale-specific timezones and that is useful information to collect.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -73,41 +68,69 @@ public void start(SparkListenerJobStart jobStart) {
...
-            .eventType("START")
-            .producer(OpenLineageClient.OPEN_LINEAGE_CLIENT_URI)
             .build();
 
     log.debug("Posting event for start {}: {}", jobStart, event);
     sparkContext.emit(event);
   }
 
-  private ParentRunFacet buildParentFacet() {
+  private List<OpenLineage.Dataset> getInputDatasets() {
```
collado-mike, line 103: Why not do the conversion to `InputDataset` here?

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -73,41 +68,69 @@ public void start(SparkListenerJobStart jobStart) {
       log.info("No execution info {}", queryExecution);
       return;
     }
-    List<Dataset> outputDatasets =
+    List<OpenLineage.Dataset> outputDatasets =
         PlanUtils.applyFirst(outputDatasetSupplier, queryExecution.optimizedPlan());
```
collado-mike, line 72: Pull out this line and the conversion to `OutputDataset` into a separate method

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -73,41 +68,69 @@ public void start(SparkListenerJobStart jobStart) {
...
+        ol.newRunEventBuilder()
+            .eventTime(toZonedTime(jobStart.time()))
+            .eventType("START")
+            .inputs(
+                inputDatasets.stream()
+                    .map(SparkSQLExecutionContext::convertToInput)
+                    .collect(Collectors.toList()))
+            .outputs(
+                outputDatasets == null
+                    ? null
```
collado-mike, line 86: We shouldn't be passing in `null` to the list of outputs- use an empty list instead

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -119,28 +142,32 @@ public void end(SparkListenerJobEnd jobEnd) {
...
+                    : outputDatasets.stream()
+                        .map(SparkSQLExecutionContext::convertToOutput)
+                        .collect(Collectors.toList()))
```
collado-mike, line 163: Same comments re: conversion of input/output datasets and the use of `null`

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/DatasetSourceVisitor.java
```diff
@@ -61,12 +61,12 @@ public boolean isDefinedAt(LogicalPlan x) {
...
 
   @Override
-  public List<LineageEvent.Dataset> apply(LogicalPlan x) {
+  public List<OpenLineage.Dataset> apply(LogicalPlan x) {
     DatasetSource datasetSource =
         findDatasetSource(x)
             .orElseThrow(() -> new RuntimeException("Couldn't find DatasetSource in plan " + x));
     return Collections.singletonList(
-        LineageEvent.Dataset.builder()
+        new OpenLineage.InputDatasetBuilder()
```
collado-mike, line 69: So is the strategy throughout to just always use `InputDataset`, then convert to the appropriate type later? Can you add a comment explicitly stating that? Add a `TODO` to fix this

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java
```diff
@@ -78,28 +81,28 @@ public boolean isDefinedAt(LogicalPlan x) {
       // make a best attempt at capturing the dataset information
       log.warn("Don't know how to extract dataset from unknown relation {}", logRel.relation());
       return Collections.singletonList(
-          LineageEvent.Dataset.builder()
+          ol.newInputDatasetBuilder()
```
collado-mike, line 84: Same comment re: the TODO for the `InputDataset` construction

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/PlanUtils.java
```diff
@@ -92,100 +92,107 @@ public static String namespaceUri(URI outputPath) {
...
-    return LineageEvent.ParentRunFacet.builder()
-        ._producer(URI.create(OpenLineageClient.OPEN_LINEAGE_CLIENT_URI))
-        ._schemaURL(URI.create(OpenLineageClient.OPEN_LINEAGE_PARENT_FACET_URI))
-        .run(LineageEvent.RunLink.builder().runId(runId).build())
-        .job(LineageEvent.JobLink.builder().name(parentJob).namespace(parentJobNamespace).build())
+    return new OpenLineage(URI.create(OpenLineageClient.OPEN_LINEAGE_CLIENT_URI))
+        .newParentRunFacetBuilder()
+        .run(
+            new OpenLineage.ParentRunFacetRunBuilder()
+                .runId(convertToUUID.apply(runId).orElse(UUID.randomUUID()))
```
collado-mike, line 189: If there's no parent runUUID, don't just make one up. The `ParentRunFacet` should just be omitted if there is no valid run id.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/PlanUtils.java
```diff
@@ -227,4 +234,16 @@ static Path getDirectoryPath(Path p, Configuration hadoopConf) {
       return p;
     }
   }
+
+  public static final Function<String, Optional<UUID>> convertToUUID = tryConvert(UUID::fromString);
```
collado-mike, line 238: This is never used as a `Function` reference. Just make it a static method. You can always use a method reference as a `Function` if you need to.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/RddExecutionContext.java
```diff
@@ -153,15 +149,22 @@ static String nameRDD(RDD<?> rdd) {
...
+            .eventType("START")
             .inputs(buildInputs(inputs))
             .outputs(buildOutputs(outputs))
-            .run(buildRun(buildRunFacets(null)))
+            .run(
+                ol.newRunBuilder()
+                    .runId(
+                        PlanUtils.convertToUUID
+                            .apply(sparkContext.getParentRunId())
+                            .orElse(UUID.randomUUID()))
```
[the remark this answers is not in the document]
OleksandrDvornik: but it's existing logic, you mean we should change it? For me it was strange that we set parentRunId to runId, but don't want to change it is scope of this ticket
```
 protected LineageEvent.Run buildRun(RunFacet facets) {
    return LineageEvent.Run.builder().runId(sparkContext.getParentRunId()).facets(facets).build();
  }
```
https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/main/java/openlineage/spark/agent/lifecycle/RddExecutionContext.java#L194

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/RddExecutionContext.java
```diff
@@ -170,15 +173,22 @@ public void start(SparkListenerJobStart jobStart) {
...
+            .eventType(getEventType(jobEnd.jobResult()))
             .inputs(buildInputs(inputs))
             .outputs(buildOutputs(outputs))
-            .run(buildRun(buildRunFacets(buildJobErrorFacet(jobEnd.jobResult()))))
+            .run(
+                ol.newRunBuilder()
+                    .runId(
+                        PlanUtils.convertToUUID
+                            .apply(sparkContext.getParentRunId())
+                            .orElse(UUID.randomUUID()))
```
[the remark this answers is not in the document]
OleksandrDvornik: ok, so we should have runId also in context as parentRunId? We should discuss that logic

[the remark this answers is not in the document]
OleksandrDvornik: should we then map jobId to some UUID and then set runId based on mapping jobId to UUID? Cause current logic is incorrect, I suppose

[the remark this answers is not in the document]
collado-mike: Hmm, the `SparkSQLExecutionContext` impl is correct- the `RddExecutionContext` is wrong. 
See https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/main/java/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java#L163 and https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/main/java/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java#L105-L108 . Most notably, the `runUuid` is only created once and passed to both the start and the end event. Otherwise, the server would be receiving a start event for one run id and an end event for another run id and there's no way to know that they are actually from the same run.

OleksandrDvornik:
@collado-mike I've put in `RddExecutionContext` the same logic for `SparkSQLExecutionContext`, generating runId on init.
Add wrappers for Input and Output Visitors.
Move visitors to separate package.
Apply changes to test data.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/PlanUtils.java
```diff
@@ -92,104 +90,100 @@ public static String namespaceUri(URI outputPath) {
...
+  public static OpenLineage.Dataset getDataset(URI outputPath, StructType schema) {
     String namespace = namespaceUri(outputPath);
-    LineageEvent.DatasetFacet datasetFacet = datasetFacet(schema, namespace);
-    return getDataset(outputPath, namespace, datasetFacet);
+    OpenLineage.DatasetFacets datasetFacet = datasetFacet(schema, namespace);
+    return getDataset(outputPath.getPath(), namespace, datasetFacet);
   }
 
   /**
    * Construct a dataset given a {@link URI}, namespace, and preconstructed {@link
```
collado-mike, line 107: Make `dataset` a `{@link Openlineage.Dataset}`. I think it's worth pointing out here that the dataset is neither an `InputDataset` nor an `OutputDataset`, but an anonymous instance since we don't know if we're creating inputs or outputs yet.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/visitor/LogicalRelationVisitor.java
```diff
@@ -78,28 +82,30 @@ public boolean isDefinedAt(LogicalPlan x) {
...
 
-  private List<LineageEvent.Dataset> handleCatalogTable(LogicalRelation logRel) {
+  private OpenLineage.DatasetFacets getDatasetFacet(OpenLineage ol, LogicalRelation logRel) {
+    return ol.newDatasetFacetsBuilder()
+        .documentation(ol.newDocumentationDatasetFacet(logRel.simpleString()))
+        .schema(PlanUtils.schemaFacet(logRel.schema()))
+        .build();
+  }
+
+  private String getDatasetName(LogicalRelation logRel) {
```
collado-mike, line 96: `get[Unknown]DatasetName`

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/visitor/DatasetSourceVisitor.java
```diff
@@ -61,15 +62,39 @@ public boolean isDefinedAt(LogicalPlan x) {
...
+   * DatasetSource is an interface that allows instrumenting an existing Relation or Datasource
+   * class with the ability to report its OpenLineage namespace and name. Relations that intend to
+   * participate in the OpenLineage reporting can implement this interface directly and they will be
+   * found in the Spark {@link LogicalPlan} during plan execution. Otherwise, {@link
+   * java.lang.instrument.ClassFileTransformer}s can be instrumented to dynamically rewrite classes
+   * to implement this interface with logic to return the necessary data. In that case, the
+   * instrumented class will still be found in the {@link LogicalPlan} during execution and the
+   * datasource details will be included in the lineage.
+   */
+  public interface DatasetSource {
```
collado-mike, line 86: Why was this moved from the original file? For developers who might plan to implement this directly, I think it makes more sense to keep it as its own file.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/wrapper/InputDatasetVisitor.java
```diff
@@ -0,0 +1,36 @@
+package openlineage.spark.agent.lifecycle.plan.wrapper;
+
+import io.openlineage.client.OpenLineage;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import scala.PartialFunction;
+import scala.runtime.AbstractPartialFunction;
+
+public class InputDatasetVisitor
```
collado-mike, line 10: I found it clearer when this was logic that was applied to the `Dataset`s returned by the visitor. It's not really a plan visitor- it's a function we apply to the visitors' results. I think we should keep the logic as it was before.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/wrapper/OutputDatasetVisitor.java
```diff
@@ -0,0 +1,36 @@
+package openlineage.spark.agent.lifecycle.plan.wrapper;
+
+import io.openlineage.client.OpenLineage;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import scala.PartialFunction;
+import scala.runtime.AbstractPartialFunction;
+
+public class OutputDatasetVisitor
```
collado-mike, line 10: Same comment here.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/wrapper/OutputDatasetWithMetadataVisitor.java
```diff
@@ -0,0 +1,66 @@
...
+    OpenLineage ol = new OpenLineage(URI.create(OpenLineageClient.OPEN_LINEAGE_CLIENT_URI));
+
+    Map<String, SQLMetric> metrics =
+        statisticsCommands.entrySet().stream()
+            .filter(e -> e.getKey().isInstance(x))
+            .map(e -> e.getValue().apply(x))
+            .findFirst()
+            .get();
+    OpenLineage.OutputStatisticsOutputDatasetFacet outputStats =
+        PlanUtils.getOutputStats(ol, metrics);
```
collado-mike, line 50: This is good logic to factor out of the command visitors, but I think it would be better off as a utility method or possibly a method in a shared base class.

integration/spark/src/test/java/openlineage/spark/agent/lifecycle/StaticExecutionContextFactory.java
```diff
@@ -10,6 +10,7 @@
 import java.util.concurrent.TimeoutException;
 import openlineage.spark.agent.OpenLineageContext;
 import openlineage.spark.agent.OpenLineageSparkListener;
+import openlineage.spark.agent.lifecycle.plan.CommonDatasetVisitors;
```
collado-mike, line 13: This doesn't really warrant its own class. I get the desire to reduce duplication, but it's really just a list of common visitors. Why not just construct the list and pass it to each of the input and output visitors?

collado-mike (commented):
That was clever use of an anonymous class to get around the Input/Output Dataset issue.  Some of the refactoring I think is unnecessary and doesn't make the code easier to understand or easier to use. I think we should revert the changes that aren't related to the OpenLineage client change and we can discuss other refactoring later.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/wrapper/OutputDatasetVisitor.java
OleksandrDvornik: We should have at least OutputDataset cause we need to set OutputStatisticsOutputDatasetFacet to it

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/wrapper/OutputDatasetWithMetadataVisitor.java
OleksandrDvornik: We should have at least OutputDataset cause we need to set OutputStatisticsOutputDatasetFacet to it. If we have Dataset, we cant't set OutputStatisticsOutputDatasetFacet to outputFacets, cause it's only in OutputDataset.
What is a downside of using wrappers? As a proc I can add that we can dynamically add functionality to it.
I see another option as it was, set outputStatistics  to addtitionalProps. and then we will try to get "stats" field from additionalProperties and then set it to outputFacets. As for me, wrapper is a better option.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/visitor/DatasetSourceVisitor.java
OleksandrDvornik: my bad, I was thinking that it's used in one place, but it's not,next time will search by name not only by usages

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/wrapper/InputDatasetVisitor.java
OleksandrDvornik: not only, you can check OutputDatasetWithMetadataVisitor there is usage of LogicalPlan, not only wrapping a result

OleksandrDvornik:
I've rollback grouping a visitors and move back interface DatasetSource. i'm still thinking that we need wrappers around Dataset. Cause before OutputStatistics come into play it was ok to have one common dataset, but right now, we need to have separation for Input and Output Datasets. 
I've made it in wrapper/decorator style which allows adding new feature to Dataset. New feature in current case is OutputStatisticsOutputDatasetFacet. 
For me it's very clear that we have Visitors which process LogicalPlan Nodes and return (basic/common) Dataset (we can even don't use Dataset but some other model NodeData?) and then we have  wrapper which also needs LogicalPlan for other meta (e.g. OutputStatistics), that's why it's also Visitor.

integration/spark/src/main/java/openlineage/spark/agent/lifecycle/plan/CommonDatasetVisitors.java
```diff
@@ -0,0 +1,35 @@
...
+import io.openlineage.client.OpenLineage;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+import openlineage.spark.agent.OpenLineageContext;
+import org.apache.spark.sql.SQLContext;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import scala.PartialFunction;
+
+public class CommonDatasetVisitors
```
collado-mike, line 12: This isn't used anymore

collado-mike (approved):
I get the reasoning for the input/output dataset visitor wrappers. Personally, I'd have written them as Functions on the returned Dataset, rather than as PartialFunction wrappers, but I think this code is going to end up being revisited anyway to support custom facets and dataset implementation-specific facets (e.g., BigQuery-specific rows/bytes read, HDFS bytes/# of files read, etc).

PR #134, opened 2021-08-03
dbt: readme

mobuchowski:

### DIFF
--- a/integration/dbt/README.md
+++ b/integration/dbt/README.md
@@ -1,0 +1,44 @@
+# OpenLineage dbt integration
+
+Wrapper script for dbt for automatic metadata collection
+
+## Features
+
+**Metadata**
+
+* Model run lifecycle
+* Model inputs / outputs
+
+## Requirements
+
+- [Python >= 3.6](https://www.python.org/downloads)
+- [dbt >= 0.20](https://www.getdbt.com/)
+
+## Installation
+
+```bash
+$ pip3 install openlineage-dbt
+```
+
+To install from source, run:
+
+```bash
+$ pip install .
+```
+
+## Configuration
+
+
+### `HTTP` Backend Environment Variables
+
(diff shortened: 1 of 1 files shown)


integration/dbt/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+### `HTTP` Backend Environment Variables
+
+`openlineage-dbt` uses OpenLineage client to push data to OpenLineage backend.
+
+OpenLineage client depends on environment variables:
+
+* `OPENLINEAGE_URL` - point to service which will consume OpenLineage events
+* `OPENLINEAGE_PRODUCER` - name of producer that client will send along with your events. This will be dropped in future versions.
```
julienledem, line 39: the PRODUCER should be something that identifies the integration itself. It is not configurable.
So defined in the code as something like:
`https://github.com/OpenLineage/OpenLineage/integration/dbt/0.0.1`

mobuchowski: Should we have full URLs, or something like `openlineage-dbt/0.0.1`?

wslulciuc: We'll want to remove support for setting the producer via `OPENLINEAGE_PRODUCER`. The producer should be a link that the integration knows how to build. Using an environment variables means that someone has to be update the value when the client version changes, which may not always happen .

julienledem: It is a URI: https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L45-L50
I would favor the full URL to a tagged version: `https://github.com/OpenLineage/OpenLineage/tree/{tag}/integration/airflow/`

mobuchowski: Removed mention of `OPENLINEAGE_PRODUCER` here. Will remove support for setting this variable in next PR.

wslulciuc (approved):
LGTM 👍

PR #137, opened 2021-08-04
dbt: use column-level catalog metadata

mobuchowski:
If file `catalog.json` is found in target directory, use it to gather column and table-level metadata, like column types and `BigQueryStatisticsDatasetFacet`.


wslulciuc (approved):
Nice work, @mobuchowski! 💯

PR #140, opened 2021-08-05
Add mechanism to enforce spec versioning

julienledem:
This modifies the code generator to:
 - always use the local version of the spec
 - issue a warning if the spec is not public at $id
 - verify that the spec matches the one published at $id if available
 - fail the build if the spec does not match to require changing the version.

This will guarantee that we don't reuse published specs.
It also removes the need for the code to be pushed to the repo for the build to work.

### DIFF
--- a/client/java/generator/src/main/java/io/openlineage/client/Generator.java
+++ b/client/java/generator/src/main/java/io/openlineage/client/Generator.java
@@ -1,7 +1,7 @@
 package io.openlineage.client;
 
 import java.io.File;
-import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.PrintWriter;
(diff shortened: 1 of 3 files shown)


spec/OpenLineage.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/0-0-1/OpenLineage.json",
+  "$id": "https://openlineage.io/spec/1-0-0/OpenLineage.json",
```
wslulciuc, line 3: 💯

wslulciuc (approved):

spec/OpenLineage.json
julienledem: I knew I would get your attention :)

client/java/generator/src/main/java/io/openlineage/client/Generator.java
```diff
@@ -55,10 +55,30 @@ public static void generate(String baseURL, InputStream input, File output) {
     try {
       ObjectMapper mapper = new ObjectMapper();
       JsonNode schema = mapper.readValue(input, JsonNode.class);
+      if (schema.has("$id") && schema.get("$id").isTextual()) {
```
wslulciuc, line 58: _Minor_: Do we need to throw an exception if the node is not textual, or is it ok to just skip / continue?

collado-mike (approved):

PR #160, opened 2021-08-11
Added original OutputStatisticFacet back to Dataset facets for backward compatibility

collado-mike:
Signed-off-by: Michael Collado <mike@datakin.com>


wslulciuc (approved):
LGTM 👍

PR #163, opened 2021-08-11
Fix snapshot builds

collado-mike:
Fixes python development builds to append `.dev${CIRCLE_BUILD_NUM}` to snapshot releases.

I tested by temporarily adding support for the `dev[0-9]` tag regex and pushing the tag `dev1`. I'll delete the tag, but the source can be viewed at https://github.com/OpenLineage/OpenLineage/commit/265e790f39684260df7d977eeabe8cbabe1c18a0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,6 @@ install_integration_common: &install_integration_common
 install_python_client: &install_python_client
   run: (cd ~/openlineage/client/python && pip install . --user)
 
-
 # Only trigger CI job on release (=X.Y.Z) with possible (rcX)
 only-on-release: &only-on-release
   filters:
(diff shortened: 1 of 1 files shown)


julienledem (approved):

wslulciuc (commented):
Should we just make the `build_tag` [param global](https://circleci.com/docs/2.0/pipeline-variables/#pipeline-parameters-in-configuration) to avoid setting the param at the job level?

collado-mike:
> Should we just make the `build_tag` [param global](https://circleci.com/docs/2.0/pipeline-variables/#pipeline-parameters-in-configuration) to avoid setting the param at the job level?
> 
> ```yaml
> parameters:
>   build-tag:
>     type: string
>     default: .dev${CIRCLE_BUILD_NUM}
> ```

I don't think you can set pipeline parameters after the build has started- based on the documentation, they have to be set when triggering the run through the API

wslulciuc:
I see, yeah then let's merge as is and optimize the CI configuration at a later point 👍

wslulciuc (approved):
LGTM 👍 💯 🥇

PR #197, opened 2021-08-13
[2 of 2] Cleanup .circleci/config.yml

wslulciuc:
This PR cleans up the `.circleci/config.yml` with the following changes:

* Standardize CI job names
* Add caching and test report to CI job `build-client-java`
* Remove the filter `only-on-main-without-release-tag` (it's not needed and we can filter only on `main`)
* Slight reorder of CI jobs to keep related jobs close and group them by context
* Removed `circleci/gradle@2.2.0`, it's not needed and already handled by `build-client-java`
* Add missing dependencies between CI jobs
* Prefix unit test CI jobs with `unit-`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ param_build_tag: &param_build_tag
         type: string
 
 jobs:
-  test-client-python:
+  unit-test-client-python:
     working_directory: ~/openlineage/client/python
     docker:
       - image: circleci/python:3.6
@@ -163,7 +163,7 @@ jobs:
           paths:
             - ~/.gradle
 
-  test-integration-common:
+  unit-test-integration-common:
     working_directory: ~/openlineage/integration/common
     docker:
       - image: circleci/python:3.6
@@ -230,7 +230,7 @@ jobs:
           paths:
             - ~/.gradle
 
-  test-integration-airflow:
+  unit-test-integration-airflow:
     working_directory: ~/openlineage/integration/airflow
     docker:
       - image: circleci/python:3.6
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -291,71 +295,72 @@ jobs:
...
+            - test-client-python
+      - test-integration-common
       - build-integration-common:
-          <<: *only-on-main-without-release-tag
+          <<: *only_on_main
           build_tag: .dev${CIRCLE_BUILD_NUM}
+          requires:
+            - test-integration-common
+      - test-integration-airflow
+      - integration-test-airflow:
```
collado-mike, line 326: While we're cleaning up, can we just rename these tasks? This is so confusing. I'd recommend `unit-tests-airflow-integration` and `integration-tests-airflow-integration` or something along those lines.

wslulciuc: 👍

wslulciuc: I prefixed the unit tests jobs with `unit-`. I think we can leave the integration test CI jobs as is with the format `integration-test-<integration>` as it's clear the function of the job. I feel `integration-test-<integration>-integration` becomes to wordy and there are sooo many _integration_ words in the job it loses meaning.

julienledem (approved):

collado-mike (approved):
Thanks for making that change. Personally, I'd rather name all the tasks `*-airflow-integration` or have them all named `*-airflow`, etc. But I think the current names are enough to reduce confusion

.circleci/config.yml
collado-mike: I agree with the latter statement- that's why I actually suggested `integration-tests-airflow` and simply drop the last `integration` from the suffix

Issue #193, opened 2021-08-13
[INTEGRATION] [dbt] job names should include the target and skip the "model" prefix. the namespace should be user provided


julienledem:
**Integration target:**
We need to improve the job naming for the dbt integration.
https://github.com/OpenLineage/OpenLineage/tree/main/integration/dbt

1) In the dbt integration, we always have "model." as a prefix for job names. This is unnecessary as they all start with model.

2) Additionally a dbt project can be run with multiple targets defined in profiles:
https://docs.getdbt.com/dbt-cli/configure-your-profile#understanding-profiles
Depending on the "target" we should have a different job name as they run on different dataset inputs and outputs.

Possibly this can be gathered form the `database` and `schema` fields in the node.

3) The NAMESPACE should be user provided.

the openlineage-dbt config should take a namespace like the openlineage-airflow config
https://github.com/OpenLineage/OpenLineage/tree/main/integration/airflow#http-backend-environment-variables

mobuchowski:
@julienledem agreed with points 1 and 3 and addressed it in the PR: https://github.com/OpenLineage/OpenLineage/pull/203

Regarding point 2, we support different targets in profiles. What's missing is support for overriding this via CLI with `-T` or `--target`. Is this what you need?

julienledem:
@mobuchowski I think so. Users will do `--target prod` or `--target dev` which will point the dbt run to a different database schema. It is important that the job has a different name in each target.

PR #203, opened 2021-08-16
[INTEGRATION] [dbt] strip model from job name, provide job namespace by env variable

mobuchowski:
This PR strips `model.` from job names, since it's unnecessary - they are all prefixed by this.
Also, this PR provides possibility to override default job namespace by providing `OPENLINEAGE_NAMESPACE` env variable.

https://github.com/OpenLineage/OpenLineage/issues/193


julienledem (approved):
This looks good to me

PR #206, opened 2021-08-18
[INTEGRATION] [dbt] parse dbt command line arguments that override defaults

mobuchowski:
dbt run has few command line arguments that change location of artifacts, like `--project-dir` or ones that override defaults, like `--profile`. This PR adds support for few of them that can affect lineage event generation. 

Solves issue 2) in https://github.com/OpenLineage/OpenLineage/issues/193

### DIFF
--- a/integration/common/openlineage/common/utils.py
+++ b/integration/common/openlineage/common/utils.py
@@ -50,3 +50,17 @@ def get_from_multiple_chains(source: Dict[str, Any], chains: List[List[str]]) ->
         if result:
             return result
     return None
+
+
+def parse_single_arg(args, keys: List[str]) -> Optional[str]:
+    """
+    In provided argument list, find first key that has value and return that value.
+    Values can be passed either as one argument {key}={value}, or two: {key} {value}
+    """
+    for key in keys:
+        for i, arg in enumerate(args):
+            if arg == key and len(args) > i:
+                return args[i+1]
+            if arg.startswith(f"{key}="):
+                return arg.split("=", 1)[1]
+    return None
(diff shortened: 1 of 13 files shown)


julienledem:
This looks good. I think the other aspect of 2) in  #193 that is not covered here is that jobs should have a different name depending on which schema they are run against.
When you pick a different `target` in your `profile` (as defined [in the dbt doc here](https://docs.getdbt.com/dbt-cli/configure-your-profile#understanding-profiles) ) the whole project will run on a different `dbname` and `schema`. Which means that the jobs should have a different name depending on which db/schema they are run against (as they have a different lineage).

mobuchowski:
@julienledem addressed that in the following commit.

Job names are now `{database}.{schema}.{model_name}` without `model.`

julienledem (approved):
awesome!

integration/common/openlineage/common/utils.py
```diff
@@ -50,3 +50,17 @@ def get_from_multiple_chains(source: Dict[str, Any], chains: List[List[str]]) ->
         if result:
             return result
     return None
+
+
+def parse_single_arg(args, keys: List[str]) -> Optional[str]:
```
wslulciuc, line 55: Why not use [`argparse`](https://docs.python.org/3/library/argparse.html)?

wslulciuc (approved):
Left a minor comment on how we are parsing the cmd line args, otherwise 👍

integration/common/openlineage/common/utils.py
mobuchowski: I thought about it, but it's far too complex for this simple use case and requires disabling a lot of options, like help, exit on failure or usage. This function does exactly what we need, which is transparent searching for keys in passed list.

julienledem:
@mobuchowski We should default the namespace to the same namespace as the output dataset (as you have in the 0.1.0 release).

Issue #218, opened 2021-08-23
Unable to clear tasks in Google Cloud Composer


collado-mike:
The 0.1.0 release causes failures when attempting to clear a task in Google Cloud Composer. It's likely a failure in any environment that uses the CeleryExecutor.

collado-mike:
Here's the stack trace when I attempt to clear a task in the Airflow UI

![image](https://user-images.githubusercontent.com/40346148/130503402-f0004234-d018-4128-a4d1-4ace735b386e.png)

collado-mike:
Fixed in https://github.com/OpenLineage/OpenLineage/pull/255

PR #219, opened 2021-08-23
Remove instance references to extractors from DAG and avoid copying log property for serializability

collado-mike:
This addresses the issue outlined in https://github.com/OpenLineage/OpenLineage/issues/218 . I moved the Extractors map from an instance variable to a static one outside the DAG instance- this seemed reasonable, as the extractors themselves aren't stateful and don't need to be reinstantiated for each instance of the DAG. Removing them from the DAG's properties means we don't have to worry about an Extractor holding a property that can't be serialized with the DAG.

The other issue I identified with the `_log` property of the `LoggingMixin`. The logger itself isn't serializable. The parent class already inherits the `LoggingMixin`, so there's no need for us to also mixin that property. The parent class doesn't have a problem being serialized because the logger is instantiated lazily and the parent DAG doesn't make a call to the logger until after the DAG has been serialized when clearing a task in the UI. However, the OpenLineage DAG executes a log statement immediately in the constructor. I overrode the `__deepcopy__` method to skip the `_log` property when serializing.

Between these two changes, I was able to restore the Clear functionality in the Airflow UI

As a separate change, I included some logging statements for debugging purposes. This included writing OpenLineage log lines to stdout in the integration tests at DEBUG level.

### DIFF
--- a/integration/airflow/openlineage/airflow/dag.py
+++ b/integration/airflow/openlineage/airflow/dag.py
@@ -10,14 +10,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import time
+import copy
 from typing import List, Union, Optional
 
 import airflow.models
 from airflow.models import DagRun
-from airflow.utils.db import provide_session
+from airflow.utils.db import create_session
 from airflow.utils.state import State
-# Handling of import of different airflow versions
-from airflow.version import version as AIRFLOW_VERSION
 from openlineage.airflow.extractors import StepMetadata, BaseExtractor
 from openlineage.airflow.extractors.extractors import Extractors
 from openlineage.airflow.utils import (
@@ -27,24 +26,15 @@ from openlineage.airflow.utils import (
     get_custom_facets,
     new_lineage_run_id
 )
-from pkg_resources import parse_version
-
-if parse_version(AIRFLOW_VERSION) >= parse_version("2.0.0"):
-    # Corrects path of import for Airflow versions below 1.10.11
-    from airflow.utils.log.logging_mixin import LoggingMixin
-elif parse_version(AIRFLOW_VERSION) >= parse_version("1.10.11"):
-    from airflow import LoggingMixin
-else:
-    # Corrects path of import for Airflow versions below 1.10.11
-    from airflow.utils.log.logging_mixin import LoggingMixin
 
 from openlineage.airflow.adapter import OpenLineageAdapter
 
 _ADAPTER = OpenLineageAdapter()
+extractor_mapper = Extractors()
+extractors = {}
 
 
-@provide_session
-def lineage_run_id(run_id, task, session=None):
+def lineage_run_id(run_id, task):
     """
     Macro function which returns the generated run id for a given task. This
     can be used to forward the run id from a task to a child run so the job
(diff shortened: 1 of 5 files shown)


integration/airflow/openlineage/airflow/dag.py
```diff
@@ -27,24 +26,15 @@
...
-from pkg_resources import parse_version
-
-if parse_version(AIRFLOW_VERSION) >= parse_version("2.0.0"):
-    # Corrects path of import for Airflow versions below 1.10.11
-    from airflow.utils.log.logging_mixin import LoggingMixin
-elif parse_version(AIRFLOW_VERSION) >= parse_version("1.10.11"):
-    from airflow import LoggingMixin
-else:
-    # Corrects path of import for Airflow versions below 1.10.11
-    from airflow.utils.log.logging_mixin import LoggingMixin
```
wslulciuc, line 39: Nice that we no longer have to check packages based on the Airflow version 💯

integration/airflow/openlineage/airflow/dag.py
```diff
@@ -60,44 +50,38 @@ def lineage_run_id(run_id, task, session=None):
...
-        return ""
-    elif isinstance(ids, list):
-        return "" if len(ids) == 0 else ids[0]
-    else:
-        return str(ids)
-
-
-class DAG(airflow.models.DAG, LoggingMixin):
-    def __init__(self, *args, extractor_mapper=None, **kwargs):
+    with create_session() as session:
```
wslulciuc, line 55: What the benefit of using `create_session()` vs `@provide_session`?

integration/airflow/tests/integration/airflow/config/log_config.py
```diff
@@ -1,4 +1,4 @@
-# Licensed to the Apache Software Foundation (ASF) under one
+git # Licensed to the Apache Software Foundation (ASF) under one
```
wslulciuc, line 1: Typo?

integration/airflow/tests/integration/airflow/config/log_config.py
```diff
@@ -86,6 +85,11 @@
         }
     },
     'loggers': {
+        'openlineage.airflow': {
```
wslulciuc, line 88: 💯

wslulciuc (approved):
Left some minor comments, otherwise nice work @collado-mike! 🔥

integration/airflow/openlineage/airflow/dag.py
collado-mike: TBH, there's not really any. I worried that the function was being wrapped with one that held an instance of the already-created session and that the session was causing serialization issues- the user_defined_macro functions are stored as instance fields of the DAG. Looking back now, the return value of the `@provide_session` wrapper is actually just a function that only invokes `create_session` when the function is called- effectively the same as what is done here.

integration/airflow/tests/integration/airflow/config/log_config.py
collado-mike: 👍🏽

integration/airflow/openlineage/airflow/dag.py
wslulciuc: Sounds good. Perhaps being more explicit to be safe isn't a bad thing 👍

Issue #221, opened 2021-08-25
[PROPOSAL] standarize expectation assertion facet


mobuchowski:
**Purpose:**
Standarize dataset facet representing result of data test ran on dataset or some of it's columns.
This allows capturing results of data quality frameworks like [dbt-expectations](https://github.com/calogica/dbt-expectations/), [deequ](https://github.com/awslabs/deequ) or [Great Expectations](https://github.com/great-expectations/great_expectations).

Implementation is based on Great Expectations's 0.1 custom facet.

Issue #147, opened 2021-08-05
Publish automatically the spec to openlineage.io


julienledem:
When the build runs on the `main` branch, we will publish the `spec/OpenLineage.json` file to [the URL defined in the `$id` field](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L3) using [github pages](https://github.com/OpenLineage/OpenLineage.github.io).. 

See versioning specification: https://github.com/OpenLineage/OpenLineage/blob/main/spec/Versioning.md

PR #222, opened 2021-08-25
[PROPOSAL] add expectations assertions facet

mobuchowski:
This PR standarizes expectation assertion facet, as described in https://github.com/OpenLineage/OpenLineage/issues/221.

This allows using standarized facet for both Great Expectations and dbt-expectations. 

### DIFF
--- a/spec/OpenLineage.json
+++ b/spec/OpenLineage.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/1-0-0/OpenLineage.json",
+  "$id": "https://openlineage.io/spec/1-0-1/OpenLineage.json",
   "definitions": {
     "RunEvent": {
       "type": "object",
@@ -142,6 +142,9 @@
               "properties": {
                 "dataQualityMetrics" : {
                   "$ref": "#/definitions/DataQualityMetricsInputDatasetFacet"
+                },
+                "dataQualityAssertions": {
+                  "$ref": "#/definitions/DataQualityAssertionsDatasetFacet"
                 }
               },
               "additionalProperties": {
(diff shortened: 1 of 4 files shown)


julienledem (dismissed):
Looking good! 
I made a few comments. In particular we need to add a field in the InputDatasetFacets

spec/OpenLineage.json
```diff
@@ -443,6 +443,41 @@
...
+                "properties": {
+                  "expectationType": {
+                    "type": "string",
+                    "description": "Type of expectation test that dataset is subjected to",
+                    "example":  "not_null"
+                  },
+                  "success": {"type":  "bool"},
+                  "column": {
+                    "type":  "string",
+                    "description": "Column that expectation is testing. If column field is empty, then expectation refers to whole dataset.",
```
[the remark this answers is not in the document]
mobuchowski: Done.

spec/OpenLineage.json
```diff
@@ -443,6 +443,41 @@
         }
       ]
     },
+    "ExpectationsAssertionsDatasetFacet": {
```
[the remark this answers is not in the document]
mobuchowski: Done.

spec/OpenLineage.json
```diff
@@ -443,6 +446,41 @@
...
```
collado-mike, line 449: Maybe minor consideration- expectation and assertion seem redundant to me. The `ExpectationsAssertion` wording seems to reference the Great Expectations framework specifically, whereas I think something like just `DataQualityAssertionsDatasetFacet` is more general. Similarly, below, I'd change `expectationType` to `assertion`- reserve `expectation` for fields that contrast `expected` and `actual`, as we'd likely encounter them in any testing framework.

spec/OpenLineage.json
```diff
@@ -443,6 +446,41 @@
...
+            "assertions": {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "expectationType": {
+                    "type": "string",
+                    "description": "Type of expectation test that dataset is subjected to",
+                    "example":  "not_null"
+                  },
```
collado-mike, line 468: Where possible, I think it would be great if the assertion had the expected value

spec/OpenLineage.json
julienledem: agreed, the name `DataQualityAssertions` is more intuitive. The field above would become `dataQualityAssertions`

julienledem (commented):
This looks good to me once @collado-mike's comments are adressed.
Once it is merged on main we need to publish 0.1.1 to https://github.com/OpenLineage/OpenLineage.github.io/tree/main/spec
Until we implement: https://github.com/OpenLineage/OpenLineage/issues/147

spec/OpenLineage.json
mobuchowski: Do you mean adding additional field that contains argument to given expectation? For example, for expectation `expect_table_row_count_to_equal` the value could be `7`. 

I think if we want to do that, the better approach would be to accept untyped dictionary with arguments - the argument count and type can wildly differ between expectations, for example: https://github.com/calogica/dbt-expectations#expect_column_values_to_be_within_n_moving_stdevs

spec/OpenLineage.json
mobuchowski: Done

julienledem:
Looks like this needs to be rebased or conflicts addressed. 
@collado-mike: is it looking good to you?

collado-mike (approved):

PR #223, opened 2021-08-25
[INTEGRATION] [dbt]: emit expectations assertions facet from dbt test run

mobuchowski:
Capture results of `dbt test` invocation as `ExpectationsAssertionsDatasetFacet`.

From each tested model, generates pair of start/complete events. On complete event, tested model is treated as input `Dataset` without any output datasets. Results of all test runs for one model are attached as one facet.


wslulciuc (approved):
I would suggest we add docs to public methods going forward as well eventually want to publish docs for the client, otherwise 👍

Issue #229, opened 2021-08-27
[INTEGRATION][DBT] Failed runs are reported as `COMPLETE`


wslulciuc:
When executing the `dbt-ol run` cmd with a missing decency (=`dbt-bigquery`), a dbt runtime error was thrown with lineage events emitted, but `COMPLETE` event types where received on the HTTP backend:

```diff
$ dbt-ol run
Running with dbt=0.20.1
Encountered an error while reading profiles:
  ERROR Runtime Error
  Credentials in profile "default", target "dev" invalid: Runtime Error
    Could not find adapter type bigquery!
Defined profiles:
 - default

For more information on configuring profiles, please consult the dbt docs:

https://docs.getdbt.com/docs/configure-your-profile

Encountered an error:
Runtime Error
  Could not run dbt
+Emitted 6 openlineage events
```

Using [Marquez](https://github.com/MarquezProject/marquez) as the HTTP backend, I queried the `lineage_events` table with the follow event types returned:

```sql
marquez=> SELECT run_id, event_type FROM lineage_events;
```
```
                run_id                | event_type
--------------------------------------+------------
 de68fe80-bcd2-444e-92e3-14eb0dfb47ea | START
 d43244de-ac91-4106-9f33-ad5a5215c7dc | START
 693963e2-5cc8-423b-a6de-5b90e353ea06 | START
 de68fe80-bcd2-444e-92e3-14eb0dfb47ea | COMPLETE
 d43244de-ac91-4106-9f33-ad5a5215c7dc | COMPLETE
 693963e2-5cc8-423b-a6de-5b90e353ea06 | COMPLETE
```
I would expect either `FAIL` event types would be emitted or no events at all.

PR #232, opened 2021-08-30
[INTEGRATION][dbt] do not emit events if run_result file was not updated

mobuchowski:
Dbt runs (and tests) that execute sql on selected target save run completion metadata to `run_result.json` file. 
If file was not updated by dbt after recent run - file modification time was in the past before run - do not emit OL events.

This does not impact or emit fail events - it represents situation where job did not start to run. It could be compared to situation where Airflow DAG fails to parse when it's not valid python. 

Fixes https://github.com/OpenLineage/OpenLineage/issues/229


wslulciuc (approved):
LGTM 👍

Issue #211, opened 2021-08-20
Great Expectations integration doesn't work in Google Cloud Composer


collado-mike:
The current implementation of the OL integration doesn’t work in Cloud Composer. It’s likely that it doesn’t work with any installation of Airflow using the CeleryExecutor. The current implementation relies on the ability to wrap the execute method of the GreatExpectationsOperator and to capture the returned result before it is passed to the Airflow DAG. For whatever reason, this “[monkey-patching](https://en.wikipedia.org/wiki/Monkey_patch)” doesn’t work in Cloud Composer. It does work correctly when running Airflow locally on a single host. I haven't been able to verify that it works in the existing airflow integration test (GreatExpectations functionality is not exercised in the test).

### XCom
Fortunately, the GreatExpecationsOperator does publish validation results into Airflow’s cross-communication mechanism (XCom), meaning we can fetch the results from the worker host at runtime. The response from the worker host is a json blob that we need to parse. The existing code already navigates this structure. However, the existing solution requires users to call a static set_dataset_info method on the GreatExpectationsExtractor class, passing in the instance of the GreatExpectationsOperator. The operator is then monkey-patched with the additional information. It is unclear whether this method of naming the dataset will work in Cloud Composer.

Alternatively, we can look at the JSON returned from the operator at runtime to determine the dataset name. Unfortunately, the arguments passed to the operator aren’t guaranteed to give an easily parseable dataset name. Two example JSON blobs are below- one for BigQuery and one for GCS:

BigQuery
```json
"meta": {
 "great_expectations_version": "0.13.26",
 "expectation_suite_name": "orders_suite",
 "run_id": {
   "run_name": "20210818T051320.186669Z",
   "run_time": "2021-08-18T05:13:20.186669+00:00"
 },
 "batch_kwargs": {
   "project": "bq-project",
   "table": "orders",
   "dataset": "bq_dataset",
   "datasource": "food_delivery_db"
 },
 "batch_markers": {
   "ge_load_time": "20210818T051320.055448Z"
 },
 "batch_parameters": null,
 "validation_time": "20210818T051320.192586Z",
 "expectation_suite_meta": {
   "BasicSuiteBuilderProfiler": {
     "batch_kwargs": {
       "bigquery_temp_table": "bq-project.bq_dataset.ge_tmp",
       "data_asset_name": "orders",
       "datasource": "food_delivery_db",
       "limit": 1000,
       "schema": null,
       "table": "orders"
     },
[...]

PR #234, opened 2021-08-31
Implement OpenLineageValidationAction for Great Expectations

collado-mike:
Implementation for issue outlined in https://github.com/OpenLineage/OpenLineage/issues/211

This includes an `OpenLineageValidationAction`. The original GreatExpectations extractor has been stripped of basically everything and the accompanying facets have been moved to the `common` package. As of this point, the ValidationAction has to be explicitly configured by users: e.g., add the following to their Great Expectations configuration:
```yaml
validation_operators:
  action_list_operator
    class_name: ActionListValidationOperator
    action_list
      - name: openlineage
        action
          class_name: OpenLineageValidationAction
          module_name: openlineage.common.provider.great_expectations
          openlineage_host: "https://staging.dev-usw2-datakin.dk-infra.com/",
          openlineage_namespace: food_delivery,
          job_name: validate_pandas
```

Dataset parsing is implemented for SQL and Pandas datasets- this should cover most cases, I think. The unit tests includes a `sqlite` database, but I also tested this on a BigQuery table and it ran fine. Similarly, the pandas test hits the local filesystem, but I was able to run this on a parquet dataset in GCS.

It attempts to handle custom sql, but there's not a test for that yet- I'll add one in

The original facets from the airflow module were moved wholesale. I also included an additional `GreatExpectationsRunFacet`, which reports the runtime configuration of the suite when run, which will give users some additional insights into the configuration of the assertions run on the datasets. 

Another PR will update the GreatExpectationsExtractor to attempt updating an existing GreatExpectations configuration to augment with the additional action if possible. Currently, this implementation won't inherit the parent run id of a DAG task, so that will also be attempted in the upcoming PR.


mobuchowski (approved):
Great work :fire:

Issue #233, opened 2021-08-30
[INTEGRATION][SPARK] spark tests are flaky


mobuchowski:
Due to timeout, `testWithLogicalRdd` often fails to complete: 

```
  java.util.concurrent.TimeoutException: Unable to acquire permit within expected timeout- OpenLineageSparkListener processing may not have completed correctly
      at io.openlineage.spark.agent.lifecycle.SparkReadWriteIntegTest.testWithLogicalRdd(SparkReadWriteIntegTest.java:321)
```

Example of this behavior, requiring multiple reruns on unrelated changes:
https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage?branch=dbt%2Femit-only-success

mobuchowski:
Works well after extending timeout.

PR #244, opened 2021-09-06
[TEST] extend spark semaphore timeout

mobuchowski:
Try to fix flaky test https://github.com/OpenLineage/OpenLineage/issues/233 by extending the timeout.

### DIFF
--- a/integration/spark/src/test/java/io/openlineage/spark/agent/lifecycle/StaticExecutionContextFactory.java
+++ b/integration/spark/src/test/java/io/openlineage/spark/agent/lifecycle/StaticExecutionContextFactory.java
@@ -42,7 +42,7 @@ public class StaticExecutionContextFactory extends ContextFactory {
    * @throws InterruptedException
    */
   public static void waitForExecutionEnd() throws InterruptedException, TimeoutException {
-    boolean acquired = semaphore.tryAcquire(1, TimeUnit.SECONDS);
+    boolean acquired = semaphore.tryAcquire(5, TimeUnit.SECONDS);
     if (!acquired) {
       throw new TimeoutException(
           "Unable to acquire permit within expected timeout- "


integration/spark/src/test/java/io/openlineage/spark/agent/lifecycle/StaticExecutionContextFactory.java
```diff
@@ -44,7 +44,7 @@ public StaticExecutionContextFactory(OpenLineageContext sparkContext) {
    * @throws InterruptedException
    */
   public static void waitForExecutionEnd() throws InterruptedException, TimeoutException {
```
OleksandrDvornik, line 46: Can we instead of such method, create a spark listener which will notify us when the job would be done and we can proceed?

integration/spark/src/test/java/io/openlineage/spark/agent/lifecycle/StaticExecutionContextFactory.java
```diff
@@ -44,7 +44,7 @@ public StaticExecutionContextFactory(OpenLineageContext sparkContext) {
...
-    boolean acquired = semaphore.tryAcquire(1, TimeUnit.SECONDS);
+    boolean acquired = semaphore.tryAcquire(5, TimeUnit.SECONDS);
```
OleksandrDvornik, line 47: So, as I got the problem is with concurrent execution of tests with SparkSession, cause it should be only one, right?

Maybe we can configure gradle to run test classes in sync and fork only test classes instead of each test?

collado-mike: No, the problem is not with Gradle's concurrent execution of the tests. It is that the Spark listener always runs in a separate thread from the main Spark execution- this is the case in production as well as in tests. Since they run in separate threads, the execution of, e.g., `DataFrame.write()` may return before the listener's `onJobEnd` method completes. We have to wait for the listener's code to finish executing before we test that the client calls were made, so we use the semaphore as a way to validate that the listener code is done.

collado-mike (approved):

collado-mike (approved):

Issue #236, opened 2021-08-31
[INTEGRATION][dbt] Add dbt integration tests


julienledem:
We need an integration test that runs the dbt command line and verifies lineage.
Similar to the airflow/spark integration tests

PR #246, opened 2021-09-06
[INTEGRATION][dbt] add dbt integration test

mobuchowski:
Dbt is tested reusing existing infrastructure for airflow. Test does not use `openlineage.airflow`, Airflow is used only as execution engine for dbt here.

https://github.com/OpenLineage/OpenLineage/issues/236


integration/airflow/tests/integration/test_integration.py
```diff
@@ -186,6 +147,7 @@ def test_integration_great_expectations():
 
 if __name__ == '__main__':
     setup_db()
-    test_integration_great_expectations()
-    test_integration_postgres()
-    test_integration_bigquery()
+    test_integration('great_expectations_validation', 'requests/great_expectations.json')
```
wslulciuc, line 150: Nice approach 💯

wslulciuc (approved):
Nice work, @mobuchowski 💯 🥇

PR #255, opened 2021-09-07
Fixes for getting GreatExpectations action working in Airflow

collado-mike:
I was working on trying to automatically add lineage collection to GreatExpectationsOperators if they weren't explicitly configured to include the OpenLineageValidationAction. I wasn't successful in this attempt (the current extractor workflow doesn't hook into the right points to be able to modify the `DataContext` in the way we'd need. 

However, I did include these changes I made to support that effort. Namely

1. Change the `StepMetadata` class to return vanilla OpenLineage `InputDataset`s and `OutputDataset`s rather than the other model we use- this would enable us to extract the Dataset and RunEvent information the validation action returns via XCOM to construct a correct model to report
2. Move the OpenLineageValidationAction's log field initialization to outside of the constructor, as that causes the `Clear` functionality to fail.

### DIFF
--- a/integration/airflow/openlineage/airflow/adapter.py
+++ b/integration/airflow/openlineage/airflow/adapter.py
@@ -90,12 +90,8 @@ class OpenLineageAdapter:
             job=self._build_job(
                 job_name, job_description, code_location, sql
             ),
-            inputs=[
-                dataset.to_openlineage_dataset() for dataset in step.inputs
-            ] if step else None,
-            outputs=[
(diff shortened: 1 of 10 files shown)


mobuchowski (approved):
Looks good :+1: 

At this point, I'd get rid of intermediary format that corresponds to Marquez model. The only thing we're missing in OL model are some helper and factory methods. This is outside of the scope of this PR, of course.

PR #260, opened 2021-09-08
Fix development versions and release script

collado-mike:
Manually bump all python versions to `0.2.2`. Updates the `new-version.sh` release script to release using the current dev version, then bump to the next dev version. This enables the CI build to always be publishing snapshots of the *next* release version, rather than the previously released version as it does now.

When testing the release, I also needed to bump the schema version for the Spark integration tests. That led to noticing that the integration test data files are always rewritten during test execution, which leaves a lot of dirty files in my git workspace. I made a small change to the `LibraryTest` where those files are consumed so that they don't need to be modified during the build and release stage.


integration/spark/integrations/container/pysparkRddToTableStartEvent.json
```diff
@@ -53,25 +53,25 @@
...
           "size": 0,
-          "_schemaURL": "https://openlineage.io/spec/1-0-0/OpenLineage.json#/definitions/OutputStatisticsOutputDatasetFacet"
+          "_schemaURL": "https://openlineage.io/spec/1-0-1/OpenLineage.json#/definitions/OutputStatisticsOutputDatasetFacet"
         }
       }
     }
   ],
   "producer": "https://github.com/OpenLineage/OpenLineage/tree/$VERSION/integration/spark",
-  "schemaURL": "https://openlineage.io/spec/1-0-0/OpenLineage.json#/definitions/RunEvent"
+  "schemaURL": "https://openlineage.io/spec/1-0-1/OpenLineage.json#/definitions/RunEvent"
```
wslulciuc, line 76: _Minor:_ Would be nice if the version bumps in the test data was aligned with the schema change version bump.

new-version.sh
```diff
@@ -154,8 +153,21 @@ git fetch --all --tags
...
+# Append '-SNAPSHOT' to 'NEXT_VERSION' if a release candidate, or missing
+# (ex: '-SNAPSHOT' will be appended to X.Y.Z or X.Y.Z-rc.N)
+if [[ "${NEXT_VERSION}" == *-rc.? ||
+      ! "${NEXT_VERSION}" == *-SNAPSHOT ]]; then
+  NEXT_VERSION="${NEXT_VERSION}-SNAPSHOT"
+fi
+
 sed -i  "" "s/^version=.*/version=${NEXT_VERSION}/g" integration/spark/gradle.properties
 sed -i  "" "s/^version=.*/version=${NEXT_VERSION}/g" client/java/gradle.properties
+echo "version ${NEXT_VERSION}" > integration/spark/src/test/resources/io/openlineage/spark/agent/client/version.properties
```
wslulciuc, line 170: _Minor:_ We'll want to unify the `build.gradle` files for our java modules https://github.com/OpenLineage/OpenLineage/issues/262

wslulciuc (approved):
Thanks for cleaning up the release 💯

integration/spark/integrations/container/pysparkRddToTableStartEvent.json
collado-mike: Would be. I don't know how the schema versions are bumped now- manually? Seems like a future task.

wslulciuc: Yeah, it's currently manual, but we'll want to try and automate the schema version bumps at some point.

PR #263, opened 2021-09-09
Spark visitor refactor

collado-mike:
Mostly moving some code around. 

* I moved non-plan visitor classes outside of the `io.openlineage.spark.agent.lifecycle.plan` package. I reduced visibility in several classes as they are only ever used within the package.
* Created a `QueryPlanVisitor` abstract base class for plan visitors. This makes the class hierarchy easier to read and understand (extends `QueryPlanVisitor<AppendData>` rather than `AbstractPartialFunction<LogicalPlan, List<Dataset>>`) and also eliminates a small amount of boilerplate when visitors simply need to test that the node visited is an instance of some class or interface. I added a small test for the QueryPlanVisitor

This is purely a refactoring, so there should be no impact to the actual code execution.


OleksandrDvornik (approved):
airflow IT continuously failing, can't merge it

mobuchowski:
Guess I was faster with Great Expectations fix here too 😉

collado-mike:
> Guess I was faster with Great Expectations fix here too 😉

Ah, glad that that's fixed. I was just testing that, but I'll rebase with your change instead.

PR #266, opened 2021-09-10
Fix BigQueryNodeVisitor to use tableName() method

collado-mike:
The `LogicalRelationVisitor` was matching `BigQueryRelation`s, causing nodes to be reported as `DirectBigQueryRelation_struct<...>`. The `UnknownEntryFacet` captures information on unvisited nodes, including unknown `LogicalRelation`s, so this code is a) redundant and b) forces the node visiting to happen in a specific order to be correct. Rather than impose an order, which won't scale to extensions that might be authored outside of this library, I removed that code.

I also noticed that the existing Spark integration tests still rely on the bytebuddy/javaagent approach rather than using the Spark ListenerBus. I modified the extension to use the listener approach and updated affected tests accordingly.

### DIFF
--- a/integration/spark/src/main/java/io/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java
+++ b/integration/spark/src/main/java/io/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java
@@ -1,7 +1,6 @@
 package io.openlineage.spark.agent.lifecycle.plan;
 
 import io.openlineage.client.OpenLineage;
-import io.openlineage.spark.agent.client.OpenLineageClient;
 import io.openlineage.spark.agent.util.PlanUtils;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


integration/spark/src/main/java/io/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java
```diff
@@ -60,41 +54,22 @@ public LogicalRelationVisitor(SparkContext context, String jobNamespace) {
...
-        .documentation(ol.newDocumentationDatasetFacet(logRel.simpleString()))
-        .schema(PlanUtils.schemaFacet(logRel.schema()))
-        .build();
-  }
-
-  private String getUnknownDatasetName(LogicalRelation logRel) {
-    return logRel.relation().getClass().getSimpleName()
-        + "_"
-        + logRel.relation().schema().catalogString();
+    return Collections.emptyList();
```
julienledem, line 72: It's a problem if we end up here. if isDefinedAt is true but we don't find anything we should collect this information somehow so that we can fix the logic.

julienledem: possibly a warnings facet? Or does it go in the unknown inputs?

julienledem (commented):
This looks good to me, I made a comment related to the default case.

wslulciuc (approved):
Other than the comment left by @julienledem, looks great! Nice work on cleaning up some old code 💯 🔨

integration/spark/src/main/java/io/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java
collado-mike: I'll change this to throw an `IllegalStateException`. `isDefined` requires the `LogicalPlan` matches one of three cases and those three cases are covered here. We should never get to this line.

PR #267, opened 2021-09-11
spark# Resolves #252 - Add Node classname for serialized UnknownEntry…

### LINKED ISSUE #252: [PROPOSAL] Add Node classname for serialized UnknownEntryFacet nodes
**Purpose:**
We often need to know the Java class name of input and/or output nodes when reporting on nodes that don't have visitors to extract dataset information. 

**Proposed implementation**
The io.openlineage.spark.agent.lifecycle.plan.LogicalPlanSerializer already uses Jackson annotations on mixing classes. We should be able to add an `com.fasterxml.jackson.annotation.JsonTypeId` annotation that will automatically extract the class name of any visited nodes and include it in the serialized output.

OleksandrDvornik:
Resolves #252
Adds classname field to Node for serialized UnknownEntryFacet


wslulciuc (approved):
LGTM 👍

Issue #242, opened 2021-09-02
[PROPOSAL] Airflow integration should have easy way for users to add custom extractors 


mobuchowski:
**Purpose:**
Airflow integration supports subset of relevant Airflow operators. There should be a way for integration users to easily extend it with additional extractors.

**Proposed implementation**

One option would be to scan all packages for implementation of `BaseExtractor` and match it with relevant `operator_class`. [Like in old marquez_airflow.](https://github.com/MarquezProject/marquez-airflow/commit/9f953db54406bd04d60db249ef5256bfc58e587a#diff-b82ee1c456fc78e6baad2dc774f31960ce89311fb39fc771be53299195bd151aL20)

Another way would be through airflow config or env variables.
For example, user could specify env variable manually pointing operator to extractor, similar to how Airflow connections can be defined.

```
OPENLINEAGE_EXTRACTOR_MySqlOperator: my.path.to.extractor.MySqlExtractor
```

PR #276, opened 2021-09-14
[INTEGRATION][airflow] add possibility to register custom extractors

mobuchowski:
There are two mechanisms to add custom extractors that this PR adds. First one is done by mapping in code via `custom_extractors` argument in DAG class. This one won't work with Airflow 2 as we won't be subclassing DAG there. 

Second one is via env variables: the operator is given via key. If prefix matches `OPENLINEAGE_EXTRACTOR_`, then, `Extractors` class tries to important extractor for which import path is given via value - similar to how Airflow deals with `LineageBackend`.

Fixes https://github.com/OpenLineage/OpenLineage/issues/242

### DIFF
--- a/integration/airflow/openlineage/airflow/dag.py
+++ b/integration/airflow/openlineage/airflow/dag.py
@@ -71,6 +71,10 @@ class DAG(airflow.models.DAG):
             macros = kwargs["user_defined_macros"]
         macros["lineage_run_id"] = lineage_run_id
         kwargs["user_defined_macros"] = macros
+        if kwargs.__contains__("lineage_custom_extractors"):
+            for operator, extractor in kwargs['lineage_custom_extractors'].items():
+                extractor_mapper.add_extractor(operator, extractor)
+            del kwargs['lineage_custom_extractors']
         super().__init__(*args, **kwargs)
 
     def add_task(self, task):
--- a/integration/airflow/openlineage/airflow/extractors/base.py
+++ b/integration/airflow/openlineage/airflow/extractors/base.py
@@ -1,10 +1,9 @@
 from abc import ABC, abstractmethod
-from typing import List, Dict, Type, Union, Optional
+from typing import List, Dict, Union, Optional
 
 from openlineage.client.run import Dataset
 from pkg_resources import parse_version
 
-from airflow.models import BaseOperator
 from airflow.version import version as AIRFLOW_VERSION
 
 from openlineage.client.facet import BaseFacet
@@ -55,10 +54,8 @@ class StepMetadata:
 
 
 class BaseExtractor(ABC, LoggingMixin):
-    operator_class: Type[BaseOperator] = None
-    operator: operator_class = None
-
     def __init__(self, operator):
+        super().__init__()
         self.operator = operator
         self.patch()
 
@@ -67,13 +64,19 @@ class BaseExtractor(ABC, LoggingMixin):
         pass
 
     @classmethod
-    def get_operator_class(cls):
-        return cls.operator_class
+    def get_operator_classnames(cls) -> List[str]:
+        """
+        Implement this method returning list of operators that extractor works for.
+        Particularly, in Airflow 2 some operators are deprecated and simply subclass the new
+        implementation, for example BigQueryOperator:
+        https://github.com/apache/airflow/blob/main/airflow/contrib/operators/bigquery_operator.py
+        The BigQueryExtractor needs to work with both of them.
+        :return:
+        """
+        raise NotImplementedError()
 
     def validate(self):
-        # TODO: maybe we should also enforce the module
-        assert (self.operator_class is not None and
-                self.operator.__class__ == self.operator_class)
+        assert (self.operator.__class__.__name__ in self.get_operator_classnames())
 
     @abstractmethod
     def extract(self) -> Union[Optional[StepMetadata], List[StepMetadata]]:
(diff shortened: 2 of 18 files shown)


integration/airflow/openlineage/airflow/extractors/extractors.py
```diff
@@ -27,26 +30,36 @@ class Extractors:
...
+
+        for extractor in _extractors:
+            for operator_class in extractor.get_operator_classnames():
+                self.extractors[operator_class] = extractor
+
+        for patcher in _patchers:
+            for operator_class in patcher.get_operator_classnames():
+                self.patchers[operator_class] = patcher
+
+        for key, value in os.environ.items():
```
wslulciuc, line 44: Mind if add a comment on the environment key pattern? For example: `OPENLINEAGE_EXTRACTOR_<extractor>`

integration/airflow/openlineage/airflow/extractors/base.py
```diff
@@ -67,13 +64,11 @@ def patch(self):
         pass
 
     @classmethod
-    def get_operator_class(cls):
-        return cls.operator_class
+    def get_operator_classnames(cls) -> List[str]:
```
wslulciuc, line 67: _Minor_: Mind if we add docs to this method?

integration/airflow/openlineage/airflow/dag.py
```diff
@@ -71,6 +71,10 @@ def __init__(self, *args, **kwargs):
             macros = kwargs["user_defined_macros"]
         macros["lineage_run_id"] = lineage_run_id
         kwargs["user_defined_macros"] = macros
+        if kwargs.__contains__("custom_extractors"):
```
wslulciuc, line 74: Mind if we keep the environment name and param name the same and use `openlineage_extractors`?

integration/airflow/openlineage/airflow/extractors/base.py
```diff
@@ -55,10 +54,8 @@ def __repr__(self):
 
 
 class BaseExtractor(ABC, LoggingMixin):
```
wslulciuc, line 56: Since we will be providing a way to define custom extractors, would now be a good time to rename the legacy class `StepMetadata` to `TaskMetadata`?

integration/airflow/tests/integration/airflow/logs/latest
```diff
@@ -1 +1 @@
-/opt/airflow/logs/2021-09-03
\ No newline at end of file
+/opt/airflow/logs/2021-09-14
```
wslulciuc, line 1: Should `/logs/latest` file be checked in?

integration/airflow/tests/integration/gcloud/gcloud-service-key.json
```diff
@@ -0,0 +1 @@
+
```
wslulciuc, line 1: Is the `gcloud-service-key.json` file needed for the testing?

wslulciuc (commented):
I know this is WIP, but looks good so far with some minor comments! Excited to see custom extractor support ❤️

integration/airflow/openlineage/airflow/extractors/base.py
mobuchowski: Good idea. Mind if I do it in separate PR prior to next release? I'd do bigger cleanup of intermediate data structures and extractor API handling then.

integration/airflow/openlineage/airflow/extractors/base.py
mobuchowski: Added doc about this:

In Airflow 2 some operators are deprecated and simply subclass the new 
implementation, for example BigQueryOperator: 
https://github.com/apache/airflow/blob/main/airflow/contrib/operators/bigquery_operator.py
The BigQueryExtractor needs to work with both of them.

integration/airflow/openlineage/airflow/dag.py
mobuchowski: I've used `lineage_custom_extractors` as lack of custom suggest complete override of default list.

integration/airflow/openlineage/airflow/extractors/base.py
mobuchowski: https://github.com/OpenLineage/OpenLineage/issues/289

integration/airflow/openlineage/airflow/extractors/extractors.py
mobuchowski: Added docs.

integration/airflow/tests/integration/gcloud/gcloud-service-key.json
mobuchowski: Removed.

integration/airflow/tests/integration/airflow/logs/latest
mobuchowski: Actually, yes - we need file to indicate that logs folder has necessary permissions. The file is actually used there, so nicely fits that purpose.

integration/airflow/openlineage/airflow/extractors/base.py
wslulciuc: Sounds great!

integration/airflow/openlineage/airflow/extractors/extractors.py
wslulciuc: Thanks!

wslulciuc (approved):
LGTM 👍

PR #279, opened 2021-09-15
adding initial test for spark kafka integration

tomassatka:
Added integration test using kafka testcontainer that is setup on same network. Intention is only to spin up kafka in independent container, run spark job to read from topic1 and write to topic2.


julienledem:
Nice! Use `./gradlew spotlessApply` to apply the linter and fix the related issues in the build.

integration/spark/build.gradle
```diff
@@ -95,8 +95,9 @@ dependencies {
     testImplementation "org.testcontainers:junit-jupiter:${testcontainersVersion}"
     testImplementation "org.testcontainers:postgresql:${testcontainersVersion}"
     testImplementation "org.testcontainers:mockserver:${testcontainersVersion}"
+    testImplementation "org.testcontainers:kafka:1.16.0"
```
mobuchowski, line 98: We should use `${testcontainersVersion}`

mobuchowski (approved):
Looks good! One minor comment about dependency.

integration/spark/build.gradle
tomassatka: updated

integration/spark/src/test/java/io/openlineage/spark/agent/SparkContainerIntegrationTest.java
```diff
@@ -206,6 +226,22 @@ public void testPysparkRddToTable() throws IOException, InterruptedException {
...
+
+    pyspark =
+        makePysparkContainerWithDefaultConf(
+            "testPysparkKafkaReadWriteTest",
+            "--packages",
+            "org.apache.spark:spark-sql-kafka-0-10_2.11:" + System.getProperty("spark.version"),
+            "/opt/spark_scripts/spark_kafka.py");
+
+    pyspark.setWaitStrategy(Wait.forLogMessage(".*ShutdownHookManager: Shutdown hook called.*", 1));
+    pyspark.start();
```
julienledem, line 242: Possibly add a comment here:
```
//TODO add lineage validation test
```
We might want to discuss how to add this feature incrementally.

julienledem (commented):
This is a great start.
Can you open an issue to propose how we would capture lineage in spark-streaming?
What's the next step?

integration/spark/src/test/java/io/openlineage/spark/agent/SparkContainerIntegrationTest.java
tomassatka: @julienledem i would suggest following approach:
Before supporting spark streaming capturing what is a whole new feature. I would start with capturing lineage from various data sources as kafka, delta, hive tables, csv, ... on simple batch read/write. 

For example fixing writing to kafka : https://github.com/OpenLineage/OpenLineage/issues/280

We also consider in future switching from spark to flink with beam so also provide support those. 
What are your thoughts or approach on adding features into openlineage?

julienledem: I think that makes sense. We can easily review adding coverage for each source independently to the current batch processing coverage. For streaming, I'm making a proposal on the project slack. Please come chime in.

julienledem: So basically:
1. increasing batch datasource coverage should go as PRs to main 1 by 1. 
2. For the streaming implementation: My proposal is to start a branch and incrementally write PRs against it until we are ready to merge it in main

PR #284, opened 2021-09-16
Raise on http error rather than fail silently

limx0:
Took me a second to figure out why my events weren't hitting marquez. I think it would be better to raise here?

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -58,12 +58,14 @@ class OpenLineageClient:
         data = Serde.to_json(event)
         if log.isEnabledFor(logging.DEBUG):
             log.debug(f"Sending openlineage event {event}")
-        self.session.post(
+        resp = self.session.post(
             urljoin(self.url, 'api/v1/lineage'),
             data,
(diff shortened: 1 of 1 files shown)


mobuchowski:
This is definitely useful, but I think the behavior should be configurable. Some people wouldn't want their tasks to fail due to OL integration timeouting.

collado-mike:
> This is definitely useful, but I think the behavior should be configurable. Some people wouldn't want their tasks to fail due to OL integration timeouting.

My thinking here is that this should be configured at the integration level. The client should most definitely surface the error, but let the individual integrations (Airflow, GE, etc.) determine how to handle the exception. E.g., in the Spark integration, we catch the exceptions and log them, but the Spark tasks continue without failing.

julienledem:
So it sounds like we want to merge this. I think the airflow integration already catches exception at the highest level?

mobuchowski (approved):

collado-mike (approved):

wslulciuc (approved):
Yeah, totally makes sense to raise an exception. Thanks, @limx0!

PR #290, opened 2021-09-16
[INTEGRATION][dbt] add progressbar for sending OL events

mobuchowski:
I've used [tqdm library](https://github.com/tqdm/tqdm) to add progress bar.

Result: https://asciinema.org/a/GSTVkQA1ZEJV1AFPiCTOkIjiJ

### DIFF
--- a/integration/dbt/scripts/dbt-ol
+++ b/integration/dbt/scripts/dbt-ol
@@ -4,6 +4,7 @@ import sys
 import time
 import os
 
+from tqdm import tqdm
 from openlineage.common.provider.dbt import DbtArtifactProcessor
 from openlineage.client.client import OpenLineageClient
 
@@ -15,6 +16,9 @@ PRODUCER = f'https://github.com/OpenLineage/OpenLineage/tree/{__version__}/integ
 
 
 def main():
+    print(f"Running OpenLineage dbt wrapper version {__version__}")
+    print(f"This wrapper will send OpenLineage events at the end of dbt execution.")
+
     args = sys.argv[2:]
     target = parse_single_arg(args, ['-t', '--target'])
     project_dir = parse_single_arg(args, ['--project-dir'], default='./')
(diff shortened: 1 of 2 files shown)


integration/dbt/scripts/dbt-ol
```diff
@@ -53,9 +54,10 @@ def main():
 
     events = processor.parse().events()
 
-    for event in events:
+    for event in tqdm(events, desc="Emitting OpenLineage events:"):
+        time.sleep(0.25)
```
julienledem: why do we need to sleep here?

integration/dbt/scripts/dbt-ol
```diff
@@ -53,9 +54,10 @@ def main():
 
     events = processor.parse().events()
 
-    for event in events:
+    for event in tqdm(events, desc="Emitting OpenLineage events:"):
```
julienledem, line 57: it looks like ":" is not needed here. Looking at your recording there are two in the display

integration/dbt/scripts/dbt-ol
julienledem: was this so that the recording looks nicer? :)

integration/dbt/scripts/dbt-ol
```diff
@@ -53,9 +54,10 @@ def main():
...
+        time.sleep(0.25)
         client.emit(event)
-    print(f"Emitted {len(events)} openlineage events")
+    print(f"Emitted {len(events)} OpenLineage events")
```
julienledem, line 59: can you print a message at the very beginning that mentions OpenLineage?
- the version (just like dbt does)
- that we're going to send events at the end.

julienledem (changes requested):
this progress bar looks great!

integration/dbt/scripts/dbt-ol
mobuchowski: Yes :wink: 
I've removed it minute before you looked.

integration/dbt/scripts/dbt-ol
mobuchowski: Done

julienledem (approved):

Issue #289, opened 2021-09-16
[INTEGRATION][airflow] refactor Extractors contract


mobuchowski:
As we're adding possibility to easily add custom extractors, we should revisit and simplify Extractor contract.

wslulciuc:
@mobuchowski Mind adding more detail on the `Extractor` contract changes? I would specify what the new contract would look like.

mobuchowski:
Currently, for legacy reason, extractors output is list of `StepMetadata` objects. 
First of all, `StepMetadata` is a confusing name. I've renamed it to `TaskMetadata` as you suggested.
Second, it returns a list - currently, we just ignore the elements past one. I've removed that possibility - making it an error instead of silent data loss.
Third, it has a mess of an internal structure for legacy reasons. Some of attributes has been datasets or run/job facets, some of them were fields that were used later to convert to facets.  
So, I've moved to provide concrete list of job/run facets in extractors, that the extractor is responsible of creating, instead of moving this responsibility to core classes.

PR #291, opened 2021-09-16
[INTEGRATION][airflow] refactor extractors contract

mobuchowski:
Refactor extractor contract.
`StepMetadata`, renamed to `TaskMetadata` dropped unused fields. 
Moved to provide concrete job/run facets in extractors, instead of substitutes. 
Removed possibility to return list of `TaskMetadata` from extractors.

https://github.com/OpenLineage/OpenLineage/issues/289


julienledem:
That makes sense. The one downside is that this is not backward compatible and will break any custom extractor that users might have written.

mobuchowski:
>The one downside is that this is not backward compatible and will break any custom extractor that users might have written.

We already discarded any output beyond first element for all extractors that returned list of elements. This PR changes silently dropping data to an error.

wslulciuc (approved):
This looks great and we can finally have a class that properly defines the metadata in contains. There are some merge conflicts to resolve, but otherwise 💯 💯 💯

mobuchowski:
@wslulciuc a lot of PR getting merged today :wink:

PR #293, opened 2021-09-17
[WIP] Prefect integration

limx0:


integration/prefect/openlineage/prefect/adapter.py
```diff
@@ -0,0 +1,229 @@
...
+        # TODO - type might be dropped?
+        data = {
+            "type": dataset_type,
+            "physicalName": url_path,
+            "sourceName": source_name,
+            "fields": kwargs.pop("fields", []),
+            **kwargs,
+        }
+        r = self.client.session.put(
+            urljoin(self.client.url, f"api/v1/namespaces/{namespace}/datasets/{dataset_name}"),
```
julienledem, line 192: you don't need to create the datasets. You should just send OpenLineage RunEvents.
Same for sources.
See:
https://openlineage.io/getting-started#step-1-start-a-run

Additional information (schema) can be added in facets.

julienledem (commented):
This is a great initiative!

integration/prefect/openlineage/prefect/adapter.py
limx0: I'm using this in the my  "ExternalData" task: 

```
class ExternalDataset(Task):
    """
    Task for reading data via `fsspec` and creating an OpenLineage Dataset at the same time.

    Useful if you want a dataset to be available in OpenLineage for lineage purposes, but don't have control over the
    process writing the data, and therefore don't have lineage already set up. An example would be some files on a HTTP
    server or S3 bucket.
```

What would you recommend in this situation @julienledem ? Is there some other equivalent in openlineage at this stage?

julienledem: There is not, so this is Marquez specific. It would be worth starting a [proposal] github issue on OpenLineage to discuss this.

limx0: https://github.com/OpenLineage/OpenLineage/issues/294

limx0:
Just a bit of an update:

- I've removed the ExternalDataset task (not really required) after a discussion in https://github.com/OpenLineage/OpenLineage/issues/294
- I've cleaned up a little and added a couple of basic tests
- I've moved most of the logic off the `FlowRunner` class onto the `OpenLineageAdapter`. 
  - There's an announcement due from Prefect team at the start of Oct apparently with potentially breaking changes so this is probably safer/cleaner.
- I've added a simple [PrefectRunFacet](https://github.com/OpenLineage/OpenLineage/blob/36946ca06fd25500f7b87bbc834776b6b77648c3/integration/prefect/openlineage/prefect/facets.py#L11)

After speaking with @mobuchowski on slack and reading the airflow code a little more, I think there is significant overlap between the two integrations. I think the next steps should probably be trying to get the airflow extractors hooked into this integration.

From my perspective, I think there are just two remaining tasks for a "first pass" on this integration:
- [ ] Making the extractors work (which gives us the extensibility of custom facets as well as any already defined facets in airflow, such as BigQuery )
- [ ] Ensure the naming scheme complies with https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md

integration/prefect/openlineage/prefect/adapter.py
julienledem: Sorry for the back and forth on that issue. Yes the dataset will be created on read so that works.

felix-ht:
@limx0 any news on the prefect 2.0 to integration?

limx0:
Hey @felix-ht! No unfortunately I've been distracted with other things and lineage has moved down my list of priorities for now. 

I believe prefect were/are considering developing integrations internally, but I'm not sure how much of a priority they are right now. 

I'm still interested in the space, just don't have a huge amount of time to contribute at this stage.

felix-ht:
@ricardogaspar2 @limx0 i took a look at the code and with the current version of prefect 2 it seems to be imossible to get an openlinage integration without monkey patching as well relying on some implementation details. Thats why i stopped looking into it as well - way to hacky for my taste.

So sadly i don't see this going anywhere without some support from the prefect team.

Issue #287, opened 2021-09-16
[INTEGRATION][dbt] dbt should recognize existing parent run, and it should create it's own parent run


mobuchowski:
Dbt is not an orchestrator, yet, we model it's runs as we model airflow dag runs.
Creating run that serves as parent run for dbt model runs allows us to properly scope all those jobs together.
It also can serve as way of "notifying" that particular dbt run finished.

mobuchowski:
@julienledem do you think whether we should send `COMPLETE` or `FAIL` events if some of the dbt model runs fail? After all, this situation can result in creation of datasets.

PR #298, opened 2021-09-20
[INTEGRATION][dbt] send ParentRunFacet

mobuchowski:
If we have `OPENLINEAGE_PARENT_ID` env variable set up, use data from it as a `ParentRunFacet`.
If not, on run start create event that would serve as parent run event for events generated by `dbt-ol run`.

`lineage_run_id` airflow macro was changed, since it's original implementation was insufficient to use data in `ParentRunFacet`. It had only `runId`, when we also have to have job name and namespace.

Solves https://github.com/OpenLineage/OpenLineage/issues/287


integration/dbt/scripts/dbt-ol
```diff
@@ -42,14 +93,44 @@ def main():
...
+        profile_name=profile_name,
     )
 
+    # Always emit "wrapping event" around dbt run. This indicates start of dbt execution, since
+    # both the start and complete events for dbt models won't get emitted until end of execution.
+    parent_event = parent_run_event_start(
+        job_name=f"dbt-run-{processor.project['name']}",
+        external_run_metadata=external_run_metadata
+    )
+    parent_run_metadata = ParentRunMetadata(
```
collado-mike, line 123: Rather than `external_parent_xxx` above and `parent_xxx` here, can we use `parent_...` and `dbt_run_xxx` respectively? Whenever I see `parent_` here, I assume it's the process that kicked off the dbt run, not the dbt run itself.

collado-mike (approved):

integration/dbt/scripts/dbt-ol
mobuchowski: Fixed.

PR #302, opened 2021-09-22
[INTEGRATION] airflow tests: remove duplicate image building

mobuchowski:
We use bitnami images for airflow. We need to customize them - install particular packages and apt packages. 
Due to fact that we need to do this for three different images. it slows build down. 
This PR replaces bitnami images with official Airflow image, which needs to be customized only once.

Few additional changes are due to fact that official image's `AIRFLOW_HOME` directory has different user. By default, this does not allow to write to those directories, which trips up some integrations like `dbt` and `great_expectations`.  

### DIFF
--- a/integration/airflow/tests/integration/docker-compose.yml
+++ b/integration/airflow/tests/integration/docker-compose.yml
@@ -1,4 +1,37 @@
 version: "3.7"
+
+x-airflow-base: &airflow-base
+  build:
+    context: .
+    target: airflow
+  environment:
+    DB_BACKEND: postgresql+psycopg2
+    DB_HOST: postgres
+    DB_PORT: 5432
+    AIRFLOW_FERNET_KEY: Z2uDm0ZL60fXNkEXG8LW99Ki2zf8wkmIltaTz1iQPDU=
+    AIRFLOW_DATABASE_HOST: postgres
+    AIRFLOW_DATABASE_NAME: airflow
+    AIRFLOW_DATABASE_USERNAME: airflow
+    AIRFLOW_DATABASE_PASSWORD: airflow
+    AIRFLOW_EXECUTOR: CeleryExecutor
+    AIRFLOW_LOAD_EXAMPLES: "no"
+    AIRFLOW__LINEAGE__BACKEND: openlineage.lineage_backend.OpenLineageBackend
+    AIRFLOW_CONN_FOOD_DELIVERY_DB: postgres://food_delivery:food_delivery@postgres:5432/food_delivery
+    AIRFLOW_CONN_BQ_CONN: google-cloud-platform://?extra__google_cloud_platform__project=openlineage-ci&extra__google_cloud_platform__key_path=%2Fopt%2Fconfig%2Fgcloud%2Fgcloud-service-key.json
+    AIRFLOW__CELERY__BROKER_URL: redis://redis:6379/0
+    AIRFLOW__CORE__LOGGING_CONFIG_CLASS: log_config.LOGGING_CONFIG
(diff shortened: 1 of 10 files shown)


integration/common/openlineage/common/provider/great_expectations/action.py
```diff
@@ -36,8 +37,9 @@ class OpenLineageValidationAction(ValidationAction):
 
     Openlineage host parameters can be passed in as constructor arguments or environment variables
     will be searched. Job information can optionally be passed in as constructor arguments or the
-    great expectations suite name and batch identifier will be used as the job name
-    (the namespace should be passed in as either a constructor arg or as an environment variable).
+    great expectations suite name and either run_name or batch identifier will be used as the job
+    name - the namespace should be passed in as either a constructor arg or as an environment
+    variable.
```
wslulciuc, line 42: I would provide examples for GE the `suite name`, `run_name` and `batch identifier` (and how they are used to build the job name). I'd also document the environment variable for setting the namespace.

integration/airflow/tests/integration/docker/up.sh
```diff
@@ -56,4 +60,5 @@ pytest==6.2.2
 EOL
 
 docker-compose down
-docker-compose up -V --build --force-recreate --exit-code-from integration
+docker-compose up -V --build --abort-on-container-exit airflow_init postgres
+docker-compose up --build --exit-code-from integration --scale airflow_init=0
```
wslulciuc, line 64: I would chain these methods with `&&`, but also wondering why we need both build commands? Would be great to add some comments here ;)

wslulciuc (dismissed):
I would remove the `.gitkeep` file and `/placeholder` dir

integration/airflow/tests/integration/docker-compose.yml
```diff
@@ -1,4 +1,36 @@
 version: "3.7"
+
+x-airflow-common: &airflow-common
+  build:
+    context: .
+    target: airflow
+  environment: &airflow-common-env
```
wslulciuc, line 7: Do you need to set `&airflow-common-env`?

integration/airflow/tests/integration/docker-compose.yml
```diff
@@ -1,4 +1,36 @@
 version: "3.7"
+
+x-airflow-common: &airflow-common
```
wslulciuc, line 3: Should this be `airflow-common` instead of `x-airflow-common`?

wslulciuc: Minor: I'd use `airflow-base` as everything else _extends_ / adds on top of the config ;)

wslulciuc:
@mobuchowski: Looks solid! I left some minor comments, but was wondering what the current state of this PR might be?

integration/airflow/tests/integration/docker-compose.yml
mobuchowski: `x-` is because

```
ERROR: The Compose file './docker-compose.yml' is invalid because:
Invalid top-level property "airflow-base". Valid top-level sections for this Compose file are: version, services, networks, volumes, secrets, configs, and extensions starting with "x-".
```

So, moved to `x-airflow-base`.

integration/airflow/tests/integration/docker-compose.yml
mobuchowski: Removed.

integration/common/openlineage/common/provider/great_expectations/action.py
mobuchowski: Removed that part of PR.

mobuchowski:
Rebased this branch on main - it's ready for review.

collado-mike (approved):

PR #307, opened 2021-09-29
Feature/split spark source tree

### LINKED ISSUE #251: [PROPOSAL] Split source trees to support Spark 2 and Spark 3 specific code
**Purpose:**
Some of the logical plan visitors in Spark are not binary compatible with both Spark 2 and Spark 3. This is due to some method signatures changing (e.g., `org.apache.spark.sql.execution.datasources.LogicalRelation#simpleString()`), while some classes have been deprecated and removed (e.g., `DataSourceV2`). 

**Proposed implementation**
While most code is common to both Spark versions, we need to have support for Spark 2 and Spark 3 specific sources. I propose multiple source trees- one for common code, one for Spark 2 and one for Spark 3. Gradle can support multiple [dependency configurations](https://docs.gradle.org/current/userguide/declaring_dependencies.html), meaning we can configure multiple compilation phases with different dependencies in each. As an example, the `LogicalRelationVisitor` can be supported with the following source files:

```
src/
  main/
    common/
      java/io/openlineage/spark/agent/lifecycle/plan/AbstractLogicalRelationVisitor.java
    spark2/
      java/io/openlineage/spark2/agent/lifecycle/plan/LogicalRelationVisitor.java
    spark3/
      java/io/openlineage/spark3/agent/lifecycle/plan/LogicalRelationVisitor.java
```
The code that is common to both will be in the `AbstractLogicalRelationVisitor`, while the code that is version dependent will be in some abstract method implementation (in this case, [`getDatasetFacet`](https://github.com/OpenLineage/OpenLineage/blob/b0f5000254b1719d25bb9f74176934145beeda21/integration/spark/src/main/java/io/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java#L88-L93) might be the abstract method, as it has a dependency on the `simpleString` method). Most existing plan visitors are common to both Spark versions, so don't need an abstract base class with concrete implementations. Some visitors will only be applicable to one version or another (e.g., a `DataSourceV2` specific visitor has no corresponding concrete implementation in Spark 3).

The package names differ between the source trees. This is because the compiled code will all be located in the same jar. The Gradle build can be configured to execute two compilation phases- one for `common` and `spark2` source trees and one for the `spark3` source tree (the generated `.class` files from the first compilation will need to be on the class path when compiling the `spark3` source tree). The resulting compilation will output the following compiled class file tree

```
build/
  classes/
    main/
      java/
        io/
          openlineage/
            spark/agent/lifecycle/plan/AbstractLogicalRelationVisitor.class
            spark2/agent/lifecycle/plan/LogicalRelationVisitor.class
            spark3/agent/lifecycle/plan/LogicalRelationVisitor.class
```
The jar phase will bundle all compiled classes.

At runtime, to know which plan visitors to use, we can use the `org.apache.spark.sql.SparkSession#version` property to determine whether to load the Spark 2 visitors or the Spark 3 ones. Rather than referencing each class name as a string from the common code (which would be error prone), a single factory interface can be used to return the list of input and output visitors for the appropriate Spark version. E.g.,

```java
private static final String SPARK2_FACTORY_NAME = "io.openlineage.spark2.agent.lifecycle.plan.VisitorFactoryImpl";
private static final String SPARK3_FACTORY_NAME = "io.openlineage.spark3.agent.lifecycle.plan.VisitorFactoryImpl";

private VisitorFactory getFactoryForSparkSession(SparkSession session) {
    try {
        if (session.version().startsWith("2.")) {
            return Class.forName(SPARK2_FACTORY_NAME).newInstance();
        } else {
            return Class.forName(SPARK3_FACTORY_NAME).newInstance();
        }
    } catch (Exception e) {
        // handle exception
    }
}
```

The `VisitorFactory` interface might have two methods- `getInputVisitors(SparkSession)` and `getOutputVisitors(SparkSession)`. The individual concrete implementations will need to be updated for each visitor that is added to one of the version-specific source trees- similar to how the `InputDatasetVisitors` and `OutputDatasetVisitors` currently list each visitor that should be included. A future task might add some class path scanning to improve the maintainability of this.

OleksandrDvornik:
Resolves #251
Splitting a source tree on spark2 and spark3 specific.
For now we have only one class which should be split it's a DatasetSourceVisitor. 
Implementation of Factory classes is based on issue description.
We decided to don't go with splitting a test source tree for now
Main changes are files renaming due to moving them under `common` folder


mobuchowski (approved):

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/VisitorFactory.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Provides Visitors for iterating on {@link LogicalPlan}.
+ *
+ * <p>All common Visitors should be grouped and passed to {@link
+ * VisitorFactory#getOutputVisitors(List)} to retrieve Visitors for {@link
+ * OpenLineage.OutputDataset}
+ */
+public interface VisitorFactory {
+
+  List<PartialFunction<LogicalPlan, List<OpenLineage.Dataset>>> getCommonVisitors();
```
collado-mike, line 17: Can we start using the `QueryPlanVisitor` type rather than referencing the `PartialFunction` type everywhere? I feel that that is clearer to readers.

collado-mike, line 17: Can the interface just include `getInputVisitors` and `getOutputVisitors` and the common visitors can just be an internal implementation detail?

integration/spark/src/test/java/io/openlineage/spark/agent/lifecycle/StaticExecutionContextFactory.java
```diff
@@ -77,16 +80,32 @@ public SparkSQLExecutionContext createSparkSQLExecutionContext(long executionId)
...
                   commonDatasetVisitors(sqlContext, sparkContext);
 
-              InputDatasetVisitors inputDatasetVisitors =
-                  new InputDatasetVisitors(commonDatasetVisitors);
-              OutputDatasetVisitors outputDatasetVisitors =
-                  new OutputDatasetVisitors(sqlContext, commonDatasetVisitors);
+              VisitorFactory common =
+                  new CommonVisitorFactory(sqlContext, sparkContext.getJobNamespace());
+              VisitorFactory versionSpecific =
+                  VersionSpecificVisitorsProvider.getInstance(SparkSession.active());
```
collado-mike, line 86: Rather than having this split here, can we just have the version-specific providers extend the common provider and return a correctly mixed list of input/output visitors?

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/VersionSpecificVisitorsProvider.java
```diff
@@ -0,0 +1,25 @@
+package io.openlineage.spark.agent.lifecycle;
+
+import io.openlineage.spark.agent.lifecycle.plan.VisitorFactory;
+import org.apache.spark.sql.SparkSession;
+
+public class VersionSpecificVisitorsProvider {
```
collado-mike, line 6: If the version-specific factory extends the common one, we can make this simply `VisitorFactoryProvider`

collado-mike (commented):
This is awesome! I think we can clean up the interface a bit with some inheritance and some encapsulation of the common/not-common visitor stuff.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/VisitorFactory.java
OleksandrDvornik: not realy, only if we redesign Visitors approach, cause we should get all common visitors and pass them to output visitors, so in our case, version specific visitors (`DatasetSourceVisitor`) should be passed to common output visitors (like to `InsertIntoDataSourceVisitor`)

OleksandrDvornik: `QueryPlanVisitor<T extends LogicalPlan> extends AbstractPartialFunction<LogicalPlan, List<OpenLineage.Dataset>>`
In our case we need` List<OpenLineage.InputDataset> ` and `List<OpenLineage.OutputDataset>`

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/VersionSpecificVisitorsProvider.java
OleksandrDvornik: sure

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/VisitorFactory.java
collado-mike: Sorry, not sure which of your statements came first. 

The implementation of `getInputVisitors` could be 
```java
List<PartialFunction<LogicalPlan, List<OpenLineage.Dataset>>> visitors = new ArrayList<>(getCommonVisitors());
visitors.addAll(inputVisitors);
return visitors;
```
Then the `getCommonVisitors` is an internal implementation detail, but not exposed in the interface.

collado-mike: We can consider changing the `QueryPlanVisitor` signature to
```java 
QueryPlanVisitor<T extends LogicalPlan, D extends OpenLineage.Dataset> extends AbstractPartialFunction<LogicalPlan, List<D>>
```

OleksandrDvornik: I've changed it

PR #310, opened 2021-10-01
ci: automatically publish spec to website

mobuchowski:
When detected that last commit contains changes to spec, copy the changes to relevant place in website repo and push them.
Changes to spec are defined as any `*.json` file in `spec/` directory and subdirectories, so main spec and facets.

This requires setting up SSH key with write access to website repo.

Solves https://github.com/OpenLineage/OpenLineage/issues/147


spec/release.sh
```diff
@@ -0,0 +1,37 @@
...
+set -e
+
+# check if there are any changes in spec in the latest commit
+if git diff --name-only --exit-code HEAD~1 HEAD 'spec/*.json' >> /dev/null; then
+  echo "no changes in spec detected, skipping publishing spec"
+  exit 0
+fi
+
+# "mangle" name to not overwrite for someone running locally
+WEBSITE_DIR=${WEBSITE_DIR:-$HOME/website-123412}
```
julienledem, line 11: maybe do something in $HOME/build or wherever the default build output is.

spec/release.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [[ -d $WEBSITE_DIR ]]; then
+  echo "deleting $WEBSITE_DIR"
+  rm -rf $WEBSITE_DIR
```
julienledem, line 15: This looks dangerous.
Do we need to delete it? Can the git command do the right thing if it exists already? If it exists and is not a git repo, we should probably fail.

.circleci/config.yml
```diff
@@ -310,6 +310,18 @@ jobs:
...
 
+  publish-spec:
+    working_directory: ~/openlineage
+    docker:
+      - image: cimg/base:2021.07
+    steps:
+      - *checkout_project_root
+      - add_ssh_keys:
+          fingerprints:
+            - "SO:ME:FIN:G:ER:PR:IN:T"
```
julienledem, line 321: is that were we need to add the key for a ci user?

spec/release.sh
mobuchowski: Changed to `$HOME/build/website`

.circleci/config.yml
mobuchowski: Added user https://github.com/openlineage-github-bot

PR #319, opened 2021-10-06
Spark3

OleksandrDvornik:
Adding support for spark3


integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/BaseVisitorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    list.add(new LogicalRelationVisitor(sqlContext.sparkContext(), jobNamespace));
+    list.add(new LogicalRDDVisitor());
+    list.add(new CommandPlanVisitor(new ArrayList<>(list)));
+    if (BigQueryNodeVisitor.hasBigQueryClasses()) {
+      list.add(new BigQueryNodeVisitor(sqlContext));
+    }
+    return list;
+  }
+
+  public abstract List<QueryPlanVisitor<? extends LogicalPlan, OpenLineage.Dataset>>
```
collado-mike, line 41: This shouldn't be public- it should only be called internally, so make it `protected`

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/BaseVisitorFactory.java
```diff
@@ -0,0 +1,74 @@
...
+import io.openlineage.spark.agent.lifecycle.plan.wrapper.InputDatasetVisitor;
+import io.openlineage.spark.agent.lifecycle.plan.wrapper.OutputDatasetVisitor;
+import io.openlineage.spark.agent.lifecycle.plan.wrapper.OutputDatasetWithMetadataVisitor;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.spark.sql.SQLContext;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+
+public abstract class BaseVisitorFactory implements VisitorFactory {
```
collado-mike, line 27: We should reduce visibility on all of these internal classes and interfaces. There's no need to expose these internals outside of the library. Let's move `VisitorFactory` into the same package as this class and make both package scoped. The Spark-version specific subclasses must also be moved into this same package and named to avoid conflicts (e.g., `Spark3VisitorFactoryImpl`).

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/VisitorFactoryProvider.java
```diff
@@ -0,0 +1,25 @@
+package io.openlineage.spark.agent.lifecycle;
+
+import io.openlineage.spark.agent.lifecycle.plan.VisitorFactory;
+import org.apache.spark.sql.SparkSession;
+
+public class VisitorFactoryProvider {
```
collado-mike, line 6: package-scoped visibility

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/BaseVisitorFactory.java
OleksandrDvornik: as I remember we agree that spark2 and spark3 will have different package name, like
```
io.openlineage.spark2.agent.lifecycle.plan
io.openlineage.spark3.agent.lifecycle.plan
```
Are you saying that we should have package `io.openlineage.spark.agent.lifecycle.plan` for both spark2 and spark3
If we have different packages I can't use package-private scope

collado-mike: We can use the `io.openlineage.spark.agent.lifecycle.plan` package for the `VisitorFactory` and its implementations. The Spark2 and Spark3 versions can use different class names, rather than relying on the package name. Otherwise, we need to make all these classes public and I'd prefer to keep them hidden so people don't build on them for whatever reason.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -141,6 +144,8 @@ public void start(SparkListenerJobStart jobStart) {
 
   @Override
   public void end(SparkListenerJobEnd jobEnd) {
+    // executing plan for injecting write metrics to optimizedPlan
+    queryExecution.executedPlan().execute();
```
collado-mike, line 148: We can't make this call- this will re-run someone's entire query.

Issue #313, opened 2021-10-02
Airflow integration start and complete events don't reflect actual task states


collado-mike:
When the Airflow DAG creates a new DagRun, it immediately posts a new `START` event for each task in the DAG. Once the entire DAG has completed, the `handle_callback` method posts a `COMPLETE` or `FAILED` event for each task, depending on the actual end state of each. An example from a simple DAG looks like this (most fields omitted)

```json
{"job":"etl_orders.if_not_exists","state":"START","inputs":[],"outputs":[]}
{"job":"etl_orders.insert","state":"START","inputs":[],"outputs":[]}
{"job":"etl_orders.if_not_exists","state":"START","inputs":[],"outputs":[]}
{"job":"etl_orders.insert","state":"START","inputs":[],"outputs":[]}
{"job":"etl_orders.insert","state":"COMPLETE","inputs":[{"name":"marquez.food_delivery.tmp_orders","namespace":"bigquery"}],"outputs":[{"name":"marquez.food_delivery.orders","namespace":"bigquery"}]}
{"job":"etl_orders.if_not_exists","state":"COMPLETE","inputs":[],"outputs":[{"name":"marquez.food_delivery.orders","namespace":"bigquery"}]}
{"job":"etl_orders.insert","state":"COMPLETE","inputs":[{"name":"marquez.food_delivery.tmp_orders","namespace":"bigquery"}],"outputs":[{"name":"marquez.food_delivery.orders","namespace":"bigquery"}]}
{"job":"etl_orders.if_not_exists","state":"COMPLETE","inputs":[],"outputs":[{"name":"marquez.food_delivery.orders","namespace":"bigquery"}]}
```

The result of this is a correct and complete lineage graph, but the runs show incorrect dataset versions consumed. For example, if the `etl_orders.insert` job creates version X of the `marquez.food_delivery.orders` dataset and I add a GreatExpectationsOperator to the DAG, I expect the `DataQuality` facet to be appended to version X of the dataset. Instead, since the `OpenLineageValidationAction` posts a `COMPLETE` event immediately, without waiting for the rest of the DAG to complete, the Marquez server sees that the validation job completed before the `etl_orders.insert` job, therefore it assumes that the version of the dataset validated is the version prior (version W).

[...]

PR #327, opened 2021-10-11
airflow: modify LineageBackend approach to be used in 1.10 when available

mobuchowski:
In Airflow 1, if `OpenLineageBackend` is detected in Airflow configuration, then `openlineage.DAG` stops emitting `COMPLETE` events and leave that to `OpenLineageBackend`. Similarly, when `1.10` is detected, `OpenLineageBackend` stops emitting `START` events.

Due to Airflow 1 import conflicts, `OpenLineageBackend` can't be located in `openlineage.airflow` namespace. If you have idea about better place for this than `openlineage.lineage_backend.OpenLineageBackend` please let me know.

Closes https://github.com/OpenLineage/OpenLineage/issues/313 

### DIFF
--- a/integration/airflow/openlineage/lineage_backend/__init__.py
+++ b/integration/airflow/openlineage/lineage_backend/__init__.py
@@ -1,18 +1,16 @@
 import logging
 import uuid
 import time
-from typing import Optional
+from pkg_resources import parse_version
 
 from airflow.lineage.backend import LineageBackend
-
-from openlineage.airflow.adapter import OpenLineageAdapter
-from openlineage.airflow.extractors.extractors import Extractors
-from openlineage.airflow.extractors.base import TaskMetadata, BaseExtractor
-from openlineage.airflow.utils import DagUtils, get_location, get_custom_facets
+from airflow.version import version as AIRFLOW_VERSION
 
 
-class OpenLineageBackend(LineageBackend):
+class Backend:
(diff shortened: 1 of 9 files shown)


integration/airflow/openlineage/lineage_backend/__init__.py
```diff
@@ -141,3 +139,15 @@ def _get_location(task):
     @staticmethod
     def _now_ms():
         return int(round(time.time() * 1000))
+
+
+class OpenLineageBackend(LineageBackend):
+    # Airflow 1.10 uses send_lineage as staticmethod, so just construct class
+    # instance on first use and delegate calls to it
```
collado-mike, line 146: How does Airflow 2.0 work with this? Will it also call this as a static method? Or do users need to start configuring lineage with `openlineage.lineage_backend.Backend`?

mobuchowski: There's no difference in Airflow 2.
In Python, you can call static methods as class methods or as instance methods:

```
Python 3.9.2 (default, May  5 2021, 14:01:15) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class A:
...     @staticmethod
...     def a():
...             return 2
... 
>>> A.a()
2
>>> A().a()
2
```

PR #328, opened 2021-10-11
- Added support for Redshift profile

AlessandroLollo:
- Added support for Redshift profile
- Added support for handling env vars in profiles.yml

Please note that I had to enclose `{{ config(...) }}` between `raw` Jinja tags, otherwise it complains about an unexpected '.'

Signed-off-by: Alessandro Lollo <alessandro.lollo@cloudacademy.com>

### DIFF
--- a/integration/airflow/tests/integration/Dockerfile
+++ b/integration/airflow/tests/integration/Dockerfile
@@ -1,26 +1,28 @@
 FROM openlineage-airflow-base:latest AS build
 
-FROM bitnami/airflow-scheduler:1.10.12 AS scheduler
+FROM bitnami/airflow-scheduler:1.10.15 AS scheduler
 COPY --from=build /app/wheel /whl
 USER root
 RUN apt-get update && \
     apt-get install -y git build-essential
 USER 1001
+RUN cd /opt/bitnami/airflow/ && . ./venv/bin/activate && python -m pip install --upgrade pip
 
-
-FROM bitnami/airflow-worker:1.10.12 AS worker
+FROM bitnami/airflow-worker:1.10.15 AS worker
 COPY --from=build /app/wheel /whl
 USER root
 RUN apt-get update && \
     apt-get install -y git build-essential
 USER 1001
+RUN cd /opt/bitnami/airflow/ && . ./venv/bin/activate && python -m pip install --upgrade pip
 
-FROM bitnami/airflow:1.10.12 AS airflow
+FROM bitnami/airflow:1.10.15 AS airflow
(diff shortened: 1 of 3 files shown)


integration/common/setup.py
```diff
@@ -23,6 +23,7 @@
 
 requirements = [
     "attrs>=19.3.0",
+    "Jinja2",
```
mobuchowski, line 26: We'll be using this only for dbt and will get this as a transitive dependency there - no need to add it to requirements.

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -417,14 +438,16 @@ def node_to_output_dataset(self, node: ModelNode, has_facets: bool = False) -> O
                 node.catalog_node,
                 [
                     ['stats', 'num_bytes', 'value'],  # bigquery
-                    ['stats', 'bytes', 'value']  # snowflake
+                    ['stats', 'bytes', 'value'],  # snowflake
+                    ['stats', 'size', 'value']  # redshift (Note: size = count of 1MB blocks)
```
mobuchowski, line 442: We need to multiply this then - it needs to be consistent with rest of databases

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -172,10 +174,29 @@ def load_catalog(cls, path: str) -> Optional[Dict]:
...
+    def env_var(var: str, default: Optional[str] = None) -> str:
+        """The env_var() function. Return the environment variable named 'var'.
+        If there is no such environment variable set, return the default.
+
+        If the default is None, raise an exception for an undefined variable.
+        """
+        if var in os.environ:
+            return os.environ[var]
+        elif default is not None:
+            return default
```
mobuchowski, line 187: Shouldn't we throw an exception here? If we don't have that variable, it will fail on dbt execution

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -515,6 +538,8 @@ def extract_namespace(self, profile: Dict) -> str:
             return f"snowflake://{profile['account']}"
         elif profile['type'] == 'bigquery':
             return "bigquery"
+        elif profile['type'] == 'redshift':
+            return f"redshift://{profile['host']}:{profile['port']}"
         else:
             raise NotImplementedError(
                 f"Only 'snowflake' and 'bigquery' adapters are supported right now. "
```
mobuchowski, line 545: And `redshift` :)

PR #334, opened 2021-10-13
Feature/sanitize jdbc urls

### LINKED ISSUE #314: Spark JDBCRelation visitor doesn't redact passwords from JDBC URL
The Spark LogicalRelationVisitor sometimes sends complete JDBC URLs, including passwords and some invalid characters, such as semi-colons which are sometimes used as parameter separators. The URL should omit everything after the database name and should ensure that credentials are not present when they are sent to the server.

OleksandrDvornik:
Resolves #314 
Clean up credentials from jdbc urls.
Checks the user password pair: _user/password_ and _user:password_ 
and in key-value style: `user=user` _password=pwd_


integration/spark/src/main/java/io/openlineage/spark/agent/lifecycle/plan/JdbcUrlSanitizerImpl.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Generic implementation of {@link JdbcUrlSanitizer} Removes username and password from jdbc url
+ */
+public class JdbcUrlSanitizerImpl implements JdbcUrlSanitizer {
```
collado-mike, line 16: This is a lot of code for a couple of regex matches. I can pass all the tests with the following
```
    jdbcUrl = jdbcUrl.substring(5);
    return jdbcUrl.replaceAll(SLASH_DELIMITER_USER_PASSWORD_REGEX, "@")
            .replaceAll(COLON_DELIMITER_USER_PASSWORD_REGEX, "")
            .replaceAll("(?<=[?,;&:)=])\\(?(?i)(?:user|username|password)=[^;&,)]+(?:[;&;)]|$)", "");
```

Maybe we can just make this a single utility method and get rid of the interface and public class?

integration/spark/src/test/java/io/openlineage/spark/agent/util/PlanUtilsTest.java
```diff
@@ -0,0 +1,99 @@
...
+  @ParameterizedTest
+  @MethodSource("provideJdbcUrls")
+  void testJdbcUrlSanitizer(String jdbcUrl, String expectedResult) {
+
+    String actualResult = PlanUtils.sanitizeJdbcUrl(jdbcUrl);
+
+    assertEquals(expectedResult, actualResult);
+  }
+
+  private static Stream<Arguments> provideJdbcUrls() {
```
collado-mike, line 21: This is a SUPER thorough list of test cases!

PR #335, opened 2021-10-13
dbt: render jinja values only in resulting dictionary values

mobuchowski:
Instead of rendering directly loaded config file, we should render only dictionary string values after loading yaml file.
This way, we avoid rendering stuff like commented templates.

This fixes integration tests on main. 

### DIFF
--- a/integration/common/openlineage/common/provider/dbt.py
+++ b/integration/common/openlineage/common/provider/dbt.py
@@ -1,13 +1,13 @@
 import datetime
 
-from jinja2 import Environment
+from jinja2 import Environment, Undefined
 import json
 
 import yaml
 import os
 import uuid
 import collections
-from typing import List, Tuple, Dict, Optional
+from typing import List, Tuple, Dict, Optional, TypeVar
 
 import attr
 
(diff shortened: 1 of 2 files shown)


integration/common/openlineage/common/provider/dbt.py
```diff
@@ -193,14 +212,43 @@ def env_var(var: str, default: Optional[str] = None) -> str:
...
-            # When using env vars for Redshift port, it must be "{{ env_var('PORT') | as_number }}"
-            # otherwise Redshift driver will complain, hence the need to add the "as_number" filter
-            env.filters.update({"as_number": lambda x: x})
-            templated_yaml = env.from_string(f.read())
-        rendered_yaml = templated_yaml.render(env_var=DbtArtifactProcessor.env_var)
-        return yaml.load(rendered_yaml, Loader=yaml.FullLoader)
+            return yaml.load(f, Loader=yaml.FullLoader)
+
+    @staticmethod
+    def setup_jinja():
```
fm100, line 218: _minor_: return type

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -193,14 +212,43 @@ def env_var(var: str, default: Optional[str] = None) -> str:
...
+            return parsed
+        elif isinstance(value, list):
+            parsed = []
+            for elem in value:
+                parsed.append(cls.render_values_jinja(environment, elem))
+            return parsed
+        elif isinstance(value, str):
+            return environment.from_string(value).render()
+        else:
+            return value
```
fm100, line 251: This looks working, but not sure if it will pass static analysis such as `mypy`, which we don't have yet btw. I like the flexibility of python's duck typing but sometimes it's confusing particularly when it is combined with type hinting. Here, the return type of `load_yaml_with_jinja` is `Dict` and it calls `render_values_jinja` but the return type of `render_values_jinja` is `Any`. It is taken care of by `if isinstance` but it will require extra caution when someone else start working on this code base.

One possible solution for this is using `multipledispatch` module which allows method overloading. However, it requires extra dependency and it could possibly change overall coding style, we may want to have further discussion about it.

fm100 (approved):

integration/common/openlineage/common/provider/dbt.py
mobuchowski: Here, I guess one thing we need is generic [typevar](https://docs.python.org/3/library/typing.html#typing.TypeVar), and change type signature from `(Environment, Any) -> Any` to `(Environment, T) -> T`. This way we know we'll get result of the same type we've passed in. 

Python 3.10 will have structural pattern matching, which will allow us to express such statements more clearly - although it might be years since we get to that point. We still support Python 3.6.

Issue #336, opened 2021-10-14
[PROPOSAL] Add a design doc process for proposals


julienledem:
**Purpose:**
Add a design doc section for proposals. The process would be the following:
 - open an issue with the "PROPOSAL" template
 - if the scope warrants it create a PR to add a proposal doc.
 - have the PR approved
 - Once the proposal is accepted you can open another issue for the implementation.

**Proposed implementation**
- Proposals live under OpenLineage/OpenLineage/proposals
- one folder per design (includes images and the proposal document)
- the proposal is keyed by the github issue id: "/proposals/{github issue id}/{short description}.md"
- each doc has a header with:
   - Status: APPROVED | IN PROGRESS | RELEASED
   - Author:
   - Created: 
   - issue tracker: #xxx
- assignees for reviews will be in the issue itself
- community can vote on the issue
```
---
Status: Proposal
Author: 
Created: 
Issue: 
---
```

collado-mike:
I like the process. Some suggestions I would make are

1. If the design doc PR is in a separate branch, can we assume that `APPROVED` status is implied when the doc is merged into `main`?
2. Github [milestones](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) could help track the status of the proposal automatically. Perhaps the proposal, once accepted, can link to a milestone where all relevant tasks are collected and we can use a badge (e.g., ![this badge](https://badgen.net/github/milestones/chrislgarry/Apollo-11/1) )

julienledem:
> We should skip CI if pull request modifies only files under `/proposal` directory.

That makes sense: Do you know how to do that?

julienledem:
> I like the process. Some suggestions I would make are
> 
> 1. If the design doc PR is in a separate branch, can we assume that `APPROVED` status is implied when the doc is merged into `main`?
> 2. Github [milestones](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones) could help track the status of the proposal automatically. Perhaps the proposal, once accepted, can link to a milestone where all relevant tasks are collected and we can use a badge (e.g., ![this badge](https://camo.githubusercontent.com/3c31a5bb010ed9673d122680e26ed4a4aa39855ed5d920ad947efc2436d58f3e/68747470733a2f2f62616467656e2e6e65742f6769746875622f6d696c6573746f6e65732f63687269736c67617272792f41706f6c6c6f2d31312f31) )

[...]

Issue #269, opened 2021-09-14
[PROPOSAL] Add a Transport abstraction to the OpenLineage clients


julienledem:
**Purpose:**
The goal is to facilitate sending OpenLineage events over various protocols. (HTTP, queues, ...) independently of the logic.
This can be configured directly in the client or in the `ProxyBackend`

**Proposed implementation**
Have a backend abstraction to send a `Lineage event`
```java
interface Backend {
  void send(LineageEvent event);
}
```
The Backend class is configured given a backend name (ex: http).
the default implementation sends the events over http.

collado-mike:
I do prefer the abstraction `Transport`- the receiver may be a "backend" service, or it could be something like a PubSub or SQS queue.

mobuchowski:
Same backend or transport should have same configuration between different language clients - same naming of env variables etc.

The backend, if not http, should be explicitely specified by some configuration variable.

mobuchowski:
There are few issues with proposed interface regarding potential errors. 

1. Currently, integration default behavior with client is "fire and forget" - we don't want to cancel or interfere with underlying job even when client fails to send the event. Hovewer, this is not the case for client libraries. The underlying integration should be notified of failure to send event and decide how to react. This behavior is useful for example for logging. The client has no knowledge how integration is logging failures - however, Spark or Airflow know how underlying systems use logging.
2. Hovewer, different languages handle errors differently. Existing Java or Python implementations have no problems with current signature, since they usually communicate errors with exceptions. If we're going to do that, we should standarize this exception per language - to not make integration implementers guess whether they should catch some underlying libraries exception. Overthinking this, we might eventually have clients in languages that do not communicate errors via exceptions - but via some kind of `Result<T, E>` interface like Rust or error value like Go.
3. The implementations of `Transport` could use asynchronous semantics. In that case, we can't really notify integration with error value.

collado-mike:
> * Overthinking this, we might eventually have clients in languages that do not communicate errors via exceptions - but via some kind of `Result<T, E>` interface like Rust or error value like Go.

Personally, I like the `Either` response pattern, but I don't think it's really idiomatic in Java or Python. I really like the pattern in Go, but most Java devs would look at it funny.

> * The implementations of `Transport` could use asynchronous semantics. In that case, we can't really notify integration with error value.

[...]

PR #300, opened 2021-09-21
server model

julienledem:
Signed-off-by: Julien Le Dem <julien@apache.org>

PR #344, opened 2021-10-18
[PROPOSAL] Transport proposal document

mobuchowski:
This document summarizes existing discussion on `Transport` abstraction from issue https://github.com/OpenLineage/OpenLineage/issues/269, following method proposed here: https://github.com/OpenLineage/OpenLineage/issues/336


proposals/269/transport.md
```diff
@@ -0,0 +1,74 @@
...
+The goal is to facilitate sending `OpenLineage` events over various protocols. 
+(HTTP, queues, ...) independently of the logic via exposing `Transport` interface.
+
+## Proposed interface
+
+In pseudocode:
+```
+interface Transport {
+  send(LineageEvent event);
+}
```
collado-mike, line 19: Thinking of the ProxyBackend, it would be nice if this interface didn't require the service to deserialize the event just so it could be serialized again. I'm tempted to say this should just accept a String argument and let the callers be responsible for handling the JSON serialization. However, this would prohibit Transports from using some more efficient serialization, (e.g., Kafka supports Avro streams). I suppose if scale becomes an issue in the future, we can add a new method to allow more efficient passthrough.

proposals/269/transport.md
```diff
@@ -0,0 +1,74 @@
...
+Integration code instantiation is usually controlled by system for which they are 
+written - for example Airflow spawns `LineageBackend` instance. Then, integration
+code would spawn `Client`, which would be responsible for spawning relevant 
+`Transport`.  
+This is contrary to usual dependency injection approach, and creates two issues:   
+that `Client` needs to be aware of which `Transport` to spawn;
+and that `Transport` instantiation should be lightweight and not depend on variables
+passed to it's constructor.  
+Former issue can be solved by standarizing `OPENLINEAGE_TRANSPORT` environment 
+variable notifying `Client` of which `Transport` to spawn.
```
collado-mike, line 54: I think we can go ahead and define a `TransportFactory` interface or something that is responsible for determining which client to instantiate. Can we flesh out some details here? Are users to supply the fully qualified class name of the transport (e.g., `OPENLINEAGE_TRANSPORT=io.openlineage.transport.KafkaTransport`)? Will the class names need to be the same across languages or is it ok to differ? or should we use simpler aliases (e.g., `OPENLINEAGE_TRANSPORT=kafka`) and let the factory figure out the correct class to instantiate?

proposals/269/transport.md
```diff
@@ -0,0 +1,74 @@
...
+should assume that `HTTP` transport is used and it utilizes `OPENLINEAGE_URL` 
+or `OPENLINEAGE_API_KEY` environment variables.
+
+## Relation to ProxyBackend
+
+`ProxyBackend` can utilize Transports to route `LineageEvents` to one or more systems.
+It's written in Java, so it can utilize only `Transports` that are written in 
+that language.  
+`ProxyBackend` could not depend on proposed environment variables, but have it's own 
+configuration system.
```
collado-mike, line 73: I think the `ProxyBackend` ought to end up using the same `TransportFactory` mechanism used at the library level. I agree, though, that it doesn't have to feed that factory parameters from env variables.

mobuchowski: >I agree, though, that it doesn't have to feed that factory parameters from env variables.

Yes, that's what I ment, seeing as usually Dropwizard uses YAML config.

proposals/269/transport.md
mobuchowski: Updated it again.

proposals/269/transport.md
```diff
@@ -0,0 +1,92 @@
+---
+Status: IN PROGRESS  
+Author: Julien Le Dem (original issue) / Maciej Obuchowski (this document)  
+Created: 2021-10-18  
+Issue: [269](https://github.com/OpenLineage/OpenLineage/issues/269)
+---
+
+##Purpose
```
pawel-big-lebowski, line 8: Isn't the whitespace missing here?

proposals/269/transport.md
```diff
@@ -0,0 +1,92 @@
...
+
+The goal is to facilitate sending `OpenLineage` events over various protocols. 
+(HTTP, queues, ...) independently of the logic via exposing `Transport` interface.
+
+## Proposed interface
+
+In pseudocode:
+```
+interface Transport {
```
pawel-big-lebowski, line 17: It's really nice when, in case of misconfiguration, a system notifies the issue immediately after the start not waiting for `send` method to be called. Perhaps some sort of `init` or `verify` method would help with this.

Imagine you provide Kafka producer configuration with broker urls, encoder classes, SSL stuff to authenticate, etc. You really would like to see if the configuration is correct before job executes and real event is emitted.

mobuchowski: I agree, this should be done on init, but I think `TransportFactory` described below is a better place for that.

proposals/269/transport.md
wslulciuc: I agree there would a (de)serialization cost, but it be beneficial to quantify. Having a `RunEvent` makes it clear the metadata emitted along with the event for a given run. The other advantage is that the underlying method _controls_ the serialization, while also avoiding callers passing a string that may or may not be invalid.

As mentioned, methods can be added by modifying the interface of the `Transport` to support other serializations.

proposals/269/transport.md
wslulciuc: Though not ideal, it's common to error on first request. Take for example a simple http client. Usually, errors will be thrown on the first use of the client. This can be for a number of reasons: the host or port aren't valid, typos, etc. That said, I don't see anything wrong with adding a step to verify that the transport properties are valid and the backend is reachable, it just comes at a cost of maintaining such code. I'd feel more natural to know if something is wrong with the transport on first use.

wslulciuc:
@mobuchowski: I left more thorough comments on the `Transport` in your initial issue: https://github.com/OpenLineage/OpenLineage/issues/269#issuecomment-1005273574

proposals/269/transport.md
wslulciuc: Yes!

mobuchowski: Fixed.

proposals/269/transport.md
wslulciuc: That's a good point on using the transport in the `ProxyBackend`. The only issue here would be which models are used? The `ProxyBackend` should use the OpenLineage [server models](https://github.com/OpenLineage/OpenLineage/pull/300)(or _string_, which I'm not a complete fan of), similar to Marquez. Since the transport accepts a `RunEvent`, we'd either have to support a _string_ type, or a server model object in the `Transport` interface.

proposals/269/transport.md
mobuchowski: I think our options are

1) don't have special handling of first request
2) optionally, make "special" request on init unique to each transport to find if backend is reachable and configuration (like apiKey) is okay
3) have some special handling on first request

I don't think 3) makes a lot of sense. 2) depends very much on Transport, so I think it's best left out for particular implementations. I don't think we can do this in general case, for example for HTTP.

proposals/269/transport.md
```diff
@@ -0,0 +1,92 @@
...
+`OPENLINEAGE_TRANSPORT=http` sets transport to `HTTP`.
+
+Default `TransportFactory` can load non-buildin transports based on fully-qualified class names. 
+The fully-qualified name can differ between languages. For example, Java implementation
+of `Apache Kafka Transport` can be `io.openlineage.transport.KafkaTransport`.
+
+The default implementation also passes all environment
+variables fulfilling `OPENLINEAGE_TRANSPORT_*` as `Config`.
+Different `TransportFactories` could utilize other configuration methods, for example
+YAML files.
```
collado-mike, line 78: Personally, I'd vote for the yml option being the default, as in @wslulciuc's comment [here](https://github.com/OpenLineage/OpenLineage/issues/269#issuecomment-1005273574)

wslulciuc: I agree, but I think we can provide (initially) 3 options for configuring the transport:

1. An `openlineage.yml` at the root of the classpath
2. An `openlineage.yml` file in a `.openlineage/` folder in the user's home directory (~/.openlineage/openlineage.yml
3. Environment variables

wslulciuc (approved):
LGTM 🚀

Issue #350, opened 2021-10-20
CI job `release-integration-spark` fails on release


wslulciuc:
The CI job `release-integration-spark` fails on release as it depends on the `openlineage-java` artifact to be published to maven (or locally), before it can be triggered, see full logs [](https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/1115/workflows/84b93c18-8a34-4b18-b1d5-d419927b24f4/jobs/7280/parallel-runs/0/steps/0-102):

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':shadowJar'.
> Could not resolve all dependencies for configuration ':runtimeClasspath'.
   > Could not find io.***********:***********-java:0.2.3.
     Searched in the following locations:
       - file:/home/circleci/.m2/repository/io/***********/***********-java/0.2.3/***********-java-0.2.3.pom
       - https://repo.maven.apache.org/maven2/io/***********/***********-java/0.2.3/***********-java-0.2.3.pom
```

PR #351, opened 2021-10-20
Cache `.m2/` for java client release

wslulciuc:
### Problem

The CI job `release-integration-spark` fails on release as it depends on the `openlineage-java` artifact to be published to maven before it can be triggered.

Closes: https://github.com/OpenLineage/OpenLineage/issues/350

### Solution

Add CI job `release-client-java` as a dependency for `release-integration-spark`,  ensuring  the `openlineage-java` artifact is published to local maven. When running the `gradle publish` task, it also runs the `publishPubNamePublicationToRepoNameRepository` tasks. We can then save / load the cached `~/.m2` with the artifact.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,10 +109,6 @@ jobs:
       - image: cimg/openjdk:11.0
     steps:
       - *checkout_project_root
-      - restore_cache:
-          keys:
-            - v1-release-client-java-{{ .Branch }}-{{ .Revision }}
-            - v1-release-client-java-{{ .Branch }}
       - run: |
           # Get, then decode the GPG private key used to sign *.jar
           export ORG_GRADLE_PROJECT_signingKey=$(echo $GPG_SIGNING_KEY | base64 -d)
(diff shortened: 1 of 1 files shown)


collado-mike:
> I guess it just depends if we wanted to use a local release vs what's published?

It's embedding the code, so as long as the byte code is identical, it doesn't really matter. The published jar is simply an upload of the locally installed one, so I don't see a lot of point in uploading, then downloading the same jar just to unpack it and copy the class files.

RELEASING.md
```diff
@@ -10,12 +10,14 @@
    > **Tip:** Use `--help` to see script usage
 
 3. Visit [CI](https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage?branch=main) to see the progress of the release! :rocket:
-4. Visit [sonatype](https://oss.sonatype.org) to promote _java_ artifacts
-5. Create a [new project](https://github.com/OpenLineage/OpenLineage/projects/new) board for the _next_ release using the _automated kanban_ template:
+4. Visit [sonatype](https://oss.sonatype.org) to promote `openlineage-java` artifact
+5. In CI, approve the _hold_ for the job to release the spark integration
```
collado-mike, line 14: This is incorrect now, right?

Issue #373, opened 2021-10-28
[INTEGRATION] Spark - separate spark 2 and spark 3 tests


pawel-big-lebowski:
According to the https://github.com/OpenLineage/OpenLineage/issues/249#issuecomment-953858708 having spark 2 and spark 3 tests within a single project introduces problems for the Spark classes that are present only within one Spark version. Although tests can be run per Spark version, single codebases leads to importing classes that are not present. 

As an example, Spark 3 class org.apache.spark.sql.connector.catalog.Table cannot be used within tests, as it is missing for Spark 2 and the codebase need to be compiled for Spark 2 either.

Issue #249, opened 2021-09-07
[INTEGRATION] [Spark] Add unit test for iceberg


mobuchowski:

mobuchowski:
Tested this with Spark 3 using SQL.
Does not work OOB.
With changes, work with both local table and on GCS.

```sql
CREATE TABLE `remote-gcs`.database.table (id bigint, data string) USING iceberg;
INSERT INTO `remote-gcs`.database.table VALUES (4, 'a'), (5, 'b'), (6, 'c');
```

Dirty impl here: https://github.com/OpenLineage/OpenLineage/compare/main...poc/iceberg

To make it work, `DatasetSourceVisitor` needs to migrate to use data from `org.apache.spark.sql.connector.catalog.Table`.
Iceberg provides `location` in properties that's sufficient to get proper dataset namespace and name. By the way, quick look at Delta suggests that the same would work there:  

https://github.com/delta-io/delta/blob/4d405ff32fe7622add1c1f43abae850e065322ec/core/src/main/scala/org/apache/spark/sql/delta/catalog/DeltaTableV2.scala

PR #381, opened 2021-11-02
separeted tests between Spark 2&3 versions plus single Spark3 test added

pawel-big-lebowski:
### Problem

According to the #249 (comment) having spark 2 and spark 3 tests within a single project introduces problems for the Spark classes that are present only within one Spark version. Although tests can be run per Spark version, single codebases leads to importing classes that are not present. 

The pull request resolves the issue by separating Spark 2 and Spark 3 tests.

Closes: https://github.com/OpenLineage/OpenLineage/issues/373

### Solution

Modify `integration/spark` so that Spark tests can be separated to common, spark2 and spark3 submodules exactly in a a way it is done for main code. All existing tests have been moved to common. A new Spark 3 test has been created and put into Spark3 module. 

### Checklist


integration/spark/build.gradle
```diff
@@ -154,6 +156,7 @@ dependencies {
 
 sourceSets {
     main.java.srcDirs = ["src/main/common/java", "src/main/spark2/java"]
+    test.java.srcDirs = ["src/test/common/java"]
```
mobuchowski, line 159: Shouldn't we have `spark2` here, just as in `main.java.srcDirs`? It looks like it works though.

integration/spark/build.gradle
```diff
@@ -163,9 +166,19 @@ sourceSets {
             destinationDirectory.set(file("$buildDir/classes/java/main/"))
         }
     }
+
+    spark3Test {
+        java {
+            srcDirs = ["src/test/spark3/java"]
```
mobuchowski, line 172: Similar: shouldn't we have common here?

mobuchowski (approved):
Looks good - just doublecheck the comments I've made

integration/spark/build.gradle
pawel-big-lebowski: I was thinking about it but at the moment we do not have any Spark2 test that fails for Spark3 and all the existing tests are put to `common`. So we could add this here but spark2 submodule will not be pushed to repo as it would be just an empty directory. Also, I was not sure if we are going to have any new Spark2 tests in future.

pawel-big-lebowski: All the current tests are within a common submodule as we do not have any Spark2 specific tests (tests that do not  compile for Spark 3)

collado-mike (approved):

Issue #370, opened 2021-10-28
[INTEGRATION][Spark] Cover Command in org.apache.spark.sql.execution.command


mobuchowski:
Cover Commands in org.apache.spark.sql.execution.command

- [x]                     CreateDataSourceTableAsSelectCommand
- [x]                     CreateDataSourceTableCommand
- [x]                     CreateTableCommand
- [x]                     CreateTableLikeCommand
- [x]                     DropTableCommand
- [x]                     AlterTableChangeColumnCommand
- [x]                     AlterTableAddPartitionCommand
- [x]                     AlterTableRenamePartitionCommand
- [x]                     AlterTableDropPartitionCommand
- [x]                     AlterTableRecoverPartitionsCommand
- [x]                     InsertIntoDataSourceDirCommand
- [x]                     ~~AlterTableRenameCommand~~ (solved within #369)
- [x]                     ~~AlterTableAddColumnsCommand~~ (solved within #369)
- [x]                     LoadDataCommand
- [ ]                     TruncateTableCommand

mobuchowski:
` InsertIntoDataSourceDirCommand` already covered.

PR #393, opened 2021-11-05
[INTEGRATION][Spark] Increase visitor coverage

mobuchowski:
Add visitors for 

- CreateHiveTableAsSelectCommandVisitor
- CreateDataSourceTableAsSelectCommandVisitor
- CreateDataSourceTableCommandVisitor
- CreateTableLikeCommandVisitor
- CreateTableAsSelectVisitor
- LoadDataCommandVisitor

Closes: some amount of tasks from mostly https://github.com/OpenLineage/OpenLineage/issues/370

### Checklist


integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/V2CatalogParser.java
```diff
@@ -0,0 +1,25 @@
...
+import io.openlineage.spark.agent.util.PlanUtils;
+import lombok.SneakyThrows;
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.hadoop.fs.Path;
+import org.apache.spark.sql.connector.catalog.CatalogPlugin;
+import org.apache.spark.sql.execution.datasources.jdbc.JDBCOptions;
+import org.apache.spark.sql.execution.datasources.v2.jdbc.JDBCTableCatalog;
+
+public class V2CatalogParser {
+  public static Path getCatalogLocation(CatalogPlugin catalog) {
```
collado-mike, line 12: I think the `PlanUtils.getDataset(CatalogTable)` route that the other visitors use will offer some consistency over how we treat the datasources from a catalog. E.g., I don't know that "catalog location" makes sense for a lot of catalogs- e.g., in memory ones. And in Hive, the table location itself is used to determine the dataset, not the catalog's location.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/V2CatalogParser.java
```diff
@@ -0,0 +1,25 @@
...
+    //     TODO: add more catalogs (iceberg?)
+    if (catalog instanceof JDBCTableCatalog) {
+      return parseJdbc((JDBCTableCatalog) catalog);
+    }
```
collado-mike, line 16: I think this is a great example of what's missing in the `PlanUtils` handling of the `CatalogTable`. Rather than add this Catalog parser, can we pass the `CatalogTable` (which we can extract with `command.catalog().loadTable(command.tableName())` ) to the `PlanUtils` method and add some extra handling there for JDBC URLs?

mobuchowski: Sometimes, we cannot do `loadTable(command.tableName())` since the table does not exist yet - or we can, but only on complete event.

In this case, also, we have `Table` interface introduced in Spark 3, so we need to have something like `PlanUtils3` for this particular case.

integration/spark/src/main/common/java/io/openlineage/spark/agent/util/PlanUtils.java
```diff
@@ -139,6 +141,39 @@ public String getName() {
...
+   * org.apache.spark.sql.catalyst.catalog.CatalogTable} If table does not contain location, use
+   * passed authority as a backup.
+   *
+   * @param table
+   * @return
+   */
+  public static OpenLineage.Dataset getDataset(CatalogTable table, String authority) {
+    Path path = getPath(table.location(), table.qualifiedName(), authority);
+    return getDataset(path.toUri(), table.schema());
+  }
```
collado-mike, line 171: I noticed this does funky stuff with Hive tables. The table location is typically the path to the data in storage- e.g., `s3://path/to/the/data`. This path is unique to the table and is, oftentimes, enough to identify the data. With this Dataset construction, we'll end up with `namespace=s3://path/to/the/data` and `name=the_data`, which I don't think we want. I think we probably want either the table location _or_ the table name, but not both. Though attaching the table name as a facet on the dataset is definitely a useful approach.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/CreateHiveTableAsSelectCommandVisitor.java
```diff
@@ -0,0 +1,38 @@
...
+ * the output {@link OpenLineage.Dataset} being written.
+ */
+public class CreateHiveTableAsSelectCommandVisitor
+    extends QueryPlanVisitor<CreateHiveTableAsSelectCommand, OpenLineage.Dataset> {
+
+  @Override
+  public List<OpenLineage.Dataset> apply(LogicalPlan x) {
+    CreateHiveTableAsSelectCommand command = (CreateHiveTableAsSelectCommand) x;
+    CatalogTable table = command.tableDesc();
+    Path path = PlanUtils.getPath(table.location(), table.qualifiedName(), "");
```
pawel-big-lebowski, line 27: `table.location()` is going to throw an exception if `locationUri` is not specified.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/CreateDataSourceTableCommandVisitor.java
```diff
@@ -0,0 +1,24 @@
...
+import org.apache.spark.sql.catalyst.catalog.CatalogTable;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import org.apache.spark.sql.execution.command.CreateDataSourceTableCommand;
+
+/**
+ * {@link LogicalPlan} visitor that matches an {@link CreateDataSourceTableCommand} and extracts the
+ * output {@link OpenLineage.Dataset} being written.
+ */
+public class CreateDataSourceTableCommandVisitor
+    extends QueryPlanVisitor<CreateDataSourceTableCommand, OpenLineage.Dataset> {
```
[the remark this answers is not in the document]
pawel-big-lebowski: We're not able as `OptimizedCreateHiveTableAsSelectCommand` is available since Spark `2.4.8`

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/V2CatalogParser.java
mobuchowski: Addressed in https://github.com/OpenLineage/OpenLineage/pull/416 and https://github.com/OpenLineage/OpenLineage/pull/419

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -61,29 +61,64 @@
...
+    log.debug("SparkListenerSQLExecutionStart - executionId: {}", startEvent.executionId());
+    startEvent(startEvent.time());
+  }
 
-  public void end(SparkListenerSQLExecutionEnd endEvent) {}
+  public void end(SparkListenerSQLExecutionEnd endEvent) {
+    log.debug("SparkListenerSQLExecutionEnd - executionId: {}", endEvent.executionId());
+    // TODO: can we get failed event here?
+    // If not, then we probably need to use this only for LogicalPlans that emit no Job events.
+    // Maybe use QueryExecutionListener?
```
collado-mike, line 91: In Spark 3.x, the `SparkListenerSQLExecutionEnd` event has an `executionFailure` property - https://github.com/apache/spark/blob/v3.1.2/sql/core/src/main/scala/org/apache/spark/sql/execution/ui/SQLListener.scala#L68-L69 . The `QueryExecutionListener` is now just a different interface that responds to the same `SparkListenerSQLExecutionEnd` events - https://github.com/apache/spark/blob/v3.1.2/sql/core/src/main/scala/org/apache/spark/sql/util/QueryExecutionListener.scala#L139-L158 .

collado-mike (approved):

Issue #376, opened 2021-11-01
[INTEGRATION][dbt] support dbt built


julienledem:
dbt build is a superset of the dbt run and dbt test commands.
We should apply the same logic there.
https://docs.getdbt.com/reference/commands/build

PR #398, opened 2021-11-09
[INTEGRATION][dbt] Support dbt build command

mobuchowski:

Dbt 0.21.0 introduced new command, `[dbt build](https://docs.getdbt.com/reference/commands/build)`. From our perspective, this command combines `dbt run` and `dbt test`, so should events generated by `dbt-ol build`. 

Main difference in behavior between running those commands separately and with `dbt-ol build` is that `dbt-ol test` incantation generated events with `DataQualityAssertionsDatasetFacet` attached to input datasets. With `dbt-ol build` the command does attach those facets to output datasets - so the end event looks like `dbt-ol run` event, but with assertions facet attached.

Some refactoring was required to add the command efficiently: `dbt-ol build` and `dbt-ol run` share the same parse command. The parsing data quality assertions was separated, and all the commands share event generation method. 

Closes: https://github.com/OpenLineage/OpenLineage/issues/376

### Checklist

- [x] You've updated the [`CHANGELOG.md`](https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md) with details about your change under the "Unreleased" section (_if relevant, depending on the change, this may not be necessary_)

### DIFF
--- a/integration/common/openlineage/common/provider/dbt.py
+++ b/integration/common/openlineage/common/provider/dbt.py
@@ -48,21 +48,11 @@ class DbtRun:
     started_at: str = attr.ib()
     completed_at: str = attr.ib()
     status: str = attr.ib()
-    inputs: List[ModelNode] = attr.ib()
-    output: Optional[ModelNode] = attr.ib()
+    inputs: List[Dataset] = attr.ib()
+    output: Optional[Dataset] = attr.ib()
     job_name: str = attr.ib()
     namespace: str = attr.ib()
     run_id: str = attr.ib(factory=lambda: str(uuid.uuid4()))
-
-
-@attr.s
-class DbtEvents:
-    starts: List[RunEvent] = attr.ib()
-    completes: List[RunEvent] = attr.ib()
-    fails: List[RunEvent] = attr.ib()
-
-    def events(self):
-        return self.starts + self.completes + self.fails
 
 
 @attr.s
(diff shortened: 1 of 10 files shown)


integration/common/openlineage/common/provider/dbt.py
```diff
@@ -373,139 +380,135 @@ def parse_test(
...
-                inputs=[Dataset(namespace, name, facets={
+            events.add(self.to_openlineage_events(
+                "success",
+                started_at,
+                completed_at,
+                self.get_run(run_id),
+                Job(self.job_namespace, job_name),
+                [Dataset(namespace, name, facets={
                     "dataQualityAssertions": assertion_facet
                 })],
```
collado-mike, line 402: We need to nail down what it means for a facet to be in `Dataset.facets` vs in `InputDataset.inputFacets` or `OutputDataset.outputFacets`. Currently the GreatExpectations integration attaches the `GreatExpectationsAssertionsDatasetFacet` and `DataQualityMetricsInputDatasetFacet` as input facets (see [here](https://github.com/OpenLineage/OpenLineage/blob/main/integration/common/openlineage/common/provider/great_expectations/action.py#L222-L227)), whereas the dbt integration attaches them as `Dataset.facets`. I see in the test case, the output dataset can have the assertions facet attached. I understand that's due to this refactoring, but it may be confusing for server-side consumers who don't know where to look for a given facet. Is the assertions facet really part of the output? or just describing the Dataset that was output? or is it describing the Dataset after the output was complete?

mobuchowski: >Currently the GreatExpectations integration attaches the GreatExpectationsAssertionsDatasetFacet and DataQualityMetricsInputDatasetFacet as input facets (see here), whereas the dbt integration attaches them as Dataset.facets.

It did attach them as input facets before, and still does when `dbt test` is a separate command.
Here, the creation of the dataset and testing it are parts of the same run - the dataset is in outputs of the job. 

If we really want to have it as an input facet, then the solution would be to split this run into two runs - the first one having this dataset as an output without the assertions, and having second run, with input dataset, with just the assertions. It would be like running `dbt run` and `dbt test` separately.

mobuchowski: We've discussed this offline with @collado-mike and decided to go with two separate events, each corresponding to `dbt run` and `dbt test` invocation.

julienledem:
This looks good to me. 
@collado-mike any other comment or shall we merge?

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -175,13 +201,15 @@ def parse(self) -> DbtEvents:
...
-        if run_result['args']['which'] == 'run':
-            return self.parse_run(manifest, run_result, catalog, nodes)
-        elif run_result['args']['which'] == 'test':
-            return self.parse_test(manifest, run_result, catalog, nodes)
+        context = DbtRunContext(manifest, run_result, catalog)
+
+        if self.command in ['run', 'build']:
+            return self.parse_execution(context, nodes)
+        elif self.command == 'test':
+            return self.parse_test(context, nodes)
```
collado-mike, line 209: Minor issue, but I would keep the `parse_execution` method ignorant of the `build` semantics and move that logic here so the workflow of each command is handled in one central place. E.g., something like

```python
events = DbtEvents()
if self.command in ['run', 'build']:
    events.add(self.parse_execution(context, nodes))
if self.command in ['test', 'build']:
    events.add(self.parse_test(context, nodes))
return events
```

collado-mike (approved):

integration/common/openlineage/common/provider/dbt.py

Issue #152, opened 2021-08-09
Proposal for a ProxyBackend


mandy-chessell:
This issue is to track the discussion and development of a proxy backend.  The diagram below shows the basic idea.

![image](https://user-images.githubusercontent.com/12354195/128734055-2a08372e-8d1b-426e-9c25-c06aeb551e36.png)

julienledem:
This initial design looks good to me. I'm looking forward to see it.

wslulciuc:
Fixed #400

PR #400, opened 2021-11-10
Proxy Backend implementation

mandy-chessell:
### Problem

This is the code that support the proposal for a proxy backend. (#152)

### Solution

This is a simple server that supports the OpenLineage REST API endpoint {baseURL}/api/v1/lineage.  It uses a yml file to define where incoming events can go.  Currently writing to the console and Apache Kafka is supported.


proxy/src/main/java/io/openlineage/proxy/api/models/LineageStream.java
```diff
@@ -0,0 +1,55 @@
...
+{
+  /**
+   * The Type enum (and JsonSubTypes above) are extended for each new type of destination that the proxy backend supports.
+   * There is a subtype class for each of these destination types.
+   */
+  enum Type
+  {
+    CONSOLE,
+    KAFKA,
+    KINESIS
```
julienledem, line 31: I would remove the KINESIS until it gets implemented.

julienledem (commented):
This looks good to me.
Pending @wslulciuc's review

proxy/src/main/java/io/openlineage/proxy/api/ProxyResource.java
```diff
@@ -0,0 +1,55 @@
...
+public class ProxyResource {
+  private final ProxyService service;
+
+  public ProxyResource(@NonNull final ProxyService service) {
+    this.service = service;
+  }
+
+  @POST
+  @Consumes(APPLICATION_JSON)
+  public void emit(@Valid String event, @Suspended final AsyncResponse asyncResponse) {
```
wslulciuc, line 42: So, fully aware that I initially used the named  `ProxyResource.emit()`, but perhaps `ProxyResource.proxyEvent()` would better convey the intent of the endpoint / method?

proxy/src/main/java/io/openlineage/proxy/api/models/ConsoleLineageStream.java
```diff
@@ -0,0 +1,34 @@
...
+{
+  public ConsoleLineageStream()
+  {
+    super(Type.CONSOLE);
+  }
+
+  @Override
+  public void collect(String event)
+  {
+    System.out.println(event);
```
wslulciuc, line 32: _Minor_: Should we use the `logger` here (`log.info()`), instead of `println()`?

proxy/src/main/java/io/openlineage/proxy/api/models/LineageStream.java
wslulciuc: Yeah, good point +1 (Was initially added to illustrate multiple writes to lineage streams).

proxy/src/main/java/io/openlineage/proxy/service/ProxyService.java
```diff
@@ -0,0 +1,75 @@
...
+import io.openlineage.proxy.api.models.KafkaLineageStream;
+import io.openlineage.proxy.api.models.LineageEvent;
+import io.openlineage.proxy.api.models.LineageStream;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import lombok.NonNull;
+
+public final class ProxyService
+{
+  private static final Logger log = LoggerFactory.getLogger(ProxyService.class);
```
wslulciuc, line 32: You can use `@Slf4j` an omit instantiating the logger yourself.

proxy/src/main/java/io/openlineage/proxy/service/ProxyService.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+
+  /**
+   * process an incoming event by sending it to all configured lineage streams.
+   *
+   * @param event incoming event
+   * @return completion future
+   */
+  public CompletableFuture<Void> emitAsync(@NonNull String event)
```
wslulciuc, line 65: Similar comment as above for  `ProxyResource.emit()`. I think `ProxyService.proxyEventAsync()` would better convey the intent of the service method.

proxy/src/main/java/io/openlineage/proxy/api/models/LineageEvent.java
```diff
@@ -0,0 +1,34 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+
+// NOTE: Only used for demonstration purposes, as of OpenLineage 0.3.0  please use server models
+// instead (see: https://github.com/OpenLineage/OpenLineage/pull/300)
+@EqualsAndHashCode
+@ToString
+public final class LineageEvent {
```
wslulciuc, line 27: Should we remove this class as it's no longer referenced / used?

proxy/src/main/java/io/openlineage/proxy/api/models/KafkaLineageStream.java
```diff
@@ -0,0 +1,75 @@
...
+public class KafkaLineageStream extends LineageStream
+{
+  private static final Logger log = LoggerFactory.getLogger(KafkaLineageStream.class);
+
+  private final String                   localServerId;
+  private final Properties               producerProperties;
+  private KafkaProducer<String, String>  producer;
+  private String                         topicName;
+
+  public KafkaLineageStream(String     localServerId,
```
wslulciuc, line 37: Mind if we define a `KafkaConfig` object? A class would make the configurable params more manageable as we add them.

```java
public KafkaLineageStream(KafkaConfig){...}
```

proxy/proxy.example.yml
```diff
@@ -0,0 +1,55 @@
...
+#
+consoleLog : true
+
+#
+# Kafka Configuration:
+#  - kafkaTopicName - name of topic to publish to - avoid mixing "-" and "." as spacers in the name - use one or other.
+#  - kafkaBootstrapServerURL - location of the kafka broker - typically localhost:9092
+#  - kafkaProperties - see [Apache Kafka producer configurations](http://kafka.apache.org/0100/documentation.html#producerconfigs) for more information and options
+#
+kafkaTopicName: openlineage.topic
```
wslulciuc, line 43: To easily allow users to add support for additional lineage streams, I would use a factory method to create a lineage stream from an array:

```yaml
# Enables lineage streams (default: console)
lineageStreams:
  - type: kafka
      topicName: my-topic
      bootstrapServerURL: localhost:9092
      ...
```

proxy/proxy.example.yml
```diff
@@ -0,0 +1,55 @@
...
+kafkaBootstrapServerURL : localhost:9092
+kafkaProperties:
+    acks: all
```
wslulciuc, line 46: Can we standardize on camelcase?

proxy/src/main/java/io/openlineage/proxy/service/ProxyService.java
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Constructor reviews the content of the config file and sets up the appropriate lineage streams.
+   *
+   * @param config configuration properties supplied to the application
+   */
+  public ProxyService(final ProxyConfig config)
+  {
+    /*
+     * There are two supported lineage streams.  The first is logging to the console.  This is activated by default or if requested in the config.
```
wslulciuc, line 45: This can be simplified if we define a `LineageStreamsFactory`, see [earlier](https://github.com/OpenLineage/OpenLineage/pull/400/files#r764421224) comment

proxy/src/main/java/io/openlineage/proxy/ProxyConfig.java
```diff
@@ -0,0 +1,103 @@
...
+{
+  @NotEmpty
+  private String     lineageSourceName;
+
+  private boolean    consoleLog = false;
+
+  @NotEmpty
+  private String     kafkaTopicName;
+  @NotEmpty
+  private String     kafkaBootstrapServerURL;
```
wslulciuc, line 39: Can we use the `URL` type here?

proxy/proxy.example.yml
```diff
@@ -0,0 +1,55 @@
...
+# routing of the lineage events without needing to unpack the event contents. In this example, it identifies that the
+# lineage event has gone through the proxy backend.  However, if a different proxy backend is serving each engine that
+# is producing lineage events, this value can be set up to represent the source engine.
+#
+lineageSourceName: openLineageProxyBackend
+
+#
+# Set to false to prevent lineage events being printed to stdout.
+#
+consoleLog : true
```
wslulciuc, line 35: We can eliminate the need for this flag, if by default, we [configure the console](https://github.com/OpenLineage/OpenLineage/blob/wip/proxy/proxy/src/main/java/io/openlineage/proxy/ProxyConfig.java#L26) lineage stream. Or, the user appends the type to the `lineageStreams` config (see below)

proxy/proxy.example.yml
```diff
@@ -0,0 +1,55 @@
...
+      # layout:
+      #   type: json
+
+#
+# This name is used to identify the lineage source in the different distribution mechanisms.  If can be used in further
+# routing of the lineage events without needing to unpack the event contents. In this example, it identifies that the
+# lineage event has gone through the proxy backend.  However, if a different proxy backend is serving each engine that
+# is producing lineage events, this value can be set up to represent the source engine.
+#
+lineageSourceName: openLineageProxyBackend
```
wslulciuc, line 30: I like this idea, but does each lineage stream config this different? That is, `server.id` is used for kafka, but what about other streams?

wslulciuc (approved):
@mandy-chessell, as discussed with @julienledem offline. We'll be merging your PR (great work btw!), with follow up PRs to address some of the feedback.

PR #437, opened 2021-12-15
Add proposal doc for making Spark event handling code more extensible

collado-mike:
### Problem

Proposal doc written for https://github.com/OpenLineage/OpenLineage/issues/168 . Follows the design proposal process in https://github.com/OpenLineage/OpenLineage/issues/336 .

### Checklist

### DIFF
--- a/proposals/168/making_spark_visitors_extensible.md
+++ b/proposals/168/making_spark_visitors_extensible.md
@@ -51,47 +51,41 @@ public interface CustomFacetBuilder<T, F> {
 public interface OpenLineageEventListenerFactory {
 
     /**
-     * Set the {@link OpenLineageContext} to be used during construction of the below builders
-     * @param context
-     */
-    default void setOpenLineageContext(OpenLineageContext context) {} // does nothing by default
-
-    /**
      * @return a collection of {@link io.openlineage.spark.api.QueryPlanVisitor}s that can construct `InputDataset`s.
      */
-    default Collection<QueryPlanVisitor<LogicalPlan>> createInputDatasetQueryPlanVisitors() {
+    default Collection<QueryPlanVisitor<LogicalPlan>> createInputDatasetQueryPlanVisitors(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<QueryPlanVisitor<LogicalPlan>> createOutputDatasetQueryPlanVisitors() {
+    default Collection<QueryPlanVisitor<LogicalPlan>> createOutputDatasetQueryPlanVisitors(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<PartialFunction<Object, List<OpenLineage.InputDatasetBuilder>>> createInputDatasetBuilder() {
+    default Collection<PartialFunction<Object, List<OpenLineage.InputDatasetBuilder>>> createInputDatasetBuilder(OpenLineageContext context) {
         return PartialFunction.empty();
     }
 
-    default Collection<PartialFunction<Object, List<OpenLineage.InputDatasetBuilder>>> createOutputDatasetBuilder() {
+    default Collection<PartialFunction<Object, List<OpenLineage.InputDatasetBuilder>>> createOutputDatasetBuilder(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<CustomFacetBuilder<Object, OpenLineage.InputDatasetFacet>> createInputDatasetFacetBuilders() {
+    default Collection<CustomFacetBuilder<Object, OpenLineage.InputDatasetFacet>> createInputDatasetFacetBuilders(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<CustomFacetBuilder<Object, OpenLineage.OutputDatasetFacet>> createOutputDatasetFacetBuilders() {
+    default Collection<CustomFacetBuilder<Object, OpenLineage.OutputDatasetFacet>> createOutputDatasetFacetBuilders(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<CustomFacetBuilder<Object, OpenLineage.DatasetFacet>> createDatasetFacetBuilders() {
+    default Collection<CustomFacetBuilder<Object, OpenLineage.DatasetFacet>> createDatasetFacetBuilders(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<CustomFacetBuilder<Object, OpenLineage.RunFacet>> createRunFacetBuilders() {
+    default Collection<CustomFacetBuilder<Object, OpenLineage.RunFacet>> createRunFacetBuilders(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
-    default Collection<CustomFacetBuilder<Object, OpenLineage.JobFacet>> createJobFacetBuilders() {
+    default Collection<CustomFacetBuilder<Object, OpenLineage.JobFacet>> createJobFacetBuilders(OpenLineageContext context) {
         return Collections.emptyList();
     }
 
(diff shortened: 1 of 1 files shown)


mobuchowski:
Thanks for the proposal @collado-mike!

I have few questions.

`InputDatasetFacets and OutputDatasetFacets will be attached to any InputDatasetBuilder or OutputDatasetBuilder found for the event.`

Does that mean that the found facets will be attached also to datasets constructed outside that particular factory implementation?

About this part:

```
For each SparkListener event, an OpenLineage.RunEvent will be constructed by
```

where will facets created by `datasetFacetBuilders` be attached? To both of them?


`OpenLineageContext` - we already have class named like that. One of those needs to be renamed.

collado-mike:
> `OpenLineageContext` - we already have class named like that. One of those needs to be renamed.
Oops- forgot to include in the doc that I intend to rename that class. Something like `OpenLineageEventLogger` or something, as that more correctly reflects what it does.

proposals/168/making_spark_visitors_extensible.md
```diff
@@ -0,0 +1,271 @@
...
+5. _If_ any `InputDatasetBuilder`s are returned, invoke the `datasetFacetBuilder`s and attach any returned facets to 
+    the `InputDatasetBuilder`s
+6. Construct `OutputDatasetBuilder`s by invoking each of the `outputDatasetBuilder`s.
+7. _If_ any `OuptutDatasetBuilder`s are returned, invoke the `outputDatasetFacetBuilder`s and attach any returned facets to
+   the `OutputDatasetBuilder`s
+5. _If_ any `OutputDatasetBuilder`s are returned, invoke the `datasetFacetBuilder`s and attach any returned facets to
+   the `OutputDatasetBuilder`s
+
+`InputDatasetFacet`s and `OutputDatasetFacet`s will be attached to _any_ `InputDatasetBuilder` or `OutputDatasetBuilder`
+found for the event. This is because facets may be constructed from generic information that is not specifi**cally tied
```
collado-mike, line 162: > Does that mean that the found facets will be attached also to datasets constructed outside that particular factory implementation?

Yes - the CustomFacetBuilder may return a very generic facet, such as `OutputStatistics`, which is applicable to any `OutputDataset`. I see two approaches:

1. The Dataset builders must be responsible for building the Datasets and the facets. This means the Dataset builder must always match on the highest level event - e.g., the Stage or the ActiveJob - and then descend down to find the RDD or whatever is needed to construct the Dataset and all of the associated facets. This also means that only facets defined by the Dataset builder can be attached to that Dataset.
2. Each builder is responsible for constructing exactly one component - Dataset builders only construct Datasets and Facet builders only construct Facets. This means each builder can be built specific to the event that actually contains the data necessary to construct the Dataset/Facet (i.e., the Stage or the RDD) while the logic of moving from high-level event to low-level event can be extracted out and that facets can be attached to any dataset regardless of whether the Dataset builder is aware of that facet. 

The approach outlined is the latter, as I feel it is more extensible and allows authors to attach facets to Datasets that are perhaps custom to the author (e.g., there might be some business metadata I want to attach to the specific BigQuery Dataset that I'm processing even though the BigQuery library author doesn't know anything about my business metadata). From my experimentation, there is generally only one input Dataset attached to a Stage, so the facets, I think, can reliably be attached to the correct input dataset. The OutputDataset must typically be found in the QueryExecution and there's only one output dataset per QueryExecution, so any output dataset facets can be reliably attached to that. 

Let me know if this reasoning makes sense to you.

proposals/168/making_spark_visitors_extensible.md
```diff
@@ -0,0 +1,271 @@
...
+the `CustomFacet`s returned are already immutable.
+
+For each `SparkListener` event, an `OpenLineage.RunEvent` will be constructed by 
+1. Construct `RunFacet`s by invoking each of the `runFacetBuilders`
+2. Construct `JobFacet`s by invoking each of the `jobFacetBuilders`
+3. Construct `InputDatasetBuilder`s by invoking each of the `inputDatasetBuilder`s.
+4. _If_ any `InputDatasetBuilder`s are returned, invoke the `inputDatasetFacetBuilder`s and attach any returned facets to
+   the `InputDatasetBuilder`s
+5. _If_ any `InputDatasetBuilder`s are returned, invoke the `datasetFacetBuilder`s and attach any returned facets to 
+    the `InputDatasetBuilder`s
```
collado-mike, line 154: > where will facets created by datasetFacetBuilders be attached? To both of them?

This is a good question. From my experimenting, a `Stage` that reads the raw data will only contain one input dataset and no output datasets. A job will have only one output dataset.

A possible exception to this includes dynamic datasets, such as dynamic partitions in Hive or dynamic topics in Kafka. E.g., a record can contain the name of the Kafka topic as a column and the Kafka topic _should_ be counted as a distinct dataset. However, we currently don't support this behavior and I think adding support for this would not allow us to rely on facets created by inspecting events on the driver. E.g., the `recordsWritten` field in the `TaskMetrics` would be the sum total of all records written to Kafka, not those written to a specific topic. To collect the actual number written to a specific topic would involve counting those records on the executor. This is work that would be very custom - it is supported by this framework, but the implementation work would likely be done by an application owner, not us.

wjohnson:
Yes!  This is awesome!!  How might this work for an existing command like the DataSourceV2Visitor.

Right now, OpenLineage checks based on the provider property if it's an Iceberg or Delta provider.

Ideally, we'd be able to extend the list of providers or have a custom "CosmosDbDataSourceV2Visitor" that knew how to work with a custom DataSourceV2.

Would that cause any conflicts if the base class is already accounted for in OpenLineage?

collado-mike:
> Right now, OpenLineage checks based on the provider property if it's an Iceberg or Delta provider.
> 
> Ideally, we'd be able to extend the list of providers or have a custom "CosmosDbDataSourceV2Visitor" that knew how to work with a custom DataSourceV2.

Good question, and this points to a larger theme that I think is missing from the description. I think the existing visitor needs to be updated to only explicitly support the DataSources it knows how to handle. I made this same error when writing the `LogicalRelationVisitor` in this [PR](https://github.com/OpenLineage/OpenLineage/commit/d114bc31e9b596211e4173a0685ee23a71c42ffc#diff-85a1ce33e29ca82a01294f6c30083b7b04f1f67120ed1fb74f2bf4ef680706a9). I ended up removing the support for unknown `LogicalRelation`s because it turned out to conflict with the `BigQueryNodeVisitor`, which supports `BigQueryRelation`s. As a rule, we should avoid having fallback behavior unless we know absolutely that there won't be another handler for a given node type. I think the existing `DataSourceV2RelationVisitor` ought to be altered to eliminate its "unknown" provider support and allow other visitors the opportunity to fill in.

There probably ought to be a way to add in fallback logic for `LogicalPlan` nodes and other events, but I haven't included that in this design.

mobuchowski (approved):

proposals/168/making_spark_visitors_extensible.md
```diff
@@ -0,0 +1,271 @@
...
+usually be constructed by walking the `RDD` dependency tree for that `Stage` and finding a `FileScanRDD` or other 
+concrete implementation. But while there is typically only one `InputDataset` read in a given stage, there's no 
+guarantee of that and the `TaskMetrics` in the `StageInfo` won't disambiguate. If a facet needs to be attached to a
+specific dataset, the user must take care to construct both the Dataset and the Facet in the same builder.
+
+For convenience, the following abstract base classes are defined:
+
+```java
+// accepts an argument of type T and returns an InputDatasetBuilder
+public abstract class AbstractInputDatasetBuilder<T>{}
```
wslulciuc, line 176: _Minor_: But is the `Abstract*` prefix redundant?

wslulciuc (approved):
Great proposal @collado-mike! Well thought out and written 💯. I left a minor comment, but otherwise 🥇

proposals/168/making_spark_visitors_extensible.md
```diff
@@ -0,0 +1,271 @@
...
+     * the name of the facet and the facet itself.
+     */
+    void accept(T event, BiConsumer<String, ? super F> consumer);
+
+}
+
+/**
+ * Factory for the builders that generate OpenLineage components and facets from Spark events.
+ */
+public interface OpenLineageEventListenerFactory {
```
wslulciuc, line 51: Based on https://github.com/OpenLineage/OpenLineage/pull/448, this is now `OpenLineageEventHandlerFactory`?

collado-mike: Yeah, that's right. I have to go back to the proposal and update for changes necessitated by implementation

proposals/168/making_spark_visitors_extensible.md
collado-mike: This is pretty standard Java convention. E.g., `AbstractMap`, `AbstractList`, `AbstractCollection`, `AbstractFunction0`, `AbstractPartialFunction`, ... I think it helps readers of the concrete subclass to know that the subclass is extending an in complete base class without having to go look at the source code of the super class.

Issue #363, opened 2021-10-28
[INTEGRATION] dbt spark adapter


dvirgiln:
**Integration target:**
https://pypi.org/project/dbt-spark

NotImplementedError: Only 'snowflake', 'bigquery', and 'redshift' adapters. Passed spark

**Proposed implementation**
It would be good to have dbt spark adapter and as well Athena. But specially spark one, as Athena is more experimental.

In our company we are thinking to use OpenLineage, but we use DBT Spark adapter mainly. We are starting using as well Athena adapter.

julienledem:
I think it makes sense to support the dbt adapters in the dbt integration.
@dvirgiln should we close #364 as a duplicate?

julienledem:
@mobuchowski eventually this would also be the parent to a spark run that would setup the spark integration but we are not there yet.

mobuchowski:
I agree. We don't have an option to automatically run data yet.

Issue #435, opened 2021-12-13
[INTEGRATION][Spark] provide physical location of a table facet in case we're using logical table name


mobuchowski:
When dealing with a logical Hive table, we've decided to use logical Hive table name as a dataset name. In some other cases, some other job - not necessarily Spark one - might deal with this dataset by directly knowing it's physical location. In this case we'll have the same dataset known by two different names in different namespaces.

To make sure those datasets can be "linked" by OpenLineage consumers, when available, we should provide additional facet (`PhysicalLocationFacet`?) that indicates physical location at a storage system (like `hdfs`, `s3` etc) or other underlying storage mechanism.

Make sure to update Naming.md to standarize how we're naming hive dataset.

wjohnson:
This is a great idea and would be very helpful for our work integrating OpenLineage with Apache Atlas.

pawel-big-lebowski:
One could also consider extending `StorageDatasetFacet` to contain a field with such an information -> https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/StorageDatasetFacet.json

georgeliang0821:
I have lot of customers using Delta Table on Spark (Databricks), we met the challenge that the OpenLineage might not be able to support SQL table name. May I draw your immediate attention please to consider prioritize this enhancement, thanks.

wjohnson:
@mobuchowski - In what cases does OpenLineage currently report the hive table name?

I couldn't find an example in the unit / integration tests and the [Naming](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md) guide does not seem to include Hive.

`spark.sql("SELECT * FROM tableName") ` appears to return the physical path rather than the Hive table name. Is that expected or am I just getting a different experience because I'm using Databricks?

mobuchowski:
@wjohnson I think this is also about changing that 😉 a valid point though  - modified issue to indicate that.

@georgeliang0821 this issue shouldn't modify anything regarding Delta Lake AFAIK, just hive for now

mobuchowski:
Also, as you see, this is not a fresh issue - it was made around time when we were filling gaps in Hive coverage with @pawel-big-lebowski - would be easier to do back then rather than now when you're using it already 😞

collado-mike:
[...]

PR #453, opened 2022-01-03
[INTEGRATION][dbt] dbt-spark integration

mobuchowski:
Supporting dbt-spark is equivalent to other adapters with only one, but large difference. 
Naming datasets properly is difficult since we don't actually know real dataset location, only the cluster URI.
The solution is to provide spark "logical" dataset namespace with spark address dependent on connection method here https://docs.getdbt.com/reference/warehouse-profiles/spark-profile . This dataset namespace won't not match one that would be created if this dataset would be referred inside spark cluster with spark integration.
 

For now, I think it's the best way - after https://github.com/OpenLineage/OpenLineage/issues/435 "linking" datasets should be solution for this.

Another theoretically possible option would be to use credentials passed to dbt to connect to cluster, and get dataset information there. Complexity of that would be very large, and it's not guaranteed to work.

Closes: https://github.com/OpenLineage/OpenLineage/issues/363

### Checklist


integration/dbt/README.md
```diff
@@ -14,7 +14,7 @@ Wrapper script for dbt for automatic metadata collection
 - [Python >= 3.6](https://www.python.org/downloads)
 - [dbt >= 0.20](https://www.getdbt.com/)
 
-Right now, `openlineage-dbt` only supports `bigquery`, `snowflake` and `redshift` dbt adapters.
+Right now, `openlineage-dbt` only supports `bigquery`, `snowflake`, `spark` and `redshift` dbt adapters.
```
wslulciuc, line 17: Maybe we should switch to a table format to list adapters? With information on:

* Version supported
* Future adopter support
* Etc.

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -633,9 +633,21 @@ def extract_namespace(self, profile: Dict) -> str:
             return "bigquery"
         elif profile['type'] == 'redshift':
             return f"redshift://{profile['host']}:{profile['port']}"
+        elif profile['type'] == 'spark':
+            if profile['method'] in ['thrift', 'odbc', 'http']:
+                port = ""
```
wslulciuc, line 638: _Minor_: Should we provide a default port if not specified by the user?

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -633,9 +633,21 @@ def extract_namespace(self, profile: Dict) -> str:
             return "bigquery"
         elif profile['type'] == 'redshift':
             return f"redshift://{profile['host']}:{profile['port']}"
+        elif profile['type'] == 'spark':
+            if profile['method'] in ['thrift', 'odbc', 'http']:
```
wslulciuc, line 637: _Minor_: Can we define the method types as enums?

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -633,9 +633,21 @@ def extract_namespace(self, profile: Dict) -> str:
...
+                return f"spark://{profile['host']}" + port
+            else:
+                raise NotImplementedError(
+                    f"Connection method `{profile['method']}` is not "
+                    f"supported for spark adapter."
+                )
         else:
             raise NotImplementedError(
-                f"Only 'snowflake', 'bigquery', and 'redshift' adapters are supported right now. "
+                f"Only 'snowflake', 'bigquery', 'redshift' and "
```
wslulciuc, line 649: If we define an `Adaptor` enum, we can list the values and build the error string based on the list. This will avoid always (and maybe sometimes forgetting) to update the error message.

integration/dbt/README.md
mobuchowski: Good idea. There's a lot of possible adapters, and this will make implementing future ones much clearer.

integration/common/openlineage/common/provider/dbt.py

integration/common/openlineage/common/provider/dbt.py

integration/common/openlineage/common/provider/dbt.py
mobuchowski: Done. Those are different depending on method: `thrift` is `10001`, `odbc` and `http` are `443`.

integration/common/openlineage/common/provider/dbt.py
```diff
@@ -627,15 +664,34 @@ def extract_job_namespace(self, profile: Dict):
...
+            elif profile['method'] == SparkConnectionMethod.THRIFT.value:
+                port = '10001'
+
+            if profile['method'] in SparkConnectionMethod.methods():
+                return f"spark://{profile['host']}{port}"
+            else:
+                raise NotImplementedError(
+                    f"Connection method `{profile['method']}` is not "
+                    f"supported for spark adapter."
+                )
```
collado-mike, line 691: I'd invert the if condition and move it up to exit early if the `profile['method']` isn't supported.

collado-mike (approved):

Issue #455, opened 2022-01-03
[BUG][AIRFLOW] openlineage-airflow uses scheduled dag execution date as a event time of the job start.


fm100:
https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/openlineage/lineage_backend/__init__.py#L54

For airflow 2.x integration, it uses `dagrun.execution_date` for the `eventTime` of the start event, which is scheduled time of the dag execution. It is possible that the actual start time could be quite later than scheduled time and the information could be misleading. Following is the information from example airflow task:
```
execution_date | 2022-01-03 21:00:00+00:00
start_date | 2022-01-03, 22:00:10
end_date | 2022-01-03, 22:00:11
```

From the above example, the task was scheduled to run at 21 but ran at 22. I believe the `eventTime` of the start event is supposed to be the actual start time, and `nominalStartTime` in facets captures the scheduled time.

PR #456, opened 2022-01-04
[INTEGRATION][AIRFLOW] Use task instance start date for event time of the start event instead of dag execution date.

fm100:
Signed-off-by: Minkyu Park <minkyu@datakin.com>

### Problem
From the airflow 2.x integration, the `eventTime` of the start event uses `dagrun.execution_date` which is scheduled time of the dag but not the actual task start time.

Closes: #455 

### Solution
This PR changes to use `task_instance.start_date` instead of `dagrun.execution_date`.

### Checklist

### DIFF
--- a/integration/airflow/openlineage/lineage_backend/__init__.py
+++ b/integration/airflow/openlineage/lineage_backend/__init__.py
@@ -51,7 +51,7 @@ class Backend:
                 run_id=run_id,
                 job_name=job_name,
                 job_description=dag.description,
-                event_time=DagUtils.get_start_time(dagrun.execution_date),
+                event_time=DagUtils.get_start_time(task_instance.start_date),
                 parent_run_id=dagrun.run_id,
                 code_location=self._get_location(operator),
                 nominal_start_time=DagUtils.get_start_time(dagrun.execution_date),
--- a/integration/airflow/tests/integration/test_integration.py
+++ b/integration/airflow/tests/integration/test_integration.py
@@ -96,6 +96,16 @@ def check_matches_ordered(expected_events, actual_events) -> bool:
     return True
 
 
+def check_event_time_ordered(actual_events) -> bool:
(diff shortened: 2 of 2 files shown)


integration/airflow/openlineage/lineage_backend/__init__.py
```diff
@@ -51,7 +51,7 @@ def send_lineage(
                 run_id=run_id,
                 job_name=job_name,
                 job_description=dag.description,
-                event_time=DagUtils.get_start_time(dagrun.execution_date),
+                event_time=DagUtils.get_start_time(task_instance.start_date),
```
wslulciuc, line 54: Is there a way we can test the fix?

wslulciuc (approved):
Minor comment on testing the fix, otherwise 👍

integration/airflow/openlineage/lineage_backend/__init__.py
wslulciuc: I think one reasonable approach would be to update our Airflow integration test, see [`test_integration.py`](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/tests/integration/test_integration.py). For each DAG run, we call [`wait_for_dag()`](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/tests/integration/test_integration.py#L39) with the DAG ID that queries the `dag_run` table. You can do something similar by checking the start time, nominal times, etc for each DAG run on completion:

```python
check_dag_start_time(DAG_ID)
check_dag_nominal_start_time(DAG_ID)
.
.
```

fm100: Thank you for the feedback. I've added a check to the integration test case, simply checking the values and orders of `eventTime` that is actually generated. The issue was not on the DAG information but was the `eventTime` of the openlineage event, so checking the generated event time would be sufficient.

Issue #452, opened 2021-12-28
[PROPOSAL][SPARK] Support Databricks Synapse Connector


wjohnson:
**Purpose:**

We would like to extend OpenLineage's Spark integration to work with the [Azure Synapse (formerly known as SQL Data Warehouse)](https://docs.microsoft.com/en-us/azure/databricks/data/data-sources/azure/synapse-analytics) in the Databricks Environment. Azure Databricks offers this proprietary package only in the Databricks runtime (across at least Azure and AWS).

This would enable people using Databricks, OpenLineage, and Azure Synapse to observe the input and output to their spark jobs.

**Proposed implementation**

Databricks' Synapse connector extends a Base Relation which works with a JDBC connection to Azure Synapse. As a result, I am proposing that we add a SqlDWRelationVisitor much in the same way as there is a BigQueryNodeVisitor.

A new namespace of "synapsedatabricks" would be added.  Why "synapsedatabricks" and not "synapse"? Because this connector is only found in Databricks and [other Synapse connectors are available in different environments](https://docs.microsoft.com/en-us/azure/synapse-analytics/spark/synapse-spark-sql-pool-import-export). I think the namespace "synapse" should be reserved for a Microsoft derived connector if it ever is implemented for DataSourceV2Relations.

A few caveats:

* The package is proprietary and the SqlDWRelation is a private class.
* As a result, a small use of reflection is necessary to access the underlying `tableNameOrSubquery` class that contains the actual table name.
* In addition, we'd want to extract out the JDBC URL to get the server and database name for the Synapse database. This also requires a bit of reflection.

Lastly, this effort would be made easier with the acceptance / completion of the proposal in #437 .

collado-mike:
This sounds good to me. Re: the reference to #437, is the proposal to include this visitor in the core OpenLineage Spark library or to create an extension package?

wjohnson:
Ah!  Good call, @collado-mike I have a PR ready to go that would be included in the core OpenLineage Spark library (since I was working on it prior to 437 being complete).

In addition, since this is *Databricks* I think it would be relevant for more than just Microsoft customers (although it's geared toward using Databricks to connect with an Azure / Microsoft product).

I would ask for your input.  Would you rather see an extension package (then I guess I wouldn't actually put out a PR to you guys here, right) or keep this as part of OpenLineage's Databricks support?

collado-mike:
[...]

PR #460, opened 2022-01-08
[INTEGRATION][SPARK]Support for SqlDWRelation on Databricks' Azure Synapse/SQL DW Connector

wjohnson:
Defines a SqlDWDatabricksVisitor that extracts a namespace of `sqlserver://<server>;database=<db>;`
and name of `<table name>` or `(<query>) q`.

Signed-off-by: Will Johnson <will@willj.co>

### Problem

[Databricks' Azure Synapse connector](https://docs.microsoft.com/en-us/azure/databricks/data/data-sources/azure/synapse-analytics) supports connecting Azure Databricks environments to Azure Synapse (formerly Sql Data Warehouse). The classes are all proprietary / closed source but can be accessed via reflection. The SqlDWRelation extends a Base Relation which works with a JDBC connection to Azure Synapse. As a result, I am proposing that we add a SqlDWRelationVisitor much in the same way as there is a BigQueryNodeVisitor.

Closes: #452

### Solution

Implement a SqlDWRelationVisitor that extracts the jdbc url and table name using reflection since the class is private.

The namespace will be sqlserver://server;database=databaseName; and the name will be either the table name or query (a user may submit a query or table name and Azure SQL DW / Synapse will execute the query or select all records from the table name and then return the results).

**Scope**: This solution has been tested on Azure Databricks Runtimes 8.3 and 9.1 LTS

### Checklist

- [ ] Your changes are accompanied by tests (_if relevant_) **Happy to add additional tests but would appreciate guidance on what we should do for cloud infra**

### DIFF
--- a/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
@@ -1,0 +1,127 @@
+package io.openlineage.spark.agent.lifecycle.plan;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.spark.api.DatasetFactory;
+import io.openlineage.spark.api.OpenLineageContext;
+import io.openlineage.spark.api.QueryPlanVisitor;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Collections;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.commons.lang3.reflect.MethodUtils;
+import org.apache.spark.api.java.Optional;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import org.apache.spark.sql.execution.datasources.LogicalRelation;
+import org.apache.spark.sql.sources.BaseRelation;
+
+/**
+ * {@link LogicalPlan} visitor that matches SqlDWRelation that comes from an Azure Databricks
+ * environment. This function extracts a {@link OpenLineage.Dataset} from the SQL DW/ Synapse table
+ * referenced by the relation. The convention used for a namespace is a URI of <code>
+ * sqlserver://&lt;server&gt;.&lt;.datasetId&gt;.&lt;tableName&gt;</code> . The name for Sql Dw
+ * tables may be table name (e.g. "exampleInputA") or a multi-part name (e.g.
+ * "[dbo].[exampleInputA]"). If the data source is a query (e.g. <code>
+ * ((select \"id\" FROM dbo.exampleInputA WHERE postalCode != '55555') q)</code>) then the name will
+ * be <code>COMPLEX</code>.
+ */
+@Slf4j
+public class SqlDWDatabricksVisitor<D extends OpenLineage.Dataset>
+    extends QueryPlanVisitor<LogicalPlan, D> {
+  private final DatasetFactory<D> factory;
+  private static final Pattern dbJdbcPattern = Pattern.compile("database=([^;]*);?");
+  private static final Pattern serverJdbcPattern = Pattern.compile("jdbc:([^;]*);?");
+  private static final String DATABRICKS_CLASS_NAME = "com.databricks.spark.sqldw.SqlDWRelation";
+
+  public SqlDWDatabricksVisitor(OpenLineageContext context, DatasetFactory<D> factory) {
+    super(context);
+    this.factory = factory;
+  }
+
+  public static boolean hasSqlDWDatabricksClasses() {
+    try {
+      SqlDWDatabricksVisitor.class.getClassLoader().loadClass(DATABRICKS_CLASS_NAME);
+      return true;
+    } catch (Exception e) {
+      // swallow- we don't care
+    }
+    return false;
+  }
+
+  protected boolean isSqlDwRelationClass(LogicalPlan plan) {
+    return plan instanceof LogicalRelation
+        && ((LogicalRelation) plan).relation().getClass().getName().equals(DATABRICKS_CLASS_NAME);
+  }
+
+  @Override
+  public boolean isDefinedAt(LogicalPlan plan) {
+    return isSqlDwRelationClass(plan);
+  }
+
+  private Optional<String> getName(BaseRelation relation) {
+    String tableName;
+    try {
+      tableName = (String) FieldUtils.readField(relation, "tableNameOrSubquery", true);
+    } catch (IllegalAccessException e) {
+      log.warn("Unable to discover SQLDW tableNameOrSubquery property");
+      return Optional.empty();
+    }
+    // The Synapse connector will return a table name wrapped in double quotes
+    // or you could have a query string (e.g. (SELECT * FROM table)q)
+    if (tableName.startsWith("\"") && tableName.endsWith("\"")) {
+      tableName = tableName.replace("\"", "");
+    }
+    // TODO If there is a query, we should ultimately parse the SQL but
+    // returning COMPLEX to be consistent with other implementations.
+    if (tableName.startsWith("(")) {
+      tableName = "COMPLEX";
+    }
+    return Optional.of(tableName);
+  }
+
+  private Optional<String> getNameSpace(BaseRelation relation) {
+
+    String jdbcUrl;
+    try {
+      Object fieldDetails = FieldUtils.readField(relation, "params", true);
+      jdbcUrl = (String) MethodUtils.invokeMethod(fieldDetails, true, "jdbcUrl");
+    } catch (IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
+      log.warn("Unable to discover SQLDW jdbcUrl Parameters");
+      return Optional.empty();
+    }
+
+    Matcher serverMatcher = serverJdbcPattern.matcher(jdbcUrl);
+    boolean serverIsFound = serverMatcher.find();
+    Matcher dbMatcher = dbJdbcPattern.matcher(jdbcUrl);
+    boolean dbIsFound = dbMatcher.find();
+
+    if (!(dbIsFound && serverIsFound)) {
+      log.warn("Unable to discover SQLDW database name or server name from jdbc url");
+      return Optional.empty();
+    }
+
+    String databaseSubString = dbMatcher.group(1);
+    String serverSubString = serverMatcher.group(1);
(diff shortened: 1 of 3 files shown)


integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.spark.sql.execution.datasources.LogicalRelation;
+import org.apache.spark.sql.sources.BaseRelation;
+
+/**
+ * {@link LogicalPlan} visitor that matches SqlDWRelation that comes from an Azure Databricks
+ * environment. This function extracts a {@link OpenLineage.Dataset} from the SQL DW/ Synapse table
+ * referenced by the relation. The convention used for a namespace is a URI of <code>
+ * sqlserver://&lt;server&gt;.&lt;.datasetId&gt;.&lt;tableName&gt;</code> . The name for Sql Dw
+ * tables may be a query (e.g. <code>
+ * ((select \"id\" FROM dbo.exampleInputA WHERE postalCode != '55555') q)</code>) or a table name
```
collado-mike, line 26: Hmm, I actually don't know if this will work for Marquez. For JDBC relations, we currently just insert `COMPLEX` as the string name (admittedly ugly), with a TODO to parse the SQL to extract the table names being read - https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/LogicalRelationVisitor.java#L106-L120 . The GreatExpectations impl actually does the SQL parsing - https://github.com/OpenLineage/OpenLineage/blob/main/integration/common/openlineage/common/provider/great_expectations/action.py#L180-L186 .

Rather than posting the SQL query, I'd vote for either of the above solutions - a TODO for now (with or without the ugly substitute) or actually parse the SQL to extract the real sources.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+    return sqlDwSupplier(x)
+        .map(
+            s -> {
+              BaseRelation relation = s.get();
+              String namespace = getNameSpace(s);
+              String name = getName(s);
+              return Collections.singletonList(
+                  factory.getDataset(name, namespace, relation.schema()));
+            })
+        .orElse(null);
```
collado-mike, line 123: Convention is to return an empty list if not applicable

integration/spark/src/main/common/java/io/openlineage/spark/agent/transformers/SqlDWRelationTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * {@link ClassFileTransformer} that rewrites both the {@link
+ * com.google.cloud.spark.bigquery.BigQueryRelation} class as well as the v2 {@link
+ * com.google.cloud.spark.bigquery.v2.BigQueryDataSourceReader} class to implement the {@link
+ * DatasetSource} interface by introducing the {@link DatasetSource#name()} method, returning the
+ * fully qualified table name (&lt;projectId&gt;.&lt;dataset&gt;.&lt;table&gt;), and the {@link
+ * DatasetSource#namespace()} method, returning the literal string "bigquery".
+ */
+@Slf4j
+public class SqlDWRelationTransformer implements ClassFileTransformer {
```
collado-mike, line 23: Sorry, I should have mentioned that this isn't necessary. The agent approach is going away and isn't advertised anymore. We can skip merging this code.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private Optional<Supplier<BaseRelation>> sqlDwSupplier(LogicalPlan plan) {
+    if (!(plan instanceof LogicalRelation)) {
+      return Optional.empty();
+    }
+
+    Class relationClass = ((LogicalRelation) plan).relation().getClass();
+    if (plan instanceof LogicalRelation
+        && relationClass.toString().equals("class com.databricks.spark.sqldw.SqlDWRelation")) {
```
collado-mike, line 64: This should be `relationClass.getName().equals("com.databricks.spark.sqldw.SqlDWRelation")`

collado-mike: If you extract this line into a `protected` method, you can write tests for this code. e.g., 

```java
protected boolean isSqlDwRelationClass(LogicalRelation plan) {
  return plan instanceof LogicalRelation && 
    ((LogicalRelation) plan).relation().getClass().getName().equals("com.databricks.spark.sqldw.SqlDWRelation");
}
```
allows you to write a test class like

```java
class TestSqlDwDatabricksVisitor extends SqlDwDatabricksVisitor {
  @Override
  protected boolean isSqlDwRelationClass(LogicalRelation plan) {
    return true;
  }
}
```
Include a test `BaseRelation` that looks something like the `SqlDwRelation` will allow you to test the regexes and the reflection code.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
wjohnson: Thank you, Mike! I will defer to the "COMPLEX" name for now. I don't think I feel qualified to choose a T-SQL parsing engine for the project!

integration/spark/src/main/common/java/io/openlineage/spark/agent/transformers/SqlDWRelationTransformer.java
wjohnson: I will drop it from the PR!  Thank you!

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
wjohnson: Thank you for the guidance, Mike!  I was able to create a couple tests based off of this insight.

wjohnson:
Okay!  We were able to account for all of the feedback provided and added a couple tests to confirm the regex and reflection works as intended.  Thank you @collado-mike for the great feedback already and we look forward to a (hopefully) final review.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+  public boolean isDefinedAt(LogicalPlan plan) {
+    return isSqlDwRelationClass(plan);
+  }
+
+  private String getName(BaseRelation relation) {
+    String tableName;
+    try {
+      tableName = (String) FieldUtils.readField(relation, "tableNameOrSubquery", true);
+    } catch (IllegalAccessException e) {
+      throw new RuntimeException("Unable to discover SQLDW tableNameOrSubquery property");
```
collado-mike, line 65: I would not throw an exception here. Exceptions can cause the overall lineage event construction to fail (there should be try/catches, but...), but really only this component is missing some details. I'd log a warning and return an empty list from the `apply` method if we can't get the information we need.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+    Matcher serverMatcher = serverJdbcPattern.matcher(jdbcUrl);
+    boolean serverIsFound = serverMatcher.find();
+    Matcher dbMatcher = dbJdbcPattern.matcher(jdbcUrl);
+    boolean dbIsFound = dbMatcher.find();
+
+    if (!(dbIsFound && serverIsFound)) {
+      throw new RuntimeException(
+          "Unable to discover SQLDW database name or server name from jdbc url");
+    }
+
```
collado-mike, line 99: Same comment as above - log a warning and let's avoid the exception

collado-mike (commented):
This is great! My only concern is the exceptions being thrown - if we can log warnings instead, we can get this merged.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * {@link LogicalPlan} visitor that matches SqlDWRelation that comes from an Azure Databricks
+ * environment. This function extracts a {@link OpenLineage.Dataset} from the SQL DW/ Synapse table
+ * referenced by the relation. The convention used for a namespace is a URI of <code>
+ * sqlserver://&lt;server&gt;.&lt;.datasetId&gt;.&lt;tableName&gt;</code> . The name for Sql Dw
+ * tables may be table name (e.g. "exampleInputA") or a multi-part name (e.g.
+ * "[dbo].[exampleInputA]"). If the data source is a query (e.g. <code>
+ * ((select \"id\" FROM dbo.exampleInputA WHERE postalCode != '55555') q)</code>) then the name will
+ * be COMPLEX.
```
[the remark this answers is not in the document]
wjohnson: Fixed in the latest update!

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
wjohnson: Got it!  I had seen a few other visitors actually throw an exception but I totally agree with and understand this intention. Thank you for explaining.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/SqlDWDatabricksVisitor.java
wjohnson: Thank you again for reviewing! Resolved in latest update.

collado-mike (approved):

wjohnson:
Thank you so much!!!

PR #474, opened 2022-01-14
Update proxy config

wslulciuc:
### Problem

With the [`openlineage-proxy`](https://github.com/OpenLineage/OpenLineage/tree/main/proxy) backend recently merged (see #400), there was some follow up work on stream configuration that needed to be addressed before the OpenLineage `0.5.0` release.

### Solution

This PR makes the following changes to [`proxy.yml`](https://github.com/OpenLineage/OpenLineage/blob/main/proxy/proxy.example.yml) when configuring proxy streams:

* A `ProxyStreamFactory` class has been added to simplify code, but also ease of proxy stream configuration
* A `ProxyStreamConfig` interface has been added to allow for defining configuration on a per stream type basis. For example, the `KafkaConfig` class defines properties specific to kafka. The config class is then used when calling `ProxyStreamFactory.build()` to return `ImmutableSet<LineageStream>` for each proxy stream configured

In the `proxy.yml` file, we now have a `proxy` property that enables the user to set the `source`, and one or more `streams`:

```yaml
proxy:
  source: # name of source (ex: openLineageProxyBackend)
  streams:
    - type: Console
    - type: Kafka
      # Kafka specific config (topic, bootstrapServerUrl, etc)
```

This PR also adds the `build-proxy-backend` and `release-proxy-backend` CI jobs. We also set the response status code for `POST` `/api/v1/lineage` to `200` when an event has been received (the semantics here is: _the event was received and the request was successfully handled_)

### Checklist


proxy/src/main/java/io/openlineage/proxy/ProxyStreamFactory.java
```diff
@@ -0,0 +1,42 @@
...
+@Slf4j
+public final class ProxyStreamFactory {
+  private static final String DEFAULT_PROXY_LINEAGE_SOURCE = "openLineageProxyBackend";
+  private static final Set<ProxyStreamConfig> DEFAULT_STREAMS =
+      Sets.newHashSet(new ConsoleConfig());
+
+  @Getter @Setter private String source = DEFAULT_PROXY_LINEAGE_SOURCE;
+  @Getter @Setter private Set<ProxyStreamConfig> streams = DEFAULT_STREAMS;
+
+  public ImmutableSet<LineageStream> build() {
```
collado-mike, line 24: I'd make this a `List` rather than `Set`, just so users have the ability to set the order of the target systems this broker is sending messages to. E.g., I may want to make sure the Kafka system always gets the messages first, in case a subsequent recipient is down or sending fails for some reason.

collado-mike (approved):

proxy/src/main/java/io/openlineage/proxy/ProxyStreamFactory.java
wslulciuc: Valid point 👍

proxy/src/main/java/io/openlineage/proxy/ProxyStreamConfig.java
```diff
@@ -0,0 +1,13 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import io.openlineage.proxy.api.models.ConsoleConfig;
+import io.openlineage.proxy.api.models.KafkaConfig;
+
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = ConsoleConfig.class, name = "Console"),
+  @JsonSubTypes.Type(value = KafkaConfig.class, name = "Kafka")
+})
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+public interface ProxyStreamConfig {}
```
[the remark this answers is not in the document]
wslulciuc: I added the javadocs to `ProxyStreamFactory`

wslulciuc:
@collado-mike: This PR is getting a bit large, I'll open a follow up PR to add the release config for the `proxy` jar.

PR #425, opened 2021-12-04
[Integration][Spark] Support adding arbitrary parameters to OpenLineage URL

wjohnson:
Support adding arbitrary parameters to OpenLineage URL via the `spark.openlineage.url` configuration variable or the newly added pattern of `spark.openlineage.url.param.xxx` where xxx is the name of the url parameter to be included.

Signed-off-by: Will Johnson <will@willj.co>

### Problem

For some services consuming OpenLineage metadata, we require having a specific url parameter for authentication purposes. We would like to have the open lineage url to include any arbitrary parameter and value we provide. This may also be useful for tracking purposes (e.g. a session id) or passing additional (static) data to the service consuming open lineage data.

Closes: #407

### Solution

Users can now pass along additional query / url parameters to the openlineage url that is used when emitting lineage. This is useful for passing additional parameters necessary for a non-marquez destination of OpenLineage metadata.

Supports extracting parameters from config: spark.openlineage.url
Supports extracting parameters from config: spark.openlineage.url.param.xxx

If using spark.openlineage.url.param.xxx, xxx represents the name of the url parameter you want to include in the lineage url. Any configuration variable passed in beginning with spark.openlineage.url.param. will be used as a url parameter in the lineage url.

These config settings will ignore api_key if it is specified as a url.param to avoid conflicting with the spark.openlineage.apiKey config.

If your change requires a schema change, please describe the schema modifcation(s) and whether it's a _backwards-incompatible_ or _backwards-compatible_ change, then select one of the following:

> **Note:** All schema changes require discussion. Please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for context.

- [ ] Your change modifies the [core](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json) OpenLineage model
- [ ] Your change modifies one or more OpenLineage [facets](https://github.com/OpenLineage/OpenLineage/tree/main/spec/facets)

If you're contributing a new integration, please specify the scope of the integration and how/where it has been tested (e.g., Apache Spark integration supports `S3` and `GCS` filesystem operations, tested with AWS EMR).

### Checklist

- [x] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [x] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
[...]

PR #480, opened 2022-01-18
Add http client for Java

wslulciuc:
### Problem

Closes: #ISSUE-NUMBER

### Checklist


client/java/build.gradle
```diff
@@ -45,12 +46,22 @@ configurations {
...
-    implementation "com.fasterxml.jackson.core:jackson-core:$jackson_version"
-    implementation "com.fasterxml.jackson.core:jackson-annotations:$jackson_version"
-    implementation "com.fasterxml.jackson.core:jackson-databind:$jackson_version"
-    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jackson_version"
+    implementation "com.fasterxml.jackson.core:jackson-core:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jdk8:${jacksonVersion}"
+    implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jacksonVersion}"
+    implementation 'com.google.code.findbugs:jsr305:3.0.2'
```
mobuchowski, line 54: Do we need this in runtime?

wslulciuc: I think we want to use `implementation`, instead of `api`? (see [API and implementation separation](https://docs.gradle.org/current/userguide/java_library_plugin.html))

mobuchowski: Yep, you're right.

mobuchowski:
Are we going to support arbitrary query params, as Spark HTTP client now does? https://github.com/OpenLineage/OpenLineage/pull/425

client/java/build.gradle
wslulciuc: That lib `com.google.code.findbugs` introduces the package `javax.annotation.*`. Meaning, `com.google.code.findbugs` is the reference implementation of the `JSR-305`.

wslulciuc:
> Are we going to support arbitrary query params, as Spark HTTP client now does? #425

Yeah, I think we should. Should the params be configured when creating the client, or provide it as an option when emitting an event?

```java
client.emit(event, queryParams)
```

mobuchowski:
@wslulciuc I think `client.emit(event)` should be an overload of `client.emit(event, queryParams)` that fills in default query params from environment.

collado-mike:
> Yeah, I think we should. Should the params be configured when creating the client, or provide it as an option when emitting an event?

Personally, I think setting the query params at construction time is the way to go. I don't imagine the query params changing from one `emit` call to another.

wslulciuc:
@mobuchowski / @collado-mike: I've documented the following the in the [`README.md`](https://github.com/OpenLineage/OpenLineage/blob/feature/http-client-java/client/java/README.md), but here's the approach I recommend we use to configure the client with query params appended on each HTTP request:

```java
URI uri = new URIBuilder("http://localhost:5000")
  .addParameter("param0", "value0")
  .addParameter("param1", "value2")
  .build();

OpenLineageClient client = Clients.newClient(uri.toURL());
```

client/java/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Installation
+
+Maven:
+
+```xml
+<dependency>
+    <groupId>io.openlineage</groupId>
+    <artifactId>openlineage-java</artifactId>
+    <version>0.5.1</version>
```
mobuchowski, line 13: 0.6 - should match what we have in different parts

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Support for dbt-spark adapter [@mobuchowski](https://github.com/mobuchowski)
 * **New** `backend` to proxy OpenLineage events to one or more event streams 🎉 [@mandy-chessell](https://github.com/mandy-chessell) [@wslulciuc](https://github.com/wslulciuc)
 * Add Spark extensibility API with support for custom Dataset and custom facet builders [@collado-mike](https://github.com/collado-mike)
+* Add HTTP client to `openlineage-java` lib [@wslulciuc](https://github.com/wslulciuc)
```
mobuchowski, line 14: Wrong version 😉

client/java/README.md
```diff
@@ -0,0 +1,69 @@
...
+</dependency>
+```
+
+or Gradle:
+
+```groovy
+implementation 'io.openlineage:openlineage-java:0.5.1'
```
mobuchowski, line 20: 0.6.0

client/java/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+## Usage
+
+```java
+// Connect to http://localhost:8080
+OpenLineageClient client = Clients.newClient();
+
+// Define a simple OpenLineage START or COMPLETE event
+OpenLineage.Run startOrCompleteRun = ...
```
mobuchowski, line 30: OpenLineage.RunEvent

client/java/README.md
```diff
@@ -0,0 +1,69 @@
...
+* `OPENLINEAGE_API_KEY`: the API key to be set on each HTTP request
+
+You can override the default configuration of the client via environment variables by specifying the URL and API key when
+creating a new client:
+
+```java
+String url = "http://localhost:5000";
+String apiKey = "f38d2189-c603-4b46-bdea-e573a3b5a7d5";
+
+OpenLineageClient client = Clients.newClient(url, apiKey);
```
mobuchowski, line 50: Won't work with `String` url, it needs to be `URI` to use `Clients.newClient`.

client/java/src/main/java/io/openlineage/client/OpenLineageHttp.java
```diff
@@ -0,0 +1,152 @@
...
+  static OpenLineageHttp create(@NonNull final URL url, @Nullable final String apiKey) {
+    final CloseableHttpClient http = HttpClientBuilder.create().build();
+    return new OpenLineageHttp(http, url, apiKey);
+  }
+
+  String post(@NonNull URL url, @NonNull String json) {
+    log.debug("POST {}: {}", url, json);
+    try {
+      final HttpPost request = new HttpPost();
+      request.setURI(url.toURI());
```
mobuchowski, line 70: We're converting URIs to URLs as params to  `Clients` or `OpenLineageClient`, then convert to URI again.
We should just use URI class everywhere I think.

Especially since URL is evil 😉 
https://blog.cyberborean.org/2007/08/22/urls-in-java-another-pitfall

mobuchowski (changes requested):
@wslulciuc overall looks good, few comments to review

CHANGELOG.md
wslulciuc: Ahh, good catch 👍

client/java/src/main/java/io/openlineage/client/Utils.java
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Adds the provided {@code apiKey} as an {@code Authorization} HTTP header to the specified
+   * {@link HttpRequestBase} object.
+   */
+  public static void addAuthTo(
+      @NonNull final HttpRequestBase request, @NonNull final String apiKey) {
+    request.addHeader(AUTHORIZATION, "Bearer " + apiKey);
+  }
```
collado-mike, line 89: I'd add a `TokenProvider` interface that the `HttpClient` can use that will allow users to supply tokens beyond just API keys. This would allow users to supply their own JWT provider that refreshed tokens on a periodic basis.

client/java/src/main/java/io/openlineage/client/OpenLineageClient.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  /**
+   * Emit the given run event to HTTP backend. The method will return successfully after the run
+   * event has been emitted, regardless of any exceptions thrown by the HTTP backend.
+   *
+   * @param runEvent The run event to emit.
+   */
+  public void emit(@NonNull OpenLineage.RunEvent runEvent) {
+    http.post(http.url("/lineage"), Utils.toJson(runEvent));
```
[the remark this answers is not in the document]
collado-mike: I'd leave the serialization to the underlying `Transport`, per the design doc on that interface

client/java/README.md
wslulciuc: I agree, but the version in the `README.md` should represent the current version of the java client. That said, I should update the version to `0.5.1` (which is the [current](https://github.com/OpenLineage/OpenLineage/releases/tag/0.5.2) published release)

client/java/README.md
wslulciuc: Ahh, right right 👍

client/java/src/main/java/io/openlineage/client/Clients.java
```diff
@@ -0,0 +1,27 @@
+package io.openlineage.client;
+
+import java.net.URL;
+import javax.annotation.Nullable;
+import lombok.NonNull;
+
+/** Factory class for creating new {@link OpenLineageClient} objects. */
+public final class Clients {
```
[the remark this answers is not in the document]
wslulciuc: Yeah, I removed the client factory methods in favor of using client builder; though I think in the future we could have:

```
Clients.newClient(ClientOptions)
```

client/java/src/main/java/io/openlineage/client/Utils.java
collado-mike: How we get the API key is an implementation detail that doesn't really impact the structure of the code that sets the API key. e.g., in your example of the client builder, the underlying code doesn't really care where the `apiKey` comes from
```java
OpenLineageClient client = OpenLineageClient.builder()
  .url("http://localhost:5000")
  .apiKey("f38d2189-c603-4b46-bdea-e573a3b5a7d5")
  .build();
```
The key can come from an environment variable or a yaml config or whatever. 

Underneath, though, we can construct a `TokenProvider` implementation, such as a `StaticTokenProvider`, that simply returns the same string. Then at some point, we can support an `OauthTokenProvider` or something that generates a refreshable token so the value in the `Authorization` header can be whatever the `TokenProvider` provides. It's quick to implement as part of the original client implementation and allows us to support a concrete use case we know we'll need to support in the future.

client/java/build.gradle
mobuchowski: Update: we actually want to use `compileOnly` and shade the libraries we're using.

`implementation` prevents compile-time dependency on the exposed library, but it still is required at runtime, and can conflict with other versions.

client/java/src/main/java/io/openlineage/client/transports/KafkaConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@NoArgsConstructor
+@ToString
+public final class KafkaConfig implements TransportConfig {
+  @Getter @Setter private String topicName;
+  @Getter @Setter private String localServerId;
+  @Getter @Setter private String bootstrapServerUrl;
```
pawel-big-lebowski, line 14: Is there an option to provide multiple bootstrap servers? 
It's really useful to have such. 
It can be comma separated list of bootstrap servers.

client/java/src/main/java/io/openlineage/client/transports/KafkaTransport.java
```diff
@@ -0,0 +1,35 @@
+package io.openlineage.client.transports;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.client.Utils;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerRecord;
```
pawel-big-lebowski, line 8: Readme mentions other Kafka options like `acks` or key/value serializers. Are they respected?

client/java/src/test/java/io/openlineage/client/OpenLineageTest.java
```diff
@@ -49,26 +47,27 @@ public void jsonSerialization() throws JsonProcessingException {
...
-    assertEquals(name,read.getJob().getName());
-    assertEquals(namespace,read.getJob().getNamespace());
-    assertEquals(runStateUpdate.getEventType(),read.getEventType());
-    assertEquals(runStateUpdate.getEventTime(),read.getEventTime());
+    assertEquals(producer, read.getProducer());
+    assertEquals(runId, read.getRun().getRunId());
+    assertEquals(name, read.getJob().getName());
+    assertEquals(namespace, read.getJob().getNamespace());
+    assertEquals(runStateUpdate.getEventType(), read.getEventType());
+    assertEquals(runStateUpdate.getEventTime(), read.getEventTime());
```
pawel-big-lebowski, line 70: Which piece of transport feature is this test trying to cover?

integration/spark/src/test/common/java/io/openlineage/spark/agent/lifecycle/MatchesMapRecursively.java
```diff
@@ -19,6 +20,7 @@
  *
  * @see AbstractObjectAssert#usingRecursiveComparison()
  */
+@Slf4j
 public class MatchesMapRecursively extends Condition<Map<String, Object>> {
```
pawel-big-lebowski, line 24: Is this class necessary?

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -0,0 +1,161 @@
...
+
+@Slf4j
+public final class HttpTransport extends Transport implements Closeable {
+  private static final String API_V1 = "/api/v1";
+
+  private final HttpClient http;
+  private final URL url;
+  private @Nullable final TokenProvider tokenProvider;
+
+  public HttpTransport(@NonNull final HttpConfig httpConfig) {
```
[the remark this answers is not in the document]
mobuchowski: Done

client/java/src/main/java/io/openlineage/client/transports/KafkaTransport.java
mobuchowski: Basically we only pass through properties for Kafka. Everything that Kafka client supports should be supported.

client/java/src/test/java/io/openlineage/client/OpenLineageTest.java
mobuchowski: This PR for some reason reformatted existing tests for OpenLineage event serialization.

pawel-big-lebowski (approved):

mobuchowski (approved):

Issue #484, opened 2022-01-19
[PROPOSAL][Spark] allow QueryPlanVisitors to know by which event they were triggered


mobuchowski:
@collado-mike 's refactor of new extension model allows facet builders to react on only select events, such as `SparkListener*` events, RDDs and more. 

However, QueryPlanVisitors do not have that possibility. They are defined strictly for some `LogicalPlan`. In some cases, like `DatasetVersionDatasetFacet`, this matters. There needs to be different behavior for input and output datasets - which is already provided - and different behavior for particular moment in run, there expressed by `SparkListenerSQLExecutionStart` and `SparkListenerSQLExecutionStart` events. 

For input datasets, we need to be able to find the versions at start. If we did this additionally on end, then the possibly different version would override the first, proper one. 

Opposite reason is for output dataset. The dataset version on start would be misleading - only the version that is the result of job execution matters. 

So, how do we can solve this? 

One way would be to change signature of `QueryPlanVisitor`'s `isDefinedAt` and `apply` to accept some "runtime context" - different from current, immutable `OpenLineageContext`. Problem with that is that we'd break the interface that we just released. 

Another way would be to have "higher" interface that wraps `QueryPlanVisitor` and accepts this runtime context. We could have "wrapping visitor" for visitors that do not care about this concept at all, that would just extract current `LogicalPlan` and pass it to actual visitor.

Yet another way would be to forget about this "runtime context" idea, and add "current event" field to `OpenLineageContext`. The problem with that approach is that `OpenLineageContext` currently is immutable entity. Hovewer, it already assumes some of the fields are mutable, like visitor lists. 

I believe last option is the best - no breaking compatibility, way less glue code, just one question - can we break something by having mutable field inside? Isn't SparkListener's single thread guarantee enough?

pawel-big-lebowski:
My first thought was to add `runContext` parameter to `apply` and `isDefinedAt` methods. We could provide within `QueryPlanVisitor` class  `default apply` and `default isDefinedAt`  methods that skips the extra 'runContext' parameter. This would give a possibility for some visitors to support `runContext`, while leaving other unaware of that. 

However, this approach is not doable without breaking the existing contracts as we rely now on `PartialFunction<LogicalPlan, List<Dataset>`. We would need change this into something like `PartialFunction<Tuple2<LogicalPlan, RunContext>, List<Dataset>` and this change should be spread all over the codebase including public builders and factories. 

[...]

PR #485, opened 2022-01-19
Add and implement (output) DatasetVersionDatasetFacet for Spark

mobuchowski:
This commit adds `DatasetVersionDatasetFacet` to OpenLineage spec.

It implements this facet for v2 writes in Apache Spark for output datasets, where provider is Iceberg or Delta.
In both cases, we use their implementation of `TableCatalog` to get their internal dataset versions. 

The implementation utilizes new facet builder API for output dataset facets. This allows the main class of this commit: `DatasetVersionOutputDatasetFacetBuilder` to listen only on `SparkListenerSQLExecutionEnd` event, which is emitted only when writing to dataset, both data and metadata is complete. Other events, like `SparkListenerJobEnd` do not have this property. 

This PR will someday be followed with implementation of this facet for input datasets. The main blocker is that we can't use current structure of `CustomFacetBuilder` that attaches resulting facet to any dataset. There can be multiple input datasets, and they can have different versions, or have no explicit version. Issue is described here: https://github.com/OpenLineage/OpenLineage/issues/484

### DIFF
--- a/integration/spark/integrations/container/pysparkWriteIcebergTableVersionEnd.json
+++ b/integration/spark/integrations/container/pysparkWriteIcebergTableVersionEnd.json
@@ -1,0 +1,29 @@
+{
+  "eventType" : "COMPLETE",
+  "job" : {
+    "namespace" : "testWriteIcebergTableVersion",
+    "name" : "open_lineage_integration_iceberg.append_data"
+  },
+  "outputs" : [ {
+    "namespace" : "file",
+    "name" : "/tmp/write_iceberg_table_version/db.table",
+    "facets" : {
+      "schema" : {
+        "fields" : [ {
+          "name" : "a",
+          "type" : "integer"
+        }, {
+          "name" : "b",
+          "type" : "integer"
+        } ]
+      },
+      "tableProvider" : {
+        "provider" : "iceberg",
+        "format" : "parquet"
+      }
+    },
+    "outputFacets" : {
+      "datasetVersion" : {}
+    }
+  } ]
+}
--- a/integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/facets/builder/DatasetVersionDatasetFacetUtils.java
+++ b/integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/facets/builder/DatasetVersionDatasetFacetUtils.java
@@ -1,0 +1,60 @@
+package io.openlineage.spark3.agent.facets.builder;
+
+import static io.openlineage.spark.agent.util.ScalaConversionUtils.asJavaOptional;
+
+import io.openlineage.spark3.agent.lifecycle.plan.catalog.CatalogUtils3;
+import java.util.Map;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+import org.apache.spark.sql.delta.files.TahoeLogFileIndex;
+import org.apache.spark.sql.execution.datasources.HadoopFsRelation;
+import org.apache.spark.sql.execution.datasources.LogicalRelation;
+import org.apache.spark.sql.execution.datasources.v2.DataSourceV2Relation;
+
+@Slf4j
+class DatasetVersionDatasetFacetUtils {
+
+  /**
+   * Check if we have all the necessary properties in DataSourceV2Relation to get dataset version.
+   */
+  public static Optional<String> extractVersionFromDataSourceV2Relation(
+      DataSourceV2Relation table) {
+    if (table.identifier().isEmpty()) {
+      log.warn("Couldn't find identifier for dataset in plan " + table);
+      return Optional.empty();
+    }
+    Identifier identifier = table.identifier().get();
+
+    if (table.catalog().isEmpty() || !(table.catalog().get() instanceof TableCatalog)) {
+      log.warn("Couldn't find catalog for dataset in plan " + table);
+      return Optional.empty();
+    }
+    TableCatalog tableCatalog = (TableCatalog) table.catalog().get();
+
+    Map<String, String> tableProperties = table.table().properties();
+    return CatalogUtils3.getDatasetVersion(tableCatalog, identifier, tableProperties);
+  }
+
+  /**
+   * Delta uses LogicalRelation's HadoopFsRelation as a logical plan's leaf. It implements FileIndex
+   * using TahoeLogFileIndex that contains DeltaLog, which can be used to get dataset's snapshot.
+   */
+  public static Optional<String> extractVersionFromLogicalRelation(
+      LogicalRelation logicalRelation) {
+    if (logicalRelation.relation() instanceof HadoopFsRelation) {
+      HadoopFsRelation fsRelation = (HadoopFsRelation) logicalRelation.relation();
+      asJavaOptional(logicalRelation.catalogTable());
+      if (logicalRelation.catalogTable().isDefined()
+          && logicalRelation.catalogTable().get().provider().isDefined()
+          && logicalRelation.catalogTable().get().provider().get().equalsIgnoreCase("delta")) {
+        if (fsRelation.location() instanceof TahoeLogFileIndex) {
+          TahoeLogFileIndex fileIndex = (TahoeLogFileIndex) fsRelation.location();
+          return Optional.of(Long.toString(fileIndex.getSnapshot().version()));
+        }
+      }
+    }
+    return Optional.empty();
+  }
+}
--- a/integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/Spark3EventHandlerFactory.java
+++ b/integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/Spark3EventHandlerFactory.java
@@ -1,0 +1,57 @@
+package io.openlineage.spark3.agent.lifecycle.plan;
+
+import com.google.common.collect.ImmutableList;
+import io.openlineage.client.OpenLineage;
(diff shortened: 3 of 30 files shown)


integration/spark/build.gradle
```diff
@@ -191,7 +191,7 @@ sourceSets {
     spark3Test {
         java {
             srcDirs = ["src/test/common/java", "src/test/spark3/java"]
-            compileClasspath = configurations.spark3Test + sourceSets.main.output
+            compileClasspath = configurations.spark3Test + sourceSets.main.output + sourceSets.spark3.output
```
mobuchowski, line 194: This fixes code completion for Spark 3 tests in IntelliJ. I can make separate PR with that if needed.

integration/spark/integrations/container/pysparkWriteIcebergTableVersionEnd.json
```diff
@@ -0,0 +1,29 @@
...
+          "type" : "integer"
+        } ]
+      },
+      "tableProvider" : {
+        "provider" : "iceberg",
+        "format" : "parquet"
+      }
+    },
+    "outputFacets" : {
+      "datasetVersion" : {}
```
mobuchowski, line 26: Iceberg's versions are apparently random longs - just checking if we have the facet here, since I can't deterministically tell which one it will be.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/facets/builder/DatasetVersionDatasetFacetUtils.java
```diff
@@ -0,0 +1,60 @@
...
+   * using TahoeLogFileIndex that contains DeltaLog, which can be used to get dataset's snapshot.
+   */
+  public static Optional<String> extractVersionFromLogicalRelation(
+      LogicalRelation logicalRelation) {
+    if (logicalRelation.relation() instanceof HadoopFsRelation) {
+      HadoopFsRelation fsRelation = (HadoopFsRelation) logicalRelation.relation();
+      asJavaOptional(logicalRelation.catalogTable());
+      if (logicalRelation.catalogTable().isDefined()
+          && logicalRelation.catalogTable().get().provider().isDefined()
+          && logicalRelation.catalogTable().get().provider().get().equalsIgnoreCase("delta")) {
```
pawel-big-lebowski, line 51: Is it possible to move this logic into handler classes? So far we managed to avoid any provider specific logic outside provider handlers.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
```diff
@@ -81,14 +88,37 @@ public static void includeProviderFacet(
...
-      return Collections.emptyList();
+    return datasetIdentifier
+        .map(
+            value ->
+                Collections.singletonList(
+                    datasetFactory.getDataset(value, relation.schema(), facets)))
+        .orElse(Collections.emptyList());
+  }
+
+  public static Optional<DataSourceV2Relation> getDataSourceV2Relation(LogicalPlan x) {
```
pawel-big-lebowski, line 103: When moving code here, consider moving unit tests to another class.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/Spark3EventHandlerFactory.java
```diff
@@ -0,0 +1,57 @@
...
+import io.openlineage.spark3.agent.facets.builder.DatasetVersionOutputDatasetFacetBuilder;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.BiConsumer;
+import lombok.AllArgsConstructor;
+import org.apache.spark.package$;
+
+public class Spark3EventHandlerFactory implements OpenLineageEventHandlerFactory {
+  @Override
+  public List<CustomFacetBuilder<?, ? extends OutputDatasetFacet>> createOutputDatasetFacetBuilders(
```
pawel-big-lebowski, line 20: Are there any tests for this class?

integration/spark/src/test/common/java/io/openlineage/spark/agent/SparkContainerIntegrationTest.java
```diff
@@ -333,6 +333,30 @@ public void testOptimizedCreateAsSelectAndLoad() {
     verifyEvents("pysparkOCTASStart.json", "pysparkOCTASEnd.json");
   }
 
+  @Test
+  @EnabledIfSystemProperty(named = "spark.version", matches = SPARK_3) // Spark version >= 3.*
```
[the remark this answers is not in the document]
mobuchowski: This can be v2 test, but for `input` Delta uses LogicalRelation/HadoopFsRelation with v1 CatalogTable.

After input is done, this will need to be a separate, non-v2 test as well.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/facets/builder/DatasetVersionDatasetFacetUtils.java
mobuchowski: Would be done this way if this was v2 catalog 😉

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/Spark3EventHandlerFactory.java
mobuchowski: [1) We have test that checks if service provider interface classes are properly loaded](https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/src/test/common/java/io/openlineage/spark/agent/lifecycle/InternalEventHandlerFactoryTest.java)
2) Integration tests check that we properly import and use this one by having proper output dataset facet there. 

Do you think we need unit test for that? I personally don't really know how that test would look like, as this class is useless without being included in OpenLineageRunEventBuilder.

DatasetVersionFacetConverter could be tested separately though...

collado-mike:
> There can be multiple input datasets, and they can have different versions, or have no explicit version. Issue is described here:

My thinking here was that the facet builder could use the input `Stage`'s `RDD` to trigger when the facet should be generated for an input dataset.

integration/spark/src/test/resources/spark_scripts/spark_write_delta_table_version.py
```diff
@@ -0,0 +1,15 @@
...
+spark = SparkSession.builder \
+    .master("local") \
+    .appName("Open Lineage Integration Delta") \
+    .config("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog") \
+    .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension") \
+    .config("packages", "io.delta:delta-core_2.12:1.0.0") \
+    .getOrCreate()
+
+spark.sql("CREATE TABLE versioned_table (a int, b int) USING delta LOCATION '/tmp/write_delta_table_version'")
+spark.sql("INSERT INTO versioned_table VALUES (1, 2)")
```
collado-mike, line 15: I think it's important to add some input datasets here to validate that the `datasetVersion` is not attached to those datasets.

collado-mike (commented):
This is cool. I'm a bit concerned about the input dataset handling - it would be great to add test coverage to verify everything is working correctly.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/Spark3EventHandlerFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    if (package$.MODULE$.SPARK_VERSION().startsWith("2")) {
+      return Collections.emptyList();
+    }
+    return ImmutableList.of(
+        new DatasetVersionFacetConverter<>(
+            context, new DatasetVersionOutputDatasetFacetBuilder(context)));
+  }
+
+  @AllArgsConstructor
+  private static class DatasetVersionFacetConverter<T>
```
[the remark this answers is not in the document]
mobuchowski: Because I wanted to 

1.  have `DatasetVersionDatasetFacet` as a core facet that's attachable to both input and output datasets. If we did it the other way, I'd have to have two exactly same facets in core spec. It could be solved by making `OutputDatasetFacet` and `InputDatasetFacet` extend `DatasetFacet`.
2. Original version of this PR had also `DatasetVersionInputDatasetFacetBuilder` that utilized the same converter feature.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
mobuchowski: Unit tests in `TableContentChangeVisitorTest` still mostly test `TableContentChangeVisitor` I think. Unless you want to test `PlanUtils3.getDataSourceV2Relation` directly, but that I think doesn't make a lot of sense, this is just `instanceof` matching and test would only test ability to correctly write mock corresponding to mocked class 😉

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/Spark3EventHandlerFactory.java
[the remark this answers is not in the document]
collado-mike: I submitted [this PR](https://github.com/OpenLineage/OpenLineage/pull/513), which I think will help. If we can make the `DatasetVersionDatasetFacet` implement both the `InputDatasetFacet` and the `OutputDatasetFacet`, I think we can simplify this code so that only a single facet builder is needed.

Issue #500, opened 2022-01-28
[BUG] SnowflakeOperator no longer reporting Lineage in Airflow 2.2


conorbev:
In recent Airflow versions (certainly in `2.2`, but I think possibly `2.1.4` also) if you use the `SnowflakeOperator` with `OpenLineageBackend` `0.5.1` you'll get an exception in the log like:

```
ERROR - Failed to extract metadata 'SnowflakeOperator' object has no attribute 'get_hook'
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/openlineage/lineage_backend/__init__.py", line 83, in _extract_metadata
    task_metadata = self._extract(extractor, task_instance)
  File "/usr/local/lib/python3.9/site-packages/openlineage/lineage_backend/__init__.py", line 104, in _extract
    task_metadata = extractor.extract_on_complete(task_instance)
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/extractors/base.py", line 61, in extract_on_complete
    return self.extract()
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/extractors/postgres_extractor.py", line 65, in extract
    authority=self._get_authority(),
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/extractors/snowflake_extractor.py", line 38, in _get_authority
    return self.operator.get_hook()._get_conn_params()['account']
AttributeError: 'SnowflakeOperator' object has no attribute 'get_hook'
```

This [PR](https://github.com/apache/airflow/pull/17741) changed the method from `get_hook()` to `get_db_hook()` so I believe OpenLineage will have to check for the presence of either method to work across Airflow versions.

wslulciuc:
Assigned to [@denimalpaca](https://github.com/denimalpaca)

PR #505, opened 2022-01-31
[INTEGRATION][AIRFLOW] Fix breaking changes from OSS Airflow

denimalpaca:
### Problem

A few months back, the SnowflakeOperator was changed (...by me)
to use `get_db_hook()` instead of `get_hook` in order to make it
inheritable by BaseSQLOperator. This allowed SnowflakeCheckOperators
to inherit from BaseSQLOperator and simplify backend connections. This
change broke the extractor in OpenLineage, and this PR aims to fix it.

Closes: #500 

### Solution

Please describe your change as it relates to the problem, or bug fix, as well as any dependencies. If your change requires a schema change, please describe the schema modifcation(s) and whether it's a _backwards-incompatible_ or _backwards-compatible_ change, then select one of the following:

The solution is to change instances of `get_hook()` in the Snowflake Extractor to `get_db_hook()`.

### Checklist

### DIFF
--- a/integration/airflow/openlineage/airflow/extractors/snowflake_extractor.py
+++ b/integration/airflow/openlineage/airflow/extractors/snowflake_extractor.py
@@ -32,13 +32,13 @@ class SnowflakeExtractor(PostgresExtractor):
         return 'snowflake'
 
     def _get_database(self) -> str:
-        return self.operator.get_hook()._get_conn_params()['database']
+        return self.operator.get_db_hook()._get_conn_params()['database']
 
     def _get_authority(self) -> str:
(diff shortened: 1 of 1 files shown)


denimalpaca:
I read the contributing page after creating this, so I'll make changes to things like the branch name tomorrow.

mobuchowski (changes requested):
I don't think we can do that as simply as we're trying here, since we still want to be compatible with older Airflow versions. We'd need to check whether `get_db_hook` or `get_hook` is available on that operator (probably just using `hasattr`) and call the one that's available.

Issue #488, opened 2022-01-20
Add AirflowPlugin utilizing TaskListener API


mobuchowski:
Use https://github.com/apache/airflow/pull/20443 API to create plugin that receives task instance started, completed and failed information.

PR #508, opened 2022-02-01
Airflow: integrate using TaskInstance listener interface

mobuchowski:
This PR implements getting lineage from Airflow using TaskInstance listener interface added to Airflow in upcoming 2.3.0 version: https://github.com/apache/airflow/pull/20443

We provide entrypoint-registered `OpenLineagePlugin` that provides listener module that airflow imports, and notifies it when TaskInstance state changes to `RUNNING`, `COMPLETE` or `FAILED`. 

To have consistent support three different methods of integration, extraction of datasets was abstracted from particular integration methods. Also, OpenLineageBackend is effectively no-op in Airflow 2.3.0+ - having it only duplicates potential events.

In addition to implementing this API, added matrix integration test support that runs on different supported Airflow versions. Tests are running on Airflow 2.1.3, Airflow 2.2.3 and on custom Airflow 2.3.0.dev0 build from Airflow main - to test new integration. This feature created few problems - tasks were concurrently operating on the same bigquery resources, so needed to add airflow version prefixing to table names that are used in integration tests.   

Closes: https://github.com/OpenLineage/OpenLineage/issues/488


.circleci/config.yml
```diff
@@ -307,7 +307,7 @@ jobs:
             - ./dist/*.whl
             - ./dist/*.tar.gz
 
-  integration-test-integration-airflow:
+  integration-test-integration-airflow-1:
```
wslulciuc, line 310: It'd be great if we can use the following CI job names to easily know which version of Airflow we were running integration tests for:

```
integration-test-integration-airflow-1.x
integration-test-integration-airflow-2.x
```

Note sure of these names are valid though ;)

.circleci/config.yml
```diff
@@ -453,7 +456,10 @@ workflows:
           filters:
             branches:
               ignore: /pull\/[0-9]+/
-      - integration-test-integration-airflow-2:
+      - integration-test-integration-airflow:
+          matrix:
+            parameters:
+              airflow-image: ['apache/airflow:2.1.3', 'apache/airflow:2.2.3', 'gcr.io/speedy-vim-308516/apache/airflow:latest']
```
wslulciuc, line 462: Nice 💯

wslulciuc: Does `gcr.io/speedy-vim-308516/apache/airflow` test Airflow on GCP?

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased](https://github.com/OpenLineage/OpenLineage/compare/0.5.1...HEAD)
+### Added
+* Airflow integration uses new TaskInstance listener API for Airflow 2.3+ [@mobuchowski](https://github.com/mobuchowski)
```
wslulciuc, line 5: Can we link out to any helpful Airflow docs in the changelog on the `TaskInstance` listener?

integration/airflow/README.md
```diff
@@ -41,25 +41,14 @@ To install from source, run:
...
-
-1. On DAG **start**, collect metadata for each task using an `Extractor` if it exists for given operator.
-2. Collect task input / output metadata (`source`, `schema`, etc)
-3. Collect task run-level metadata (execution time, state, parameters, etc)
-4. On DAG **complete**, also mark the task as _complete_ in OpenLineage
+Integration automatically registers itself for Airflow 2.3 if it's installed on Airflow worker's python.
+This means you don't have to do anything besides configuring it, which is described in Configuration section.
 
-### Airflow 2.1+ (*experimental*)
+### Airflow 2.1 - 2.2
```
wslulciuc, line 51: _Minor_: We'll eventually want to discourage users from using the `LineageBackend` as it's not actually supported and has limitations.

integration/airflow/openlineage/airflow/extractors/manager.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class ExtractorManager:
+    """Class abstracting management of custom extractors."""
+    def __init__(self):
+        self.extractors = {}
+        self.extractor_mapper = Extractors()
+        self.log = logging.getLogger()
+
+    def extract_metadata(self, dagrun, task, complete: bool, task_instance=None) -> TaskMetadata:
```
wslulciuc, line 15: _Minor_: Would it be sensible to default complete to `False`?

integration/airflow/openlineage/airflow/extractors/manager.py
```diff
@@ -0,0 +1,56 @@
...
+
+from openlineage.airflow.extractors import TaskMetadata, BaseExtractor, Extractors
+from openlineage.airflow.utils import get_job_name
+
+
+class ExtractorManager:
+    """Class abstracting management of custom extractors."""
+    def __init__(self):
+        self.extractors = {}
+        self.extractor_mapper = Extractors()
```
wslulciuc, line 12: _Minor_: To be a bit more clear on the mapping, I'd use `task_to_extractor` instead of `extractor_mapper`

integration/airflow/openlineage/airflow/dag.py
```diff
@@ -120,7 +117,7 @@ def __init__(self, *args, **kwargs):
         kwargs["user_defined_macros"] = macros
         if kwargs.__contains__("lineage_custom_extractors"):
             for operator, extractor in kwargs['lineage_custom_extractors'].items():
-                extractor_mapper.add_extractor(operator, extractor)
+                extractor_manager.extractor_mapper.add_extractor(operator, extractor)
```
wslulciuc, line 120: I would encapsulate adding extractors within `ExtractorManager`:

```
extractor_manager.add_extractor()
```

That is, extractor map should be internal to and managed by `ExtractorManager`

wslulciuc: Well, maybe not. Looks like that's a pattern used throughout. I would have expected the `ExtractorManager` to define an interface for interacting with `extractors` and the underlying `extractor_mapper`

integration/airflow/openlineage/airflow/extractors/manager.py
```diff
@@ -0,0 +1,56 @@
...
+        self.log = logging.getLogger()
+
+    def extract_metadata(self, dagrun, task, complete: bool, task_instance=None) -> TaskMetadata:
+        extractor = self._get_extractor(task)
+        task_info = f'task_type={task.__class__.__name__} ' \
+            f'airflow_dag_id={task.dag_id} ' \
+            f'task_id={task.task_id} ' \
+            f'airflow_run_id={dagrun.run_id} '
+        if extractor:
```
wslulciuc, line 21: _Minor_: I would briefly document the default behavior if an extractor is not defined for a task / operator.

integration/airflow/openlineage/lineage_backend/__init__.py
```diff
@@ -70,72 +68,6 @@ def send_lineage(
             task=task_metadata
         )
 
-    def _extract_metadata(self, dag_id, dagrun, task, task_instance=None):
```
wslulciuc, line 73: Nice to see so much _red_ 🥇

integration/airflow/openlineage/lineage_backend/__init__.py
```diff
@@ -148,6 +80,9 @@ class OpenLineageBackend(LineageBackend):
 
     @classmethod
     def send_lineage(cls, *args, **kwargs):
+        # Do not use LineageBackend approach when we can use plugins
```
wslulciuc, line 83: I think this make sense, but is this a reasonable expectation for the user? That is, if I as a user configure the `LineageBackend`, but the plugin get's configured without me knowing, that might be confusing.

Can we log a warning instead? I.e. recommending users to use the plugins based on the version of Airflow they are running?

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
...
+from openlineage.airflow.extractors import ExtractorManager
+from openlineage.airflow.utils import DagUtils, get_task_location, get_job_name, get_custom_facets
+
+if TYPE_CHECKING:
+    from airflow.models import TaskInstance, BaseOperator
+    from sqlalchemy.orm import Session
+
+
+@attr.s(frozen=True)
+class RunData:
```
wslulciuc, line 20: `RunData` contains data specific to an active run, so I'd use `ActiveRun` instead

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
...
+    run_id: str = attr.ib()
+    task: "BaseOperator" = attr.ib()
+
+
+class RunDataHolder:
```
wslulciuc, line 25: To be consistent with `ExtractorManager` I would name this class `ActiveRunManager` or `ActiveTaskRunManager` with the interface:

```
add_active_run()
get_active_run()
```

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
...
+    """Class that stores run data - run_id and task in-memory. This is needed because Airflow
+    does not always pass all runtime info to on_task_instance_success and
+    on_task_instance_failed that is needed to emit events. This is not a big problem since
+    we're only running on worker - in separate process that is always spawned (or forked) on
```
wslulciuc, line 29: With the details for an active task run being transient, are we concerned about workers restarting?

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
...
+        return ti.dag_id + ti.task_id + ti.run_id
+
+
+run_data_holder = RunDataHolder()
+extractor_manager = ExtractorManager()
+adapter = OpenLineageAdapter()
+log = logging.getLogger('airflow')
+
+
+def run_thread(target: Callable, kwargs=None):
```
wslulciuc, line 54: _Minor_: I'd use similar naming as underlying method you're calling `start_thread()` (though semantically the same thing)

integration/airflow/openlineage/airflow/extractors/manager.py
```diff
@@ -0,0 +1,56 @@
+import logging
+from typing import Optional
+
+from openlineage.airflow.extractors import TaskMetadata, BaseExtractor, Extractors
+from openlineage.airflow.utils import get_job_name
+
+
+class ExtractorManager:
```
wslulciuc, line 8: Do we have a test for this class?

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
+import logging
```
wslulciuc, line 1: Do we have unit tests for this `listener`?

integration/airflow/openlineage/lineage_backend/__init__.py
mobuchowski: Plugin's functionality is a superset of what LineageBackend can do, so I don't really see use case for that 🙂 

We're not turning plugin off when LineageBackend is configured, so this would result in duplicate events emitted - which I had for a bit while testing.

integration/airflow/openlineage/airflow/dag.py
mobuchowski: Ok. Here I was just 'extracting' common functionality between three integrations, but I guess slight refactor won't hurt.

integration/airflow/README.md
mobuchowski: We want to, but it's the only stuff that's working below 2.3. Maybe add some additional disclaimer?

.circleci/config.yml
mobuchowski: No, it just uses Google Container Registry as a source of Airflow 2.3's docker image.

.circleci/config.yml
mobuchowski: `integration-test-integration-airflow-1.10` - ok.

For Airflow 2's tests, the image used is appended to the end of job name - it looks like that: 
https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage?branch=airflow%2Ftasklistener&filter=all

integration/airflow/openlineage/airflow/listener.py
mobuchowski: Depends on the failure mode. If we fail inside operator, it will get caught by changing state to `FAILED`. 

If we crash - for example due to OOM, then worker will restart, and another job will start - and change state to `RUNNING` again. If this succeeds, we will capture complete data about this run.

If we wanted to track retries, then we'd need to listen to events on scheduler, because particular worker has no option to get this information. Tracking lifecycle of task instance closer would require additional work on Airflow side.

integration/airflow/openlineage/airflow/listener.py
mobuchowski: It's not a class 😉 

Also, it's hard to unit test this until we can run tests on released 2.3. 
Contrary to `ExtractorManager` it assumes things about Airflow and is not imported below 2.3.

integration/airflow/openlineage/airflow/listener.py
mobuchowski: 👍

.circleci/config.yml
mobuchowski: We can't use dots in job names:
```
**ERROR IN CONFIG FILE:[#/jobs/integration-test-integration-airflow-1.10] string [integration-test-integration-airflow-1.10] does not match pattern ^[A-Za-z][A-Za-z\s\d_-]*$**
```

CHANGELOG.md
mobuchowski: Fixed.

integration/airflow/openlineage/airflow/extractors/manager.py
mobuchowski: Done.

integration/airflow/openlineage/airflow/dag.py
mobuchowski: Done.

integration/airflow/openlineage/airflow/extractors/manager.py
mobuchowski: Done.

integration/airflow/README.md
wslulciuc: Great, thanks for adding the disclaimer 👍

integration/airflow/openlineage/airflow/listener.py
wslulciuc: We don't have a `RETRY` run state, though I think we should eventually support tracking job retries. I think capturing the `FAIL` run state and setting the job to `RUNNING` again make sense and is good enough for now.

.circleci/config.yml
wslulciuc: I think appending the image tag to the job name is better than simply having `-2` appended to the job name. The only think now is we get job names like:

```
integration-test-integration-airflow-gcr.io/speedy-vim-308516/apache/airflow:latest
integration-test-integration-airflow-apache/airflow:2.2.3
...
```

which are long, but also descriptive so that isn't necessarily a bad thing.

.circleci/config.yml
wslulciuc: Shouldn't the tag be `2.3.x` instead of `latest` then?

integration/airflow/openlineage/airflow/extractors/manager.py
wslulciuc: Thanks!

integration/airflow/openlineage/lineage_backend/__init__.py
wslulciuc: Makes sense, but should we still log a `info` or `warn` message to the user that the plugin will be used to emit events instead of the backend?

That is:
```
if parse_version(AIRFLOW_VERSION) >= parse_version("2.3.0.dev0"):
            return
```

returns on Airflow `>=` `2.3.0`, so I would still let the user know that the plugin will be used instead and maybe recommend that the user disable the backend because it's beed deprecated.

integration/airflow/tests/extractors/test_extractor_manager.py
```diff
@@ -0,0 +1,59 @@
+from typing import Any, Optional, List
```
wslulciuc, line 1: 💯

wslulciuc:
Amazing work, @mobuchowski! To merge, this PR depends on Airflow `2.2.3` which has been delayed. The new date is sometime in March.

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
...
+        self.run_data = {}
+
+    def set_active_run(self, task_instance: "TaskInstance", run_id: str):
+        self.run_data[self._pk(task_instance)] = ActiveRun(run_id, task_instance.task)
+
+    def get_active_run(self, task_instance: "TaskInstance") -> Optional[ActiveRun]:
+        pk = self._pk(task_instance)
+        if pk not in self.run_data:
+            return None
+        return self.run_data[pk]
```
kaxil, line 42:
```suggestion
        return self.run_data.get(pk)
```

This will return `None` if the object doesn't exist in the dict

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -0,0 +1,141 @@
...
+    # Join, but ignore checking if thread stopped. If it did, then we shoudn't do anything.
+    # This basically gives this thread 2 seconds to complete work, then it can be killed,
+    # as daemon=True. We don't want to deadlock Airflow if our code hangs.
+    thread.join(timeout=2)
+
+
+@hookimpl
+def on_task_instance_running(previous_state, task_instance: "TaskInstance", session: "Session"):
+    if not hasattr(task_instance, 'task'):
+        log.warning(f"No task set for TI object {id(task_instance)}")
```
kaxil, line 72: Wouldn't it be better to log `TI.task_id`, `TI.dag_id` and `TI.run_id` instead of  `id(task_instance)` ?

integration/airflow/openlineage/airflow/macros.py
```diff
@@ -0,0 +1,72 @@
...
+        task_id='render_template',
+        python_callable=my_task_function,
+        op_args=['{{ lineage_run_id(run_id, task) }}'], # lineage_run_id macro invoked
+        provide_context=False,
+        dag=dag
+    )
+
+    :param run_id:
+    :param task:
+    :return:
```
kaxil, line 31: Worth adding descriptions to these or removing them completely

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased](https://github.com/OpenLineage/OpenLineage/compare/0.6.2...HEAD)
 ### Added
+* Airflow integration uses [new TaskInstance listener API](https://github.com/apache/airflow/blob/main/docs/apache-airflow/listeners.rst) for Airflow 2.3+ [@mobuchowski](https://github.com/mobuchowski)
```
kaxil, line 5: The following doc section might need updating too

https://github.com/OpenLineage/OpenLineage/blob/f0b8b5ce30c5f82971e25674142e54ae16ac312e/integration/airflow/docs/task-lifecycle.md#openlineage-interaction

mobuchowski: Done.

integration/airflow/openlineage/airflow/listener.py
mobuchowski: Fixed.

integration/airflow/openlineage/airflow/macros.py
mobuchowski: Removed them.

integration/airflow/openlineage/airflow/listener.py
mobuchowski: Done.

pawel-big-lebowski (approved):

Bowrna:
Hello 
I am new to this airflow feature. Could you tell me how this `task listener` interface is different from `task_callback` feature supported?

PR #530, opened 2022-02-10
Http and Kafka transports for python client

mobuchowski:
This is draft implementation of Transport interface for Python proposed in #269 
Also, added very basic KafkaTransport in addition to HttpTransport.

### Checklist

- [x] You've updated the [`CHANGELOG.md`](https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md) with details about your change under the "Unreleased" section (_if relevant, depending on the change, this may not be necessary_)

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -32,7 +32,7 @@ class OpenLineageClient:
         transport: Optional[Transport] = None,
     ):
         if url:
-            # Backwards compatibility: if URL is set, use old path to initialize
+            # Backwards compatibility: if URL or options is set, use old path to initialize
             # HTTP transport.
             if not options:
                 options = OpenLineageClientOptions()
@@ -50,13 +50,10 @@ class OpenLineageClient:
         options: OpenLineageClientOptions,
         session: Session
     ):
-        self.transport = HttpTransport(HttpConfig(
+        self.transport = HttpTransport(HttpConfig.from_options(
             url=url,
-            timeout=options.timeout,
-            verify=options.verify,
-            api_key=options.api_key,
-            session=session,
-            adapter=options.adapter
+            options=options,
+            session=session
         ))
 
     def emit(self, event: RunEvent):
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -1,17 +1,15 @@
 # SPDX-License-Identifier: Apache-2.0.
 
-import os
 import logging
-from urllib.parse import urljoin
+from typing import Optional
 
 import attr
 from requests import Session
 from requests.adapters import HTTPAdapter
-from urllib3.util.url import parse_url
 
-from openlineage.client import constants
 from openlineage.client.run import RunEvent
-from openlineage.client.serde import Serde
+from openlineage.client.transport import Transport, get_default_factory
+from openlineage.client.transport.http import HttpTransport, HttpConfig
 
 
 @attr.s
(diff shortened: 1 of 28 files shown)


client/python/openlineage/client/client.py
```diff
@@ -27,55 +25,48 @@ class OpenLineageClientOptions:
...
+        options: OpenLineageClientOptions,
+        session: Session
+    ):
+        self.transport = HttpTransport(HttpConfig(
+            url=url,
+            timeout=options.timeout,
+            verify=options.verify,
+            api_key=options.api_key,
+            session=session,
+            adapter=options.adapter
```
collado-mike, line 59: I'd leave the options parsing to the `HttpConfig`, rather than passing specific keys here. Allow the `HttpConfig` to support new features without having to change the client code here.

client/python/openlineage/client/transport/http.py
```diff
@@ -0,0 +1,83 @@
...
+log = logging.getLogger(__name__)
+
+
+@attr.s
+class HttpConfig(Config):
+    url: str = attr.ib()
+    timeout: float = attr.ib(default=5.0)
+    # check TLS certificates
+    verify: bool = attr.ib(default=True)
+    api_key: Optional[str] = attr.ib(default=None)
```
collado-mike, line 26: Rather than an api key directly, let's add a `TokenProvider` type. This enables us to add some oauth or other token generator easily in the future. It is also an extension point for users who might want to add their own implementation so they're not blocked on our timeline.

collado-mike, line 26: Rather than an api key directly, let's add a `TokenProvider` type. This enables us to add some oauth or other token generator easily in the future.

client/python/openlineage/client/transport/factory.py
```diff
@@ -0,0 +1,102 @@
...
+                    config = yaml.safe_load(f)
+                    return config['transport']
+            except Exception:
+                # Just move to read env vars
+                pass
+        return
+
+    @staticmethod
+    def _find_yaml() -> Optional[str]:
+        # Check current working directory:
```
collado-mike, line 72: I'd add another option to allow the user to specify via environment variable. This saves users the hassle of figuring out what the CWD is for, .e.g, a running airflow docker image. They can just drop the config file somewhere and explicitly tell us where it is.

collado-mike, line 72: I'd add another option to allow the user to specify via environment variable. This saves users the hassle of figuring out what the CWD is for, .e.g, a running airflow docker image. They can just drop the config file somewhere and explicitly tell us where it is.

client/python/openlineage/client/transport/transport.py
```diff
@@ -0,0 +1,20 @@
+from openlineage.client.run import RunEvent
+
+
+class Config:
+    @classmethod
+    def from_dict(cls, params: dict):
+        return cls()
+
+
+class Transport:
```
collado-mike, line 10: Add some doc comments. Pointers to the factory and existing impls would be helpful

collado-mike, line 10: Add some doc comments. Pointers to the factory and existing impls would be helpful

collado-mike (changes requested):
This is great! A couple of minor changes for extensibility and ease of use, but I think it looks good!

client/python/openlineage/client/client.py
wslulciuc: In the `openlineage-java` lib, we take the approach you're suggestion here @collado-mike. That is, we instantiate the [`HttpConfig`](https://github.com/OpenLineage/OpenLineage/blob/feature/http-client-java/client/java/src/main/java/io/openlineage/client/transports/HttpConfig.java) object based on what transport is enabled in [`openlineage.yml`](https://github.com/OpenLineage/OpenLineage/tree/feature/http-client-java/client/java#configuration) file

client/python/openlineage/client/client.py
```diff
@@ -27,55 +25,48 @@ class OpenLineageClientOptions:
...
-        return resp
-
-    def _add_auth(self, api_key: str):
-        self.session.headers.update({
-            "Authorization": f"Bearer {api_key}"
-        })
+        if not isinstance(event, RunEvent):
+            raise ValueError("`emit` only accepts RunEvent class")
+        if not self.transport:
+            log.error("Tried to emit OpenLineage event, but transport is not configured.")
```
wslulciuc, line 66: I think we should default the transport to `console`, rather than throw an error and document the behavior. In the `openlineage-java` lib, we default to [`ConsoleTransport `](https://github.com/OpenLineage/OpenLineage/blob/feature/http-client-java/client/java/src/main/java/io/openlineage/client/OpenLineageClient.java#L52)

client/python/openlineage/client/transport/factory.py
wslulciuc: @collado-mike, very good point. We should add the same option to the `openlineage-java` lib. Should we keep it simple and call the environment variable `OPENLINEAGE_CONFIG`?

wslulciuc (changes requested):
Looks great, @mobuchowski! Mind if we add a readme for the `py` client on usage etc, similar to the [`README.md`](https://github.com/OpenLineage/OpenLineage/blob/feature/http-client-java/client/java/README.md) for `openlineage-java`?

client/python/openlineage/client/transport/http.py
mobuchowski: I was thinking how to do that. So, we can't have property "token_provider" with `TokenProvider` type if we actually want to both 1) instantiate this config from yaml and 2) provide it as an extension point - since we don't know what types will be instantiated.

What we can do is have something like `auth` dictionary field that has type field that determines what other values are accepted by this config and what class will be instantiated to take care of it. 

So, essentially, a second layer of what we have in this PR in `DefaultTransportFactory` 🙂

client/python/openlineage/client/transport/factory.py
mobuchowski: 👍 for `OPENLINEAGE_CONFIG`

client/python/openlineage/client/transport/http.py
mobuchowski: Duplicate comment

client/python/openlineage/client/transport/transport.py
mobuchowski: Added README.md config and top level comment in this file.

client/python/openlineage/client/client.py
mobuchowski: Added `ConsoleTransport` and made it default if there's no valid config or env variable.

mobuchowski:
@wslulciuc added info in README.md.

client/python/openlineage/client/transport/http.py
mobuchowski: Added TokenProvider.

client/python/openlineage/client/transport/factory.py
```diff
@@ -0,0 +1,120 @@
...
+            yml_config = self._try_config_from_yaml()
+            if yml_config:
+                return self._create_transport(yml_config)
+        # Fallback to setting HTTP transport from env variables
+        http = self._try_http_from_env_config()
+        if http:
+            return http
+        # If there is no HTTP transport, log events to console
+        from openlineage.client.transport.console import ConsoleTransport, ConsoleConfig
+        return ConsoleTransport(ConsoleConfig())
```
collado-mike, line 38: I'd add a log statement here just to inform users why events are being printed rather than being sent somewhere

client/python/openlineage/client/transport/factory.py
```diff
@@ -0,0 +1,120 @@
...
+    @staticmethod
+    def _find_yaml() -> Optional[str]:
+        # Check OPENLINEAGE_CONFIG env variable
+        try:
+            path = os.getenv('OPENLINEAGE_CONFIG', None)
+            if path and os.path.isfile(path) and os.access(path, os.R_OK):
+                return path
+        except Exception:
+            # We can get different errors depending on system
+            pass
```
collado-mike, line 85: If the user has defined `OPENLINEAGE_CONFIG` and we can't find it or read it, we need to log an error for the user.

client/python/openlineage/client/transport/kafka.py
```diff
@@ -0,0 +1,48 @@
...
+    # Kafka producer config
+    # https://docs.confluent.io/platform/current/clients/confluent-kafka-python/html/index.html#kafka-client-configuration
+    config: Dict[str, str] = attr.ib()
+
+    # Topic on which we should send messages
+    topic: str = attr.ib()
+
+    # Flush by default. The process that emits can be killed in some cases - for example
+    # in Airflow integration.
+    flush: bool = attr.ib(default=True)
```
collado-mike, line 23: This is the opposite of what's in the README

collado-mike (approved):
Awesome work!

client/python/openlineage/client/transport/kafka.py
mobuchowski: Clarified, both in readme and here.

client/python/openlineage/client/transport/factory.py
mobuchowski: Fixed.

client/python/openlineage/client/transport/factory.py
mobuchowski: Done.

PR #545, opened 2022-02-16
ci: run most expensive jobs on large resource class

mobuchowski:
Some tests are slow, among other things, due to being CPU starved.

Move them to large resource class with 4 CPUs compared to 2 on medium. 
Overall total duration moved from ~30 minutes to ~20 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,6 +164,7 @@ jobs:
         type: string
     working_directory: ~/openlineage/integration/spark
     machine: true
+    resource_class: large
     environment:
       TESTCONTAINERS_RYUK_DISABLED: "true"
       JDK8_HOME: /usr/lib/jvm/java-8-openjdk-amd64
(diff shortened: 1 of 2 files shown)


pawel-big-lebowski:
Great improvement :thumbsup:

pawel-big-lebowski (approved):

wslulciuc (approved):

wslulciuc:
@mobuchowski, how will this affect our circle quota?

mobuchowski:
@wslulciuc minute of larger resource class will be more expensive, but we should use less of them 🙂 

Overall I think we'll use slightly more. We're still not close to the monthly limit, so I think it's not a problem yet.

mobuchowski:
@wslulciuc also, if we start having problems with it, there are several relatively quick things we can do to improve this. For example, not running all tests always. There's no need to run Airflow integration tests when only code changes were on Java side of things. Same for documentation PRs and stuff like this. Moreover, we could turn off running integration tests on commits and enable them only on PRs.

PR #547, opened 2022-02-16
Add UnknownOperatorAttributeRunFacet and support in lineage backend

collado-mike:
### Problem
In the Spark integration, we support the `UnknownEntryFacet` which records logical plan nodes that don't have a matching visitor so we can detect nodes that need coverage. This introduces a similar facet for Airflow DAGs so we can start recording `Operators` that don't have a matching `Extractor`.

In addition to recording the operators when no matching `Extractor` is found, I also added the facet to the output of the `PythonExtractor` and `BashExtractor`, since those extractors exist only to record source code that may be of interest. Since no lineage is actually being recorded by those extractors, we should consider them *unknown*. 

### Checklist

- [X ] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [X] You've updated the [`CHANGELOG.md`](https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md) with details about your change under the "Unreleased" section (_if relevant, depending on the change, this may not be necessary_)


integration/airflow/openlineage/airflow/facets.py
```diff
@@ -27,3 +28,21 @@ def from_task(cls, task):
 @attr.s
 class AirflowRunArgsRunFacet(BaseFacet):
     externalTrigger: bool = attr.ib(default=False)
+
+
+@attr.s
+class UnknownOperatorInstance:
```
wslulciuc, line 34: I would use `UnknownOperatorClass` for the facet, as the `name` is set to the name of the operator class not currently supported.

integration/airflow/openlineage/airflow/facets.py
```diff
@@ -27,3 +28,21 @@ def from_task(cls, task):
...
+    """
+    Describes an unknown operator - specifies the (class) name of the operator and its properties
+    """
+    name: str = attr.ib()
+    properties: Dict[str, object] = attr.ib()
+    type: str = "operator"
+
+
+@attr.s
+class UnknownOperatorAttributeRunFacet(BaseFacet):
```
wslulciuc, line 44: Does the facet naming read "unknown attribute of an operator for a given run"?

integration/airflow/openlineage/airflow/facets.py
```diff
@@ -27,3 +28,21 @@ def from_task(cls, task):
...
+    """
+    RunFacet that describes unknown operators in an Airflow DAG
+    """
+    unknownItems: List[UnknownOperatorInstance] = attr.ib()
```
wslulciuc, line 48: Do we need to be generic and use `unknownItems` or can we use `unknownOperatorClasses`?

integration/airflow/openlineage/lineage_backend/__init__.py
```diff
@@ -97,8 +97,21 @@ def _extract_metadata(self, dag_id, dagrun, task, task_instance=None):
...
                 f'Unable to find an extractor. {task_info}')
         from openlineage.airflow.extractors.base import TaskMetadata
+        from openlineage.airflow.facets import UnknownOperatorAttributeRunFacet, \
+            UnknownOperatorInstance
+
         return TaskMetadata(
-            name=self._openlineage_job_name(dag_id, task.task_id)
+            name=self._openlineage_job_name(dag_id, task.task_id),
+            run_facets={
+                "unknownSourceAttribute": UnknownOperatorAttributeRunFacet(
```
wslulciuc, line 106: Does `unknownSourceAttribute` mean that the source code is unknown?

integration/airflow/openlineage/airflow/extractors/python_extractor.py
```diff
@@ -36,6 +37,20 @@ def extract(self) -> Optional[TaskMetadata]:
...
+                run_facets={
+
+                    # The BashOperator is recorded as an "unknownSource" even though we have an
+                    # extractor, as the <i>data lineage</i> cannot be determined from the operator
+                    # directly.
+                    "unknownSourceAttribute": UnknownOperatorAttributeRunFacet(
+                        unknownItems=[
+                            UnknownOperatorInstance(
+                                name="PythonOperator",
+                                properties=self.operator
```
mobuchowski, line 50: I don't think you can put everything here. Imagine `self.operator` that holds `thread` - it will fail on serialization.

integration/airflow/tests/integration/test_integration.py
```diff
@@ -46,6 +46,7 @@ def wait_for_dag(dag_id):
 
     log.info(f"DAG '{dag_id}' state set to '{state}'.")
     if state == 'failed':
+        log.warning("DAG failed - exiting")
```
mobuchowski, line 49: Fixing this stuff here... https://github.com/OpenLineage/OpenLineage/pull/550

mobuchowski (changes requested):

integration/airflow/openlineage/lineage_backend/__init__.py
collado-mike: Hmm, I should have clarified this in the docs. The `UnknownOperatorAttributeRunFacet` is an _operator-specific_ instance of the more generic `UnknownSourceAttributeRunFacet`. Thus, the field names are the same (generic `unknownItems` instead of `unknownOperators`, the `type` field hardcoded to `operator`) and the facet name here is the same. I'll update the doc comments to clarify this.
I'll add comments in the code to clarify this.

integration/airflow/openlineage/airflow/extractors/python_extractor.py
collado-mike: This has been fixed. JSON now looks like

```json
{
  "unknownSourceAttribute": {
    "_producer": "https://github.com/OpenLineage/OpenLineage/tree/0.6.0/integration/airflow",
    "_schemaURL": "https://raw.githubusercontent.com/OpenLineage/OpenLineage/main/spec/OpenLineage.json#/definitions/BaseFacet",
    "unknownItems": [
      {
        "name": "PythonOperator",
        "properties": {
          "_dag": "<<non-serializable: DAG>>",
          "_downstream_task_ids": [],
          "_inlets": {
            "auto": false,
            "datasets": [],
            "task_ids": []
          },
          "_log": "<<non-serializable: Logger>>",
          "_outlets": {
            "datasets": []
          },
          "_upstream_task_ids": [
            "just_wait"
          ],
          "depends_on_past": false,
          "do_xcom_push": true,
          "email": [
            "datascience@example.com"
          ],
          "email_on_failure": false,
          "email_on_retry": false,
          "executor_config": {},
          "inlets": [],
          "op_args": [],
          "op_kwargs": {},
          "outlets": [],
          "owner": "datascience",
          "params": {},
          "pool": "default_pool",
          "pool_slots": 1,
          "priority_weight": 1,
          "provide_context": false,
          "python_callable": "<<non-serializable: function>>",
          "queue": "default",
          "retries": 0,
          "retry_delay": "<<non-serializable: timedelta>>",
          "retry_exponential_backoff": false,
          "start_date": "<<non-serializable: datetime>>",
          "task_id": "emit_event",
          "trigger_rule": "all_success",
          "wait_for_downstream": false,
          "weight_rule": "downstream"
        }
      }
    ]
  }
}
```
Note the non-serializable fields are called out

integration/airflow/openlineage/airflow/facets.py
collado-mike: It does use the class name as the name f the item, but it's recording the fields of the instance, not just generic class properties, so... 🤷🏽‍♂️

wslulciuc: Ahh, makes sense 👍

integration/airflow/openlineage/lineage_backend/__init__.py
wslulciuc: Cool. Thanks for clarifying 👍

wslulciuc (approved):
Great work, @collado-mike 💯 🛳️

PR #548, opened 2022-02-17
enable visitors being applied on specified events

pawel-big-lebowski:

### Problem

In order to collect version of input & output datasets, visitors need to collect input version before the processing, and output version at the end of processing. This requires visitors to know whether they're collecting input or output data and what kind of event triggered the visitor. 

Closes: [#484](https://github.com/OpenLineage/OpenLineage/issues/484)

### Solution

* Inject triggering event into `QueryPlanVisitor`. 
* Pass `isInputVisitor` property when constructing some visitors. 

### Checklist


integration/spark/src/main/common/java/io/openlineage/spark/api/QueryPlanVisitor.java
```diff
@@ -103,6 +106,10 @@ public boolean isDefinedAt(LogicalPlan x) {
     return false;
   }
 
+  public void setTriggeringEvent(SparkListenerEvent triggeringEvent) {
+    this.triggeringEvent = triggeringEvent;
+  }
```
collado-mike, line 111: I'd really rather not add a mutable field - especially one visible to implementations. In an earlier iteration, there was a method that exposed the type of events the visitor should be triggered on. What happened to that method? If we keep that approach, the visitor fields never need to mutate and the filtering can happen in the run event creation process

pawel-big-lebowski: Let's start first by providing examples why is that necessary:

1. AlterTableVisitor and schema facet - The visitor need to take schema from the command for `START` event, and retrieve it from `SparkCatalog` for `COMPLETE` event. To do so, it needs to know the triggering event. 
2. Dataset version facet - The input datasets should contain `datasetVersion` only for START events. The output datasets should contain `datasetVersion` only for COMPLETE events. As an example, think of an input and output being the same dataset. In this case, input dataset version should contain previous version and this information can be only obtained for START event. 

To solve the issues above, visitors need to know:
 - What event did trigger me?
 - Am I input or output visitor?  

During previous iteration, we tried simply with filtering out some visitors based on `SparkListenerEvent`. This did not solve the issue because we do want a visitor to skip only a `datasetVersions` facet while still allowing it to collect other information. 

Current code architecture is based on `PartialFunction<LogicalPlan, List<Dataset>` which is part of the public contract, as mentioned here https://github.com/OpenLineage/OpenLineage/pull/479/files, and provides a limitation to possible solutions. 

The possible solutions I see at the moment are:
 * Rewrite code architecture from `PartialFunction<LogicalPlan, List<Dataset>` to `PartialFunction<OpenLineageRunContext, List<Dataset>`, where
 `OpenLineageRunContext` contains `LogicalPlan` and `SparkListenerEvent`. This is the cleanest solution I know, but requires significant changes in public contract. Perhaps it is the last moment to go for it. 
 * Include `currentTriggeringEvent` within `OpenLineageContext`, which still introduces mutable field.
 * Include `setTriggeringEvent` to `QueryPlanVisitor` which does not break the contract but adds mutable field.

collado-mike: > This did not solve the issue because we do want a visitor to skip only a datasetVersions facet while still allowing it to collect other information.

Why is this the case? If a visitor triggers on some events but doesn't trigger on others, we don't lose any information. The fact that most query plan visitors trigger on every event is incidental - it doesn't have to work that way. The OpenLineage backends are already responsible for accumulating data from each event (e.g., sometimes the input datasets are present at the start but output datasets aren't available until the end).

collado-mike: See https://github.com/OpenLineage/OpenLineage/pull/559 for an alternative to this approach. I think if we can keep immutability, it will be a better in the long run.

PR #553, opened 2022-02-18
[SPARK] convert to LifecycleStateChangeDatasetFacet

pawel-big-lebowski:

### Problem

Visitors make use of custom `tableStateChange` facet introduced previously. Instead, they should use `lifecycleStateChange` facet from global spec introduced recently. 

Relates to [#518](https://github.com/OpenLineage/OpenLineage/issues/518)

### Solution

Make visitors use `lifecycleStateChange`  

### Checklist


integration/spark/src/test/common/java/io/openlineage/spark/agent/SparkContainerIntegrationTest.java
```diff
@@ -143,6 +146,8 @@ private static MockServerContainer makeMockServerContainer() {
                   "spark.extraListeners=" + OpenLineageSparkListener.class.getName(),
                   "--conf",
                   "spark.sql.warehouse.dir=/tmp/warehouse",
+                  "--conf",
```
pawel-big-lebowski, line 149: This limits nicely the amount of logs produced by `delta` provider.

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
```diff
@@ -60,14 +60,14 @@ public static void includeProviderFacet(
...
-    return fromDataSourceV2Relation(datasetFactory, context, relation, new HashMap<>());
+    return fromDataSourceV2Relation(datasetFactory, context, relation, null);
   }
 
   public static <D extends OpenLineage.Dataset> List<D> fromDataSourceV2Relation(
       DatasetFactory<D> datasetFactory,
       OpenLineageContext context,
       DataSourceV2Relation relation,
-      Map<String, OpenLineage.DatasetFacet> facets) {
+      OpenLineage.LifecycleStateChangeDatasetFacet.LifecycleStateChange lifecycleStateChange) {
```
mobuchowski, line 70: Can we make this not depend on particular facet?

pawel-big-lebowski: Yeah, good point. The other facets like `schema` or `tableProvider` can be extracted within this method and do have to be passed as separate params. On the other hand, `LifecycleStateChange` is something that only the visitor can determine. Moreover, dataset object cannot be modified afterwards ie. one cannot build `OpenLineage.Dataset` and then set `lifecycleStateChange`. 

An alternative approach would be to modify a method to return a map of `DatasetIdentifier` -> `DatasetFacetsBuilder` so that visitors could inject whatever they need to the builder and call something equivalent to last line of a method which builds dataset collection.

Need to try it out.

pawel-big-lebowski: Changed the code so that method does not depend on a facet.

mobuchowski (approved):

Issue #496, opened 2022-01-25
[PROPOSAL] Integration tests to verify OpenLineage backend failure doesn't break Airflow


collado-mike:
**Purpose:**
The philosophy of this project is that failure to report lineage should not have any negative impact on the running application (aside from errors in the logs). We should add some simple integration tests to verify

* No response from server (down)
* Timed out waiting for response from server
* Error response from server (bad input, bad server state)

mobuchowski:
We should also verify that bad configuration - for example, not providing `OPENLINEAGE_URL` does not crash application.

PR #557, opened 2022-02-22
airflow: failing to emit event does not fail airflow task

mobuchowski:
Failure to emit OpenLineage event should not break any system we're integrating with. This PR fixes this, by catching possible failure when emitting events, and logs it.

Integration tests for that failure scenario are added - they run test dag with three possible OL HTTP backends - one that 500s, one that times out, and one that does not exist. Each time, dag should complete with state `success`. 

Closes: https://github.com/OpenLineage/OpenLineage/issues/496


integration/airflow/tests/failures/docker/up.sh
```diff
@@ -0,0 +1,55 @@
...
+psycopg2-binary==2.9.2
+httplib2>=0.18.1
+retrying==1.3.3
+pytest==6.2.2
+jinja2==3.0.2
+python-dateutil==2.8.2
+${OPENLINEAGE_AIRFLOW_WHL}
+EOL
+
+docker-compose -f docker-compose-2.yml down
```
pawel-big-lebowski, line 44: Would it be possible to run three tests in a single docker run and switch flask app behavior based on request content instead of URL specified at container startup?

integration/airflow/tests/failures/test_integration.py
```diff
@@ -0,0 +1,78 @@
...
+    log.info(f"DAG '{dag_id}' state set to '{state}'.")
+    if state == 'failed':
+        return False
+    elif state != "success":
+        raise Exception('Retry!')
+    return True
+
+
+def setup_db():
+    time.sleep(10)
```
pawel-big-lebowski, line 56: Can this be replaced with `retry` lib?
If not, could we add a comment on that?

integration/airflow/tests/failures/airflow/dags/wait_dag.py
```diff
@@ -0,0 +1,47 @@
...
+    'owner': 'datascience',
+    'depends_on_past': False,
+    'start_date': days_ago(7),
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'email': ['datascience@example.com']
+}
+
+
+dag = DAG(
```
pawel-big-lebowski, line 31: Could not spot difference betweem `wait_dag` and `test_dag`.
If there is a reason for that, please provide it in comment.

integration/airflow/tests/failures/server/app.py
```diff
@@ -0,0 +1,41 @@
...
+def error_lineage():
+    logger.warning("Called error endpoint")
+    return "", 500
+
+
+@app.route("/timeout/api/v1/lineage", methods=['GET', 'POST'])
+def timeout_lineage():
+    logger.warning("Called timeout endpoint")
+    time.sleep(15)
+    return jsonify({}), 200
```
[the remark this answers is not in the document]
mobuchowski: It would be server side timeout to some other resource. I want to model client side one. Doesn't matter what will we return, since it will be after 15 seconds, and client waits only 10 seconds.

integration/airflow/tests/failures/docker-compose-2.yml
```diff
@@ -0,0 +1,125 @@
+version: "3.7"
+
+x-airflow-common: &airflow-common
```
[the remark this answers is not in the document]
collado-mike: There's already a docker-compose file for Airflow 2 and for Airflow 1. I'm really talking about this whole copy of the integration test stack under `tests/failures`. Can't we just reuse the stack defined under `tests/integration`? At most, you might want to override one service (the integration test app), but that's just one docker-compose override file, not a copy of `up.sh`, `wait-for-it.sh`, `Dockerfile`, `docker-compose.yml`,... I really don't think we should copy/paste this whole directory just to run one new test suite

integration/airflow/tests/failures/test_integration.py
mobuchowski: Yes, good idea. Done.

integration/airflow/tests/failures/docker/up.sh
mobuchowski: It would be simpler, yes, but I'd have to make Airflow integration have different behavior, and it's the system under test. 

I think additional complexity in tests is better than modifying code. Those tests don't run *that* slow and I would like them to run only on main.

integration/airflow/tests/failures/docker-compose-2.yml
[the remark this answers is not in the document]
mobuchowski: @collado-mike done

collado-mike (approved):

PR #559, opened 2022-02-23
Abstract query plan builder approach to exposing SparkListenerEvents to node visitors

collado-mike:
### Problem
An alternative approach to https://github.com/OpenLineage/OpenLineage/pull/548 , this introduces a new base class `AbstractQueryPlanDatasetBuilder`, which can trigger on specific `SparkListenerEvent`s (or other events, like `StageInfo` or whatever) and can apply the node visiting logic either with or without the triggering event present. To validate potential use, I refactored the `LogicalPlanVisitor` and the `SaveIntoDataSourceCommandVisitor` to both extend the base class and updated tests accordingly.

I like this approach because it avoids adding mutable fields the QueryPlanVisitors - a base class I already think ought to be deprecated in favor of the more general `AbstractInputDatasetBuilder` and `AbstractOutputDatasetBuilder` base classes. There's not a strong reason to handle the `LogicalPlan` nodes separately from every other event except for that's how it was being done when we started. I think an approach like this will allow us to deprecate the `QueryPlanVisitors` (rather than complicating the base class and further entangling implementations with dependencies that will make it harder to migrate later), while still providing the full functionality required, which as I understand them are

1. Trigger functions only on certain Spark events
2. Expose the triggering event along with the `LogicalPlan` node matched by the partial function
3. Allow the function to produce either `InputDataset`s or `OutputDataset`s as appropriate. 

Note that this approach is incomplete - I'm leaving this PR in draft mode so we can evaluate it and decide whether to continue this approach. 

This solves #484 

### Checklist

### DIFF
--- a/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/DatasetBuilderFactory.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/DatasetBuilderFactory.java
@@ -1,0 +1,26 @@
+package io.openlineage.spark.agent.lifecycle;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.spark.api.AbstractQueryPlanDatasetBuilder;
+import io.openlineage.spark.api.OpenLineageContext;
+import java.util.Collection;
+import java.util.List;
+import scala.PartialFunction;
+
+/**
+ * Provides input and output dataset builders
+ *
+ * <p>All common {@link AbstractQueryPlanDatasetBuilder} need to be grouped and passed into {@link
+ * DatasetBuilderFactory#getInputBuilders(io.openlineage.spark.api.OpenLineageContext)} or {@link
+ * DatasetBuilderFactory#getOutputBuilders(io.openlineage.spark.api.OpenLineageContext)} in order to
+ * produce within OpenLineage event {@link OpenLineage.InputDataset} and {@link
+ * OpenLineage.OutputDataset} respectively.
+ */
+public interface DatasetBuilderFactory {
+
+  Collection<PartialFunction<Object, List<OpenLineage.InputDataset>>> getInputBuilders(
+      OpenLineageContext context);
+
+  Collection<PartialFunction<Object, List<OpenLineage.OutputDataset>>> getOutputBuilders(
+      OpenLineageContext context);
+}
--- a/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/AppendDataVisitor.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/AppendDataVisitor.java
@@ -6,6 +6,7 @@ import io.openlineage.client.OpenLineage;
 import io.openlineage.spark.agent.util.PlanUtils;
 import io.openlineage.spark.api.OpenLineageContext;
 import io.openlineage.spark.api.QueryPlanVisitor;
+import java.util.ArrayList;
 import java.util.List;
 import org.apache.spark.sql.catalyst.plans.logical.AppendData;
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
@@ -23,7 +24,8 @@ public class AppendDataVisitor extends QueryPlanVisitor<AppendData, OpenLineage.
   @Override
   public List<OpenLineage.OutputDataset> apply(LogicalPlan x) {
     // Needs to cast to logical plan despite IntelliJ claiming otherwise.
-    return PlanUtils.applyFirst(
-        context.getOutputDatasetQueryPlanVisitors(), (LogicalPlan) ((AppendData) x).table());
+    return new ArrayList<>(
+        PlanUtils.applyAll(
+            context.getOutputDatasetQueryPlanVisitors(), (LogicalPlan) ((AppendData) x).table()));
   }
 }
--- a/integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
@@ -7,6 +7,7 @@ import io.openlineage.spark.agent.util.PlanUtils;
 import io.openlineage.spark.agent.util.ScalaConversionUtils;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
@@ -122,6 +123,9 @@ public abstract class AbstractQueryPlanDatasetBuilder<T, P extends LogicalPlan, 
     Type[] typeArgs = ((ParameterizedType) genericSuperclass).getActualTypeArguments();
     if (typeArgs != null && typeArgs.length > 1) {
       Type arg = typeArgs[1];
+      if (arg instanceof TypeVariable) {
+        return false;
+      }
       return ((Class) arg).isAssignableFrom(logicalPlan.getClass());
     }
     return false;
--- a/integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
@@ -1,0 +1,96 @@
+package io.openlineage.spark.api;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.client.OpenLineage.Dataset;
+import io.openlineage.spark.agent.lifecycle.UnknownEntryFacetListener;
+import io.openlineage.spark.agent.util.ScalaConversionUtils;
+import java.lang.reflect.ParameterizedType;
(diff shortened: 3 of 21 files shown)


pawel-big-lebowski (commented):
This looks amazing!

I was not aware that `LogicalPlan` can be extracted from `OpenLineageContext`. I am really happy to see that.

I am trying to put my PR  (https://github.com/OpenLineage/OpenLineage/pull/527/files) on the top of this to make sure this solves the problem. 

I found some issue (hopefully minor): dataset builders from spark3 (for example: `DataSourceV2ScanRelationVisitor`) need to be added 
while `InternalEventHandlerFactory` is in common package. Perhaps something similar to `VisitorFactoryProvider` needs to be created. 

Thank You @collado-mike for bringing this awesome idea.

integration/spark/src/test/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        SparkSession.builder()
+            .config("spark.sql.warehouse.dir", "/tmp/warehouse")
+            .master("local")
+            .getOrCreate();
+    OpenLineage openLineage = new OpenLineage(OpenLineageClient.OPEN_LINEAGE_CLIENT_URI);
+    InputDataset expected = openLineage.newInputDataset("namespace", "the_name", null, null);
+
+    OpenLineageContext context = createContext(session, openLineage);
+    AbstractQueryPlanDatasetBuilder<SparkListenerJobEnd, LocalRelation, InputDataset> builder =
+        new AbstractQueryPlanDatasetBuilder<SparkListenerJobEnd, LocalRelation, InputDataset>(
```
pawel-big-lebowski, line 43: I think this requires creating separate classes per each type of event to capture which is kind of a drawback.

integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+    return DatasetFactory.output(context.getOpenLineage());
+  }
+
+  protected DatasetFactory<OpenLineage.InputDataset> inputDataset() {
+    return DatasetFactory.input(context.getOpenLineage());
+  }
+
+  public abstract List<D> apply(P p);
+
+  public List<D> apply(T event) {
```
pawel-big-lebowski, line 49: When invoking partial functions, we invoke `io.openlineage.spark.agent.util.PlanUtils#merge` method which worked before as visitors where defined on disjunctive types. Now the types are common and mostly are just a couple of `SparkListenerEvent's.

I am having some issues when running integration tests for my branch: https://github.com/OpenLineage/OpenLineage/pull/573/files
and I think they may relate to this.

collado-mike: Yeah, I need to finish up this change - it's incomplete and was already failing the tests. I took a look at your branch and I think I can finish up the changes in my branch and hopefully that addresses your failures. However, there are some compilation and formatting fixes that need to be made to your branch anyway. If you can get those addressed, I can re-run the tests with your branch rebased on mine after my fixes are done.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/AppendDataVisitor.java
```diff
@@ -23,7 +24,8 @@ public AppendDataVisitor(OpenLineageContext context) {
   @Override
   public List<OpenLineage.OutputDataset> apply(LogicalPlan x) {
     // Needs to cast to logical plan despite IntelliJ claiming otherwise.
-    return PlanUtils.applyFirst(
-        context.getOutputDatasetQueryPlanVisitors(), (LogicalPlan) ((AppendData) x).table());
+    return new ArrayList<>(
+        PlanUtils.applyAll(
+            context.getOutputDatasetQueryPlanVisitors(), (LogicalPlan) ((AppendData) x).table()));
```
pawel-big-lebowski, line 29: Need to keep in mind this won't work soon once we rewrite `DatasourceV2RelationVisitor` into dataset builder.
But yeah, we can leave it for now.

integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /**
+   * Similar to the logic in {@link AbstractPartial}, this reads the type from the <i>second</i>
+   * generic argument on the class, if it is present and non-null.
+   *
+   * @param logicalPlan
+   * @return
+   */
+  protected boolean isDefinedAtLogicalPlan(LogicalPlan logicalPlan) {
```
pawel-big-lebowski, line 117: I experienced some strange behavior of this method although not sure what was that. I think the behavior of this method may vary for extended classes with different generic types on the top level class. While extending class, one can omit some parametrized types or even change their ordering.

Consider making it abstract and pass implementation to extending classes, which is more explicit and is already done that way in our examples.

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/DatasetBuilderFactory.java
```diff
@@ -0,0 +1,16 @@
+package io.openlineage.spark.agent.lifecycle;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.spark.api.OpenLineageContext;
+import java.util.Collection;
+import java.util.List;
+import scala.PartialFunction;
+
+public interface DatasetBuilderFactory {
```
collado-mike, line 9: Can you add a javadoc comment explaining the need for this interface?

integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanInputDatasetBuilder.java
```diff
@@ -0,0 +1,22 @@
+package io.openlineage.spark.api;
+
+import io.openlineage.client.OpenLineage;
+import org.apache.spark.scheduler.SparkListenerEvent;
+import org.apache.spark.scheduler.SparkListenerJobStart;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import org.apache.spark.sql.execution.ui.SparkListenerSQLExecutionStart;
+
+public abstract class AbstractQueryPlanInputDatasetBuilder<P extends LogicalPlan>
```
collado-mike, line 9: I'm not really sure I understand the reasoning behind triggering this builder only on start events. Please add javadoc here explaining the intended use of this class and try to make it clear to users that only start events will trigger subclasses

integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanOutputDatasetBuilder.java
```diff
@@ -0,0 +1,21 @@
+package io.openlineage.spark.api;
+
+import io.openlineage.client.OpenLineage;
+import org.apache.spark.scheduler.SparkListenerEvent;
+import org.apache.spark.scheduler.SparkListenerJobEnd;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import org.apache.spark.sql.execution.ui.SparkListenerSQLExecutionEnd;
+
+public abstract class AbstractQueryPlanOutputDatasetBuilder<P extends LogicalPlan>
```
collado-mike, line 9: Similar to the input dataset builder, please add javadoc explaining the intended use of this class and the fact that only end events will trigger subclasses.

collado-mike (commented):
Thanks for this change - it looks great! Please add some javadoc, then I think all looks good.

As the author of the original commits, I feel funny approving this PR 😂

integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanInputDatasetBuilder.java
pawel-big-lebowski: The behavior is repeated for all Spark 3 visitors so that it gets triggered only for start/stop events. Javadoc added.

pawel-big-lebowski (approved):
Changes introduced within the branch work well for implementing dataset version facet.

integration/spark/src/main/common/java/io/openlineage/spark/api/AbstractQueryPlanDatasetBuilder.java
collado-mike: I updated the tests to be explicit about this - the generic type arguments are on the _superclass_, not on the concrete implementation. So in your example, `extends AbstractQueryPlanDatasetBuilder<SparkListenerEvent, LogicalPlan, OutputDataset>` provides the concrete type argument. In this code, we look at the second argument, which is `LogicalPlan` in the example.

I did notice that we don't handle the case when some arguments are supplied, but others aren't. I added some handling for `TypeVariable`s, which also implement `Type`, but aren't `Class`es, thus cause a `ClassCastException` without explicit handling.

Issue #35, opened 2021-03-25
[PROPOSAL] Dataset version facet


collado-mike:
**Purpose:**
Many databases have explicit versioning in place for datasets. It is highly desirable to track the specific version of the dataset if it is known at consumption/generation time. Rather than relying on the service to correctly tie job inputs to specific versions of a dataset, the client should be allowed to report that information if it is available.

**Proposed implementation**
A dataset version facet may be sufficient to report dataset version information, if it exists. Custom facets can be used to include any database specific version information that may exist, while the minimum required fields should include a versionName string field.

julienledem:
Thank you @collado-mike That makes sense. Could you provide specific examples of databases and what the facet would look like?

collado-mike:
Sure- Apache Iceberg (https://iceberg.apache.org/spark/#time-travel ) and Databricks' Delta Lake (https://docs.delta.io/0.3.0/delta-batch.html#query-an-older-snapshot-of-a-table-time-travel ) are two systems that allow interacting with a dataset as of a specific version. I've also worked with a similar internal system previously.

I suppose we may need to disambiguate between dataset structure changes and content changes. I was thinking specifically of the snapshot version- Job X produces Snapshot A, which is consumed by Job Y- without considering structural changes to the dataset. 

The proposed change for the schema would be

```json
    "Dataset": {
      "type": "object",
      "properties": {
        "namespace": {
          "description": "The namespace containing that dataset",
          "type": "string",
          "example": "my-datasource-namespace"
        },
        "name": {
          "description": "The unique name for that dataset within that namespace",
          "type": "string",
          "example": "instance.schema.table"
        },
        "facets": {
          "description": "The facets for this dataset",
          "type": "object",
          "properties": {
            "documentation": {
              "$ref": "#/definitions/DocumentationDatasetFacet"
            },
            "schema": {
              "$ref": "#/definitions/SchemaDatasetFacet"
            },
            "dataSource": {
              "$ref": "#/definitions/DatasourceDatasetFacet"
            },
            "datasetVersion": {
              "$ref": "#/definitions/DatasetVersionFacet"
            }
          },
          "additionalProperties": {
            "$ref": "#/definitions/CustomFacet"
          }
        }
      },
      "required": [
        "namespace",
        "name"
      ]
    },
    "DatasetVersionFacet": {
      "allOf": [
        {
          "$ref": "#/definitions/BaseFacet"
        },
        {
          "type": "object",
          "properties": {
[...]

PR #580, opened 2022-03-01
introduce DatasetVersion facet in spec

pawel-big-lebowski:
### Problem

Let the spec support dataset versioning which is useful for storage providers like Iceberg or Delta. 

Part of the issue: [#35](https://github.com/OpenLineage/OpenLineage/issues/35)

### Solution

Introduce `DatasetVersionFacet` in spec. 

- [X] Your change modifies the [core](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json) OpenLineage model

### Checklist


mobuchowski (approved):

spec/facets/DatasetVersionDatasetFacet.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$schema" : "https://json-schema.org/draft/2020-12/schema",
+  "$id" : "https://openlineage.io/spec/facets/1-0-0/DatasetVersionDatasetFacet.json",
+  "$defs" : {
+    "DatasetVersionDatasetFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/DatasetFacet"
+        },
```
collado-mike, line 9: Wasn't the original version of this facet capable of being treated as an input facet or an output facet? Do we not need that capability anymore?

pawel-big-lebowski: There were no changes in the facet since _original_ version I got here:
https://github.com/OpenLineage/OpenLineage/pull/485/files#diff-93680b419816322d4d4b9576e763f72d5ffc9c2c4d72569a8c7a64e1fa0c6843R8

spec/facets/DatasetVersionDatasetFacet.json
```diff
@@ -0,0 +1,33 @@
...
+        {
+          "type": "object",
+          "properties": {
```
julienledem, line 12: do we want to add a field to capture what type of version it is? (iceberg, deltalake, ...)

julienledem: (possibly the wrong place to put this comment)

julienledem (commented):
This looks good to me. I just made a comment on being a little more precise in case user of this facet want to know where it's coming for (and act on them)

julienledem (commented):
can you add a line or two in this test case to check it generates what you want? https://github.com/OpenLineage/OpenLineage/blob/main/client/java/src/test/java/io/openlineage/client/OpenLineageTest.java

integration/spark/src/main/common/java/io/openlineage/spark/agent/lifecycle/plan/InsertIntoDataSourceVisitor.java
```diff
@@ -48,6 +48,7 @@ public InsertIntoDataSourceVisitor(OpenLineageContext context) {
                         .newDatasetFacets(
                             ds.getFacets().getDocumentation(),
                             ds.getFacets().getDataSource(),
+                            ds.getFacets().getVersion(),
```
wslulciuc, line 51: To get the dataset version, the method calls would look like:

```java
dataset.getFacets().getVersion().getDatasetVersion()
```

which, is a bit awkward. Another option would be to reverse the method calls (i.e. naming):

```
dataset.getFacets().getDatasetVersion().getVersion()
```

In the future, we might have a job version facet that can follow the same naming strategy:

```
job.getFacets().getJobVersion().getVersion()
```

spec/facets/DatasetVersionDatasetFacet.json
```diff
@@ -0,0 +1,33 @@
...
+          ]
+        }
+      ],
+      "type": "object"
+    }
+  },
+  "type" : "object",
+  "properties" : {
+    "version" : {
+      "$ref" : "#/$defs/DatasetVersionDatasetFacet"
```
wslulciuc, line 30: From my comment [above](https://github.com/OpenLineage/OpenLineage/pull/580/files#r819366769), I would reverse the facet property naming to go from specific to more general naming strategy which is a bit more intuitive when chaining the method calls:

```java
dataset.getFacets().getDatasetVersion().getVersion()
```

spec/facets/DatasetVersionDatasetFacet.json
wslulciuc: Wouldn't we know that based on the dataset? If not, I'd prefer to have the dataset type at a higher level. Meaning, knowing the type may not just be useful for the _version_ facet, but potentially other facets we'll be defining in the future.

pawel-big-lebowski: We do have a `tableProvider` facet for Spark which contains information on that like iceberg on delta. I do not think we should duplicate this information. 

The only issue is that `tableProvider` is currently a custom facet within Spark integration only. Perhaps we should generalize it to something like `storageProvider` and promote to global spec. Anyway, this is out of the scope of this PR.

spec/facets/DatasetVersionDatasetFacet.json
pawel-big-lebowski: On one hand, from the generated Java client code perspective, this looks better. 
However, there used to be some discussion about that in the past:
https://github.com/OpenLineage/OpenLineage/issues/35#issuecomment-1004741548
that resulted in `DatasetVersionDatasetFacet` over `VersionDatasetFacet`
which looks better from JSON spec files point of view as one instantly gets to know what `version` is the facet about. 

I can't see any significant benefits of one approach over the another, and would leave it like that.

spec/facets/DatasetVersionDatasetFacet.json
julienledem: That makes sense.
Yes, it sounds like a good idea to make it standard. Do you think this is already provided by DatasourceDatasetFacet?
https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/DatasourceDatasetFacet.json

PR #607, opened 2022-03-11
Support Spark 3.2.1

pawel-big-lebowski:

### Problem

Support Spark version 3.2. 

### Solution

 * Include separate integration test run for Spark 3.2.1
 * Change docker image provider to `bitnami/spark` as previously used `godatadriven/pyspark` does not support latest spark versions. 
 * Match iceberg & delta versions in integrations test with Spark version. There are different iceberg/delta jars for Spark 3.1 and Spark 3.2
 * Exclude `alter table` integration test for Spark 3.2. Spark 3.2 introduced several alter table features. Supporting them in OpenLineage requires extra visitors/dataset builders which is outside the scope of this PR and should be done later. [616]

Please describe your change as it relates to the problem, or bug fix, as well as any dependencies. If your change requires a schema change, please describe the schema modifcation(s) and whether it's a _backwards-incompatible_ or _backwards-compatible_ change, then select one of the following:

### Checklist

### DIFF
--- a/integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
@@ -140,8 +140,8 @@ public class OpenLineageSparkListener extends org.apache.spark.scheduler.SparkLi
 
   /** called by the SparkListener when a spark-sql (Dataset api) execution starts */
   private static void sparkSQLExecStart(SparkListenerSQLExecutionStart startEvent) {
-    ExecutionContext context = getSparkSQLExecutionContext(startEvent.executionId());
-    context.start(startEvent);
+    getSparkSQLExecutionContext(startEvent.executionId())
+        .ifPresent(context -> context.start(startEvent));
   }
 
   /** called by the SparkListener when a spark-sql (Dataset api) execution ends */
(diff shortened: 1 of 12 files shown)


wslulciuc (commented):
Changes look great, @pawel-big-lebowski! But, the CI tests for spark `3.2.1` fails with the error:

```
ERROR EventEmitter: Could not emit lineage ... OpenLineageHttpException(code=null, message=, details=null)
```

Given `OpenLineageHttpException` is thrown, looks like there's an error on the test server?

mobuchowski (approved):

integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -140,8 +141,13 @@ public void onOtherEvent(SparkListenerEvent event) {
 
   /** called by the SparkListener when a spark-sql (Dataset api) execution starts */
   private static void sparkSQLExecStart(SparkListenerSQLExecutionStart startEvent) {
+    if (SQLExecution.getQueryExecution(startEvent.executionId()) == null) {
+      return;
+    }
```
collado-mike, line 146: No point in adding this here - the manipulation of the underlying map happens in a different thread. It may well be visible here, but gone by the time we try to get it in the `ContextFactory`. Remove this and just rely on the `if (context != null)` check below.

integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -186,11 +192,12 @@ public void onJobStart(SparkListenerJobStart jobStart) {
                                     .map(ds -> ds.jobIdToActiveJob().get(jobStart.jobId()))
                                     .flatMap(ScalaConversionUtils::asJavaOptional))
                         .map(job -> getSqlExecutionId(job.properties())))
-            .map(
-                id -> {
-                  long executionId = Long.parseLong(id);
-                  return getExecutionContext(jobStart.jobId(), executionId);
-                })
+            .map(id -> Long.parseLong(id))
```
collado-mike, line 195:
```java
.map(Long::parseLong)
```

integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -186,11 +192,12 @@ public void onJobStart(SparkListenerJobStart jobStart) {
...
+            .filter(
+                id ->
+                    sparkSqlExecutionRegistry.containsKey(id)
+                        || SQLExecution.getQueryExecution(id) != null)
+            .map(id -> getExecutionContext(jobStart.jobId(), id))
```
collado-mike, line 200: I'd remove the `filter` here, for the same reason as above - the map is changed in a different thread. Checking before getting will likely lead to bugs. Just call `getExecutionContext`, which will first check the map, then invoke the `ContextFactory` to try to create the `ExecutionContext`. We should update the `ContextFactory` method to actually verify the `QueryExecution` we get isn't null. Here, we can filter _after_ the map call to ensure `Objects::nonNull`.
I.e., 
```java
.map(Long::parseLong)
.map(id -> getExecutionContext(jobStart.jobId(), id))
.filter(Objects::nonNull)
.orElseGet(() -> getExecutionContext(jobSTart.jobId());

**Review by collado-mike — APPROVED** [review]
Nice!

Some very minor issue with the NPE handling you added (thanks for addressing that, btw).

File: integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java (hunk 1 above)
Comment (pawel-big-lebowski) [reply integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java:146]: When I remove this, the integration test `spark_write_delta_table_version.py` fails due to the exception:
```
java.lang.NullPointerException
        at io.openlineage.spark.agent.lifecycle.ContextFactory.createSparkSQLExecutionContext(ContextFactory.java:43)
```
This looks like delta triggers Spark listener event with `executionId` that won't work within `SQLExecution.getQueryExecution(executionId)`. That's why the check is added here.

File: integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java (hunk 3 above)
Comment (pawel-big-lebowski) [reply integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java:200]: Rewriting this to `filter(Objects::nonNull)` makes the delta test fail with NullPointerExceptions, the same as in the comment above.

File: integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java [hunk 4]
```diff
@@ -155,6 +156,17 @@ private static void sparkSQLExecEnd(SparkListenerSQLExecutionEnd endEvent) {
   /** called by the SparkListener when a job starts */
   @Override
   public void onJobStart(SparkListenerJobStart jobStart) {
+    if (jobStart
```
pawel-big-lebowski, line 159: @collado-mike Need your help to find out optimal way to implement filtering like this here.

pawel-big-lebowski: Created new issue to improve Spark & Delta integration:
[631](https://github.com/OpenLineage/OpenLineage/issues/631)

mobuchowski (approved):

Issue #619, opened 2022-03-21
[SPEC] Introduce provider facet in global Spec


pawel-big-lebowski:
Spark integration collects table provider information like `delta` or `iceberg` within custom facet. This information becomes useful together with a dataset version collected as version is retrieved from the provider. To make it consistent, if a `version` is within a global spec, the provider information should also be put there, however a more generic name should be considered like `storageProvider`, `backendStorageProvider`, `storageProperties` with fields: `format` or `storageProvider`

PR #620, opened 2022-03-21
add StorageDatasetFacet to spec

pawel-big-lebowski:

### Problem

Spark integration collects table provider information like delta or iceberg within custom facet. This information becomes useful together with a dataset version collected as version is retrieved from the provider. To make it consistent, if a version is within a global spec, the provider information should also be put there, however a more generic name should be considered like storageProvider, backendStorageProvider, storageProperties with fields: format or storageProvider

Closes: #619

### Solution

Add `datasetProvider` facet to spec.

### Checklist

- [] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)


spec/facets/DatasetProviderDatasetFacet.json
```diff
@@ -0,0 +1,37 @@
...
+            "datasetProvider"
+          ]
+        }
+      ],
+      "type": "object"
+    }
+  },
+  "type" : "object",
+  "properties" : {
+    "provider" : {
```
julienledem, line 33: To be consistent, the type should be "ProviderDatasetFacet"

spec/facets/DatasetProviderDatasetFacet.json
```diff
@@ -0,0 +1,37 @@
...
+  "$defs" : {
+    "DatasetProviderDatasetFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/DatasetFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "datasetProvider": {
```
julienledem, line 13: Since we are already in the "provider" facet, this would be "name".
We should document a list of known providers to ensure consistency. Would that apply to other things than Table Formats? Possibly a name more specific than "provider" would be appropriate.

spec/facets/DatasetProviderDatasetFacet.json
```diff
@@ -0,0 +1,37 @@
...
+              "description": "Dataset provider like iceberg, delta-lake, etc.",
+              "type": "string"
+            },
+            "datasetFormat": {
```
julienledem, line 17: Similar comment: we should document known formats

julienledem (commented):
Thanks for starting this

spec/facets/DatasetProviderDatasetFacet.json
mobuchowski: Add to https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md perhaps?

spec/facets/DatasetProviderDatasetFacet.json
pawel-big-lebowski: `provider` is consistent with other facets like `schema` or 'version` which are fields within `DatasetFacets`.

spec/facets/DatasetProviderDatasetFacet.json
pawel-big-lebowski: Given the discussion we have, term `provider` seems to be kind of misleading. 
Official iceberg and delta definitions are:
 * `Iceberg is a high-performance format for huge analytic tables`
 * `Delta Lake is a storage layer that brings scalable, ACID transactions to Apache Spark and other big-data engines.`

Following that, we could name it `storage` of the type `DatasetStorageDatasetFacet` with the fields:
 * `storageLayer` which can be `iceberg`, `delta`
 * `fileFormat` which can be `parquet`, `orc`, etc. 
 
Alternatively we could go with `StorageLayerDatasetFacet` which is also OK but has a limited extensibility in the future.

pawel-big-lebowski: I modified facet into `DatasetStorageDatasetFacet` as it describes better its content.

spec/facets/DatasetProviderDatasetFacet.json
julienledem: I think the description here should be more directive:
It is "iceberg" or "delta"
not "like iceberg or delta"

julienledem:
I made few last comments. Otherwise this looks great to me.

mobuchowski (approved):

spec/facets/DatasetProviderDatasetFacet.json
pawel-big-lebowski: Thanks for the comments. I've applied the changes accordingly.

collado-mike (approved):

PR #645, opened 2022-03-31
[Spark][Integration] column level lineage

pawel-big-lebowski:

### Problem

Turn on the introduced column level lineage mechanism and include it within the integration tests. 

Part of: #673

### Solution

* Modify existing visitors and dataset builders to include column level lineage when applicable.
* Modify existing integration tests to verify if column level lineage is included. 

### Checklist


xiaoyongzhu:
@pawel-big-lebowski thanks for the effort! do you have a target date for this feature to release?

pawel-big-lebowski:
@xiaoyongzhu Nope but this should not take long.

integration/spark/src/main/common/java/io/openlineage/spark/api/DatasetFactory.java
```diff
@@ -161,6 +162,7 @@ public D getDataset(DatasetIdentifier ident, StructType schema) {
    * @return
    */
   public D getDataset(
+      // OpenLineageContext context,
```
collado-mike, line 165: Unnecessary?

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/AppendDataDatasetBuilder.java
```diff
@@ -39,8 +40,16 @@ public List<OpenLineage.OutputDataset> apply(AppendData x) {
...
+      DataSourceV2Relation relation = (DataSourceV2Relation) logicalPlan;
+      List<OpenLineage.OutputDataset> outputDatasets =
+          new DataSourceV2RelationOutputDatasetBuilder(context, factory)
+              .apply((DataSourceV2Relation) logicalPlan);
+
+      return ColumnLevelLineageUtils.buildColumnLineageDatasetFacet(context, relation.schema())
+          .filter(facet -> outputDatasets.stream().findFirst().isPresent())
+          .map(facet -> ColumnLevelLineageUtils.rewriteOutputDataset(outputDatasets.get(0), facet))
+          .map(el -> Collections.singletonList(el))
+          .orElse(outputDatasets);
```
collado-mike, line 52: Wouldn't it make sense to move this logic to the `OpenLineageRunEventBuilder` so we don't have to include it in every OutputDataset builder? I think it's better to keep these kinds of internal APIs private so that extenders don't need to add these calls to every extension visitor they want to add - it'll make the visitor API difficult for others to maintain, as we may change the signatures or functionality going forward.

integration/spark/integrations/container/pysparkCTASWithColumnLineageEnd.json
```diff
@@ -0,0 +1,70 @@
...
+      "lifecycleStateChange" : {
+        "lifecycleStateChange" : "CREATE"
+      },
+      "columnLineage": {
+        "fields": {
+          "a": {
+            "inputFields": [
+              {
+                "namespace": "file",
+                "name": "/tmp/ctas_load/temp",
```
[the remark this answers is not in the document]
pawel-big-lebowski: Both `namespace` and `name` describe input datasets. 
Yes, you're right: `inputNamespace` and `inputName` would be more explicit.
I wouldn't like to add this to this PR due to its size. 
However, I can prepare another one just after this to rename it. 
What do you think?

integration/spark/src/main/spark3/java/io/openlineage/spark3/agent/lifecycle/plan/AppendDataDatasetBuilder.java
pawel-big-lebowski: Yeah, that's the great idea. Thanks for that.

integration/spark/integrations/container/pysparkCTASWithColumnLineageEnd.json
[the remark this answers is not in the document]
mobuchowski: I think combination of `namespace` and `name` are describing `dataset` everywhere else in the spec. 

Renaming `name` to `dataset` would be confusing, since `name` does not specify dataset by itself - only in combination with `namespace`. 

I think adding nesting might do what you want to do, but don't think it would add clarity.

```json
"inputFields": [{
    "dataset": {
        "namespace": "file",
        "name": "/tmp/ctas_load/temp",    
    },
    "field": "a"
}]
```

mobuchowski (approved):

PR #656, opened 2022-04-06
airflow: import extractors based on operator classnames

mobuchowski:
Some environments restrict case-sensitive environment variable names, like Astronomer.
To solve this, make extractor loading use only operator classnames that are provided by external extractors. 


rossturk (approved):
As I understand it, this will make it so that what follows `OPENLINEAGE_EXTRACTOR` in the env could be any string at all 👍

mobuchowski:
@rossturk `OPENLINEAGE_EXTRACTOR_`, but yes.

PR #664, opened 2022-04-08
Update parent_run_id to be a uuid from the dag name and run_id

collado-mike:
### Problem
Airflow DAG tasks are currently reporting ParentRunFacets, which _should_ connect a task to a DAG run. However, the values that are used for the ParentRunFacet don't match what one would expect. For one, the `runId` is typically the schedule time of the Airflow DAG (a non-unique value). The job name of the parent is also the same as the name of the task, so connecting multiple tasks to the same DAG run is difficult, as they don't share a parent job name. 

### Solution
This uses a [v3 uuid](https://www.uuidtools.com/uuid-versions-explained#version-3-version-5) to generate a _deterministic_ runId for the DAG so that the parent `runId` for each task is the same. It also uses the DAG's `dag_id` field to fill the job name so that all tasks have the same parent job name.

### Checklist

- [X] You've updated the [`CHANGELOG.md`](https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md) with details about your change under the "Unreleased" section (_if relevant, depending on the change, this may not be necessary_)


wslulciuc (approved):
LGTM 🛳️

Issue #649, opened 2022-04-01
[INTEGRATION][AIRFLOW] Need to set BigQuery location when queries happen in non-default region


collado-mike:
If I run a query in a non-default location, the BigQuery client configured by OpenLineage doesn't find the query execution information, as it request the info from the default location.

I ran a query on a couple of datasets in the `us-west-1` region, but got the following in my airflow logs
```
[2022-04-01, 22:16:36 UTC] {bigquery.py:125} ERROR - Cannot retrieve job details from BigQuery.Client. 404 GET https://bigquery.googleapis.com/bigquery/v2/projects/bq-airflow-marquez/jobs/airflow_564654654654654_6d6ba7b094b081d64604ce30361ce77e?projection=full&prettyPrint=false: Not found: Job bq-airflow-marquez:airflow_564654654654654_6d6ba7b094b081d64604ce30361ce77e
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/openlineage/common/provider/bigquery.py", line 101, in get_facets
    job = self.client.get_job(job_id=job_id)
  File "/usr/local/lib/python3.9/site-packages/google/cloud/bigquery/client.py", line 2066, in get_job
    resource = self._call_api(
  File "/usr/local/lib/python3.9/site-packages/google/cloud/bigquery/client.py", line 782, in _call_api
    return call()
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry.py", line 286, in retry_wrapped_func
    return retry_target(
  File "/usr/local/lib/python3.9/site-packages/google/api_core/retry.py", line 189, in retry_target
    return target()
  File "/usr/local/lib/python3.9/site-packages/google/cloud/_http.py", line 484, in api_request
    raise exceptions.from_http_response(response)
```

PR #691, opened 2022-04-20
bigquery: use client from Airflow hook when possible

mobuchowski:
This PR makes bigquery integration for Airflow reuse hook code that provides BigQuery client.

By doing that, we use "proven" connection instead of relying on `GOOGLE_APPLICATION_CREDENTIALS` which are one of the few different methods of specifying BigQuery connection in Airflow - for example, providing keyfile json in UI.

Additionally, this solves issue of using proper query location that @collado-mike raised.

Closes: https://github.com/OpenLineage/OpenLineage/issues/649

### DIFF
--- a/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py
+++ b/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py
@@ -89,6 +89,10 @@ class BigQueryExtractor(BaseExtractor):
         # Get client using Airflow hook - this way we use the same credentials as Airflow
         if hasattr(self.operator, 'hook') and self.operator.hook:
             hook = self.operator.hook
+            return hook.get_client(
+                project_id=hook.project_id,
+                location=hook.location
+            )
         elif parse_version(AIRFLOW_VERSION) >= parse_version("2.0.0"):
             BigQueryHook = try_import_from_string(
                 'airflow.providers.google.cloud.operators.bigquery.BigQueryHook'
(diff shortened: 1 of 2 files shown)


integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py
```diff
@@ -64,7 +67,10 @@ def extract_on_complete(self, task_instance) -> Optional[TaskMetadata]:
                 }
             )
 
-        stats = BigQueryDatasetsProvider().get_facets(bigquery_job_id)
+        print(self.operator.__dict__)
```
collado-mike, line 70: Remove?

collado-mike (approved):
Nice

integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py
mobuchowski: 😅

integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py
```diff
@@ -80,6 +85,28 @@ def extract_on_complete(self, task_instance) -> Optional[TaskMetadata]:
             job_facets=job_facets
         )
 
+    def _get_client(self):
+        # Get client using Airflow hook - this way we use the same credentials as Airflow
+        if hasattr(self.operator, 'hook') and self.operator.hook:
+            hook = self.operator.hook
```
pankajkoti, line 91: I can see we're returning `Client()` on line 108. What do we do with this `hook`?

Issue #670, opened 2022-04-14
spark: Provide a ParentRunFacet using spark.sql.execution.parent property


wjohnson:
In the recent support for Databricks' Synapse connector (#452), lineage is truncated due to the way the Databricks connector is written when a Synapse table is the output destination.

There are two separate run ids being reported during a normal execution.

* Run Id 1 comes from the execution start and displays the correct lineage
* Run Id 2 comes from the job start and job end events in Spark 
    * Reports the correct inputs 
    * Reports a user defined staging location as the output

The individual job start events do provide a property called `spark.sql.execution.parent` which links Run Id 2 to Run Id 1. In this case, the value in `spark.sql.execution.parent` is the execution id of the parent spark sql execution start event.

Ideally, we would be able to use the [ParentRunFacet](https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/ParentRunFacet.json) to link these two run ids and correctly represent the lineage.

I would like to propose that this property should be detected and a ParentRunFacet should be added to the OpenLineage payload that refers to this parent run id.

collado-mike:
Hey, @wjohnson, this would be fantastic! Do you have a link or reference to source code that shows this property being set? I don't see any reference to it in the Spark source code. Is it an Azure-specific setting? Or is it Databricks-specific? I'd love to know the specifics of when we can use this property.

wjohnson:
> Hey, @wjohnson, this would be fantastic! Do you have a link or reference to source code that shows this property being set? I don't see any reference to it in the Spark source code. Is it an Azure-specific setting? Or is it Databricks-specific? I'd love to know the specifics of when we can use this property.

@collado-mike I dug around too and I cannot find this in the particular jar that Databricks is using to write the Synapse tables. I'm guessing it's buried somewhere in the proprietary code of Databricks' spark?  If that's the case, I could add this to the Databricks environment facet as a stopgap for US but it won't help pursue a general solution for all of spark.

wjohnson:
I'm investigating how I might add this for JUST Databricks and I'm a bit stuck.

I've put together a sample of what I want to accomplish and I can get it to generate the ParentRunFacet on #707 with a hard coded UUID and parent job name but I could use some help thinking about how to go from execution id to RunId and Parent Job Name when I'm in the `SparkSqlExecutionContext`.

@collado-mike  and @mobuchowski  if you had a cycle or two, I'd appreciate any thoughts.

mobuchowski:
[...]

PR #707, opened 2022-04-27
Support ParentRunFacet for Databricks' Synapse DW Connector in Spark

wjohnson:
### Problem

The Datbaricks Synapse Data Warehouse Connector does not report lineage cleanly when Synapse is the output destination. The SparkSQLExecutionStart contains the correct output but the subsequent JobStart events contain lineage for a temporary folder. However, the JobStart event does report a `spark.sql.execution.parent` property which can be used to identify which parent execution id called this JobStart.

It doesn't close but it is a candidate for #670 for the Synapse DW Connector specifically.

### Solution

I plan to modify `SparkSQLExecutionContext.start(jobStart)` to do the following:
* Check if the Job Start event has the `spark.sql.execution.parent` property.
* If so, get the OpenLineage RunId for this execution Id. **Need help here**
* If there is a parent run, modify the call to `runEventBuilder.buildRun` and the `buildParentFacet` to accept:
  * Optional<UUID> parentRunID (this is provided by the above steps if I can figure them out)
  * String parentJobName (**Need help here**)
    * job Name seems to be buried in `SparkSqlExecutionContext.buildJob`  
    ```
    sparkContext.appName().replaceAll(CAMEL_TO_SNAKE_CASE, "_$1").toLowerCase(Locale.ROOT)
    + "."
    + node.nodeName().replaceAll(CAMEL_TO_SNAKE_CASE, "_$1").toLowerCase(Locale.ROOT));
    ```

### Checklist

### DIFF
--- a/integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
+++ b/integration/spark/src/main/common/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
@@ -87,7 +87,6 @@ public class OpenLineageSparkListener extends org.apache.spark.scheduler.SparkLi
 
   /** called by the SparkListener when a spark-sql (Dataset api) execution starts */
   private static void sparkSQLExecStart(SparkListenerSQLExecutionStart startEvent) {
-    log.info("WILLJ: SparkListenerSQLExecutionStart event occurred");
     getSparkSQLExecutionContext(startEvent.executionId())
         .ifPresent(context -> context.start(startEvent));
   }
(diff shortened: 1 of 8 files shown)


collado-mike:
Seems a good time to start refactoring the `ParentRunFacet` to be generated by an `AbstractRunFacetBuilder` rather than treating it as a special case the way we do. Currently, we hold that `parentRunId` in the `EventEmitter`, as we parse the Spark configuration at startup. Instead, we can move that logic to a new `AbstractRunFacetBuilder`. A tricky consideration is how to allow the default behavior when there is no databricks-specified parent execution id, but execute the custom builder when there is. I hate to rely on the order of execution (let the default run first, then override the facet map when the custom builder runs). Franky, I'm not opposed to putting all the logic in one facet builder; it just means that _only_ the databricks case is handled. Maybe that's ok, as we don't know of any other cases to handle.

wjohnson:
> Seems a good time to start refactoring the `ParentRunFacet` to be generated by an `AbstractRunFacetBuilder` ... A tricky consideration is how to allow the default behavior when there is no databricks-specified parent execution id, but execute the custom builder when there is. ... I'm not opposed to putting all the logic in one facet builder; it just means that _only_ the databricks case is handled. Maybe that's ok, as we don't know of any other cases to handle.

I've been exploring this route for a bit and my one challenge is getting the Parent Run Id.

* JobStart contains the `spark.sql.execution.parent` and the execution id of the jobStart.
* I need to get the UUID for the ParentRun.
* I can only get the ParentRun by getting the SparkSqlExectutionContext which has an OpenLineageContext which has an OpenLineage which has the RunId but parts of this chain are consider private fields.
* To get that SparkSqlExecutionContext, I need to work with the `sparkSqlExecutionRegistry` field in the `OpenLineageSparkListener`.

I can put together a `DatabricksParentRunFacetBuilder` but I don't know how to extract that parent run id when it's stuck inside the OpenLineageSparkListener.

collado-mike:
> I can put together a `DatabricksParentRunFacetBuilder` but I don't know how to extract that parent run id when it's stuck inside the OpenLineageSparkListener.

I was thinking to eliminate all references to the parent in the `OpenLineageSparkListener` - remove the spark configuration parsing, remove it from the `SparkSQLExecutionContext`, the `OpenLineageRunEventBuilder`... everything. Make a facet builder something like the following
```java
public class ParentRunFacetBuilder extends AbstractRunFacetBuilder<SparkListenerEvent> {
  private static final Map<Long, UUID> executionIdToRunIdMap = new ConcurrentHashMap<>();

  public ParentRunFacetBuilder(@NonNull OpenLineageContext context) {
    super(context);
  }

  @Override
  protected void build(SparkListenerEvent event, BiConsumer<String, ? super RunFacet> consumer) {
    context.getSparkSession()
        .ifPresent(sess -> {
          Optional<ParentRunFacet> parentFacet = buildParentFacet(sess);
          parentFacet.ifPresent(facet -> consumer.accept("parent", facet));
        });
  }

  private Optional<ParentRunFacet> buildParentFacet(SparkSession sess) {
    String parentRunId = SparkConfUtils.findSparkConfigKey(sess.sparkContext().conf(),
        "openlineage.parentRunId",
        null);
    String parentJobName = SparkConfUtils.findSparkConfigKey(
        sess.sparkContext().conf(),
        "openlineage.parentJobName",
        null
    );
    String namespace = SparkConfUtils.findSparkConfigKey(
        sess.sparkContext().conf(),
        "openlineage.namespace",
        null
    );
    if (parentRunId != null && parentJobName != null && namespace != null) {
      return Optional.of(context.getOpenLineage().newParentRunFacetBuilder()
          .run(context.getOpenLineage()
              .newParentRunFacetRunBuilder()
              .runId(UUID.fromString(parentRunId))
              .build())
          .job(context.getOpenLineage()
              .newParentRunFacetJobBuilder()
              .namespace(namespace)
              .name(parentJobName)
              .build())
          .build());
    }
    return Optional.empty();
  }
}
```
This will find the parent id and set it into the run facets.

What this _doesn't_ do is track the run ids of previous jobs and make it available for lookup. We could do something "hacky" and add a Map that tracks execution ids and the run ids associated with them - somewhat similar to how the unknown facet listener works. There's just not an easy way to make two different run facet builders - one for general purpose use and one for Databricks. But again, I don't know of any other scenario we need to handle, so maybe just a single facet builder with the logic for both scenarios is enough.

wjohnson:
> What this _doesn't_ do is track the run ids of previous jobs and make it available for lookup. We could do something "hacky" and add a Map that tracks execution ids and the run ids associated with them - somewhat similar to how the unknown facet listener works.
@collado-mike I ran with an idea @mobuchowski had on the issue. This still doesn't quite do the trick because I don't provide namespace or name of the parent job.  Mind taking a look at the approach?  It does not conform to your approach 😰

I ended up modifying the ExecutionContext interface so that I could get a RunId from it.  And I use the sparkSqlExecutionRegistry to get the parent Run Id that is happening internally. I'm thinking I would do the same thing for job name (namespace name should be the same as normal I think) but haven't gotten that far yet.

wjohnson:
Abandoning this one in favor of #849

PR #708, opened 2022-04-27
Airflow integ tests refactor

collado-mike:
### Problem

Currently, integration testing Airflow requires 
1. building a base Docker image that has the compiled wheels for each OL library
2. building an Airflow image that copies the compiled wheels from the base image
3. building an integration test image that copies the compiled wheels from the base image
4. running all images with docker-compose and checking test results.

Because the images built in steps 2 and 3 are dependent on the artifacts provided by the image in step 1, the Docker build operations in those steps can never be cached. This includes a lot of downloading/installation of apt packages as well as many pypi packages. This is especially a problem during local development, but also causes slow builds in CI.

### Solution

I've updated the build process to eliminate step 1 and instead copy the OpenLineage source packages to Docker volumes which can be mounted and the packages installed without dependencies at startup. All apt and pypi dependencies are installed during the image build so the same images can be used again and again as code changes are made to the OpenLineage packages. This makes running the integration tests locally much faster and gives us the option of turning on docker image caching in CI.

**Note** I used Docker volumes rather than just mounting the local directories because mounting local directories isn't supported if a remote docker host is used, which is typical in CI setups and also how I personally do local development.

I also removed the separate `integration-test-integration-airflow-failure` step in CI and changed the failure tests to run in the same step as the Airflow 2 integration tests. This allowed me to avoid copy/pasting the same docker-compose updates to three different scripts. Ultimately, I'd like to refactor the Airflow 1/2 docker-compose configuration files so that most of the configuration can be reused and just rely on overrides for the minor differences.

### Checklist


mobuchowski:
Unfortunately, right now CI shows increase of running time to ~25m on tests from ~6m 😢

PR #718, opened 2022-04-29
ci: non-snapshot relase should release sql; airflow release does not require runtime sql dependency

mobuchowski:
Release 0.8.0 failed because `release-python`, in comparison to snapshot release, did not provide SQL dependency. 
Instead of providing it, stop requiring dependency that is used only at runtime.

https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/3164/workflows/5c9350ae-fe06-455b-abba-2a792093ccbc/jobs/30183

Also, build SQL parser on release. Even if we did not fail on ⬆️ , we'd have to release again with SQL. 


wslulciuc (approved):
Thanks for the fix, @mobuchowski!

Issue #734, opened 2022-05-06
[Bug][Airflow] SnowflakeOperator SQL sometimes is in an array


conorbev:
I have a case where the `sql` in an Airflow `taskInfo` is in an array, something like: 

```
'sql': ['BEGIN;', "DO STUFF;", 'COMMIT;']
```

No Lineage is extracted and the sql facet is also not populated. 

I _think_ that this may be because the parser is expecting a [string](https://github.com/OpenLineage/OpenLineage/blob/16b24dae3d8703e8e94b5ae2bcf006599af4a8a1/integration/common/openlineage/common/sql/parser.py#L265) and needs to also be able to deal with an array?

PR #739, opened 2022-05-10
sql: parser accepts lists of sql statements

mobuchowski:
Closes: https://github.com/OpenLineage/OpenLineage/issues/734

### DIFF
--- a/integration/common/tests/sql/test_parser.py
+++ b/integration/common/tests/sql/test_parser.py
@@ -382,3 +382,18 @@ def test_parse_multi_statement():
     )
     assert sql_meta.in_tables == [DbTableMeta('schema0.table0')]
     assert sql_meta.out_tables == [DbTableMeta('schema1.table1')]
+
+
+def test_parse_statement_list():
+    sql_meta = parse(["""
+        DROP TABLE IF EXISTS schema1.table1;
+        CREATE TABLE schema1.table1(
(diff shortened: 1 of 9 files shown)


integration/common/tests/sql/test_parser.py
```diff
@@ -382,3 +382,18 @@ def test_parse_multi_statement():
...
+        DROP TABLE IF EXISTS schema1.table1;
+        CREATE TABLE schema1.table1(
+          col0 VARCHAR(64),
+          col1 VARCHAR(64)
+        );""", """
+        INSERT INTO schema1.table1(col0, col1)
+          SELECT col0, col1
+            FROM schema0.table0;
+        """])
+    assert sql_meta.in_tables == [DbTableMeta('schema0.table0')]
```
pawel-big-lebowski, line 398: Shouldn't we treat this as a list of separete queries and return list of `sql_meta`?
Going this way can cause a mixup of the inputs & outputs of different queries. 
We also lose the information about the query ordering.

mobuchowski: Then, somewhere we need to "join" it as users of parser expect single result.

I think if you have distinct jobs and want to get distinct results, you can call parser in a loop

pawel-big-lebowski: yes, indeed.

PR #774, opened 2022-05-24
java: spark to use Java client directly

mobuchowski:
This PR builds on top of @pawel-big-lebowski's PR https://github.com/OpenLineage/OpenLineage/pull/759 and makes Spark integration use OpenLineage Java client directly.

After this, Java client owns it's own, shaded Jackson version that
1. is the same version for every Spark version
2. does not collide with Spark's own Jackson

This PR largely deals with this: everything that serializes OpenLineage events should use OpenLineage provided Jackson, where serializing Spark stuff (like LogicalPlanSerializer) should use Spark provided one.


integration/spark/src/test/common/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
```diff
@@ -119,7 +119,7 @@ public void testRdd(SparkSession spark) throws IOException {
     List<OpenLineage.RunEvent> events = lineageEvent.getAllValues();
     assertEquals(2, events.size());
 
-    ObjectMapper objectMapper = OpenLineageClient.getObjectMapper();
+    ObjectMapper objectMapper = Utils.newObjectMapper();
```
pawel-big-lebowski, line 122: We already do have different kinds of `Utils` classes. 
If leaving `OpenLineageClient` with static method `getObjectMapper` and producer uri is not good enough (this would limit amount of files changed), consider more explicit class name like `EmitterUtils`, `OpenlineageClientUtils`, etc.

integration/spark/src/test/common/java/io/openlineage/spark/agent/lifecycle/SparkReadWriteIntegTest.java
```diff
@@ -253,14 +243,16 @@ public void testReadFromFileWriteToJdbc(@TempDir Path writeDir, SparkSession spa
             .isNotEmpty()
             .first();
     completionEvent
-        .extracting(RunEvent::getInputs, InstanceOfAssertFactories.list(InputDataset.class))
+        .extracting(
+            RunEvent::getInputs, InstanceOfAssertFactories.list(OpenLineage.InputDataset.class))
```
pawel-big-lebowski, line 247: minor: some changes get rid of OpenLineage prefix while other introduce it.

pawel-big-lebowski (approved):
Shading Jackson is a huge improvement.

integration/spark/src/test/common/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
mobuchowski: I can rename the class in follow-up PR - this way the PR won't get even more complicated, ok?

Issue #725, opened 2022-05-04
[python][client] Re-implement messages in DEBUG on client.py


howardyoo:
In client/python/openlineage/client/client.py , there was a code that created a INFO log that says,
`Constructing openlineage client to send events to {server_url}` in this part of the code: https://github.com/OpenLineage/OpenLineage/blob/16b24dae3d8703e8e94b5ae2bcf006599af4a8a1/client/python/openlineage/client/client.py#L69

However, in recent PR that got merged, the code was removed, and we no longer see the message from happening.
However, it was found out that the message was somewhat important for debugging and troubleshooting in Airflow, so would like to see if the code ca be reapplied, with DEBUG level log, right before this line of code: https://github.com/OpenLineage/OpenLineage/blob/16b24dae3d8703e8e94b5ae2bcf006599af4a8a1/client/python/openlineage/client/client.py#L34 to let the user know that the openlineage client was constructed.

Also might want to reintroduce
```
        if log.isEnabledFor(logging.DEBUG):
            log.debug(f"Sending openlineage event {event}")
```
for sending event info at: https://github.com/OpenLineage/OpenLineage/blob/16b24dae3d8703e8e94b5ae2bcf006599af4a8a1/client/python/openlineage/client/client.py#L65

PR #775, opened 2022-05-24
client: add debug messages on transport construction

mobuchowski:
This PR adds helpful debug messages on client constructor that would indicate where the message is going.

Closes https://github.com/OpenLineage/OpenLineage/issues/725


pawel-big-lebowski (approved):
LGTM. 
Minor: not sure if debug level is required for console transport.

Issue #726, opened 2022-05-04
[FEATURE] Support for  'disabling' openlineage


howardyoo:
In OpenLineage, we can set `OPENLINEAGE_URL` as the endpoint for sending openlineage data. However, there is a time when user may not choose to emit openlineage metrics - due to various reasons (testing, maintenance, troubleshooting, etc.). Currently, there is no way to stop lineage from being emitted, as long as the proper OPENLINEAGE_URL is specified.

Therefore, I would like to propose an additional env variable that can control whether the OPENLINEAGE should be enabled/disabled, by having the new variable `OPENLINEAGE_ENABLED`.

By default, if no such variables are found, it should be set to `True` where openlineage will continue to emit lineage data. However, when the variable is set to `OPENLINEAGE_ENABLED=False`, then even thought there may be a valid URL given, the client should not emit any lineage data to the server.

wslulciuc:
+1 as I've heard others in the wild requesting a similar option to toggle metadata collection for our integrations. My first thought would have been to add toggling support at the client level (`java`, `py` ), but that would mean the integration would extract metadata, but the client `emit()` call would be a no-op. So, with `OPENLINEAGE_ENABLED` we'll want to essentially disable the integration all together, meaning the integration would skip over any metadata extraction code.

howardyoo:
> +1 as I've heard others in the wild requesting a similar option to toggle metadata collection for our integrations. My first thought would have been to add toggling support at the client level (`java`, `py` ), but that would mean the integration would extract metadata, but the client `emit()` call would be a no-op. So, with `OPENLINEAGE_ENABLED` we'll want to essentially disable the integration all together, meaning the integration would skip over any metadata extraction code.

+1 on this, when `OPENLINEAGE_ENAGLED` is set to `false`, it should NOT ONLY stop emitting the events, but also should NOT DO any extractions or operations at all.

julienledem:
agreed with Willy as well. When disabled it should be a no-op that does nothing

kaxil:
> agreed with Willy as well. When disabled it should be a no-op that does nothing

+1 I agree

PR #780, opened 2022-05-25
clients: add OPENLINEAGE_DISABLED environment variable which overrides config to NoopTransport

mobuchowski:
Currently, configured OpenLineage integrations will always try to emit events. 
This PR adds environment variable OPENLINEAGE_DISABLE that when set to `true` overrides this - and will make integration use (added) NoopTransport when enabled. 

This does not override clients that were manually provided with other transports - just overrides the default transport factory construction.

Closes https://github.com/OpenLineage/OpenLineage/issues/726

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/Clients.java
+++ b/client/java/src/main/java/io/openlineage/client/Clients.java
@@ -1,5 +1,6 @@
 package io.openlineage.client;
 
+import io.openlineage.client.transports.NoopTransport;
 import io.openlineage.client.transports.Transport;
 import io.openlineage.client.transports.TransportFactory;
 
@@ -13,6 +14,10 @@ public final class Clients {
   }
 
   public static OpenLineageClient newClient(ConfigPathProvider configPathProvider) {
+    String isDisabled = Environment.getEnvironmentVariable("OPENLINEAGE_DISABLED");
+    if (Boolean.parseBoolean(isDisabled)) {
+      return OpenLineageClient.builder().transport(new NoopTransport()).build();
+    }
     final OpenLineageYaml openLineageYaml = Utils.loadOpenLineageYaml(configPathProvider);
     final TransportFactory factory = new TransportFactory(openLineageYaml.getTransportConfig());
     final Transport transport = factory.build();
--- a/client/python/openlineage/client/transport/noop.py
+++ b/client/python/openlineage/client/transport/noop.py
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: Apache-2.0.
+import logging
+
+from openlineage.client.run import RunEvent
+from openlineage.client.serde import Serde
+from openlineage.client.transport.transport import Transport, Config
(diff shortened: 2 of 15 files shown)


integration/airflow/openlineage/airflow/plugin.py
```diff
@@ -6,7 +8,12 @@
 # Provide empty plugin for older version
 from openlineage.airflow.macros import lineage_parent_id, lineage_run_id
 
-if parse_version(AIRFLOW_VERSION) < parse_version("2.3.0.dev0"):
+
+def _is_disabled():
+    return os.getenv("OPENLINEAGE_DISABLED", None) in [True, 'true', "True"]
```
collado-mike, line 13: Since we know that extractors can have an impact on the runtime, can we add similar logic to the `LineageBackend` so that no code is executed when OPENLINEAGE_DISABLED is true?

client/java/src/main/java/io/openlineage/client/Clients.java
```diff
@@ -13,6 +14,10 @@ public static OpenLineageClient newClient() {
   }
 
   public static OpenLineageClient newClient(ConfigPathProvider configPathProvider) {
+    String isDisabled = Environment.getEnvironmentVariable("OPENLINEAGE_DISABLED");
+    if (isDisabled != null && (isDisabled.equals("true") || isDisabled.equals("True"))) {
```
pawel-big-lebowski, line 18: consider using `equalsIgnoreCase`

client/python/openlineage/client/transport/noop.py
```diff
@@ -0,0 +1,24 @@
...
+class NoopConfig(Config):
+    pass
+
+
+class NoopTransport(Transport):
+    kind = 'noop'
+    config = NoopConfig
+
+    def __init__(self, config: NoopConfig):
+        log.debug("OpenLineage client is disabled. NoopTransport.")
```
pawel-big-lebowski, line 21: why not making it an `info` log?

client/java/src/main/java/io/openlineage/client/Clients.java
collado-mike: Use `Boolean.parseBoolean` - it does the null check and the case-insensitive comparison

mobuchowski: Thanks - `Boolean.parseBoolean` is exactly what I need.

integration/airflow/openlineage/airflow/plugin.py
mobuchowski: Done

client/python/openlineage/client/transport/noop.py
mobuchowski: Done.

mobuchowski:
Added disabling Spark.

collado-mike (approved):

Issue #807, opened 2022-06-01
[Integration][Airflow] Snowflake extractor retrieves connection info from secrets backends multiple times per task


josh-fell:
During task execution in Airflow with the latest version of OpenLineage configured, there are several log entries for retrieving connection information directly from the secrets backend. The `SnowflakeOperator` in Airflow [only does this retrieval once](https://github.com/apache/airflow/blob/d7dbfb7e26a50130d3550e781dc71a5fbcaeb3d2/airflow/providers/snowflake/hooks/snowflake.py#L311) via a [call to `SnowflakeHook._get_conn_params()`](https://github.com/apache/airflow/blob/d7dbfb7e26a50130d3550e781dc71a5fbcaeb3d2/airflow/providers/snowflake/hooks/snowflake.py#L155) when executing a query. This method in turn [calls `BaseHook.get_connection()`](https://github.com/apache/airflow/blob/7f264f09dda86fbf5ce2b57141745453f27d57bd/airflow/hooks/base.py#L58) to retrieve the connection information from the secrets backend. However, the Airflow task logs can have at least 4 other instances of this retrieval.

For example, this is an abbreviated and redacted example Airflow task log. Note that the log line `{base.py:68} INFO - Using connection ID 'galaxy_snowflake_etl' for task execution.` appears 5 times. The log line of `Using connection ID ... for task execution.` occurs each time connection information is retrieved from an Airflow secrets backend. This could be the Airflow metadatabase, a third-party secrets store, or environment variables.

```bash
...
*** Reading local file: /usr/local/airflow/logs/dag_id=cal***/run_id=scheduled__2022-05-28T01:00:00+00:00/task_id=cal***/attempt=8.log
[2022-05-28, 02:26:11 UTC] {taskinstance.py:1160} INFO - Dependencies all met for <TaskInstance: cal***** scheduled__2022-05-28T01:00:00+00:00 [queued]>
[2022-05-28, 02:26:11 UTC] {taskinstance.py:1160} INFO - Dependencies all met for <TaskInstance: cal***** scheduled__2022-05-28T01:00:00+00:00 [queued]>
[2022-05-28, 02:26:11 UTC] {taskinstance.py:1357} INFO -
--------------------------------------------------------------------------------
[2022-05-28, 02:26:11 UTC] {taskinstance.py:1358} INFO - Starting attempt 8 of 10
[2022-05-28, 02:26:11 UTC] {taskinstance.py:1359} INFO -
[...]

PR #820, opened 2022-06-03
Made listener thread use a queue rather than spinning up many new thr…

collado-mike:
### Problem

The Airflow task listener collects and emits lineage information in a separate thread that's spun up on each state change event. The thread is terminated after a couple of seconds so that a) the thread doesn't prevent the process from terminating and b) we try to give the thread a chance to actually emit the lineage information.

Unfortunately, as described in #807, the extra thread introduces the possibility of deadlock, especially since we terminate the thread manually without the opportunity to release any locks held.

Closes: #807

### Solution

This change alters the way we manage the lineage thread, starting a single thread at the beginning of the process and using a `Queue` to manage tasks that need to be executed in that thread. It also adds an exit hook that waits for the queue to be empty and then terminates the thread (with a short timeout). Since it's an exit hook, we can be sure that the Airflow task itself has completed before we attempt to terminate the thread. This should completely eliminate the possibility of deadlock.

### Checklist

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -436,7 +436,7 @@ jobs:
       - attach_workspace:
           at: .
       - run: docker build -f airflow/Dockerfile.tests -t openlineage-airflow-base .
-      - run: ./airflow/tests/integration/docker/up-failure.sh
+      - run: AIRFLOW_IMAGE=apache/airflow:2.3.1-python3.7 ./airflow/tests/integration/docker/up-failure.sh
       - store_artifacts:
           path: airflow/tests/integration/failures/airflow/logs
           destination: airflow-logs
--- a/integration/airflow/openlineage/airflow/listener.py
+++ b/integration/airflow/openlineage/airflow/listener.py
@@ -1,6 +1,10 @@
 import logging
 import threading
 import uuid
+from queue import Queue, Empty
+
+import atexit
+
 import attr
 
 from typing import TYPE_CHECKING, Optional, Callable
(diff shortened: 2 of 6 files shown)


mobuchowski (approved):
Looks good. I think there are two additional things. First, would be nice to document assumptions we have about behavior of threading mechanisms. Like

1. Sometimes extractors do things - like network or database calls - that live significantly longer than main thread. They can internally start non-daemonic threads.
2. We register `atexit` function, that run as part of Python's cleanup process when main thread exits. However, before it's called, it tries to join all the non-daemonic threads.
3. That means `TaskRunner.terminate` won't be called until those finish their job and terminate.
4. After they terminate and `TaskRunner.terminate` method is called, the OpenLineage integration has 2 seconds to emit events before it's torn down. If we didn't do this, Python would kill the daemonic thread indiscriminately - it does not care about properly finishing daemonic threads.

Also, I would love some kind of test for this. 
For example, to start non-daemon thread in extractor, that could sleep for like 10 seconds, then return results.

JDarDagran (approved):

collado-mike:
> Also, I would love some kind of test for this. For example, to start non-daemon thread in extractor, that could sleep for like 10 seconds, then return results.

I did go ahead and add a test for this - the sleeping executor was a good idea. I added a 30 second sleep and included the test in the `failures` suite. Despite the Github report, the latest build did pass - https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/3542/workflows/e06be856-0f9d-40d2-8f48-014df603a85e

PR #855, opened 2022-06-16
Fix Java client jar to stop bundling dependencies

collado-mike:
### Problem
A recent change made Jackson an `implementation` rather than `compileOnly` dependency. I actually think this is a good change, as I think that bundling dependencies is a strategy that should be used sparingly. I think it makes a lot of sense when deploying to a platform whose dependency versions are fixed and can't be easily upgraded (e.g., running on a Spark cluster or deploying a Spring/Dropwizard service). Outside of that, I don't think general-purpose libraries ought to bundle transitive dependencies, as it bloats jars and duplicates code on the classpath.

### Solution

Currently, the java client both includes Jackson as a transitive dependency _and_ bundles all the Jackson classes in the client jar. This change removes the bundling and solely includes Jackson as a transitive dependency. 

### Checklist

### DIFF
--- a/client/java/build.gradle
+++ b/client/java/build.gradle
@@ -65,8 +65,8 @@ dependencies {
     implementation 'org.apache.httpcomponents:httpclient:4.5.13'
     implementation 'commons-logging:commons-logging:1.2'
     implementation 'org.yaml:snakeyaml:1.30'
+    implementation 'org.slf4j:slf4j-api:1.7.36'
     compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
-    compileOnly 'org.slf4j:slf4j-api:1.7.36'
     compileOnly 'org.apache.kafka:kafka-clients:3.1.0'
     compileOnly "org.projectlombok:lombok:${lombokVersion}"
     annotationProcessor "org.projectlombok:lombok:${lombokVersion}"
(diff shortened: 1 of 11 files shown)


wjohnson:
Is this different than #853 ?

collado-mike:
> Is this different than #853 ?

Yeah, it's only half a fix. #853 changes how we bundle the dependencies in the Spark library - this only changes the bundling in the Java client.

pawel-big-lebowski:
Yeah, that's a tradeoff. You either have 3,4MB jar with all the dependencies bundled inside and hidden (shadowed) not to interfer with other client's classes on classpath. As an opposite, we can go with tiny 175KB jar which was build and tested in one context (like slf4j-api:1.7.36, jackson 2.13.2, etc...) and is run within different context as it gets the classes available on the classpath in different versions. Not to mention missing classes as Spark's `addJar` does not verify nor add transitive dependencies. 

So the tradeoff is: the size of jar vs the risk of version conflicts.

mobuchowski:
This will get us back to problem that shading dependencies solved: one integration, like Spark, can bundle different versions of some `provided` dependencies, some of which are not working with features of provided library that we use. And those problems are ones of the most unpleasant to solve. Having single version that we target, allows us to target only that version of the library - and make it easily testable.

However, we could do something like Spark Bigquery - publish two versions, one without dependencies for general use, second, shaded, for "special" use cases like `openlineage-java` for Spark.

collado-mike:
As a rule, I think general-purpose libraries should not bundle their dependencies. The OL Java client may be used anywhere by anyone. Bundling and renaming dependencies (especially one as popular as Jackson) introduces problems for users downstream. E.g., a downstream user can't use the off-the-shelf Jackson to interact with OL code - the annotations, etc., won't be recognized by their Jackson dependency, models won't implement the Jackson interfaces, etc.

I think the Spark build should handle the Spark-specific library issues. A given Spark version has a fixed Jackson version. If that Jackson version is incompatible with our needs, we can customize the build as necessary without worrying about affecting anyone who isn't using the Spark integration.

mobuchowski:
I'm okay with this @collado-mike as long as we're not relying on version provided by Spark. We've tried that, and it introduces too much headaches. Or drop Spark 2...

wslulciuc (approved):
LGTM 👍

PR #881, opened 2022-06-23
Reduce payload size by excluding local data and including output node in start events

collado-mike:
### Problem

We see consistently high payload sizes, almost always due to the `spark_unknown` facet. This tackles two of the root causes

1. The LocalRelation has a `data` field that includes the base64 encoded serialized binary of the local data sent to the executors. This can be many MBs of data.
2. The `AbstractQueryPlanOutputDatasetBuilder` is, by default, only defined for `SparkListenerSqlExecutionEnd` and `SparkListenerJobEnd` events. I believe the intention here was to only supply dataset version information _after_ the dataset has been changed. However, the result is that all nodes handled by subclasses of that builder are now serialized in the `spark_unknown` facet for start events.

### Solution
1. Filter the `data` field from LogicalPlans so we don't send unnecessary binary
2. Change `AbstractQueryPlanOutputDatasetBuilder` to be defined for all events, but override the `apply(SparkListenerEvent, LogicalPlan)` method so that implementations can selectively include dataset version information while still marking the node as _visited_.

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/AppendDataDatasetBuilder.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/AppendDataDatasetBuilder.java
@@ -34,12 +34,7 @@ public class AppendDataDatasetBuilder extends AbstractQueryPlanOutputDatasetBuil
   }
 
   @Override
-  public List<OpenLineage.OutputDataset> apply(AppendData x) {
-    return Collections.emptyList();
-  }
-
-  @Override
-  public List<OpenLineage.OutputDataset> apply(SparkListenerEvent event, AppendData appendData) {
+  protected List<OpenLineage.OutputDataset> apply(SparkListenerEvent event, AppendData appendData) {
     LogicalPlan logicalPlan = (LogicalPlan) (appendData).table();
 
     return delegate(
--- a/integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2RelationDatasetBuilderTest.java
+++ b/integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2RelationDatasetBuilderTest.java
@@ -21,13 +21,9 @@ import io.openlineage.spark3.agent.utils.DatasetVersionDatasetFacetUtils;
 import io.openlineage.spark3.agent.utils.PlanUtils3;
 import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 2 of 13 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/AppendDataDatasetBuilder.java
```diff
@@ -34,12 +34,7 @@ public boolean isDefinedAtLogicalPlan(LogicalPlan logicalPlan) {
   }
 
   @Override
-  public List<OpenLineage.OutputDataset> apply(AppendData x) {
-    return Collections.emptyList();
-  }
-
-  @Override
-  public List<OpenLineage.OutputDataset> apply(SparkListenerEvent event, AppendData appendData) {
+  protected List<OpenLineage.OutputDataset> apply(SparkListenerEvent event, AppendData appendData) {
```
wslulciuc, line 37: _Minor_: Since `apply()` was scoped as `public` in `AppendDataDatasetBuilder`, any concerns in limiting the scope to `protected`? We'll want to document the method scope change of the class in our changelog as this would be a breaking change.

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2RelationDatasetBuilderTest.java
```diff
@@ -106,50 +109,4 @@ private static Stream<Arguments> provideBuilders() {
             factory,
             openLineage));
   }
-
-  @ParameterizedTest
-  @MethodSource("provideBuildersWithSparkListeners")
-  public void testIsDefinedStSparkListenerEvent(
```
wslulciuc, line 112: Minor: It's not obvious why `testIsDefinedStSparkListenerEvent()` was removed.

wslulciuc (approved):
Left a couple minor comments, but nice work on identifying how/where to reduce the overall OL payload size 💯

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2RelationDatasetBuilderTest.java
collado-mike: It's always true now. I don't see a point to a test

pawel-big-lebowski:
Great job and thanks for spotting `AbstractQueryPlanOutputDatasetBuilder ` issue.  :100:

Issue #446, opened 2021-12-20
[PROPOSAL] Add generic ErrorMessageRunFacet


mobuchowski:
On execution failure, multiple systems provide text-based error messages, like dbt:

```
22:52:07  Database Error in model customers (models/customers.sql)
22:52:07    Syntax error: Expected "(" or keyword SELECT or keyword WITH but got identifier "PLEASE_REMOVE" at [56:12]
22:52:07    compiled SQL at target/run/jaffle_shop/models/customers.sql
```

Airflow:

```
airflow.exceptions.AirflowException: Bash command failed. The command returned a non-zero exit code.
```

Spark:
```
org.apache.spark.sql.AnalysisException: Table or view not found: wrong_table_name; line 1 pos 14;
```

We should add simple `ErrorMessageRunFacet` to capture human-readable messages. Currently, we have some custom facets for those for `BigQuery` and `Spark`

Proposed schema would look like that:

```json
{
  "$schema" : "https://json-schema.org/draft/2020-12/schema",
  "$id" : "https://openlineage.io/spec/facets/1-0-0/ErrorRunFacet.json",
  "$defs" : {
    "ErrorMessageRunFacet" : {
      "allOf" : [ {
        "$ref" : "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/RunFacet"
      }, {
        "type" : "object",
        "properties" : {
          "message" : {
            "description" : "A human-readable string representing error message generated by observed system",
            "type" : "string",
            "example" : "org.apache.spark.sql.AnalysisException: Table or view not found: wrong_table_name; line 1 pos 14"
          },
          "stackTrace" : {
            "description" : "A language-specific stack trace generated by observed system",
            "type" : "string",
            "example" : "Exception in thread \"main\" java.lang.RuntimeException: A test exception\nat io.openlineage.SomeClass.method(SomeClass.java:13)\nat io.openlineage.SomeClass.anotherMethod(SomeClass.java:9)"
          }
        },
        "required" : [ "message" ]
      } ],
      "type" : "object"
    }
  },
  "type" : "object",
  "properties" : {
    "error" : {
      "$ref" : "#/$defs/ErrorMessageRunFacet"
    }
  }
}
```

collado-mike:
[...]

PR #897, opened 2022-06-27
spec: add ErrorMessageRunFacet

mobuchowski:
Add generic `ErrorMessageRunFacet` as in proposal https://github.com/OpenLineage/OpenLineage/issues/446

This will be used instead of custom facets in Spark, Flink and other integrations that can report errors thrown by observed systems.


spec/facets/ErrorMessageRunFacet.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema" : "https://json-schema.org/draft/2020-12/schema",
+  "$id" : "https://openlineage.io/spec/facets/1-0-0/ErrorMessageRunFacet.json",
+  "$defs" : {
+    "ErrorMessageRunFacet" : {
```
julienledem, line 5: Please also add a blurb here: https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#standard-facets

integration/spark/shared/src/test/java/io/openlineage/spark/agent/OpenLineageRunEventTest.java
```diff
@@ -41,10 +41,12 @@ void testSerializeRunEvent() throws IOException, URISyntaxException {
...
-        ol.newRunFacets(
-            ol.newParentRunFacet(
-                ol.newParentRunFacetRun(runId), ol.newParentRunFacetJob("namespace", "jobName")),
-            null);
+        ol.newRunFacetsBuilder()
+            .parent(
+                ol.newParentRunFacet(
+                    ol.newParentRunFacetRun(runId),
+                    ol.newParentRunFacetJob("namespace", "jobName")))
+            .build();
```
julienledem, line 49: did you want to add use of this facet in this test case?

julienledem (changes requested):
This looks good to me. Thanks @mobuchowski 
I made some comments inline

collado-mike (approved):

client/python/openlineage/client/facet.py
```diff
@@ -195,3 +195,17 @@ def _get_schema() -> str:
 class ExternalQueryRunFacet(BaseFacet):
     externalQueryId: str = attr.ib()
     source: str = attr.ib()
+
+
+@attr.s
+class ErrorMessageRunFacet(BaseFacet):
+    """
+    This facet represents error message that was the result of a job run
```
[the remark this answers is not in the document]
mobuchowski: This is actually _the_ hardest part of _an_ English language, to remember to add those little words that do not change the meaning of _a_ message at all 😉

integration/spark/shared/src/test/java/io/openlineage/spark/agent/OpenLineageRunEventTest.java
mobuchowski: Added.

spec/facets/ErrorMessageRunFacet.json
mobuchowski: Added.

spec/facets/ErrorMessageRunFacet.json
```diff
@@ -0,0 +1,38 @@
...
+        "properties" : {
+          "message" : {
+            "description" : "A human-readable string representing error message generated by observed system",
+            "type" : "string",
+            "example" : "org.apache.spark.sql.AnalysisException: Table or view not found: wrong_table_name; line 1 pos 14"
+          },
+          "programmingLanguage" : {
+            "description" : "Programming language the observed system uses.",
+            "type" : "string",
+            "example" : "JAVA"
```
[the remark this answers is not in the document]
mobuchowski: I would decide on lowercase or uppercase, without having explicit enum as there are 100s of programming languages.

Having optional version is a good idea. We can "steal" more ideas later: https://github.com/OpenLineage/OpenLineage/issues/446#issuecomment-1006811072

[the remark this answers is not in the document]
mobuchowski: BTW, kinda philosophical question: maybe `programmingLanguage` and `languageVersion` should belong to separate facet? It still could be attached only when job fails. Do we want more "composable" facets or every possible information available in one?

spec/facets/ErrorMessageRunFacet.json
```diff
@@ -0,0 +1,38 @@
...
+          },
+          "stackTrace" : {
```
pawel-big-lebowski, line 21: Are we able to limit the size of `stakTrace` directly within a Java client to make sure this does not exceed X characters?

mobuchowski: As with all facets, client itself does not provide that functionality as facets code is autogenerated. We can add it on integration side.

spec/facets/ErrorMessageRunFacet.json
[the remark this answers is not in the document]
pawel-big-lebowski: Yeah, separate facet would be a better match. Perhaps this could be something like `platform`, so that it could include frameworks like Spark or Flink. This can be done in separate PR.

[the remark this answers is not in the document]
mobuchowski: We know framework/integration from `producer` field right now.

PR #914, opened 2022-07-08
airflow: Allow lineage metadata to flow through inlets and outlets

### LINKED ISSUE #384: airflow: Consider using manual inlets/outlets as a fallback for automatic extractor mechanism
There are way more operators than we can reasonably cover. Consider adding manual fallback mechanism that would cover cases where we don't have any extractors. 

Advantage of this approach is that it's generally easier for users to annotate dag than to modify it to use OpenLineage client manually. 

Disadvantage is that it might not work with future mechanism: https://github.com/apache/airflow/issues/17984

fenil25:
Signed-off-by: Fenil Doshi <fdoshi@slack-corp.com>

### Problem

Currently, in Airflow integration, lineage is only captured via a few extractors which greatly limits its use case. Airflow allows users to manually annotate lineage for the tasks using inlets and outlets. These can be a great fallback mechanism when extractors cannot automatically extract the input and output Datasets. 
The PR closes #384 

Addresses: #384 

### Solution

Airflow sends inlets and outlets information. Inlets and Outlets can be one of [these](https://github.com/apache/airflow/blob/main/airflow/lineage/entities.py) entities. The code only currently supports `Table` entity because OpenLineage does not track column-level lineage yet. 
- Check the TaskMetadata object that is created after corresponding extractors extract the lineage data and see whether inputs and outputs exist
- If TaskMetadata object does not have inputs/outputs, use Airflow's inlets and outlets.
- Get all the Table entities and convert them to Dataset entities that are suitable for TaskMetaData object

### Checklist


integration/airflow/openlineage/airflow/extractors/converters.py
```diff
@@ -0,0 +1,16 @@
+from openlineage.client.run import Dataset
+from airflow.lineage.entities import Table
+
+
+def table_to_dataset(table: Table):
+    return Dataset(
+        namespace=f"{table.cluster}",
+        name=f"{table.database}.{table.name}",
+        facets={
+            "tags": table.tags,
```
mobuchowski, line 10: Those should either be standard facets defined in OpenLineage standard, or defined somewhere custom facets.

We can do just `namespace` and `name` for now, and add this stuff later.

mobuchowski (changes requested):
We're missing integration test, but I guess this can be added la

mobuchowski (approved):

JDarDagran (approved):

PR #918, opened 2022-07-12
sql: explicitely specify to build source distribution

mobuchowski:
New Maturin release broke build: https://github.com/PyO3/maturin/blob/main/Changelog.md

```
Breaking Change: Don't build source distribution by default in maturin build command in https://github.com/PyO3/maturin/pull/955, --no-sdist option is replaced by --sdist
```


pawel-big-lebowski (approved):
👍

Issue #920, opened 2022-07-12
airflow: Mask secrets when capturing SQL


conorbev:
#870 allows capturing SQL once templates have been expanded. Airflow has a [mechanism](https://github.com/apache/airflow/blob/main/airflow/utils/log/secrets_masker.py) for redacting secrets from task logs and would should apply something similar/same in how we capture SQL in OpenLineage.

PR #921, opened 2022-07-13
airflow: redact secrets in UnknownOperatorAttributeRunFacet and taskInfo

mobuchowski:
After rendering templates, in those facets we can include fields that should be redacted. This PR uses Airflow mechanism for masking secrets to replace values of those fields with stars.

Closes: #920


integration/airflow/openlineage/airflow/extractors/manager.py
```diff
@@ -6,6 +6,20 @@
...
 
 
+def mask_secrets(task):
+    log = logging.getLogger()
+    try:
+        from airflow.utils.log.secrets_masker import redact
+
+        for field_name in task.template_fields:
+            rendered_value = getattr(task, field_name)
+            log.error(f"Masking field {field_name} with value {rendered_value}")
```
JDarDagran, line 16: Why is that on error level?

integration/airflow/tests/integration/tests/airflow/dags/secrets.py
```diff
@@ -0,0 +1,53 @@
...
+"""This shouldn't have extractor - we're testing if we'll see password in UnknownSourceAttribute"""
+class SecretsOperator(BaseOperator):
+    template_fields = ['password']
+
+    def __init__(self, password, **kwargs):
+        super().__init__(**kwargs)
+        self.password = password
+
+    def execute(self, context):
+        logging.getLogger(__name__).warning(f"Password! {self.password}")
```
JDarDagran, line 28: This brings me to thought it could be also unit tested with logs. See how SecretsMasker is tested in Airflow as an example with `caplog` fixture [here](https://github.com/apache/airflow/blob/main/tests/utils/log/test_secrets_masker.py)

integration/airflow/tests/integration/requests/secrets.json
```diff
@@ -0,0 +1,44 @@
...
+            "airflow_version": {
+                "taskInfo": "{{ not_match(result, '1500100900') }}"
+            },
+            "unknownSourceAttribute": {
+                "unknownItems": [
+                    {
+                        "name": "SecretsOperator",
+                        "properties": {
+                            "password": "***"
+                        }
```
JDarDagran, line 18: This test is fine but does not make us 100% sure that all secrets have been redacted from properties of the operator. This could be checked somewhere in unit test if that's easier to do?

JDarDagran (changes requested):

integration/airflow/openlineage/airflow/extractors/manager.py
mobuchowski: Debug info leftover :)

Issue #616, opened 2022-03-18
[INTEGRATION][SPARK] Implement dataset builders for alter table commands


pawel-big-lebowski:
Spark 3.2.1 introduced new alter table commands. In order to collect this information, new dataset builders need to be created. 

https://github.com/apache/spark/blob/v3.2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/v2AlterTableCommands.scala

mobuchowski:
Can we do it without introducing separate Spark 3.2 tree? Just with `hasClasses`?

tnazarew:
hi, please assign this to me.

PR #927, opened 2022-07-15
added dataset builder for AlterTableCommand

tnazarew:

### Problem

in spark 3.2 the AlterTable class functionality was extended and moved to AlterTableCommand. There was no dataset builder for this class in our integration.

Closes: #616 

### Solution

AlterTableCommandDatasetBuilder class was created to obtain data drom AlterTableCommand

### DIFF
--- a/integration/spark/spark32/build.gradle
+++ b/integration/spark/spark32/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     testAnnotationProcessor "org.projectlombok:lombok:${lombokVersion}"
 
     implementation(project(":shared"))
+    implementation(project(":spark3"))
 
     compileOnly "com.google.cloud.spark:spark-bigquery_2.12:${bigqueryVersion}"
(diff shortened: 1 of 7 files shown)


integration/spark/spark32/build.gradle
```diff
@@ -53,6 +53,7 @@ dependencies {
     testAnnotationProcessor "org.projectlombok:lombok:${lombokVersion}"
 
     implementation(project(":shared"))
+    implementation(project(":spark3"))
```
mobuchowski, line 56: I think it's better to specify dependencies manually here - it simplifies our internal dependency graph

tnazarew: can we do it different way? here we need stuff from `io.openlineage.spark3.agent.lifecycle.plan.catalog` and `io.openlineage.spark3.agent.utils`

mobuchowski (approved):

Issue #934, opened 2022-07-19
spark: Unable to find submodules when adding spark integration to other projects via gradle


varuntestaz:
I am trying to add openlineage-spark to my project with a similar line in `build.gradle`:
```
dependencies {
    api 'io.openlineage:openlineage-spark:0.11.0'
}
```

But it gives this error when I run `gradle shadowJar` or `gradle build`:
```
* What went wrong:
Could not determine the dependencies of task ':shadowJar'.
> Could not resolve all dependencies for configuration ':runtimeClasspath'.
   > Could not find spark:app:0.11.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/spark/app/0.11.0/app-0.11.0.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project : > io.openlineage:openlineage-spark:0.11.0
   > Could not find spark:shared:0.11.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/spark/shared/0.11.0/shared-0.11.0.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project : > io.openlineage:openlineage-spark:0.11.0
   > Could not find spark:spark2:0.11.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/spark/spark2/0.11.0/spark2-0.11.0.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project : > io.openlineage:openlineage-spark:0.11.0
   > Could not find spark:spark3:0.11.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/spark/spark3/0.11.0/spark3-0.11.0.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project : > io.openlineage:openlineage-spark:0.11.0
```

This used to work fine with version 0.9.0 and older.

varuntestaz:
Hi, thank you for the speedy solution.

[...]

PR #937, opened 2022-07-20
remove module metadata file from published artifacts

tnazarew:

### Problem

gradle module metadata file has our submodules as dependencies which causes problems when spark integration is used as dependency. 

Closes: #934

### Solution

For now we should be ok relying only on pom file so .module can be removed from published artifacts

### DIFF
--- a/integration/spark/build.gradle
+++ b/integration/spark/build.gradle
@@ -130,6 +130,10 @@ jacocoTestReport {
     }
 }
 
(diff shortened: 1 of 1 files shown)


collado-mike (approved):
After this is merged to main, let's test it by adding a snapshot dependency from https://datakin.jfrog.io/ui/repos/tree/General/maven-public-libs-snapshot-local

PR #939, opened 2022-07-20
Spark: add Kusto relation visitor

hmoazam:
### Problem

Openlineage does not support Azure Kusto (aka Azure Data Explorer) as an input or output.

### Solution

Implemented the KustoRelationVisitor to support lineage for [Azure Kusto's spark connector](https://github.com/Azure/azure-kusto-spark). In the process of implementing the class, we found the following: 

1. There are two paths in OpenLineage to the KustoRelationVisitor:
    * Write events reach the SaveIntoDataSourceCommandVisitor, so we included a condition to route to the KustoRelationVisitor's `createKustoDatasets` function. 
    * Read events trigger KustoRelationVisitor's `getNameSpace` function via the list of visitors instantiated in BaseVisitorFactory. 
2. When developing the `hasKustoClasses` method, we discovered that the function used by the other relation visitors to load the class - [`class.getClassLoader()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getClassLoader--) did not work as expected in all scenarios. It gives the expected result if the connector jar and its dependencies are added to the classpath, but this does not automatically happen when a library is installed on Spark, and requires an additional step from the user. For example, a user can install the Kusto Spark connector, and successfully interact with their Kusto data source, but would not see any lineage unless they had updated the classpath or copied the jars to a location already on the classpath. This behaviour is not specific to the Kusto connector - we also observed it with other connectors e.g. Apache Iceberg. We addressed this issue in `hasKustoClasses` by using [`Thread.currentThread().getContextClassLoader`](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#getContextClassLoader--) in addition to [`class.getClassLoader()`](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getClassLoader--). The former allows us to load all installed classes, regardless of whether or not they are on the classpath. [Apache Spark](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/util/Utils.scala#L208-L237) also uses the same approach.

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
@@ -1,0 +1,181 @@
+/*
+/* Copyright 2018-2022 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.spark.agent.lifecycle.plan;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.spark.api.DatasetFactory;
+import io.openlineage.spark.api.OpenLineageContext;
+import io.openlineage.spark.api.QueryPlanVisitor;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.reflect.FieldUtils;
+import org.apache.spark.api.java.Optional;
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
+import org.apache.spark.sql.execution.datasources.LogicalRelation;
+import org.apache.spark.sql.sources.BaseRelation;
+import org.apache.spark.sql.sources.CreatableRelationProvider;
+import org.apache.spark.sql.types.StructType;
+
+/**
+ * {@link LogicalPlan} visitor that matches the KustoRelation that comes from the Azure Kusto (Azure
+ * Data Factory). The KustoRelation is used to extract the table name and database name to populate
+ * the {@link OpenLineage.Dataset} during Kusto read operations.
+ */
+@Slf4j
+public class KustoRelationVisitor<D extends OpenLineage.Dataset>
+    extends QueryPlanVisitor<LogicalRelation, D> {
+
+  private final DatasetFactory<D> factory;
+  private static final String KUSTO_CLASS_NAME =
+      "com.microsoft.kusto.spark.datasource.KustoRelation";
+
+  private static final String KUSTO_PROVIDER_CLASS_NAME =
+      "com.microsoft.kusto.spark.datasource.DefaultSource";
+
+  private static final String KUSTO_URL_SUFFIX = ".kusto.windows.net";
+
+  private static final String KUSTO_PREFIX = "azurekusto://";
+
+  public KustoRelationVisitor(OpenLineageContext context, DatasetFactory<D> factory) {
+    super(context);
+    this.factory = factory;
+  }
+
+  protected boolean isKustoClass(LogicalPlan plan) {
+    try {
+      Class c = Thread.currentThread().getContextClassLoader().loadClass(KUSTO_CLASS_NAME);
+      return (plan instanceof LogicalRelation
+          && c.isAssignableFrom(((LogicalRelation) plan).relation().getClass()));
+    } catch (Exception e) {
+      // swallow - not a kusto class
+    }
+    return false;
+  }
+
+  public static boolean isKustoSource(CreatableRelationProvider provider) {
+    try {
+      Class c = Thread.currentThread().getContextClassLoader().loadClass(KUSTO_PROVIDER_CLASS_NAME);
+      return c.isAssignableFrom(provider.getClass());
+    } catch (Exception e) {
+      // swallow - not a kusto source
+    }
+    return false;
+  }
+
+  public static boolean hasKustoClasses() {
+    /**
+     * Checking the Kusto class with both KustoRelationVisitor.class.getClassLoader.loadClass and
+     * Thread.currentThread().getContextClassLoader().loadClass. The first checks if the class is
+     * present on the classpath, and the second one is a catchall which captures if the class has
+     * been installed. This is relevant for Azure Databricks where jars can be installed and
+     * accessible to the user, even if they are not present on the classpath.
+     */
+    try {
+      KustoRelationVisitor.class.getClassLoader().loadClass(KUSTO_PROVIDER_CLASS_NAME);
+      return true;
+    } catch (Exception e) {
+      // swallow - we don't care
+    }
+    try {
+      Thread.currentThread().getContextClassLoader().loadClass(KUSTO_PROVIDER_CLASS_NAME);
+      return true;
+    } catch (Exception e) {
+      // swallow - we don't care
+    }
+
+    return false;
+  }
+
+  @Override
+  public boolean isDefinedAt(LogicalPlan plan) {
+    return isKustoClass(plan);
+  }
+
+  private static Optional<String> getName(BaseRelation relation) {
+    String tableName = "";
+    try {
+      Object query = FieldUtils.readField(relation, "query", true);
+      tableName = (String) query;
+    } catch (IllegalAccessException | IllegalArgumentException e) {
+      log.warn("Unable to discover Kusto table property");
+      return Optional.empty();
+    }
+
+    if ("".equals(tableName)) {
+      log.warn("Unable to discover Kusto table property");
+      return Optional.empty();
+    }
+    // Check if the query is complex
+    if (StringUtils.countMatches(tableName, "|") > 0) {
+      tableName = "COMPLEX";
+    }
+
+    return Optional.of(tableName);
+  } // end of getName
+
+  private static Optional<String> getNamespace(BaseRelation relation) {
+    String url;
+    String databaseName;
+    String kustoUrl;
+    try {
+      Object kustoCoords = FieldUtils.readField(relation, "kustoCoordinates", true);
+      Object clusterUrl = FieldUtils.readField(kustoCoords, "clusterUrl", true);
+      Object database = FieldUtils.readField(kustoCoords, "database", true);
+
+      kustoUrl = (String) clusterUrl;
+      kustoUrl = kustoUrl.replace("https://", "");
+      databaseName = (String) database;
+      url = String.format("%s%s/%s", KUSTO_PREFIX, kustoUrl, databaseName);
+    } catch (IllegalAccessException | IllegalArgumentException e) {
+      log.warn("Unable to discover clusterUrl or database property");
+      return Optional.empty();
+    }
+    if ("".equals(url)) {
+      return Optional.empty();
+    }
+
+    return Optional.of(url);
+  }
+
+  public static <D extends OpenLineage.Dataset> List<D> createKustoDatasets(
+      DatasetFactory<D> datasetFactory,
+      scala.collection.immutable.Map<String, String> options,
+      StructType schema) {
+    // Called from SaveIntoDataSourceCommandVisitor on Kusto write operations.
+    List<D> output;
+
+    Map<String, String> javaOptions =
+        io.openlineage.spark.agent.util.ScalaConversionUtils.fromMap(options);
+
+    String name = javaOptions.get("kustotable");
+    String database = javaOptions.get("kustodatabase");
+    String kustoCluster = javaOptions.get("kustocluster");
+
+    String namespace =
+        String.format("%s%s%s/%s", KUSTO_PREFIX, kustoCluster, KUSTO_URL_SUFFIX, database);
+    output = Collections.singletonList(datasetFactory.getDataset(name, namespace, schema));
+    return output;
+  }
+
+  @Override
+  public List<D> apply(LogicalPlan x) {
+    BaseRelation relation = ((LogicalRelation) x).relation();
+    List<D> output;
+    Optional<String> name = getName(relation);
+    Optional<String> namespace = getNamespace(relation);
+    if (name.isPresent() && namespace.isPresent()) {
+      output =
+          Collections.singletonList(
+              factory.getDataset(name.get(), namespace.get(), relation.schema()));
+    } else {
+      output = Collections.emptyList();
+    }
+    return output;
+  }
+}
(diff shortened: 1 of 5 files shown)


wjohnson:
In this case, we don't have a test in place given the requirement for integration with an Azure cloud service called "Azure Data Explorer".  We'd love to contribute an integration test if we had an Azure environment to point to.

In addition, Hanna and I did a lot of research and were assisted by @mobuchowski,  @pawel-big-lebowski  and @collado-mike along with Microsoft engineers to uncover the `Thread.currentThread().getContextClassLoader` solution for packages instealled on Azure Databricks.  You can see much of the detective work [on stackoverflow](https://stackoverflow.com/questions/72985892/how-to-check-if-a-class-is-installed-on-databricks-through-a-spark-listener/73043752#73043752)

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
```diff
@@ -96,9 +96,17 @@ public List<OutputDataset> apply(SparkListenerEvent event, SaveIntoDataSourceCom
           command.schema());
     }
 
+    KustoRelationVisitor.isKustoSource(command.dataSource());
```
pawel-big-lebowski, line 99: Is this line necessary?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+    String database = javaOptions.get("kustodatabase");
+    String kustoCluster = javaOptions.get("kustocluster");
+
+    String namespace = KUSTO_PREFIX + kustoCluster + KUSTO_URL_SUFFIX + "/" + database;
+    output = Collections.singletonList(datasetFactory.getDataset(name, namespace, schema));
+    return output;
+  }
+
+  @Override
+  public List<D> apply(LogicalPlan x) {
```
pawel-big-lebowski, line 147: Would it be possible to provide a unit test for that method?

pawel-big-lebowski (commented):
Great to see we onboard Azure Kusto in Openlineage 💯 
I think `apply` method is worth writing a unit test.
Besides some really minor remarks to be considered.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      KustoRelationVisitor.class
+          .getClassLoader()
+          .loadClass("com.microsoft.kusto.spark.datasource.DefaultSource");
+      return true;
+    } catch (Exception e) {
+      // swallow - we don't care
+    }
+    try {
+      Thread.currentThread()
```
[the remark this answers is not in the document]
hmoazam: We kept both just in case there is a setup we didn't consider in which the thread classloader misses the class, but the class.getClassLoader captures it, in which case we would still return true. But happy to switch it to the single check using Thread.currentThread.getClassLoader if you recommend that

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+      return Optional.empty();
+    }
+
+    if (tableName == "") {
+      log.warn("Unable to discover Kusto table property");
+      return Optional.empty();
+    }
+    // Check if the query is complex
+    int n = tableName.length();
+    int countPipes = 0;
```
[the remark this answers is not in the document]
hmoazam: Will update to use countMatches. Could you please explain a bit more about the classpath? Should we check the classpath to see if the table name is given there?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
hmoazam: No sorry, my mistake!

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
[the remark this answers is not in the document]
mobuchowski: Spark uses commons-lang3, so we don't need to add dependency to it in our gradle.build.

mobuchowski:
To pass checks, make sure that DCO checks pass: https://github.com/OpenLineage/OpenLineage/pull/939/checks?check_run_id=7439211869 

and new files have correct copyright notice: https://github.com/OpenLineage/OpenLineage/runs/7439320110?check_suite_focus=true

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+    try {
+      Thread.currentThread()
+          .getContextClassLoader()
+          .loadClass("com.microsoft.kusto.spark.datasource.DefaultSource");
+      return true;
+    } catch (Exception e) {
+      // swallow - we don't care
+    }
+
+    return false;
```
collado-mike, line 67: We should generalize this so we don't have to keep copy/pasting the same boilerplate.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final String KUSTO_PREFIX = "azurekusto://";
+
+  public KustoRelationVisitor(OpenLineageContext context, DatasetFactory<D> factory) {
+    super(context);
+    this.factory = factory;
+  }
+
+  public static boolean isKustoClass(LogicalPlan plan) {
+    return plan instanceof LogicalRelation
+        && ((LogicalRelation) plan).relation().getClass().getName().equals(KUSTO_CLASS_NAME);
```
collado-mike, line 42: You might consider loading the class (as you do in `hasKustoClasses`), then using `Class.isAssignableFrom()` - this allows you to support subclasses in the future.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+    String tableName = "";
+    try {
+      Object query = FieldUtils.readField(relation, "query", true);
+      tableName = (String) query;
+    } catch (IllegalAccessException | IllegalArgumentException e) {
+      log.warn("Unable to discover Kusto table property");
+      return Optional.empty();
+    }
+
+    if (tableName == "") {
```
collado-mike, line 85: Use `tableName.equals("")`

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+      kustoUrl = (String) clusterUrl;
+      kustoUrl = kustoUrl.replace("https://", "");
+      databaseName = (String) database;
+      url = KUSTO_PREFIX + kustoUrl + "/" + databaseName;
+
+    } catch (IllegalAccessException | IllegalArgumentException e) {
+      log.warn("Unable to discover clusterUrl or database property");
+      return Optional.empty();
+    }
+    if (url == "") {
```
collado-mike, line 121: Use `"".equals(url)`

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,161 @@
...
+    List<D> output;
+
+    Map<String, String> javaOptions =
+        io.openlineage.spark.agent.util.ScalaConversionUtils.fromMap(options);
+
+    String name = javaOptions.get("kustotable");
+    String database = javaOptions.get("kustodatabase");
+    String kustoCluster = javaOptions.get("kustocluster");
+
+    String namespace = KUSTO_PREFIX + kustoCluster + KUSTO_URL_SUFFIX + "/" + database;
```
collado-mike, line 141: I like `String.format()` for these kinds of concatenations

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
```diff
@@ -96,9 +96,17 @@ public List<OutputDataset> apply(SparkListenerEvent event, SaveIntoDataSourceCom
...
+    if (KustoRelationVisitor.isKustoSource(command.dataSource())) {
+      return KustoRelationVisitor.createKustoDatasets(
+          outputDataset(), command.options(), command.schema());
+    }
+
     StructType schema = getSchema(command);
     LifecycleStateChange lifecycleStateChange =
         (SaveMode.Overwrite == command.mode()) ? OVERWRITE : CREATE;
+    // Look at the datasource coming through
+    log.error("command.dataSource: " + command.dataSource());
```
collado-mike, line 109: Is this meant to be checked in?

collado-mike (changes requested):

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
hmoazam: That's a great idea. Do you suggest we do that as part of this PR or a separate one @collado-mike? Seeing as it would affect the other datasources' relationVisitor implementations too.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
hmoazam: No it's not, thank you for spotting it!

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
hmoazam: One quick question on this point @collado-mike - if we want to go this route, we would need to first import the class right? Would we need to use compileOnly for that?

wjohnson: @collado-mike Are you saying  we will support subclassing as in if someone else has created a subclass of the Kusto spark connector?  While it's possible, I'm less confident someone would actually take the time to do so given it's a product and not entirely an open source compute engine.  That's not to say we don't want to discourage Kusto users from writing whatever code they need to but I've never seen a company modify / extend one of Microsoft's pre-built SDKs.

collado-mike: @hmoazam, no you don't need to import it - you can load the class as you do in `hasKustoClasses`, ie.,
```
Class c = getClass().getClassLoader()
          .loadClass("com.microsoft.kusto.spark.datasource.DefaultSource")
```
then use the `isAssignableFrom` method on the class object
```
return c.isAssignableFrom(plan.relation().getClass());
```

@wjohnson , I was actually thinking if the Kusto library itself ends up using subclasses for whatever reason. It's not uncommon for a library to end up using subclasses as part of a refactoring or adding new functionality. By using `isAssignableFrom` instead of `equals`, we just have a little more resilience to those changes in the future.

wjohnson: (insert Wayne's World 'We're Not Worthy' gif here).  Excellent insights as usual, Mike.  Thank you!

Fingers crossed that approach works from the ContextClassLoader workaround we have to do for Databricks.

hmoazam:
@pawel-big-lebowski, @mobuchowski, @collado-mike, thank you all for your helpful feedback! We've updated the code based on your suggestions, and thanks to @tnazarew's work with the unit tests, we also have a test for KustoRelationVisitor's `apply` method.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
wjohnson: @collado-mike - This one is marked as changes requested.  Do you still think it's necessary for this PR to amend other data sources or can we mark this as a separate issue and Hanna and I can look into doing a larger cleanup?

pawel-big-lebowski (approved):

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
collado-mike: I'll mark this statement resolved

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,181 @@
...
+    String tableName = "";
+    try {
+      Object query = FieldUtils.readField(relation, "query", true);
+      tableName = (String) query;
+    } catch (IllegalAccessException | IllegalArgumentException e) {
+      log.warn("Unable to discover Kusto table property");
+      return Optional.empty();
+    }
+
+    if ("".equals(tableName)) {
```
collado-mike, line 110: Since you're already using commons-lang, you can use `StringUtils.isBlank()`

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KustoRelationVisitor.java
```diff
@@ -0,0 +1,181 @@
...
+      log.warn("Unable to discover Kusto table property");
+      return Optional.empty();
+    }
+    // Check if the query is complex
+    if (StringUtils.countMatches(tableName, "|") > 0) {
+      tableName = "COMPLEX";
+    }
+
+    return Optional.of(tableName);
+  } // end of getName
```
collado-mike, line 120: unnecessary comment

collado-mike (approved):

wjohnson:
Congrats, @hmoazam for your second contribution!

Issue #714, opened 2022-04-29
Spark Library not linking path when using partitions


mdiazmonzon:
Currently spark operator saves the file path in S3 as the dataset name (mypath/schemaname/tablename). However when we read this path using a partition in another spark call the dataset name saves as mypath/schemaname/tablename/partition. We are unable to link these tasks and create lineage even that they are using the same filepath/tablename.

collado-mike:
Link to https://github.com/OpenLineage/OpenLineage/issues/435

dolfinus:
There is a discussion for altering this behavior: #3306

PR #940, opened 2022-07-21
[SPEC] include partitioning facet

pawel-big-lebowski:

### Problem

OpenLineage specification does not provide any suitable place for partitioning information

Part of: #714

### Solution

Introduce `PartitioningFacet` that can contain:
 * List of partition columns in case of table creation, altering, etc. 
 * List of partitions where each partition is defined as an array of `column` -> `value`. This can be useful when writing to a certain partitions. 

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
@@ -159,6 +159,7 @@ public class SaveIntoDataSourceCommandVisitor
                           ds.getFacets().getSchema(),
                           null,
                           null,
+                          null,
                           context
                               .getOpenLineage()
                               .newLifecycleStateChangeDatasetFacet(lifecycleStateChange, null));
--- a/spec/facets/PartitionDatasetFacet.json
+++ b/spec/facets/PartitionDatasetFacet.json
@@ -1,0 +1,63 @@
+{
+  "$schema" : "https://json-schema.org/draft/2020-12/schema",
+  "$id" : "https://openlineage.io/spec/facets/1-0-0/PartitionDatasetFacet.json",
+  "$defs" : {
+    "PartitionDatasetFacet" : {
+      "allOf" : [ {
+        "$ref" : "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/DatasetFacet"
+      }, {
+        "type" : "object",
+        "properties" : {
+          "columns": {
+            "type": "array",
+            "items": {
+              "type": "object",
+              "properties": {
+                "columnName": {
+                  "description": "Partitioning column name.",
+                  "type": "string",
+                  "example": "business_date"
+                }
+              }
+            }
+          },
+          "partitions": {
+            "type": "array", 
+            "items": {
+              "type": "object",
+              "properties": {
+                "identifier" : {
+                "description" : "List of partition keys and values describing a single partition.",
+                "type" : "array",
+                "items" : {
+                  "type" : "object",
+                  "properties" : {
+                    "column" : {
+                      "description" : "Partitioning column name.",
+                      "type" : "string",
+                      "example" : "business_date"
+                    },
+                    "value" : {
+                      "description" : "Partition value.",
+                      "type" : "string",
+                      "example" : "20220801"
+                    }
+                  },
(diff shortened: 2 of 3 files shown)


spec/facets/PartitionDatasetFacet.json
```diff
@@ -0,0 +1,63 @@
...
+        "$ref" : "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/DatasetFacet"
+      }, {
+        "type" : "object",
+        "properties" : {
+          "columns": {
+            "type": "array",
+            "items": {
+              "type": "object",
+              "properties": {
+                "columnName": {
```
collado-mike, line 16: Given the containing property is already `columns`, I'd just go with `name` here.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/SaveIntoDataSourceCommandVisitor.java
```diff
@@ -159,6 +159,7 @@ public List<OutputDataset> apply(SparkListenerEvent event, SaveIntoDataSourceCom
                           ds.getFacets().getSchema(),
                           null,
                           null,
+                          null,
```
mobuchowski, line 162: Maybe refactor this to a builder finally 😉

spec/facets/PartitionDatasetFacet.json
```diff
@@ -0,0 +1,63 @@
...
+              }
+            }
+          },
+          "partitions": {
+            "type": "array", 
+            "items": {
+              "type": "object",
+              "properties": {
+                "identifier" : {
+                "description" : "List of partition keys and values describing a single partition.",
```
collado-mike, line 30: Indenting here is off - I had a hard time parsing this.

Rather than wrapping this in an `identifier` with `column` and `value` properties, we could structure the partition records as just `column=value` properties. e.g., something like
```json
"partitions": {
    "type": "array",
    "items": {
        "type": "object",
        "additionalProperties": true
    }
}
```
which would allow you send items like
```json
{"partitions": [
    {"business_date": "2022-07-20"},
    {"countryCode": "US"}
]}
```

mobuchowski (commented):
Can we have more context about problem you're solving?

Are we going to guarantee that particular partitions aren't treated as "full" datasets anywhere, especially when a job reads from single partition?

spec/facets/PartitionDatasetFacet.json
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema" : "https://json-schema.org/draft/2020-12/schema",
+  "$id" : "https://openlineage.io/spec/facets/1-0-0/PartitionDatasetFacet.json",
+  "$defs" : {
+    "PartitionDatasetFacet" : {
+      "allOf" : [ {
+        "$ref" : "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/DatasetFacet"
```
collado-mike, line 7: I think the `DatasetFacet`s typically describe the nature of the dataset, rather than the portion of the dataset being processed. I like having the partition columns described in the dataset definition, but maybe the actual partition values belong in the `InputDataset` or `OutputDataset` facets

spec/facets/PartitionDatasetFacet.json
julienledem: +1

spec/facets/PartitionDatasetFacet.json
julienledem: are those the partitions that have been written by the run? If there are many this could be a bit verbose. Since the columns are captured above, maybe just an array of arrays?

julienledem (commented):
This is looking good. I put a few comments in there.
Could you add a description to the facet to clarify whether it is capturing the partitions written by the run?
Thanks

spec/facets/PartitionDatasetFacet.json
mobuchowski:
```
Rather than wrapping this in an identifier with column and value properties, we could structure the partition records as just column=value properties. e.g., something like
```

Unfortunately we lose static typing on the generated client side then.

collado-mike: It would be generated as a `Map<String, String>`, right?

pawel-big-lebowski:
It turns out that implementing partitioning concept within OpenLineage can be more complex than adding a single facet. 
I've created a separate proposal for that -> https://github.com/OpenLineage/OpenLineage/pull/1006

PR #949, opened 2022-07-26
flink: add failed job event

mobuchowski:
Add tests that check if job execution succeeds even if `OpenLineageFlinkJobListener` fails.


pawel-big-lebowski (commented):
It's great we test behavior of the app when endpoint is down. 
Based on our Spark integration experience, we know that lots of unpredictable things can happen. Especially for Flink integration jar built and tested within one Flink version which can be deployed with different Flink versions on a client side. Different versions may mean different Flink object and classes, etc.

I think it's worth having a single place like this:
```
 try {
      doMyFlinkStuff()
    } catch (Exception | NoClassDefFoundError | NoSuchMethodError e) {
      log.error("oops", e);
    }
```
that not only checks for the exceptions but also for the Java errors related to possible version mismatch.

pawel-big-lebowski (approved):

PR #951, opened 2022-07-26
add Apache Flink integration

mobuchowski:
This PR adds new experimental Apache Flink integration that has been in work for the last few months.

The integration is limited in scope: for now, it works only for Application mode jobs that use Kafka or Iceberg connectors.
There is a way of making it work with any custom connector by implementing `LineageProvider` interface that allows Sources or Sinks to return datasets by itself.

More about mechanism by which the integration works at https://openlineage.io/docs/integrations/flink


pawel-big-lebowski (commented):
Fantabulous PR 💯  🥇 

Fortunately, all the commits contained within PR were previously reviewed one after another before merging to `feature/flink`, so it does not make much sense looking into this. 

Three things that came to my mind: 
 * Do we have pmd turned for Flink? It's worth having it, not to lose the compliance badge.
 * Do we have jar publish stage for Flink in circle-cid? I couldn't spot it. 
 * Let's create a separate issue to split `openlineage-java` into smaller chunks. It has already grown too big. It does not make much sense to start Spark integration tests when doing Flink change.

pawel-big-lebowski (approved):

Issue #964, opened 2022-07-29
Databricks 10.4 Uses com.databricks.sql.managedcatalog.UnityCatalogV2Proxy


wjohnson:
Databricks Runtime 10.4 LTS uses `com.databricks.sql.managedcatalog.UnityCatalogV2Proxy` when querying a Hive table.

This causes an exception to occur and no input / output is returned for the relevant end of the query.

pawel-big-lebowski:
@wjohnson I was solving some databricks issue which was similar to this and merged it yesterday. It affected `create table as select` syntax. Did you check it directly on main?

wjohnson:
@pawel-big-lebowski  that's exactly what I'm experiencing.  I just confirmed even after the recent PR you made #950 the issue persists on Databricks Runtime 10.4.

PR #969, opened 2022-07-29
Support the new Databricks Unity V2 Proxy

wjohnson:
The Databricks Unity V2 Handler is a copy paste of the DatabricksDeltaHandler but uses the unity class name.

I believe it's worth separating out because we have no idea what Unity will turn into and it should be handled independently from Delta.

Signed-off-by: Will Johnson <will@willj.co>

### Problem

Starting with Databricks Runtime 10.4, Databricks introduced a UnityV2Proxy class that sits in front of the normal hive metastore. We now have to handle this new catalog layer.

Closes: #964 

### Solution

The Databricks Unity V2 Handler is a copy paste of the DatabricksDeltaHandler but uses the unity class name.

I believe it's worth separating out because we have no idea what Unity will turn into and it should be handled independently from Delta.

### Checklist

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java
@@ -1,0 +1,101 @@
+/*
+/* Copyright 2018-2022 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.spark3.agent.lifecycle.plan.catalog;
+
+import io.openlineage.spark.agent.util.DatasetIdentifier;
+import io.openlineage.spark.agent.util.PathUtils;
+import io.openlineage.spark.api.OpenLineageContext;
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.reflect.MethodUtils;
+import org.apache.hadoop.fs.Path;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.catalyst.TableIdentifier;
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+import scala.Option;
+
+/**
+ * The AbstractDatabricksHandler is intended to support Databricks' custom proxies which supersede
+ * the open source Delta Catalog. The proprietary class name of
+ * com.databricks.sql.transaction.tahoe.catalog.DeltaCatalog is used by Databricks instead of the
+ * open source class name of org.apache.spark.sql.delta.catalog.DeltaCatalog. It is used in the same
+ * way as the {@link DeltaHandler}.
+ */
+@Slf4j
+public abstract class AbstractDatabricksHandler implements CatalogHandler {
+  final String DATABRICKS_CLASS_NAME_STRING;
+
+  protected final OpenLineageContext context;
+
+  public AbstractDatabricksHandler(OpenLineageContext context) {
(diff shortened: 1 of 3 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DatabricksUnityV2Handler.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  @Override
+  public boolean isClass(TableCatalog tableCatalog) {
+    return "com.databricks.sql.managedcatalog.UnityCatalogV2Proxy"
+        .equals(tableCatalog.getClass().getCanonicalName());
+  }
+
+  @Override
+  public DatasetIdentifier getDatasetIdentifier(
```
pawel-big-lebowski, line 60: If it is a copy/paste from `DatabricksDeltaHandler` (as described in PR), please consider creating an `AbstractDatabricksHandler` class with `isClass` and `getDatasetIdentifier` methods implemented and let classes `DatabricksUnityV2Handler` and `DatabricsDeltaHandler` extend this abstract class.

wjohnson: @pawel-big-lebowski can do.  My only concern is that we have no idea how Unity is going to evolve.  Is it still worth handling if Unity suddenly grows to be a completely different beast than the way Databricks Delta catalog was being handled?

pawel-big-lebowski: We make use of non-public API. There is no guarantee on how do they evolve and there is little we can do about that. For the future versions we can create new Handlers within `spark32`, `spark33`.... projects.

wjohnson: Got it!  Thank you so much for the guidance, Pawel!  I really appreciate the perspective and assurance that we can handle this as they grow.

wjohnson: @pawel-big-lebowski  per your suggestion, I've refactored this to use an `AbstractDatabricksHandler` instead and noted their relationships.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * The AbstractDatabricksHandler is intended to support Databricks' custom proxies which supersede
+ * the open source Delta Catalog. The proprietary class name of
+ * com.databricks.sql.transaction.tahoe.catalog.DeltaCatalog is used by Databricks instead of the
+ * open source class name of org.apache.spark.sql.delta.catalog.DeltaCatalog. It is used in the same
+ * way as the {@link DeltaHandler}.
+ */
+@Slf4j
+public abstract class AbstractDatabricksHandler implements CatalogHandler {
+  final String DATABRICKS_CLASS_NAME_STRING;
```
pawel-big-lebowski, line 33: Uppercase with underscore is a naming convention for constants in Java, while `DATABRICKS_CLASS_NAME_STRING` is not such. Consider renaming it to `databricksClassNameString`.

pawel-big-lebowski (approved):
LGTM 💯 👍 
A single minor comment added.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java
wjohnson: Changes made!  Thank you so much, for the excellent guidance as usual, Pawel!

PR #973, opened 2022-08-01
flink: copy files in tests instead of db mount

mobuchowski:
This allows to run the Flink integration tests locally without needing to copy files to `/tmp/warehouse` first.


mzareba382:
Since `./gradlew examples:stateful:build` needs to be executed before integration tests, maybe we can include this step in the task or create new one just for testing on local machine?

mzareba382 (approved):

pawel-big-lebowski (approved):

PR #980, opened 2022-08-02
enable tests for app, fix bugs, disable LibraryTest.testRdd

tnazarew:

### Problem

`app` module was only running integration tests

Original solution assumed that the test will be enabled and all potential bugs fixed. That worked for all tests except `LibraryTest.testRdd`, which has two problems:
1. `  java.lang.NoSuchMethodError: io.openlineage.client.OpenLineageClientUtils.newObjectMapper()Lcom/fasterxml/jackson/databind/ObjectMapper;`
2. when ObjectMapper was changed there is issue with serialization, if test is run with `./gradlew app:test` the tests are passing but if it's run with `./gradlew app:build` the `OpenLineage.RunFacets.additionalProperties` is serialized as additional field and test is failing 

EDIT:
It turns out there are also issues with Delta.

Closes: potentially -> https://openlineage.slack.com/archives/C01CK9T7HKR/p1658868064009839

### Solution

It seems to be a blocking issue so the proposed solution for now is to disable `LibraryTest.testRdd` and `DeltaDataSourceTest.testInsertIntoDeltaSource` so everything is passing and in the meantime investigate both issues

### DIFF
--- a/integration/spark/app/build.gradle
+++ b/integration/spark/app/build.gradle
@@ -92,6 +92,7 @@ dependencies {
     testFixturesApi "org.apache.spark:spark-sql_${versions.scala}:${sparkVersion}"
     testFixturesApi "org.apache.spark:spark-hive_${versions.scala}:${sparkVersion}"
     testFixturesApi 'commons-beanutils:commons-beanutils:1.9.3'
+    testFixturesApi "io.delta:delta-core_2.12:1.0.0"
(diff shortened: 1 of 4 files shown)


pawel-big-lebowski (approved):
Please mention new issue related to disabled tests. 
Even though there are such tests, it's still better to merge it and turn on tests on `app`.

Issue #449, opened 2021-12-22
[PROPOSAL][SPARK] Support Azure Cosmos DB Spark Connector


wjohnson:
**Purpose:**

We would like to extend OpenLineage's Spark integration to work with the [Azure Cosmos DB Spark Connector for Spark3](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/cosmos/azure-cosmos-spark_3-1_2-12).

This would enable people using Spark, OpenLineage, and Cosmos DB to observe the input and output to their spark jobs.

**Proposed implementation**

The Cosmos DB Spark connector is implemented as a DataSourceV2Relation and DataSourceV2ScanRelation. As a result, I am proposing that the DataSourceV2RelationVisitor be extended much in the same way as it has been for Iceberg and Delta data types.

A new namespace and provider of "AzureCosmos" would be added.

The `Spark3VisitorFactoryImpl.java` must be updated to add `DataSourceV2RelationVisitor` as a common visitor since the `AppendDataVisitor` takes an input of only the common visitors and tries to cast the AppendData to one of the common visitors.

I believe this touches on #368 (AppendData and it depends on Provider).

A few caveats:

* The Cosmos DB relation does not use the provider property.
* The namespace must be derived from the table name which begins with "com.azure.cosmos".
* Reading a Cosmos DB is done so with a DataSourceV2ScanRelation which carries with it a relation property.
* The underlying relation is a DataSourceV2Relation

Lastly, this effort would be made easier with the acceptance / completion of the proposal in #437 .

PR #983, opened 2022-08-03
spark3: add Azure Cosmos Handler

hmoazam:
### Problem

Openlineage does not support Azure Cosmos DB as an input or output.

Closes: #449 

### Solution

We discovered that the [Cosmos DB spark connector](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/cosmos/azure-cosmos-spark_3_2-12) does not have TableCatalog, Identifier, TableProperties objects instantiated, but we can extract the table name and namespace from the relation's table object. In order to keep Cosmos specific code within a CosmosHandler, we defined a new interface - RelationHandler, the functions of which require only the DataSourceV2Relation. The CosmosHandler implements the RelationHandler interface. The RelationHandler interface can also be used to implement support for any future Spark 3 data sources which do not have TableCatalog, Identifier and TableProperties set. 

We have tested the implementation on Azure Databricks.

### Checklist

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
@@ -49,7 +49,7 @@ public class CosmosHandler implements RelationHandler {
 
     if (tableParts.length != expectedParts) {
       name = relationName;
-      namespace = relationName;
+      namespace = "azurecosmos";
     } else {
       namespace =
           String.format(
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
@@ -1,3 +1,8 @@
+/*
+/* Copyright 2018-2022 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
 package io.openlineage.spark3.agent.lifecycle.plan.catalog;
 
 import io.openlineage.spark.agent.util.DatasetIdentifier;
@@ -9,15 +14,22 @@ public class CosmosHandler implements RelationHandler {
 
   @Override
   public boolean hasClasses() {
+    String COSMOS_CATALOG_NAME = "com.azure.cosmos.spark.CosmosCatalog";
+
     try {
-      CosmosHandler.class.getClassLoader().loadClass("com.azure.cosmos.spark.CosmosCatalog");
+      CosmosHandler.class.getClassLoader().loadClass(COSMOS_CATALOG_NAME);
       return true;
     } catch (Exception e) {
-      log.info("Exception raised in CosmosHandler hasClasses {}", e);
       // swallow- we don't care
     }
-    return true; // TODO: figure out how to get the cosmos class - classpath issue that Kusto also
-    // has.
+    try {
+      Thread.currentThread().getContextClassLoader().loadClass(COSMOS_CATALOG_NAME);
+      return true;
+    } catch (Exception e) {
+      // swallow- we don't care
+
+    }
+    return false;
   }
 
   @Override
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
@@ -1,0 +1,54 @@
+package io.openlineage.spark3.agent.lifecycle.plan.catalog;
+
+import io.openlineage.spark.agent.util.DatasetIdentifier;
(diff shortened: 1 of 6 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
```diff
@@ -67,6 +67,19 @@ public static Optional<DatasetIdentifier> getDatasetIdentifier(
     }
   }
 
+  public static Optional<DatasetIdentifier> getDatasetIdentifierFromRelation(
```
pawel-big-lebowski, line 70: What's the benefit of having this as a separate `public` method? From the API perspective, it does the same as `getDatasetIdentifier` method above. Perhaps this could be a private static method called from the `getDatasetIdentifier`

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CosmosHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    String name;
+    String namespace;
+
+    String relationName = relation.table().name().replace("com.azure.cosmos.spark.items.", "");
+    int expectedParts = 3;
+    String[] tableParts = relationName.split("\\.", expectedParts);
+
+    if (tableParts.length != expectedParts) {
+      name = relationName;
+      namespace = relationName;
```
pawel-big-lebowski, line 52: Wouldn't be a namespace `azurecosmos` more gentle for the backend?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
hmoazam: Thank you for the feedback - we just updated it like you suggested.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
```diff
@@ -78,10 +98,26 @@ public static <D extends OpenLineage.Dataset> List<D> fromDataSourceV2Relation(
...
+      di = PlanUtils3.getDatasetIdentifier(context, relation);
+      if (!di.isPresent()) {
+        return Collections.emptyList();
+      } else {
+
+        datasetFacetsBuilder
+            .schema(PlanUtils.schemaFacet(openLineage, relation.schema()))
+            .dataSource(PlanUtils.datasourceFacet(openLineage, di.get().getNamespace()));
+
+        return Collections.singletonList(
```
pawel-big-lebowski, line 117: Do we need to return datasets like this? Wouldn't it be possible to include `else` implementation for `if (relation.identifier().isEmpty())` condition to establish `DatasetIdentifier` in a different way, and then return dataset facets list.

pawel-big-lebowski (approved):
I really like the idea of `RelationHandlers`. 
I've added a single minor comment which perhaps could improve the code. Please have a look at it.  
Besides this looks good to me. 💯

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
hmoazam: @pawel-big-lebowski, one note: updating the getDatasetIdentifier method as you suggested works well, but we do now need to update the following test: 

io.openlineage.spark3.agent.utils.PlanUtils3Test.testGetDatasetIdentifierFromV2RelationWithMissingIdentifier

It assumes that if the identifier is null, we return Optional.empty(), however with the relation handler, we now have the possibility of finding the dataset despite the identifier being null.

Would you suggest we modify the test to reflect the change?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/PlanUtils3.java
hmoazam: I moved the catalog handler specific code inside the `else` condition for `(relation.identifier().isEmpty())`, and so we have 

    ` datasetFacetsBuilder
        .schema(PlanUtils.schemaFacet(openLineage, relation.schema()))
        .dataSource(PlanUtils.datasourceFacet(openLineage, di.get().getNamespace()));
    
        datasetFactory.getDataset(
            di.get().getName(), di.get().getNamespace(), datasetFacetsBuilder.build())); `

Common to both the RelationHandler and CatalogHandler cases.

pawel-big-lebowski: Seems like some debug logging got committed. Please remove it.

hmoazam: This one `log.warn("Couldn't find identifier for dataset in plan {}", relation);`? In the previous code this was logged when no identifier was found. Should we remove it or switch it to `info` level?

pawel-big-lebowski: This one is fine. I think I saw some debugging logs. 
Anyway, they're not there anymore.

Issue #952, opened 2022-07-26
[PROPOSAL] Add application name to the specification


dpenmets:
A new parameter - Application Name needs to be added to both the job and dataset parameters.

howardyoo:
could you elaborate what this `application name` refers to?

julienledem:
I spoke with @dpenmets, the goal is to capture ownership of job and dataset.
I'm proposing to add an `ownership` facet to both `Dataset` and `Job` to capture this 
Example:
```
ownership: {
   owners: [ { name: "application:application_name"} ]
}
```
Schema:
```
{
  "$schema" : "https://json-schema.org/draft/2020-12/schema",
  "$id" : "https://openlineage.io/spec/facets/1-0-0/OwnershipDatasetFacet.json",
  "$defs" : {
    "OwnershipDatasetFacet" : {
      "allOf" : [ {
        "$ref" : "https://openlineage.io/spec/1-0-3/OpenLineage.json#/$defs/DatasetFacet"
      }, {
        "type" : "object",
        "properties" : {
          "owners" : {
            "description" : "The owners of the dataset.",
            "type" : "array",
            "items" : {
              "type" : "object",
              "properties" : {
                "name" : {
                  "description" : "The URN describing the owner of the Dataset",
                  "type" : "URN",
                  "example" : "application:app_name"
                },
                "type" : {
                  "description" : "The type of ownership (optional)",
                  "type" : "string",
                  "example" : "MAINTAINER"
                }
              },
              "required" : [ "name" ]
            }
          }
        }
      } ],
      "type" : "object"
    }
  },
  "type" : "object",
  "properties" : {
    "schema" : {
      "$ref" : "#/$defs/OwnershipDatasetFacet"
    }
  }
}
```

mobuchowski:
`OwnershipDatasetFacet` -> `DatasetOwnershipDatasetFacet` and `JobOwnershipDatasetFacet` - we had the same discussion with `DatasetVersionDatasetFacet` - it more clearly indicates what the ownership is about. Also, I think attaching the same facet to both `Run` and `Job` would require us to rework how the Java client generated code looks like.

`"type" : "URN"` - I would make that `String` - more permissive, I don't think there's anything about ownership that requires us to narrow it down to URN.

wslulciuc:
Great to finally see a proposal for dataset and job ownership! For clarity, my understanding of ownership rules are:

1. A **job** can be an **owner of** a **dataset**
[...]

PR #996, opened 2022-08-05
Create ownership facets

julienledem:
### Problem
capture ownership of jobs and datasets.

Closes: #952 

### Solution

I'm proposing to add an ownership facet to both Dataset and Job to capture this

- [x]  modifies one or more OpenLineage [facets](https://github.com/OpenLineage/OpenLineage/tree/main/spec/facets)

### Checklist


mobuchowski:
@julienledem I think description should reference `URN` only as example, as we aren't enforcing those to be in any particular format. And if we want to enforce it, then it should be explicitely specified, as well as having the reasoning explained

wjohnson:
Is there an integration that you had in mind when writing this spec?

I can see from the Databricks side that we can extract who RAN the job but that's not necessarily the owner of the job.

Do you envision the end user providing this information at runtime?

julienledem:
@mobuchowski I have updated the description. The goal is to have unique identifiers (as now mentioned in the description).

wslulciuc (approved):
Nice, @julienledem ❤️

QQ: [Marquez](https://github.com/MarquezProject/marquez) currently associates ownership to a namespace for datasets and jobs. Should we also consider adding an ownership facet for namespaces?

julienledem:
That's a good question @wslulciuc.  I'm not sure what's the right thing to do. I think we do need to capture Namespace related metadata in some way

Issue #185, opened 2021-08-12
Handling of Dataset Partitions


pomdtr:
If i have a job scheduled to push a partition to a table each hour.
And a job running daily consuming the 24 last partitions created by the previous job.
Would marquez be able to modelize this dependency ?

I don't see a way to use multiple datasetVersion as input in a request.

PR #1006, opened 2022-08-09
proposal to handle partitioning

pawel-big-lebowski:

### Problem

Dataset partitioning is not supported.

Closes: #185 

### Solution

Propose some way to handle partitioning. 

### Checklist

### DIFF
--- a/proposals/185/partitioning.md
+++ b/proposals/185/partitioning.md
@@ -1,0 +1,47 @@
+# Partitioning proposal
+
+Partitioning is a concept of dividing datasets into parts based on values of particular columns like date, city, etc. For Apache Hive and Spark partitioning affected physical location of data, as each partition is stored in a separate directory. This serves in a similar way to database indexing as accessing certain partitions requires reading only some of data directories. 
+
+It is a common usecase to have a large dataset containing data collected over years, while most of the processing require only recent data to proceed, like the last day. Based on that, partitioning is an important concept within data processing and deserves proper OpenLineage represantion.
+
+## Dataset partitioning columns
+
+Specifying partition columns is part of a dataset schema. That's why it should extend existing `SchemaDatasetFacet` and enrich it with an array of `partitionFields` containing names of some of the schema fields. Providing a `partitionFields` would give an extra information on how is the dataset organised. `partitionFields` should be collected for both input and output datasets. 
+
+## Partitions being modified and accessed
+
+Although the dataset may contain a plenty of partitions, only a small subset of them is read or written most of the time. However, it is difficult to find the proper balance for collecting such information. For example, providing an information on partitions read, may result in thousands of useless entries sent within a single OL event. On the other hand, if a certain partition is affected by an upstream issue, it is valuable to know it. 
+
+It may be then worth to collect at least information on created/dropped partiions:
+```
+ALTER TABLE table_identifier ADD PARTITION ... 
+ALTER TABLE table_identifier DROP PARTITION ...
+```
+
+This could be sent within `PartitionsDatasetFacet`:
+```
+partitions: {
+    affected: 
+            [
+                {
+                    col1: value1,
+                    col2: value2, 
+                    ...
+                }
+            ],
+    operationType: ADD|DROP|WRITE
+}
+```
+
+## Partition date as a run facet 
+
+Partitioning concept is often related with job frequency and scheduling although the term `partitioning` does not apply well within this context. If an enriched dataset is partitioned daily or hourly, then the job run to append the data is often run daily or hourly. 
+
+OpenLineage should be capable of colleting such information. Otherwise, it is tempting to include partition name in job name which can lead to several duplicate jobs. This can be solved by an adding an extra `RunParameterRunFacet` facet within `RunFacet` to store any:
+```
+{
+    key1: value1,
+    key2: value2
+}
+```
+key/value parameters describing the job run. 


proposals/185/partitioning.md
```diff
@@ -0,0 +1,47 @@
...
+                    col1: value1,
+                    col2: value2, 
+                    ...
+                }
+            ],
+    operationType: ADD|DROP|WRITE
+}
+```
+
+## Partition date as a run facet
```
collado-mike, line 36: As stated, I think the information about the partition(s) affected should be part of the Input/Output DatasetFacets. I think the run args generally captures things like the partition date - it's hard to guess which of the arguments is actually referencing a partition field (e.g., there could be country code, city name, etc) or the arguments may indirectly imply a partition (e.g., partition date could be the `startDate` minus one hour).

proposals/185/partitioning.md
```diff
@@ -0,0 +1,47 @@
...
+            ],
+    operationType: ADD|DROP|WRITE
+}
+```
+
+## Partition date as a run facet 
+
+Partitioning concept is often related with job frequency and scheduling although the term `partitioning` does not apply well within this context. If an enriched dataset is partitioned daily or hourly, then the job run to append the data is often run daily or hourly. 
+
+OpenLineage should be capable of colleting such information. Otherwise, it is tempting to include partition name in job name which can lead to several duplicate jobs. This can be solved by an adding an extra `RunParameterRunFacet` facet within `RunFacet` to store any:
```
collado-mike, line 40: I do think we're missing a general run parameters facet, so I definitely think this ought to be addressed. I don't think it's relevant to the partition information, though.

proposals/185/partitioning.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Dataset partitioning columns
+
+Specifying partition columns is part of a dataset schema. That's why it should extend existing `SchemaDatasetFacet` and enrich it with an array of `partitionFields` containing names of some of the schema fields. Providing a `partitionFields` would give an extra information on how is the dataset organised. `partitionFields` should be collected for both input and output datasets. 
+
+## Partitions being modified and accessed
+
+Although the dataset may contain a plenty of partitions, only a small subset of them is read or written most of the time. However, it is difficult to find the proper balance for collecting such information. For example, providing an information on partitions read, may result in thousands of useless entries sent within a single OL event. On the other hand, if a certain partition is affected by an upstream issue, it is valuable to know it.
```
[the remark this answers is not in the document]
pawel-big-lebowski: I've seen that: a job writing a single daily partition required reading 30 daily partitions of another table.
Perhaps this should be limited during an implementation: if there are 100 partitions read, do not include this in OL event.

proposals/185/partitioning.md
pawel-big-lebowski: The relation between partition(s) and run-args is indeed tricky but in most cases Airflow DAG scheduled knows which partition it is going to handle. And we should be able to somehow let the users pass this information to OL event and let them specify that certain run args are referencing partition fields.

proposals/185/partitioning.md
pawel-big-lebowski: In many cases we will not be able to extract partition information automatically from Spark jobs or SQLs executed. That's why we need to be able to get this information from the run parameters or even let users specify which partition is affected.

proposals/185/partitioning.md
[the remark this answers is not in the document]
collado-mike: TBH, I don't consider that even remotely as harmful or irritating as the bigger sources of excess information we get now.  And I think we can structure the partition information so that the incremental cost of each partition is small. E.g., following your `PartitionsDatasetFacet` example, something like the following adds ~50 characters for each partition.
```
{ 
    "partitionsRead": 
        [ 
            {"date": "2022-08-01T00:00:00Z", "countryCode": "CA"}
        ]
}
```
100 partitions means ~5KB of data. The big OL messages we get now are sometimes many MBs, so I'm not too worried about an extra KB or two. I think setting a limit is reasonable, but I'd set that limit pretty high.

proposals/185/partitioning.md
collado-mike: > And we should be able to somehow let the users pass this information to OL event and let them specify that certain run args are referencing partition fields.

This yes. 100%

proposals/185/partitioning.md
```diff
@@ -0,0 +1,47 @@
...
+
+Although the dataset may contain a plenty of partitions, only a small subset of them is read or written most of the time. However, it is difficult to find the proper balance for collecting such information. For example, providing an information on partitions read, may result in thousands of useless entries sent within a single OL event. On the other hand, if a certain partition is affected by an upstream issue, it is valuable to know it. 
+
+It may be then worth to collect at least information on created/dropped partiions:
+```
+ALTER TABLE table_identifier ADD PARTITION ... 
+ALTER TABLE table_identifier DROP PARTITION ...
+```
+
+This could be sent within `PartitionsDatasetFacet`:
```
julienledem, line 21: to be a bit more precise here, I would say that:
 - We have an [InputDataset facet](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#input-dataset-facets) that specifies the partitions that have been read.
 - We have an [OutputDataset facet](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#output-dataset-facets) that specifies the partitions that have been replaced, added or dropped.

proposals/185/partitioning.md
julienledem, line 36: The [nominal time facet](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.md#run-facets) is meant to capture a similar concept. It is somewhat assuming the time interval the job is scheduled for matches the output partition produced.

proposals/185/partitioning.md
```diff
@@ -0,0 +1,47 @@
...
+```
+partitions: {
+    affected: 
+            [
+                {
+                    col1: value1,
+                    col2: value2,
```
julienledem, line 28: If the output partition is derived from an input parameter, possibly you can capture this relation here. 
```col1 : { value: "value1", derivedFrom: "runParam:foo"}```
or something like that

julienledem (commented):
I think this is pretty close if we add a few more details

pawel-big-lebowski:
I am not sure if this is what we need. Closing.

PR #1029, opened 2022-08-22
[Integration][Common][GE] Limit upper GreatExpectations version.

JDarDagran:
Signed-off-by: Jakub Dardzinski <kuba0221@gmail.com>

### Problem

The issue introduced with `0.15.19` version update of Great Expectations has been fixed by a noble PR #1025 (thanks @collado-mike!). This PR is only to limit GE version.

PS. I could post a separate issue to run tests in some way to test against most recent versions of OL integrations.

### DIFF
--- a/integration/common/setup.py
+++ b/integration/common/setup.py
@@ -34,7 +34,7 @@ extras_require = {
         "pyyaml>=5.3.1"
     ],
     "great_expectations": [
-        "great_expectations>=0.13.26",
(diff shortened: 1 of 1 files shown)


julienledem (approved):
We should merge this to fix main branch build. 
IMO, We should also 2 the two following things:
 - open a separate issue to make it work with the latest version of GE
 - Have a separate test that checks the latest version of GE so that we keep getting notified when a new release breaks the integration.

Issue #1012, opened 2022-08-12
Frequent timeout while using latest version of 0.12.0 jar in spark lineage


kumar-gaurav:
The default value of transport in HTTPConfig is set as 5 sec, causing frequent timeout while sending lineage info from spark pipeline.

please increase default value to 60 sec.also, how to set this timeout from configuration?

merobi-hub:
Hello, thanks for opening an issue and welcome to the community! We'll respond as soon as possible. In the meantime, we hope you'll join us on Slack: http://bit.ly/OpenLineageSlack.

howardyoo:
Hi, could you be more specific on the case of how this timeout is occurring?
Perhaps, something that can be scripted to reproduce this issue would be helpful. Thanks!

pawel-big-lebowski:
Hi @kumar-gaurav, I think 5 secs may be good enough as default, but we do miss the ability to provide custom configuration. It is available in `openlineage-java` library but unfortunately spark integration library does not allow this. 

The solution to this would be: 
 * include `timeout` in ArgumentParser -> https://github.com/OpenLineage/OpenLineage/blob/0.12.0/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
 * Make use of the parameter when building http transport -> https://github.com/OpenLineage/OpenLineage/blob/0.12.0/integration/spark/app/src/main/java/io/openlineage/spark/agent/EventEmitter.java#L60
 * Describe parameter in readme: https://github.com/OpenLineage/OpenLineage/blob/0.12.0/integration/spark/README.md
* and describe it in docs: https://github.com/OpenLineage/docs/blob/main/docs/integrations/spark/spark.md

PR #1050, opened 2022-08-29
timeout configurable from url and spark conf

tnazarew:

### Problem

in some cases when openlineage event is send it takes longer than 5 seconds to respond, which exceeds the timeout value.

Closes: #1012

### Solution

Make timeout configurable by user.

### Checklist

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
@@ -209,6 +209,9 @@ public final class HttpTransport extends Transport implements Closeable {
       httpConfig.setUrl(uri);
       httpConfig.setAuth(tokenProvider);
       httpConfig.setTimeout(timeout);
+      if (httpClient != null) {
+        return new HttpTransport(httpClient, httpConfig);
+      }
       return new HttpTransport(httpConfig);
     }
   }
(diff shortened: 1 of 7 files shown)


client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -149,6 +149,7 @@ public static final class Builder {
 
     private @Nullable CloseableHttpClient httpClient;
     private @Nullable TokenProvider tokenProvider;
+    private @Nullable Double timeout;
```
mobuchowski, line 152: Add test for configuring this in client.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -202,9 +208,7 @@ public HttpTransport build() {
       final HttpConfig httpConfig = new HttpConfig();
       httpConfig.setUrl(uri);
       httpConfig.setAuth(tokenProvider);
-      if (httpClient != null) {
```
mobuchowski, line 205: Not sure why you removed this - if we're getting passed another, externally configured HttpClient we should use it.

mobuchowski: Also this fails tests, since we pass mock client for tests this way 😉

mobuchowski (changes requested):

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
tnazarew: yeah, that was not intentional :s

mobuchowski (approved):

PR #1053, opened 2022-08-30
sql: add macos binary package build

mobuchowski:
Currently, sql parser has binary builds only for `linux` systems. MacOS users have to install the library from source, which requires pulling whole rust toolchain and maturin build tool.

This PR adds CI step which builds MacOS universal binary wheel - it works both on x86 Intel macs and ARM M1 macs. 


mzareba382 (approved):
Tested on Intel machine with macOS Monterey 12.5.1, works correctly.

PR #1094, opened 2022-09-13
sql: add test for create table as select

mobuchowski:


pawel-big-lebowski (approved):
such tests work better than docs 🚀

PR #1107, opened 2022-09-22
dbt: remove requirement for OPENLINEAGE_URL to be set

mobuchowski:
dbt-ol erroneously required OPENLINEAGE_URL to be set. 

Now, the check is removed, and it can be initialized with openlineage.yml also.


JDarDagran (approved):
That's a great change, Maciej! Keep up doing an A1 work!

PR #1122, opened 2022-09-28
python: remove sqlparse dependency, add optional backup dependency

mobuchowski:
`sqlparse` - Python library, not `sqlparser-rs` - should not be installed by default, as it's unused with systems where Rust parser works.

This add option to install `sqlparse` separately if needed.

### DIFF
--- a/integration/airflow/dev-requirements-1.x.txt
+++ b/integration/airflow/dev-requirements-1.x.txt
@@ -1,3 +1,3 @@
 -e ../../client/python
--e ../common[sql]
+-e ../common[python-sql]
 -e .[dev,airflow-1]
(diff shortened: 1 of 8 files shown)


integration/airflow/dev-requirements-1.x.txt
```diff
@@ -1,3 +1,3 @@
 -e ../../client/python
--e ../common[sql]
+-e ../common[backup-sql]
```
collado-mike, line 2: Maybe `python-sql` to be a bit clearer?

mobuchowski: Ok, changed to that.

JDarDagran (approved):

Issue #1146, opened 2022-10-10
sql: implement Java bindings for SQL parser library


mobuchowski:
Current SQL parser library has available bindings for Python. We want to use SQL parsing capabilities in other Java libraries, such as Spark integration. Create JNI-based Java bindings to integrate parser with potential new libraries

Issue #1052, opened 2022-08-29
failure to extract information by OL integration should be indicated in facet 


mobuchowski:
If we fail to extract lineage information - for example, our sql parser chokes on some statement - we should send facet that notifies OpenLineage consumer about in, in addition to logging the error and sending "empty" `START` or `COMPLETE` event. 

This should be run facet - since we can fail to discover datasets. Example name is `OpenLineageExtractionErrorRunFacet`.

This is in contrast to sending `FAIL` event, which notifies that actual processing job has failed.

PR #1172, opened 2022-10-24
sql: rust parser interface for java

mobuchowski:
This PR implements Java interface to Rust SQL parser, with build script and native library loading mechanism.

Code is taken from https://github.com/StarostaGit/OpenLineage/pull/2, with added CI support and minor build fixes.

This code will require few follow-up changes to make it fully done:
- add CI release job
- document SQL parser contract, as it's used between languages, not only as Python integration's internal detail

Closes: #1146


integration/sql/iface-java/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "openlineage_sql_java"
+description = "Java interface for the Rust OpenLineage lineage extraction library"
+version = "0.15.0"
+edition = "2021"
```
pawel-big-lebowski, line 5: Is this an edition of a package?

integration/sql/iface-java/src/java/io/openlineage/sql/OpenLineageSql.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.nio.file.Files;
+import java.util.List;
+
+public final class OpenLineageSql {
+    public static native SqlMeta parse(List<String> sql, String dialect, String defaultSchema) throws RuntimeException;
```
pawel-big-lebowski, line 18: Would it make sense to return Optional<SqlMeta> while leaving all the exceptions within this class. 
There is little we can do with them in other class.

integration/sql/README.md
```diff
@@ -2,15 +2,15 @@
 
 Library that extracts lineage from sql statements. 
 
-### Usage
+## Usage
 
 If you're using OpenLineage integration, there's good chance that you're already using this integration.
 
-This library can be used both as Python library and as Rust library, however it's not published at Cargo yet.
+This library is implemented in Rust and provides a Python and Java interface. The Rust implementation has not yet been published to Cargo.
```
pawel-big-lebowski, line 9: Wouldn't be better to have separate libraries for python and java?

integration/sql/iface-java/src/java/io/openlineage/sql/SqlMeta.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class SqlMeta {
+    private final List<DbTableMeta> inTables;
+    private final List<DbTableMeta> outTables;
+
+    public SqlMeta(List<DbTableMeta> in, List<DbTableMeta> out) {
```
pawel-big-lebowski, line 17: will this abstraction suit well for column-lineage or is this out-of-scope of this PR ?
Asking this because we want table meta and column lineage meta in a single rust parser call and perhaps it makes sense to consider this although leaving column-lineage objects empty for now.

integration/sql/iface-java/tests/java/io/openlineage/sql/OpenLineageSqlTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    void parseError() {
+        boolean exceptionCaught = false;
+        try {
+            SqlMeta output = OpenLineageSql.parse(Arrays.asList("Definitely not an SQL statement"));
+        } catch (Exception e) {
+            exceptionCaught = true;
+            assertTrue(e instanceof RuntimeException);
```
pawel-big-lebowski, line 41: Are we able to distinguish parsing errors from errors related to library installation and platform?

integration/sql/iface-java/pom.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>io.openlineage.sql</groupId>
+  <artifactId>openlineage-sql-java</artifactId>
+  <version>1.0-SNAPSHOT</version>
```
pawel-big-lebowski, line 9: hardcoded version and url pointing to example.com.

integration/sql/iface-java/src/java/io/openlineage/sql/DbTableMeta.java
```diff
@@ -0,0 +1,69 @@
...
+        if (o == this) {
+            return true;
+        }
+
+        if (!(o instanceof DbTableMeta)) {
+            return false;
+        }
+
+        DbTableMeta other = (DbTableMeta)o;
+        return qualifiedName().equals(other.qualifiedName());
```
pawel-big-lebowski, line 58: Qualified name has no dots or other separators.
So `equals` can be true even when being not equal.

pawel-big-lebowski (commented):
This is going to be hot 🌶️ 
Put some questions to help me understand the code.

integration/sql/README.md
mobuchowski: Not sure what you mean - they are not one single binary/file that provides both interfaces. This means that the Python and Java interfaces depend on the same native code.

integration/sql/iface-java/Cargo.toml
mobuchowski: Rust language edition. Basically equivalent to Java versions, like 8 or 11: https://doc.rust-lang.org/edition-guide/editions/index.html

integration/sql/iface-java/src/java/io/openlineage/sql/SqlMeta.java
mobuchowski: This is out of scope for this PR, which just wants to achieve parity between Java and Python interfaces.
Separate PR will provide interface for column-level lineage.

integration/sql/iface-java/pom.xml
mobuchowski: Removed Maven

integration/sql/iface-java/src/java/io/openlineage/sql/OpenLineageSql.java
mobuchowski: This is `native` method and it's not fun to change it's signature as it requires to change deep bowels of Rust code here https://github.com/OpenLineage/OpenLineage/blob/337e5c719580ad3ac40874f723808d8d9ebad403/integration/sql/iface-java/src/lib.rs#L13. 

We can have wrapper methods returning Optional easily though.

integration/sql/iface-java/src/java/io/openlineage/sql/DbTableMeta.java
mobuchowski: Is there database where `database`, `schema` and `table` are separated by anything else than a dot? If yes that's a problem, but I don't know that it exists.

integration/sql/iface-java/tests/java/io/openlineage/sql/OpenLineageSqlTest.java
mobuchowski: Will be done in separate PR: https://github.com/OpenLineage/OpenLineage/issues/1052

integration/sql/iface-java/src/java/io/openlineage/sql/DbTableMeta.java
pawel-big-lebowski: you're right, I didn't notice how dots are filled within `qualifiedName` method,

pawel-big-lebowski (approved):

PR #1227, opened 2022-11-02
fix: sql bumpversion

mobuchowski:
Cargo.toml moved in the new release. Fix that for the release script.


merobi-hub (approved):
LGTM @mobuchowski , thanks for the quick fix

PR #1228, opened 2022-11-02
fix: sql macos build directory

mobuchowski:
Since package renaming, mac os CI release build was run in the wrong directory.


merobi-hub (approved):
LG, thank you @mobuchowski

PR #1229, opened 2022-11-03
proposal: automatically authorize release after it fails for technical reasons

mobuchowski:
Due to large amount of OpenLineage components and "monolith" nature of release process to Maven and Pypi, it sometimes fails. We had that happen a few times, and while fixes to this process are usually relatively simple, it requires us to repeat the release authorization process. We need to fix that to prevent broken releases.


merobi-hub:
May I suggest we add more specifics to the language? A clear time window might be helpful:
`If a release build fails or is delayed for up to three business days after initiation for a purely technical reason, the next point release is automatically authorized after the build and release process has been fixed.` (I'm not wedded to 3 days.)

rossturk (approved):

mobuchowski:
@merobi-hub updated the wording! Thanks.

merobi-hub (approved):
LGTM, thanks for suggesting this needed change

PR #1249, opened 2022-11-07
spec change to include possible facets usage

pawel-big-lebowski:

### Problem

Openlineage spec could have been more descriptive on what are the possible facets.

### Solution

Extend `facets` definition with a list of available facets. 

### Checklist

### DIFF
--- a/spec/OpenLineage.json
+++ b/spec/OpenLineage.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/1-0-3/OpenLineage.json",
+  "$id": "https://openlineage.io/spec/1-0-4/OpenLineage.json",
   "$defs": {
     "RunEvent": {
       "type": "object",
@@ -76,7 +76,12 @@
           "description": "The run facets.",
           "type": "object",
           "additionalProperties": {
-            "$ref": "#/$defs/RunFacet"
+            "anyOf": [
+              { "$ref": "#/$defs/RunFacet" },
+              { "$ref": "facets/ErrorMessageRunFacet.json" },
+              { "$ref": "facets/NominalTimeRunFacet.json" },
+              { "$ref": "facets/ParentRunFacet.json" }
+            ]
           }
         }
       },
@@ -108,7 +113,14 @@
           "description": "The job facets.",
           "type": "object",
           "additionalProperties": {
-            "$ref": "#/$defs/JobFacet"
+            "anyOf": [
+              { "$ref": "#/$defs/JobFacet" },
+              { "$ref": "facets/DocumentationJobFacet.json" },
+              { "$ref": "facets/OwnershipJobFacet.json" },
+              { "$ref": "facets/SourceCodeJobFacet.json" },
+              { "$ref": "facets/SourceCodeLocationJobFacet.json" },
+              { "$ref": "facets/SQLJobFacet.json" }
+            ]
           }
         }
       },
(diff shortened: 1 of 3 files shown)


pawel-big-lebowski:
@howardyoo I was able to push a version that does not alter client code generated and includes facets in openapi.html (see screenshot. 

Zrzut ekranu 2022-11-11 o 14 06 35

This version fully relies on `anyOf` keyword with additional `ref` being removed (previous version contained both `anyOf` and `ref` for additional properties which was the problem I believe). 

However, there is still an unclear assumption I had to made. Although `anyOf` is a part of json schema standard, it is hard to create a corresponding Java implementation code, especially in the case like here, where we don't want to change generated Openlineage code. So, I decided to generate Java code based on the first definition of `anyOf`, which is a kind of tradeoff that suits both purposes: code and openapi generation.

howardyoo:
Great! Let me take a look :-)Sent from my iPhoneOn Nov 11, 2022, at 7:15 AM, pawel.leszczynski ***@***.***> wrote:﻿
@howardyoo I was able to push a version that does not alter client code generated and includes facets in openapi.html (see screenshot.

This version fully relies on anyOf keyword with additional ref being removed (previous version contained both anyOf and ref for additional properties which was the problem I believe).
However, there is still an unclear assumption I had to made. Although anyOf is a part of json schema standard, it is hard to create a corresponding Java implementation code, especially in the case like here, where we don't want to change generated Openlineage code. So, I decided to generate Java code based on the first definition of anyOf, which is a kind of tradeoff that suits both purposes: code and openapi generation.

howardyoo:
> @howardyoo I was able to push a version that does not alter client code generated and includes facets in openapi.html (see screenshot.
> 
> Zrzut ekranu 2022-11-11 o 14 06 35
> 
> This version fully relies on `anyOf` keyword with additional `ref` being removed (previous version contained both `anyOf` and `ref` for additional properties which was the problem I believe).
> 
> However, there is still an unclear assumption I had to made. Although `anyOf` is a part of json schema standard, it is hard to create a corresponding Java implementation code, especially in the case like here, where we don't want to change generated Openlineage code. So, I decided to generate Java code based on the first definition of `anyOf`, which is a kind of tradeoff that suits both purposes: code and openapi generation.

yes, I know - the current code generation, in my opinion, isn't too great. The ideal state for the those facets to be defined should be in 'allOf', not in 'anyOf', technically. But I know we are resorting to something that can work in the middle - so we do not shake up too much of what's currently working. Maybe one day, the code generation can be fixed to be more robust and correct.

As for the state, I'm glad that you have found an interim solution that can preserve the code generation as well as make spec more readible and useful! Thank you for working on this.

howardyoo:
> This looks good to me. Good idea. Could you add a unit test to illustrate the new API this generates?

There is no change in the Java code generated by this change, so no need to unit test anything.

spec/OpenLineage.json
```diff
@@ -108,7 +113,14 @@
           "description": "The job facets.",
           "type": "object",
           "additionalProperties": {
-            "$ref": "#/$defs/JobFacet"
+            "anyOf": [
+              { "$ref": "#/$defs/JobFacet" },
+              { "$ref": "facets/DocumentationJobFacet.json" },
```
julienledem, line 118: I think this should be: 
```
{ "$ref": "facets/DocumentationJobFacet.json#/$defs/DocumentationJobFacet" },
```
same for other facets

julienledem: So that it points to this object: https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/DocumentationDatasetFacet.json#L5

PR #1285, opened 2022-11-10
dependabot: move updates to weekend

mobuchowski:
Let's move the update schedule to weekend to not get buried under updates in the middle of a workday.


merobi-hub (approved):
LGMT, thanks!

Issue #986, opened 2022-08-03
Add Airflow Task Instance attempt number facet


collado-mike:
In Airflow, a task instance can be cleared and attempted multiple times. Airflow tracks the try number and uses that to show relevant logs. We should report the try number for a run as a `RunFacet`. Consider something like the following facet definition:

```json
{
  "$schema" : "https://json-schema.org/draft/2020-12/schema",
  "$id" : "https://openlineage.io/spec/facets/1-0-0/AirflowTaskInstanceAttemptRunFacet.json",
  "$defs" : {
    "AirflowTaskInstanceAttemptRunFacet" : {
      "allOf" : [ {
        "$ref" : "https://openlineage.io/spec/1-0-2/OpenLineage.json#/$defs/RunFacet"
      }, {
        "type" : "object",
        "properties" : {
          "dagName" : {
            "description" : "The name of the DAG",
            "type" : "string",
            "example" : "my_processing_dag"
          },
          "taskName" : {
            "description" : "The name of the task.",
            "type" : "string",
            "example" : "read_data"
          },
          "dagRunId" : {
            "description" : "The identifier of the specific DagRun this task instance belongs to",
            "type" : "string",
            "example" : "scheduled_2022-08-01T12:00:00Z"
          },
          "taskAttempt": {
            "description": "The task attempt as an integer - the attempt count begins with 1",
            "type": "integer",
            "example": 2
          }
        },
        "required" : [ "dagName", "taskName", "dagRunId", "taskAttempt" ]
      } ],
      "type" : "object"
    }
  },
  "type" : "object",
  "properties" : {
    "errorMessage" : {
      "$ref" : "#/$defs/AirflowTaskInstanceAttemptRunFacet"
    }
  }
}
```

julienledem:
I believe this is resolved by: https://github.com/OpenLineage/OpenLineage/pull/1413

PR #1310, opened 2022-11-17
Add AirflowTaskRunFacet facet and deprecate task_info field from AirflowVersionRunFacet

collado-mike:
### Problem

Closes: #986 

### Solution

This adds a new `AirflowTaskRunFacet` with some descriptive information about the task instance. Originally intended to address getting the task attempt number into the OL event, I started adding the tryNumber to the `airflow_version` facet, but I felt that facet has too much information that isn't really relevant to the Airflow version. Instead, I created a new facet that contains information specific to the task and dag run, including the try number. I also fixed some of the ugly serialized fields that we get from the TaskInstance and the DagRun (such as `_BaseOperator__init_kwargs`). I think we should deprecate the `taskInfo` and `operator` fields of the `airflow_version` facet and instead move that information to the `AirflowTaskRunFacet`. 

### Checklist

----

### DIFF
--- a/integration/airflow/openlineage/airflow/facets.py
+++ b/integration/airflow/openlineage/airflow/facets.py
@@ -4,6 +4,7 @@
 from typing import List, Dict
 
 import attr
+from airflow.models import TaskInstance, DagRun
 from airflow.version import version as AIRFLOW_VERSION
 from openlineage.client.facet import BaseFacet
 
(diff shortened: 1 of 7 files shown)


mobuchowski:
I think @JDarDagran is working on the same stuff

JDarDagran:
> I think @JDarDagran is working on the same stuff

I do. I'll do my best to post a complete PR beginning of next week.

collado-mike:
@JDarDagran, If we don't go with this one, can you post something even if it's not complete? Or can we converge on a structure and name for the facet? Some of the work we have is blocked on this being done

collado-mike:
@JDarDagran any word on this? As mentioned, my current work is blocked due to the lack of task information (specifically the try number) in the existing airflow facet. I'd love to get something merged quickly.

JDarDagran:
@collado-mike, with the next change we were planning to have one composite facet each per run and job for Airflow. Therefore, you should expect following structure for Airflow-specific run facets:
```
{...,
    "run": {
        "facets": {
            "airflow": {
                "_producer": "https://github.com/OpenLineage/OpenLineage/tree/0.17.0/integration/airflow",
                "_schemaURL": "https://raw.githubusercontent.com/OpenLineage/OpenLineage/main/spec/OpenLineage.json#/definitions/AirflowRunFacet",
                "task": {
                                  ....
                            }
                       }
                   }
               }
 }
 ```

However, along with the composite facets the old will be sent as they were to keep backwards compatibility for some time. In that case I might advise you to put this into `run->facets->airflow_taskRun` (I'm guessing it could be just `airflow_task` instead, it's already under `run` key`. Then, when composite facets are introduced we'll prepare some sort of dictionary which old fields are translated to new ones. Would that work for you?

collado-mike:
> In that case I might advise you to put this into `run->facets->airflow_taskRun` (I'm guessing it could be just `airflow_task` instead, it's already under `run` key`. Then, when composite facets are introduced we'll prepare some sort of dictionary which old fields are translated to new ones. Would that work for you?

I'd prefer not to start consuming a structure that we know is going to change imminently. If there is a known structure, can you link to a doc that describes it? Can we start with a partial implementation now that matches the expected structure, but doesn't yet contain all of the fields? Then, I can start consuming a partial facet and when new fields are added, the existing processing jobs will just continue consuming the information needed.

collado-mike:
> For this I'd need some more context. Are those renamed fields going to be widely used and OL should take care of parsing them into some nicely named fields even if they change in time in Airflow? I'd rather prefer not to handle too often whenever Airflow renames some clearly internal fields.

That's a fine point. I'm specifically referring to these fields here: https://github.com/OpenLineage/OpenLineage/pull/1310/files#diff-0fbcf9a73ef2239bf62fd00088867c6e2bb6db16549fa46c21d3d614be98826cR83-R94 . The init args and the upstream/downstream tasks are all likely to be consumed by someone (the latter is definitely on our roadmap).

JDarDagran:
> That's a fine point. I'm specifically referring to these fields here: https://github.com/OpenLineage/OpenLineage/pull/1310/files#diff-0fbcf9a73ef2239bf62fd00088867c6e2bb6db16549fa46c21d3d614be98826cR83-R94 . The init args and the upstream/downstream tasks are all likely to be consumed by someone (the latter is definitely on our roadmap).

Okay then, let's stick to the cleaner names in OL.

mobuchowski:
This is superseded by https://github.com/OpenLineage/OpenLineage/pull/1410 and https://github.com/OpenLineage/OpenLineage/pull/1413 right?

JDarDagran:
> This is superseded by #1410 and #1413 right?

That's correct, I'm closing the issue.

Issue #1350, opened 2022-11-27
Spark Integration doesn't return Mount Points for Databricks Runtime 11.3 LTS


wjohnson:
Currently, the Databricks Runtime 11.3 LTS uses a new constructor for initialization of `DbFsUtils` which is different than all previous runtimes.  This causes us to lose all mount point information from the `DatabricksEnvironmentFacetBuilder` and `environment-properties` facet.

To resolve this, we need to look at whether the `DbFsUtils` has zero or one parameters in its constructor.  If it has one parameter, it's expected to be the Databricks Runtime 11.3 LTS.

PR #1351, opened 2022-11-27
Handle Databricks Runtime 11.3 changes to DbFsUtils constructor

wjohnson:
Handle Databricks Runtime 11.3 changes to DbFsUtils constructor by using reflection to check the number of and name of parameters in the constructor.

### Problem

Currently, the Databricks Runtime 11.3 LTS uses a new constructor for initialization of DbFsUtils which is different than all previous runtimes. This causes us to lose all mount point information from the DatabricksEnvironmentFacetBuilder and environment-properties facet.

To resolve this, we need to look at whether the DbFsUtils has zero or one parameters in its constructor. If it has one parameter, it's expected to be the Databricks Runtime 11.3 LTS.

Closes: #1350

### Solution

Since Databricks has changed the constructor, we test the first declared constructor to see how many parameters and the name of the parameters.  

* If there are no parameters in the constructor, it should be the older Databricks Runtimes.
* If there is one parameter named allowMountOperations, it should be the 11.3 Databricks Runtime.

This may change and need to be revisited in the future as [Databricks pushes users to the Unity Catalog / hive metastore approach](https://learn.microsoft.com/en-us/azure/databricks/data-governance/unity-catalog/best-practices#external-locations-and-storage-credentials) instead.

### Checklist

----

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
@@ -10,11 +10,15 @@ import com.databricks.dbutils_v1.DbfsUtils;
 import io.openlineage.spark.agent.facets.EnvironmentFacet;
 import io.openlineage.spark.agent.models.DatabricksMountpoint;
 import io.openlineage.spark.api.CustomFacetBuilder;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Parameter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.BiConsumer;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.spark.scheduler.SparkListenerJobStart;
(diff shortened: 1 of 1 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
```diff
@@ -71,17 +74,53 @@ private Map<String, Object> getDatabricksEnvironmentalAttributes(SparkListenerJo
...
     }
 
     return dbProperties;
   }
 
+  // Starting in Databricks Runtime 11, there is a new constructor for DbFsUtils
+  // If running on an older version, the constructor has no parameters.
+  // If running on DBR 11 or above, you need to specify whether you allow mount operations (true or
+  // false)
+  private static DbfsUtils getDbfsUtils()
```
pawel-big-lebowski, line 96: Just to make sure: couldn't we use here `com.databricks.dbutils_v1.DBUtilsHolder.dbutils().fs()` instead of creating `DbfsUtils` on our own?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
```diff
@@ -71,17 +74,53 @@ private Map<String, Object> getDatabricksEnvironmentalAttributes(SparkListenerJo
...
+      throws ClassNotFoundException, InstantiationException, IllegalAccessException,
+          IllegalArgumentException, InvocationTargetException {
+    Class dbutilsClass = Class.forName("com.databricks.dbutils_v1.impl.DbfsUtilsImpl");
+    Constructor[] dbutilsConstructors = dbutilsClass.getDeclaredConstructors();
+    if (dbutilsConstructors.length == 0) {
+      log.warn(
+          "Failed to load dbutils in OpenLineageListener as there were no declared constructors");
+      return null;
```
pawel-big-lebowski, line 104: Please consider returning `Optional<DbfsUtils>` instead of `null`.

pawel-big-lebowski (commented):
Please consider also upgrading `dbutils` library to 0.0.6.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
wjohnson: @pawel-big-lebowski - Are you thinking of using the [DBUtils API](https://learn.microsoft.com/en-us/azure/databricks/dev-tools/databricks-utils#--databricks-utilities-api-library) if so, the pre-instantiated dbutils is not available to the spark listener.  We [had this challenge](https://stackoverflow.com/questions/69859733/how-to-use-dbutils-in-a-sparklistener-on-databricks) when first instantiating this class.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
wjohnson: Done!

pawel-big-lebowski (approved):

wjohnson:
Thank you, @pawel-big-lebowski !!!!

PR #1381, opened 2022-12-05
airflow: use threadpool-based event emission on Airflow 2.5

mobuchowski:
Airflow 2.5 added proper lifecycle methods to plugins. This means we can properly start-and-stop thread pool, without risk that tasks send to it would get killed before the worker process ends.


integration/airflow/openlineage/airflow/utils.py
```diff
@@ -396,3 +397,8 @@ def _get_logger_name(self):
                 "openlineage.airflow.extractors."
                 f"{self.__class__.__module__}.{self.__class__.__name__}"
             )
+
+
+is_airflow_version_enough = lambda x: parse_version(
```
JDarDagran, line 402: You've made it too generic. It's true only for integration tests.

Issue #1412, opened 2022-12-12
[PROPOSAL] OpenLineage client side proxy as a sidecar


fm100:
# Problem Statements
Currently OpenLineage integrations are directly sending events to the configured backend, which is not flexible enough for some situations such as 
- Not responsive backend due to an outage or a network issue
- Long latency of the API that adds extra delay to the processing

The first scenario will likely cause the data loss which we would like to avoid as much as possible. The second one is less critical and could be mitigated by implementing asynchronous integrations, but it is very hard to do so for every single integrations.

# Proposal
In this proposal, I suggest to implement OpenLineage Proxy that is lightweight, fast, and can be run as a sidecar along with the processing engine. This sidecar proxy will be run in the same network as the processing engine and the network delay will be minimal. Integrations may or may not implement asynchronous call, but anyway it is simple super fast call which brings very minimal extra latency. The sidecar proxy is also responsible for dealing with the backend outage. Even if the backend is not able to respond, the sidecar will keep the OpenLineage events locally and retry them. This prevents the data loss even in the case of the backend outage, and we are able to have a full picture of the lineage.

# Requirements
- Proxy must implement the HTTP POST API which receives an OpenLineage JSON event according to the [OpenLineage spec](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.yml).
- Proxy must be able to validate the OpenLineage events against the spec including standard facets.
    - Proxy must return 200 if the event is compliant to the spec regardless of the actual OpenLineage backend service availability.
    - Proxy must pass the event to configured *FailedEventHandler* if the event is non-compliant to the spec.
        - Proxy should return 400 for this.
- Proxy must buffer the OpenLineage events locally and send them to the backend when possible.
    - If the backend is available, the proxy should send the events as soon as possible
    - If the backend is unavailable for any reasons, proxy must keep events locally until it hits the buffer limit, and send them to the backend later when it’s available.
        - Buffer is limited by following parameters, whatever hits the limit first
            - T: hours to keep events to the buffer
            - B: bytes of the messages to keep events to the buffer
- Proxy must guarantee the order of the messages to be sent to the backend as the order it is received.

## Lightweight Proxy
This proposal will not mention how much lightweight it should be such as memory requirements or the latency, but the general idea is to be minimal enough to run as a sidecar of the processing engines.

# Proposed Design

## Language Options - Go
[...]

PR #1420, opened 2022-12-16
Implement OpenLineage client side proxy skeleton

fm100:

### Problem
#1412 

### Solution
This implements the skeleton of the design proposed from #1412 and does not implement the details of event validation, consumption, and transportation yet.

### Checklist

----


wslulciuc (commented):
_Minor_ suggestion on organization: since we have the `proxy/` dir that contains our `proxy` backendAPI, I recommend:

1. We move the client proxy, under `proxy/client/`
2. Or, we breakup the `proxy/` dir and move the proxy backend, under `proxy/backend/` and proxy client under `proxy/client/`

fm100:
> _Minor_ suggestion on organization: since we have the `proxy/` dir that contains our `proxy` backendAPI, I recommend:
> 
>     1. We move the client proxy, under `proxy/client/`
> 
>     2. Or, we breakup the `proxy/` dir and move the proxy backend, under `proxy/backend/` and proxy client under `proxy/client/`
> 
> 
> Option `1` would be my suggestion as having the proxy client contained within backend codebase if perfectly reasonable.

Good point. I think it makes sense, but directory structure change will result in a large diff and it will be mixed up with the actual code change and that will be very confusing. Let me make a separate PR only for restructuring.

mobuchowski:
I like the overall structure of new code, but my Go is too ...rusty 🦀  nowadays to properly give approve here.

client-proxy/app.env
```diff
@@ -0,0 +1,6 @@
+RETENTION_HOURS=24
```
wslulciuc, line 1: I would be more specific with the `*.env` file name and use `openlineage-proxy.env` or  just `openlineage.env`..

client-proxy/app.env
```diff
@@ -0,0 +1,6 @@
...
+RETENTION_BYTES=104857600
```
wslulciuc, line 2: Should we prefix the OpenLineage proxy specific environment variables with `OPENLINEAGE_PROXY_*`?

client-proxy/consumer/consumer.go
```diff
@@ -0,0 +1,24 @@
...
+	db        database.IDatabase
+	transport transports.ITransport
+}
+
+type IConsumer interface {
+	Run()
+}
+
+func (consumer *Consumer) Run() {
+	// Not yet implemented
```
wslulciuc, line 18: With non-implemented functions, I suggested adding:

```
panic("Consumer.Run() not implemented")
```

client-proxy/database/database.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		log.Fatalln("Cannot open database:", err)
+	}
+
+	log.Println("Initializing DB")
+	InitDB(db)
+	return &Database{db: db}
+}
+
+func InitDB(db *sql.DB) {
```
wslulciuc, line 28: Dows `InitDB()` need to be a public function? Otherwise, I would make it private `initDB()`

client-proxy/database/database.go
```diff
@@ -0,0 +1,45 @@
...
+	db.Exec(`
+		CREATE TABLE IF NOT EXISTS partitions (
+			name       text     NOT NULL,
+			size       integer  NOT NULL,
+			created_at datetime NOT NULL,
+			current    boolean  NOT NULL
```
wslulciuc, line 34: _Minor_: maybe rename to `is_current`? as `current` reads more like a pointer

client-proxy/database/database.go
```diff
@@ -0,0 +1,45 @@
...
+		)
+	`)
+
+	db.Exec(`
```
wslulciuc, line 38: Do we have a way to manage db migrations?

client-proxy/main.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	lineageEventConsumer := consumer.New(db, transport)
+	go lineageEventConsumer.Run()
+
+	router := gin.Default()
+	router.Use(gin.Logger())
+	router.Use(gin.Recovery())
+	router.POST("/api/v1/lineage", lineageController.PostLineage)
+
+	log.Fatalln(router.Run(":8080"))
```
wslulciuc, line 42: Do we need to handle the shutdown case? What if we have OL events queued?

client-proxy/services/lineage/service.go
```diff
@@ -0,0 +1,37 @@
...
+	validator          validator.IEventValidator
+	failedEventHandler validator.IFailedEventHandler
+	eventLogger        logger.IEventLogger
+}
+
+type ILineageService interface {
+	CreateLineage(lineageEvent string) int
+}
+
+func (service *Service) CreateLineage(lineageEvent string) int {
```
wslulciuc, line 20: _Minor:_ Our clients and integrations use `emit()` as the method for handling OL events and would recommend to keep things aligned.

client-proxy/transports/http_transport.go
```diff
@@ -0,0 +1,10 @@
+package transports
+
+type HttpTransport struct {
+	OpenLineageURL    string
+	OpenLineageAPIKey string
+}
+
+func (transport *HttpTransport) Send(lineageEvent string) error {
```
wslulciuc, line 8: _Minor_: Similar comment as in https://github.com/OpenLineage/OpenLineage/pull/1420/files#r1054828563

client-proxy/transports/transport.go
```diff
@@ -0,0 +1,19 @@
+package transports
```
wslulciuc, line 1: We define how transports are configured as an underlying [client](https://github.com/OpenLineage/OpenLineage/tree/main/client) implementation. For example, in Java you can configure a [`Transport`](enLineage/tree/main/client/java#transports) via a [`openlineage.yml`](https://github.com/OpenLineage/OpenLineage/tree/main/client/java#configuration). This can be done as a follow up PR, but I would suggest we move the `client` code under `client/go` that would be used by the proxy; though the proxy is handling strings, the client can allow for `string` type events or (more preferably) define a `OlEvent` object with only top-level properties that are required.

client-proxy/validator/validator.go
```diff
@@ -0,0 +1,16 @@
+package validator
+
+type Validator struct{}
+
+type IEventValidator interface {
+	Validate(event string) error
+}
+
+func (Validator *Validator) Validate(event string) (err error) {
+	// Not yet implemented.
```
wslulciuc, line 10: What sort of validation on the event will be applied?

client-proxy/app.env
fm100: yeah makes sense.

client-proxy/app.env
fm100: This is already inside openlineage context. Do we need the prefix? What do we distinguish it from?

client-proxy/consumer/consumer.go
fm100: From what I understood, the `panic` will terminate the process immediately. Developers would want to try run the app multiple times of development iteration and this will block the testing. Of course the unit tests could run with `panic` here but that doesn't mean we discourage doing manual testing.

client-proxy/database/database.go
fm100: Ah, is that a go convention, starting with lower case for private? Good to know. Thank you.

client-proxy/database/database.go
fm100: yeah, that sounds better.

client-proxy/database/database.go
fm100: Yes, there is a few library options that are ready to use. From this PR, I just did `db.Exec` just to verify the query itself and for some feedback regarding the column names, types, and etc. Evaluation and adoption of migration library itself could be a separate story, and I will address it from another PR.

client-proxy/main.go
fm100: The queued events are in the sqlite3 database as well as the checkpoint. This is designed to be able to restart the app in case of crash.

client-proxy/services/lineage/service.go
fm100: Well, this is not a client, but a service that holds the business logic of the proxy. `Create` sounds a little misleading but event emission is not the only job it does. Maybe `ProcessLineage` is better? I don't know.

client-proxy/transports/transport.go
fm100: The transport is not for the client implementation but the logic for transport. It might use go client or simple http client, but the responsibility of this is not implementing the client, but is to use client in order to send events to the configured backend.

For the second point, I don't think defining an `OLEvent` is preferred. For the client library we should do, because we construct the openlineage event from the integration. We construct the event with information available, and the defined model is helpful. For the proxy, we only pass the event as is but we don't touch it. Having a defined model means that we have to deserialize and serialize again with the risk of missing some information. I would prefer not to have a model for this.

client-proxy/validator/validator.go
fm100: Validation against the open lineage event json schema.

client-proxy/services/lineage/service.go
fm100: As a second thought, the responsibility of this is adding a lineage event to the local database so that the consumer can take it and process it. I think `Create` makes sense.

client-proxy/transports/http_transport.go
fm100: Here, `Emit` makes sense.

client-proxy/app.env
wslulciuc: We wouldn't need the prefix, but my assumption is that these are environment variables? Really, my comment was related to avoiding collisions with other environment variables (with the same name) already set for other libs/applications.

client-proxy/consumer/consumer.go
wslulciuc: Ahh, I see. You may want to log that this application has started, but otherwise 👍

client-proxy/database/database.go
wslulciuc: Sure is ;)

client-proxy/database/database.go
wslulciuc: You may want to level a note, but otherwise adding / discussing migration support in a separate PR make sense

client-proxy/main.go
wslulciuc: I see, that's part of the design. Would be helpful to document that and have metrics around the `#` of queued events, etc.

client-proxy/services/lineage/service.go
wslulciuc: I'd favor `AddLineage(string)` as the event itself isn't being created but rather validated, then added to the underlying datastore (at least that's my feedback, but you can decide on what's best here!)

client-proxy/validator/validator.go
wslulciuc: That the event is valid schema? or a valid event based on the OL spec?

client-proxy/transports/transport.go
wslulciuc: > The transport is not for the client implementation but the logic for transport. It might use go client or simple http client, but the responsibility of this is not implementing the client, but is to use client in order to send events to the configured backend.

I don't think the `go` client needs to be its own package, at least not initially, but it would be great to have the `proxy` make use of a `go` client in the log term. For example, the `http` transports is simple enough, but the `proxy` expands to support `kafka` transports as a `backend`, it move the embedded `go` into its own package.

wslulciuc (approved):
As we've [discussed](https://github.com/OpenLineage/OpenLineage/pull/1420#pullrequestreview-1223303937), we'll want to re-org the `proxy/` dir as follow up to:

```
proxy/
├── backend
└── client
```

We'll also want to discuss how the poxy client will be added to CI, the release cycle, etc. But, otherwise, minus the comments, great work! Excited to see progress on this!  💯 💯 💯

client-proxy/validator/validator.go
fm100: If the event is valid based on the spec.

client-proxy/main.go
fm100: Yeah, it's not in this skeleton yet, but the design #1412 covers the metrics publication.

client-proxy/services/lineage/service.go
fm100: `AddLineage` sounds good. This PR is open too long and I saw you already approved, so I may include the change to the follow up PR.

client-proxy/transports/transport.go
fm100: I think that this is something we fundamentally have a different idea. We might need separate go client outside of this project if we implement some go integrations, but even with that I don't think it's a good idea to use the client for this use case. Using a defined model brings no benefit but only the friction.

Besides, it's definitely out of scope but I'm not even sure that it's a good idea to make client package internally deal with different transport type. At the moment, the spec only defines the HTTP endpoint and both python and java client implement just that. Do we even plan doing so for java and python clients? I don't think we have a consensus about it.

wslulciuc: > At the moment, the spec only defines the HTTP endpoint and both python and java client implement just that. Do we even plan doing so for java and python clients? I don't think we have a consensus about it.

Both the both python and java client implement `Kafka` as a transport.

client-proxy/transports/http_transport.go
```diff
@@ -0,0 +1,10 @@
+package transports
+
+type HttpTransport struct {
+	OpenLineageURL    string
+	OpenLineageAPIKey string
```
collado-mike, line 5: I'd really like to stop using API keys all over the place as the only auth mechanism. I think a token provider is a very simple interface to provide so that we can have support for OAuth tokens for anyone willing to write the token provider implementation.

client-proxy/transports/transport.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"errors"
+	"github.com/OpenLineage/OpenLineage/client-proxy/config"
+)
+
+type ITransport interface {
```
collado-mike, line 8: All the types and public functions should have some doc comments.

PR #1442, opened 2022-12-23
spec: add extraction error run facet

mobuchowski:
This PR adds `ExtractionErrorRunFacet` which is supposed to reflect internal processing errors of OpenLineage. For example, we want to distinguish SQL job that we parsed and found no datasets processed, from the one we've failed to process.

The design allows describing partial failure. This feature is mainly to use with SQL parser, since it can process multiple statements, yet still do it's job - for example, `ALTER SESSION` statements are not fully supported, yet they don't change the material outcome - we still have proper lineage. 


spec/facets/ExtractionErrorRunFacet.json
```diff
@@ -0,0 +1,51 @@
+{
+	"$schema": "https://json-schema.org/draft/2020-12/schema",
```
JDarDagran, line 2: Looking at other JSON spec files use 2-space indents.

JDarDagran: It requires reformatting anyway I think.

spec/facets/ExtractionErrorRunFacet.json
```diff
@@ -0,0 +1,51 @@
...
+                    "errors": {
+                        "type": "array",
+                        "items": {
+                            "type" : "object",
+                            "properties" : {
+                                "errorMessage" : {
+                                    "description" : "Text representation of extraction error message.",
+                                    "type" : "string"
+                                },
+                                "task": {
```
JDarDagran, line 28: How would that look exactly? Let's say parsing SQL fails - should this be a manually-named name of the stage, e.g. `SQL Parser`?

spec/facets/ExtractionErrorRunFacet.json
```diff
@@ -0,0 +1,51 @@
...
+                                    "description": "Text representation of task that failed. This can be, for example, SQL statement that parser could not interpret.",
+                                    "type" : "string"
+                                },
+                                "taskNumber": {
+                                    "description": "Order of task (counted from 0).",
```
JDarDagran, line 33: Is this count of failed tasks or all? If latter, I don't think we have a concept of task in extractors to have them counted in any manner.

JDarDagran (commented):
@mobuchowski , it would be good to add new facet to Python client as well.

I also wonder if we could simply put traceback to point with what kind of error we deal.

spec/facets/ExtractionErrorRunFacet.json
mobuchowski: In this case, this does not relate to Airflow tasks or extractors at all, but would be used as a count of "tasks" inside a single run - piece of work.

A good example is one SQL job that executes multiple statements, whether inside of outside of transaction.

spec/facets/ExtractionErrorRunFacet.json
mobuchowski: No, this should be a SQL statement that failed.

spec/facets/ExtractionErrorRunFacet.json
mobuchowski: Fixed.

JDarDagran (approved):

Issue #1440, opened 2022-12-22
OpenLineage json schema has invalid relative references of facets


fm100:
OpenLineage spec defines its [JSON schema](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json), and it has many relative references for facets:
* https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L83-L86
* https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L122-L126
* https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L158
* https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L188
* https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L223-L231

Because they are relative references, by the JSON schema standard, the reference is translated to absolute URI by combining with the location specified in `$id`. For example, if the `$id` of the main doc is `https://openlineage.io/spec/1-0-5/OpenLineage.json` and the relative reference `"$ref": "facets/ParentRunFacet.json"` the resulting absolute reference becomes `https://openlineage.io/spec/1-0-5/ParentRunFacet.json`. However, the actual ID of `ParentRunFacet` is [`https://openlineage.io/spec/facets/1-0-0/ParentRunFacet.json`](https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/ParentRunFacet.json#L3) and translated absolute reference is invalid.

Tools like JSON schema validators recursively follow the schema def and compile it, but this blocks the use of many available tools.

denimalpaca:
@fm100 is this closed wrt the above merged PR?

fm100:
> @fm100 is this closed wrt the above merged PR?

No, the PR is related with the issue, but it made a workaround to solve it. The issue still exists in spec.

pawel-big-lebowski:
My understanding of the problem is: 
 * At the beginning of the project all the facets definitions were within single schema. 
 * This lead to problems: each update of a single facet's definition causes version bump of a while Openlineage. 
 * At some point, it got changed according to the [proposal](https://github.com/OpenLineage/OpenLineage/issues/153)
[...]

PR #1469, opened 2023-01-06
Implement lineage event validator for client proxy

fm100:

### Problem
As proposed in the [design](https://github.com/OpenLineage/OpenLineage/issues/1412), The client proxy must validate the lineage event and handle it properly. This PR implements the validator logic for it.

### Solution
Due to the open issue #1440 regarding the spec, there is a limitation on using existing JSON schema validators. In this PR, we compile schemas for each facets and validate against them. For the long term, we would want to fix the main JSON schema so that we validate event at once.

### Checklist

----

### DIFF
--- a/proxy/client/validator/validator.go
+++ b/proxy/client/validator/validator.go
@@ -1,16 +1,297 @@
 package validator
 
-type Validator struct{}
+import (
+	"encoding/json"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"github.com/santhosh-tekuri/jsonschema/v5/httploader"
+	"io"
+	"path"
+)
+
+var schemaURLs = map[string]string{
+	"ErrorMessageRunFacet.json":                "https://openlineage.io/spec/facets/1-0-0/ErrorMessageRunFacet.json",
+	"ExternalQueryRunFacet.json":               "https://openlineage.io/spec/facets/1-0-0/ExternalQueryRunFacet.json",
+	"ExtractionErrorRunFacet.json":             "https://openlineage.io/spec/facets/1-0-0/ExtractionErrorRunFacet.json",
+	"NominalTimeRunFacet.json":                 "https://openlineage.io/spec/facets/1-0-0/NominalTimeRunFacet.json",
+	"ParentRunFacet.json":                      "https://openlineage.io/spec/facets/1-0-0/ParentRunFacet.json",
+	"ProcessingEngineRunFacet.json":            "https://openlineage.io/spec/facets/1-0-0/ProcessingEngineRunFacet.json",
+	"DocumentationJobFacet.json":               "https://openlineage.io/spec/facets/1-0-0/DocumentationJobFacet.json",
+	"OwnershipJobFacet.json":                   "https://openlineage.io/spec/facets/1-0-0/OwnershipJobFacet.json",
+	"SourceCodeJobFacet.json":                  "https://openlineage.io/spec/facets/1-0-0/SourceCodeJobFacet.json",
+	"SourceCodeLocationJobFacet.json":          "https://openlineage.io/spec/facets/1-0-0/SourceCodeLocationJobFacet.json",
+	"SQLJobFacet.json":                         "https://openlineage.io/spec/facets/1-0-0/SQLJobFacet.json",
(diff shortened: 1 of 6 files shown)


collado-mike:
> Due to the open issue #1440 regarding the spec, there is a limitation on using existing JSON schema validators. In this PR, we manually validates each fields and its requirements by parsing JSON document. For the long term, I think we should fix the spec and then we will be able to use the JSON schema validator.

Even though we can't use the validator from the top-level spec, can't we use the validators for the individual facets? I'd prefer we didn't have to manage all this manual field-level validation code. The relative URL issue should only affect references from the top-level spec to the facets, but I think the facet schemas themselves should be usable.

proxy/client/validator/validator.go
```diff
@@ -1,16 +1,219 @@
...
+	"errors"
+	"sort"
+	"strings"
+)
+
 type Validator struct{}
 
 type IEventValidator interface {
 	Validate(event string) error
 }
```
collado-mike, line 14: type and func doc comments should be on every public type/function

mobuchowski:
Strongly agree with @collado-mike's suggestion to not manually validate every single facet. Doing it manually would add additional place where adding new facet could break something. 

Additionally, I think this needs to work in a way that adding new facet to spec and emitting event on an older version of this proxy won't fail - not sure if it works that way with this PR.

fm100:
Updated the PR in order to use JSON schema validator.

proxy/client/validator/validator.go
```diff
@@ -1,16 +1,300 @@
...
+	return compileFacetSchemas(compiler, facets)
+}
+
+func compileFacetSchemas(compiler *jsonschema.Compiler, facets []string) (schemas []*jsonschema.Schema, err error) {
+	schemas = make([]*jsonschema.Schema, len(facets))
+	for i, schemaKey := range facets {
+		schemaURL := schemaURLs[schemaKey]
+		schemas[i], err = compiler.Compile(schemaURL)
+		if err != nil {
+			return
```
collado-mike, line 288: I know this is legal, but I find it jarring. I prefer always specifying return variables rather than just relying on the return signature (I might call this a medium-sized function in reference to https://google.github.io/styleguide/go/decisions#named-result-parameters and its advice re: Naked returns)

proxy/client/validator/validator.go
```diff
@@ -1,16 +1,300 @@
...
-	return &Validator{}
+func New() *validator {
+	jsonschema.Loaders["https"] = loadSchema
+
+	compiler := jsonschema.NewCompiler()
+	compiler.AssertFormat = true
+
+	mainSchema, err := compiler.Compile("https://openlineage.io/spec/1-0-5/OpenLineage.json")
+	if err != nil {
+		panic(err)
```
collado-mike, line 165: Avoid `panic`. Just return the error

collado-mike (commented):
Style comments only.

proxy/client/validator/validator.go
```diff
@@ -1,16 +1,300 @@
...
+// IEventValidator is an interface for the event validator.
 type IEventValidator interface {
 	Validate(event string) error
 }
 
-func (Validator *Validator) Validate(event string) (err error) {
-	// Not yet implemented.
+// Validate is the function that implements interface of IEventValidator, and it is responsible for
+// validating if the given event is compliant with the schema defined in OpenLineage spec.
+func (validator *validator) Validate(event string) error {
```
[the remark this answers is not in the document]
fm100: Thank you. I was looking for this kind of style guide and found https://go.dev/doc/effective_go#names this one, but it was a little hard to read. Let me change this.

proxy/client/validator/validator.go
```diff
@@ -1,16 +1,300 @@
...
+	var doc interface{}
+	if err := json.Unmarshal([]byte(event), &doc); err != nil {
+		return err
+	}
+
+	if err := validator.mainSchema.Validate(doc); err != nil {
+		return err
+	}
+
+	run := doc.(map[string]interface{})["run"]
```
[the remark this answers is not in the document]
fm100: Sure.

collado-mike (approved):

PR #1479, opened 2023-01-09
sql parser: do not fail whole extraction on partial failures, report ExtractionErrorRunFacet

### LINKED ISSUE #1333: SQL Parser (Snowflake) Bug
Valid SQL
```
ALTER SESSION SET QUERY_TAG = 'FOO' ;
 
INSERT
....
```

returns
```
SQL parser failed: sql parser error: Expected TABLE, found: SESSION
```

https://github.com/sqlparser-rs/sqlparser-rs/blob/main/src/parser.rs#L3160
looks like only `ALTER TABLE` is recognized

mobuchowski:
This PR adds an ability for a SQL parser to report errors to integration that's calling it, and integrations to report those error using `ExtractionErrorRunFacet`. 

Additionally, the behavior (and API) of parser changes - if it fails to parse some part of SQL statements send to it, it won't raise exception and fail whole extraction, but report partial failure from which some input and output datasets still can be extracted. PR also changes integrations to take care of that API change.

Closes: #1052, partially closes #1333

### DIFF
--- a/integration/airflow/tests/integration/requests/failed_sql_extraction.json
+++ b/integration/airflow/tests/integration/requests/failed_sql_extraction.json
@@ -1,0 +1,49 @@
+[
+  {
+    "eventType": "START",
+    "job": {
+      "facets": {
+        "sql": {
+          "query": "CREATE TYPE myrowtype AS (f1 int, f2 text, f3 numeric)"
+        }
+      },
+      "name": "failed_sql_extraction.fail"
+    },
+    "run": {
+      "facets": {
+        "extractionError": {
+          "totalTasks": 1,
+          "failedTasks": 1,
+          "errors": [{
+            "errorMessage": "Expected an object type after CREATE, found: TYPE",
+            "task": "CREATE TYPE myrowtype AS (f1 int, f2 text, f3 numeric)",
+            "taskNumber": 0
+          }]
+        }
+      }
+    }
+  }, {
+    "eventType": "COMPLETE",
+    "job": {
+      "facets": {
+        "sql": {
+          "query": "CREATE TYPE myrowtype AS (f1 int, f2 text, f3 numeric)"
+        }
+      },
+      "name": "failed_sql_extraction.fail"
+    },
+    "run": {
+      "facets": {
+        "extractionError": {
+          "totalTasks": 1,
+          "failedTasks": 1,
+          "errors": [{
+            "errorMessage": "Expected an object type after CREATE, found: TYPE",
+            "task": "CREATE TYPE myrowtype AS (f1 int, f2 text, f3 numeric)",
+            "taskNumber": 0
+          }]
+        }
+      }
+    }
+  }
+]
--- a/integration/sql/iface-java/src/main/java/io/openlineage/sql/ExtractionError.java
+++ b/integration/sql/iface-java/src/main/java/io/openlineage/sql/ExtractionError.java
@@ -1,0 +1,60 @@
+/*
+/* Copyright 2018-2022 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.sql;
+
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+public class ExtractionError {
+  private final long index;
+  private final String message;
+  private final String originStatement;
+
+  public ExtractionError(long index, String message, String originStatement) {
+    this.index = index;
+    this.message = message;
+    this.originStatement = originStatement;
+  }
+
+  private long index() {
+    return index;
+  }
+
+  public String message() {
+    return message;
+  }
+
+  public String originStatement() {
+    return originStatement;
+  }
+
+  @Override
+  public String toString() {
+    return String.format(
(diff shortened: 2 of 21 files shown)


integration/airflow/tests/integration/requests/failed_sql_extraction.json
```diff
@@ -0,0 +1,26 @@
+[
+  {
+    "eventType": "START",
```
JDarDagran, line 3: I think we should always check for both START and COMPLETE events.

JDarDagran (approved):

integration/sql/iface-java/src/test/java/io/openlineage/sql/OpenLineageSqlTest.java
```diff
@@ -144,6 +134,21 @@ void columnLevelLineageWindow() {
...
+          OpenLineageSql.parse(Collections.singletonList("NOT A STATEMENT"))
+              .get();
+      assertEquals(1, output.errors().size());
+      assertEquals(
+        Collections.singletonList(new ExtractionError(0, "Expected an SQL statement, found: NOT", "NOT A STATEMENT")),
+        output.errors());
+  }
+
+  @Test
+  void returnedMultipleErrors() {
```
pawel-big-lebowski, line 150: why is it left empty?

integration/sql/iface-java/src/main/java/io/openlineage/sql/ExtractionError.java
```diff
@@ -0,0 +1,58 @@
...
+
+  private long index() {
+    return index;
+  }
+
+  public String message() {
+    return message;
+  }
+
+  public String originStatement() {
```
pawel-big-lebowski, line 30: did you mean `originalStatement` ?

integration/sql/iface-java/src/main/java/io/openlineage/sql/ExtractionError.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Optional;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+public class ExtractionError {
+  private final long index;
+  private final String message;
+  private final String originStatement;
+
+  public ExtractionError(long index, String message, String originStatement) {
```
pawel-big-lebowski, line 16: Isn't it more `SqlParserError` than `ExtractionError`. 
In Java there is a convention that `Errors` extend `Error` class which is not the case here.

integration/sql/iface-java/src/main/java/io/openlineage/sql/SqlMeta.java
```diff
@@ -13,11 +13,13 @@ public class SqlMeta {
   private final List<DbTableMeta> inTables;
   private final List<DbTableMeta> outTables;
   private final List<ColumnLineage> columnLineage;
+  private final List<ExtractionError> errors;
 
-  public SqlMeta(List<DbTableMeta> in, List<DbTableMeta> out, List<ColumnLineage> columnLineage) {
+  public SqlMeta(List<DbTableMeta> in, List<DbTableMeta> out, List<ColumnLineage> columnLineage, List<ExtractionError> errors) {
```
pawel-big-lebowski, line 18: Wouldn't it be possible to throw Java exception `SqlParserException` instead of returning valid `SqlMeta` with error fields?

pawel-big-lebowski (commented):
My concern is that Python and Java prefer returning errors in form of exceptions and this PR adds extra `errors` field to a valid `SqlMeta` return type object.  What are the reasons to use such a pattern?

integration/sql/iface-java/src/main/java/io/openlineage/sql/ExtractionError.java
mobuchowski: Use of this facet isn't limited to SQL tbh. Maybe this could be moved to client as it's in Python.

integration/sql/iface-java/src/main/java/io/openlineage/sql/SqlMeta.java
mobuchowski: The point of this task is to allow for partial success. Look at `returnedMultipleErrors` test.

integration/airflow/tests/integration/requests/failed_sql_extraction.json
mobuchowski: Done.

integration/sql/iface-java/src/test/java/io/openlineage/sql/OpenLineageSqlTest.java
mobuchowski: Fixed.

integration/sql/iface-java/src/main/java/io/openlineage/sql/SqlMeta.java
JDarDagran: Agreed on passing those exceptions through. Same happens in Python.

JDarDagran: > passing exceptions through
With you @mobuchowski :)

pawel-big-lebowski (approved):

PR #1509, opened 2023-01-17
Add proposal for client proxy

fm100:

### Problem

Closes: #1412 

### Solution
Add proposal #1412 to the repository 

### Checklist

----


proposals/1412/client_proxy.md
```diff
@@ -0,0 +1,157 @@
...
+- Not responsive backend due to an outage or a network issue
+- Long latency of the API that adds extra delay to the processing
+
+The first scenario will likely cause the data loss which we would like to avoid as much as possible. The second one is less critical and could be mitigated by implementing asynchronous integrations, but it is very hard to do so for every single integrations.
+
+# Proposal
+In this proposal, I suggest implementing OpenLineage Proxy that is lightweight, fast, and can be run as a sidecar along with the processing engine. This sidecar proxy will be run in the same network as the processing engine and the network delay will be minimal. Integrations may or may not implement asynchronous call, but anyway it is simple super fast call which brings very minimal extra latency. The sidecar proxy is also responsible for dealing with the backend outage. Even if the backend is not able to respond, the sidecar will keep the OpenLineage events locally and retry them. This prevents the data loss even in the case of the backend outage, and we are able to have a full picture of the lineage.
+
+# Requirements
```
wslulciuc, line 11: We might want to also call out that proxy client must be highly configurable (buffer flushing, `N` retries, errors, etc)?

proposals/1412/client_proxy.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Event Logger
+*EventLogger* buffers the OpenLineage events to the temporal local storage. This is responsible for writing events to the log file, and rotating the log based on the configuration. The events must be written to the file by preserving the order of the events that the proxy received. At this point, the event is already validated.
+
+### Event Log Consumer
+This log consumer reads the log written by the *EventLogger*, and send them to the configured backend. It checkpoints the offset of the logs so that it can restart in case of proxy app restart. It is also responsible for retry in case of the backend outage. Backend might not respond to the requests for many possible reasons and this will retry until it succeeds. This will not skip an event and move on to the next in order to avoid data loss.
+
+### Metric Publisher
+*MetricPublisher* gets the data from Event Logger and Event Log Consumer, how many messages are being written to the local storage and how many are being sent to the backend, and it keeps track of the consumer lag metrics. This uses the the *statsd* client so that the metrics can be sent to any *statsd* implementation.
```
wslulciuc, line 49: Do we want to adopt [`OpenMetrics`](https://openmetrics.io/)?

proposals/1412/client_proxy.md
```diff
@@ -0,0 +1,157 @@
...
+
+# Appendix
+- [Gin](https://github.com/gin-gonic/gin) is a web framework written in Go
+- [Lumberjack](https://github.com/natefinch/lumberjack) is a logging library that support rotation based on size and ages
+- [Tail](https://github.com/hpcloud/tail) is a go library for log tailing that is designed to work with a log rotation
+- [Sqlite](https://www.sqlite.org/) is a database engine that strives to provide local data storage for applications
+
+# Side Note
+As this proposes a client side proxy, which will make a confusion with existing Kafka proxy, I would like to propose rename the Kafka proxy to the reversed proxy, as it sits in front of the backend.
```
wslulciuc, line 157: Do you mean the `proxy` [backend](https://github.com/OpenLineage/OpenLineage/tree/main/proxy/backend)?

mobuchowski:
@fm100 please rebase this PR if possible, you have unfortunately forked branch off broken CI configuration 😞

proposals/1412/client_proxy.md
fm100: This section is already mentioning about the buffer limit configuration, so I think it implies that it is configurable. For the retry, I think the proxy should indefinitely retry until the backend is available as long as the failure was retryable, so making retry count configurable doesn't make sense. I will add a `SHOULD` statement for the buffer flushing.

proposals/1412/client_proxy.md
fm100: That's a good point. The current design does not consider `OpenMetrics` but the `MetricPublisher` will have generic interface for publishing metrics. I think it's worth mentioning possible adoption of `OpenMetrics` in the future.

proposals/1412/client_proxy.md
fm100: Yes. Let me make it clear.

fm100:
> @fm100 please rebase this PR if possible, you have unfortunately forked branch off broken CI configuration 😞

proposals/1412/client_proxy.md
```diff
@@ -0,0 +1,161 @@
+# Problem Statements
+Currently, OpenLineage integrations are directly sending events to the configured backend, which is not flexible enough for some situations such as
+- Not responsive backend due to an outage or a network issue
+- Long latency of the API that adds extra delay to the processing
+
```
pawel-big-lebowski, line 5: Proposal precisely defines a problem and presents a solution to it. My concern is that a problem definition is kind of narrowed while the ultimate goal should be `Propose architecture that allows a guarantee to users that no events are lost` which strictly relates to high availability (HA) of a system. The problem here is that storing data locally on disks is a single point of failure and should not be considered as a solution to guarantee HA.

The solution for `Long latency of the API` may also depend on the nature of the problem. 
For example, if the problem is _users generate OL events on premise and would like to send them to backend run in public cloud_, then being able to buffer the events in own datacenter is cool. On the other hand, if the problem is _lineage backend is in different public cloud region_ which causes delays, then there can be many other (perhaps better) solutions to this problem than the one proposed. I think it's worth giving some examples to help users decide when the proposed solution is appropriate.

fm100: Thank you for the feedback. Of course this design is not an ultimate solution for the problem and never intend to be, so I rephrased the proposal section to say how it *helps* dealing with the problem. Also mentioned that there could be other possible solutions such as cloud based message broker but it is designed to use as minimal resources as possible. Users will be able to make their own decisions on that.

proposals/1412/client_proxy.md
```diff
@@ -0,0 +1,161 @@
...
+- Proxy must buffer the OpenLineage events locally and send them to the backend when possible.
+    - If the backend is available, the proxy should send the events as soon as possible
+    - If the backend is unavailable for any reasons, proxy must keep events locally until it hits the buffer limit, and send them to the backend later when it’s available.
+        - Buffer is limited by following parameters, whatever hits the limit first
+            - T: hours to keep events to the buffer
+            - B: bytes of the messages to keep events to the buffer
+        - Buffer should be flushed when it hits the limit in order to keep the local storage available.
+- Proxy must guarantee the order of the messages to be sent to the backend as the order it is received.
+
+## Lightweight Proxy
```
wslulciuc, line 26: Would it help to elaborate on the difference between the proxy `backend` vs the `client`? Clearly, this docs outlines the `client` design and usage, but maybe a section on when/why a user should use proxy `backend` would be helpful.

proposals/1412/client_proxy.md
wslulciuc: @pawel-big-lebowski does make a good point on calling out logstash. But, the client proxy would provide domain-specific logic on how to process / handle OL events.

wslulciuc (approved):
Great write up, @fm100! I'm approving, but would also like @pawel-big-lebowski to sign off.

PR #1526, opened 2023-01-23
ci: use ruff instead of flake8/isort/...

mobuchowski:
Similar to Airflow, move to `ruff` package that bundles several linters and formatters in one fast binary. 
This PR also includes changes made by first run of this application.
 

JDarDagran (approved):
That's a great change!
The only thing I could ask for is to include this in changelog.

Issue #1195, opened 2022-10-28
[BUG] Extractor and GE giving different `datasource_url`s


denimalpaca:
**Purpose:**
In Snowflake, at least, Great Expectations is returning a different datasource than the SnowflakeOperator's extractor is generating, causing two datasets to be created in the lineage UI, one from Snowflake, one from GE.

The GE dataset has a `datasource_url` like:
`snowflake://[account].us-east-1/BENJI.YELLOW_TRIPDATA`

while the Snowflake extractor's dataset has a `datasource_url` like:
`snowflake://[account]/SANDBOX.BENJI.YELLOW_TRIPDATA`

where the latter has the database as well as schema and table, and is missing the region found in GE's url.

**Proposed implementation**
Not sure. Probably just need to add a database option to the url string in the string builder.

mobuchowski:
I think no-`us-east-1` is clearly a bug. There is one Snowflake AWS default region, and it's `us-west-1` (Oregon): https://docs.snowflake.com/en/user-guide/admin-account-identifier.html#using-an-account-locator-as-an-identifier

`GE` one misses `.aws` I think too.

At the end, we'd prefer to have canonical - full-names, with region and cloud, since they are uniquely representing the dataset namespace.

harels:
PR #1527

PR #1527, opened 2023-01-24
great expectations, airflow: follow snowflake dataset naming rules

mobuchowski:
Normalize Snowflake dataset and datasource naming rules among dbt/airflow/great expectations.

Canonicalize old snowflake account paths around making them all full size with account, region and cloud names.

Closes: https://github.com/OpenLineage/OpenLineage/issues/1195

### DIFF
--- a/integration/airflow/tests/integration/requests/snowflake.json
+++ b/integration/airflow/tests/integration/requests/snowflake.json
@@ -13,7 +13,7 @@
 	"outputs": [{
 		"facets": {
 			"dataSource": {
-				"name": "snowflake://{{ env_var('SNOWFLAKE_ACCOUNT_ID') }}",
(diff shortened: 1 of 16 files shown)


integration/airflow/tests/integration/requests/snowflake.json
```diff
@@ -13,7 +13,7 @@
 	"outputs": [{
 		"facets": {
 			"dataSource": {
-				"name": "snowflake://{{ env_var('SNOWFLAKE_ACCOUNT_ID') }}",
+				"name": "snowflake://gp21411.us-east-1.aws",
```
JDarDagran, line 16: That would be great if we could restore templated expected result.
I have a solution why it fails: `urlparse` used in `fix_snowflake_sqlalchemy_uri` returns `hostname` which is always lowercase according to docs. Adding ` | lower` filter to Jinja template should resolve the issue.

JDarDagran (changes requested):

JDarDagran (approved):

Issue #1524, opened 2023-01-23
[GE/Common integration] GE unable to find the openlineage integration starting with 0.19.1


merobi-hub:
Beginning with OpenLineage 0.19.1 , Great Expectations is seemingly unable to find the `openlineage.common.provider.great_expectations` module. 

`$ great_expectations checkpoint run <checkpoint>` throws the following error:

```
No module named `openlineage.common.provider.great_expectations` could be found in your plugins directory.
    - Please verify your plugins directory is configured correctly.
    - Please verify you have a module named `openlineage.common.provider.great_expectations` in your plugins directory.
```

When an earlier version of the integration is used, the command executes without the error being thrown.

denimalpaca:
We investigated this issue and it is caused by a hidden new dependency: the Great Expectations integration, to run, now needs `openlineage-sql==0.19.2` installed. This may also be the case for other common providers.

Either these common providers _do_ need the SQL dependency, in which case a better error message is required, or they do not and this is a bug that is causing a breaking change.

PR #1532, opened 2023-01-26
common: add explicit SQL dependency

mobuchowski:
Address 0.19.2 break of GE dependency by including SQL dependency explicitely.

Closes: https://github.com/OpenLineage/OpenLineage/issues/1524

### DIFF
--- a/integration/airflow/tests/integration/Dockerfile
+++ b/integration/airflow/tests/integration/Dockerfile
@@ -1,11 +1,10 @@
 ARG AIRFLOW_IMAGE=apache/airflow:2.4.0-python3.7
 FROM $AIRFLOW_IMAGE AS airflow
-COPY integration/sql /app/openlineage/integration/sql
 COPY integration/dbt /app/openlineage/integration/dbt
 COPY integration/common /app/openlineage/integration/common
 COPY integration/airflow /app/openlineage/integration/airflow
-COPY client/python /app/openlineage/client/python
 COPY integration/target /app/openlineage/integration/sql/target
+COPY client/python /app/openlineage/client/python
 USER root
 RUN mkdir -p /opt/data
 RUN apt-get update && apt-get install -y git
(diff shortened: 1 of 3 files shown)


integration/airflow/tests/integration/Dockerfile
```diff
@@ -34,11 +34,15 @@ RUN mkdir -p /app && mkdir -p /opt/openlineage
...
 COPY client/python /opt/openlineage/client/python
 COPY integration/airflow/integration-requirements.txt /opt/openlineage/integration/integration-requirements.txt
 WORKDIR /app
 COPY integration/airflow/tests/integration/*.py ./
 COPY integration/airflow/tests/integration/pytest.ini ./
 COPY integration/airflow/tests/integration/requests requests
-RUN cd /opt/openlineage/integration/ && pip install --use-deprecated=legacy-resolver --user -r integration-requirements.txt
+RUN cd /app/openlineage/integration/sql && \
+    pip install openlineage-sql --no-index --find-links target/wheels --force-reinstall && \
+    cd /opt/openlineage/integration/ && pip install --use-deprecated=legacy-resolver --user -r integration-requirements.txt
```
denimalpaca, line 48: Not super familiar with the integration structure of these Dockerfiles, but shouldn't this be in the Great Expectations integration?

mobuchowski: Dependency change for common affects docker tests, and they need to be able to install locally-build package, because it's not actually on pypi 😉 
So all of this work is to build and install local deps - we want to test 100% code from local repo, and you can modify different parts of the monorepo at once.

denimalpaca (approved):

integration/airflow/tests/integration/Dockerfile
```diff
@@ -34,11 +34,15 @@ RUN mkdir -p /app && mkdir -p /opt/openlineage
 RUN apt-get update && \
     apt-get install -y python3-dev default-libmysqlclient-dev build-essential
 COPY integration/common /opt/openlineage/integration/common
+COPY integration/sql /app/openlineage/integration/sql
+COPY integration/target /app/openlineage/integration/sql/target
```
JDarDagran, line 38: `COPY integration/target /app/openlineage/integration/sql/target` is already called in base stage.
Also, there's no need to copy whole `integration/sql` as previously built wheel is used.

mobuchowski: Agreed with not copying. But, we don't import anything from base stage here now; it's effectively a second dockerfile.

JDarDagran (approved):

integration/airflow/tests/integration/Dockerfile
```diff
@@ -4,8 +4,8 @@ COPY integration/sql /app/openlineage/integration/sql
 COPY integration/dbt /app/openlineage/integration/dbt
 COPY integration/common /app/openlineage/integration/common
 COPY integration/airflow /app/openlineage/integration/airflow
-COPY client/python /app/openlineage/client/python
 COPY integration/target /app/openlineage/integration/sql/target
+COPY client/python /app/openlineage/client/python
```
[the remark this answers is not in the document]
Wtfff911: > Is that only to make Dockerfile prettier?

Issue #1419, opened 2022-12-15
[PROPOSAL] Add custom environment variables to the RunEvent


Anirudh181001:
Hi,
I would like to capture some custom environment variables and send it along with the column lineage to the server. This feature will be of great use as there is a potential use case to capture and persist some environment-specific metadata from spark clusters.
Wondering if this feature can be supported, or is on the roadmap potentially? Thanks a lot!
cc: @pawel-big-lebowski

Anirudh181001:
@pawel-big-lebowski Hi, is there a potential roadmap for this issue? Thanks!

pawel-big-lebowski:
Yeah, this seems to be pretty cool generic feature. 
It's great we have an issue for that.
I don't think it's on a roadmap soon, but I would be happy to help other contributors to tackle with this.

Anirudh181001:
Hi @pawel-big-lebowski I have extended the EnvironmentFacet class, to have a CustomEnvironmentFacetBuilder clas to support this use case for my own use. However, I would like to generalize it, to contribute back to the open source. Just bouncing off an idea to see whether this might be appealing. I was thinking if the following flow might work for a more general use case.
1. Set an env variable "CUSTOM_VARS"='["VAR1","VAR2","VAR3","VAR4"]'
2. If CUSTOM_VARS is present in env, then CustomEnvironmentFacetBuilder will collect VAR1, VAR2, VAR3, and VAR4 env variables, and pass it onto the start event.
Thanks!

PR #1545, opened 2023-02-01
Add CustomEnvironmentFacetBuilder class

Anirudh181001:
### Problem
The main idea behind this pull request is to enable a way to capture custom environment variables from a spark integration. The flow defined here, as applicable to our use case is to have a CUSTOM_VARS="VAR1,VAR2,VAR3..." environment variable that indicates which environment variables should be captured. Then, we would want to capture those variables, ie VAR1, VAR2, VAR3, ... from the spark clusters.

Closes: #1419

### Solution

The solution proposed extends the EnvironmentFacets class, to build a CustomEnvironmentFacetBuilder class. This class then is used to get the values of the environment variables and pass on to the startEvent json.

### Checklist

----


integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@link CustomFacetBuilder} that generates a {@link EnvironmentFacet} when using OpenLineage on
+ * Databricks.
+ */
+@Slf4j
+public class CustomEnvironmentFacetBuilder
+    extends CustomFacetBuilder<SparkListenerJobStart, EnvironmentFacet> {
+  private Map<String, Object> envProperties;
+
+  public static boolean isCustomEnvironmentVarCollectionEnabled() {
+    return System.getenv().containsKey("CUSTOM_VARS");
```
mobuchowski, line 32: I think longer form `CUSTOM_ENV_VARIABLES` would be preferred, even if more annoying to type. Or even `CUSTOM_OPENLINEAGE_ENV_VARIABLES`. Not strongly attached to this though.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  protected void build(
+      SparkListenerJobStart event, BiConsumer<String, ? super EnvironmentFacet> consumer) {
+    consumer.accept(
+        "environment-properties", new EnvironmentFacet(getCustomEnvironmentalAttributes(event)));
```
mobuchowski, line 39: I wonder if this won't override another `DatabricksEnvironmentFacetBuilder` that emits the same facet. Maybe there should be some merging there, or those facet builders be merged.

@pawel-big-lebowski what do you think?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
mobuchowski: Also, I think this is something that should also be specified as Spark config, as there is rest of our configuration settings.

For example, as `spark.openlineage.facets.custom_env_vars`

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
```diff
@@ -0,0 +1,108 @@
...
+  private Map<String, Object> getCustomEnvironmentalAttributes(SparkListenerJobStart jobStart) {
+    envProperties = new HashMap<>();
+    // These are useful properties to extract if they are available
+    String customVars = System.getenv().get("CUSTOM_VARS");
+    List<String> customPropertiesToExtract = Arrays.asList(customVars.split(","));
+
+    customPropertiesToExtract.stream()
+        .forEach(
+            (p) -> {
+              envProperties.put(toCamelCase(p, false, new char[] {'_'}), System.getenv().get(p));
```
mobuchowski, line 102: Why do we need to convert those to camelCase?

mobuchowski (changes requested):
Hey, most of this looks good. I've commented at few things in the code; also to get rid of test values you need to 

- rebase your PR on main, which will fix the Python test failures
- run `./gradlew spotlessApply` - make sure you're using Java 8 here, it works differently on different Java versions
- sign your commits https://github.com/OpenLineage/OpenLineage/pull/1545/checks?check_run_id=11056802342

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
Anirudh181001: Hmm yeah this is optional...more so for our specific use case, but yeah I will remove this part from the class.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
Anirudh181001: There will be an override I think...I will edit this class to merge with the custom vars instead.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/CustomEnvironmentFacetBuilder.java
Anirudh181001: Yeap that makes sense. I am now passing it in as a spark config.

pawel-big-lebowski:
The code looks good to me. 
Would it be possible to create a test (similar to this one: https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/DeltaDataSourceTest.java) which verifies presence of env variable in RunEvent? 

That's the best guarantee the feature will not be broken in future.

Anirudh181001:
> The code looks good to me. Would it be possible to create a test (similar to this one: https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/DeltaDataSourceTest.java) which verifies presence of env variable in RunEvent?
> 
> That's the best guarantee the feature will not be broken in future.

Hi @pawel-big-lebowski Thanks for the suggestion! The DeltaDataSourceTest is failing the assertions on local, and seems like it is not able to read any lineageEvent in the test. Is this expected behaviour?

mobuchowski (approved):

mobuchowski:
Thanks for the contribution @Anirudh181001

client/java/src/main/java/io/openlineage/client/transports/FacetsConfig.java
```diff
@@ -21,4 +21,8 @@ public class FacetsConfig {
   @Getter
   @JsonProperty("disabled")
   private String[] disabledFacets;
+
+  @Getter
+  @JsonProperty("custom_environment_variables")
```
pawel-big-lebowski, line 26: @Anirudh181001 I know it's a kind of a late remark, but would you be able to include description of this parameter in Spark integration docs: https://github.com/OpenLineage/OpenLineage/tree/main/integration/spark#general  ?

Issue #1160, opened 2022-10-17
[INTEGRATION] Snowflake Object Tagging


rossturk:
### What is the target system?

Snowflake

### What kind of integration is this?

- [ ] Produces OpenLineage metadata
- [X] Consumes OpenLineage metadata
- [ ] Something else

## Background

Snowflake offers a tagging mechanism that can be used to, for example, identify sensitive information in a database. We believe that it could be useful to collect these tags while observing Snowflake workloads, so users could study them in a lineage context. 

There is a hierarchy for object tagging in Snowflake: column-level, table-level, schema-level, etc. There are multiple ways to query for the existence of tags, depending on the scope, and some of them are recursive. From the Snowflake docs:

> Tag inheritance means that if a tag is applied to a table, the tag also applies to the columns in that table. This behavior is referred to as tag lineage.

![](https://docs.snowflake.com/en/_images/securable-object-hierarchy-org-to-column.png)

The query that can be used to determine the tags for a table or view “with lineage” is:

```
select *
from table(my_db.information_schema.tag_references(‘my_table’, ‘table’));
```

This will return tags on all objects *above* the table in the hierarchy, but not below it. So if there is a tag on the account containing this table, this will find it. However, if there is a tag on a column of this table, this query will not find it.

## Approach
This integration could be accomplished by:

1. Creating a new facet, `TagDatasetFacet`, with two properties: `tag_key` and `tag_value`, both strings.

3. Modifying the `OPENLINEAGE_ACCESS_HISTORY` [view](https://github.com/Snowflake-Labs/OpenLineage-AccessHistory-Setup/blob/main/open_lineage_access_history.sql) so it constructs this new facet for each of its run events, passing along both the key and value of the tag.

4. Updating the example to include a tagged table.

## Open Questions
1. Should there also be `TagColumnDatasetFacet`, where we store column-level facets? Since `OPENLINEAGE_ACCESS_HISTORY` does not include column-level lineage, does it make sense to find & emit tags for columns included in the query? Or should we do this when we implement column-level lineage in this view?
2. Should the facet also collect the tag's domain? Is it useful when looking at an OpenLineage dataset tag to know that it was actually a tag on the entire schema?

### Where should this integration be implemented?

- [ ] In the target system
- [X] In the OpenLineage repo
- [X] Somewhere else

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

mobuchowski:
@rossturk I'm looking at it.

[...]

PR #1630, opened 2023-02-14
spec: add TagDatasetFacet

mobuchowski:
Main motivation for this facet is [Snowflake object tagging mechanism.](https://github.com/OpenLineage/OpenLineage/issues/1160)

But, we want to have standardized, general mechanism to capture tags from different systems - and enable OL Consumers to do thing like propagation - show all datasets (or columns) that use some PII tagged columns.

Initial idea was to just have key-value dictionary of tags, but we want to capture column level tags too. One idea is to add separate facet for that. In contrast, this one adds `column` field. If the column field is filled, that means tag is ment for that column only. If it's empty, it means the tag concerns whole table.

### DIFF
--- a/client/python/tests/test_facet.py
+++ b/client/python/tests/test_facet.py
@@ -21,6 +21,8 @@ from openlineage.client.facet import (
     StorageDatasetFacet,
     SymlinksDatasetFacet,
     SymlinksDatasetFacetIdentifiers,
+    Tag,
+    TagDatasetFacet,
 )
 from openlineage.client.run import Dataset, Job, Run, RunEvent, RunState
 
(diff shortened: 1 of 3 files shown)


spec/facets/TagDatasetFacet.json
```diff
@@ -0,0 +1,46 @@
...
+              "properties" : {
+                "key" : {
+                  "description" : "The name of this tag.",
+                  "type" : "string",
+                  "example" : "key1"
+                },
+                "value" : {
+                  "description" : "The value of this tag.",
+                  "type" : "string",
+                  "example" : "VARCHAR|INT|..."
```
pawel-big-lebowski, line 25: Example value suggests as if this was something similar to data type.

spec/facets/TagDatasetFacet.json
```diff
@@ -0,0 +1,46 @@
...
+                },
+                "column" : {
```
pawel-big-lebowski, line 27: I think we're using `field` notion in spec.

spec/facets/TagDatasetFacet.json
```diff
@@ -0,0 +1,46 @@
...
+      }, {
+        "type" : "object",
+        "properties" : {
+          "tags" : {
+            "description" : "Tags specified on this table",
+            "type" : "array",
+            "items" : {
+              "type" : "object",
+              "properties" : {
+                "key" : {
```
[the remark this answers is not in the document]
mobuchowski: I don't see where this design forbids multiple tags for single column:
```
{
    "tags": [{
        "key": "pii",
        "value": "true",
        "field": "email"
    }, {
        "key": "something-else",
        "value": "false",
        "field": "email"
    }, {
        "key": "pii",
        "value": "true",
        "field": "pesel"
    }]
}
```

spec/facets/TagDatasetFacet.json
mobuchowski: Agree, fixed.

spec/facets/TagDatasetFacet.json
mobuchowski: Fixed.

rossturk (approved):

client/python/tests/test_facet.py
```diff
@@ -199,6 +201,65 @@ def test_ownership_job_facet():
...
+            "value": "true",
+            "field": "name"
+        }, {
+            "key": "public",
+            "value": "false"
+        }],
+        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/0.0.1/client/python",
+        "_schemaURL": "https://raw.githubusercontent.com/OpenLineage/OpenLineage/main/spec"
+                      "/OpenLineage.json#/definitions/TagDatasetFacet",
+    }
```
julienledem, line 224: Should we have a schema similar to the column lineage facet for consistency? also key-value tags sound like a good match for additional fields?
https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/ColumnLineageDatasetFacet.md
```
"tags": {
  "dataset": { 
    "public": "false"
  },
  "fields": {
    "email": {
      "pii": "true"
    },
    "name": {
      "pii": "true"
    }
  }
}
```

julienledem (commented):
I made a suggestion on the schema for those.

client/python/tests/test_facet.py
mobuchowski: Having separate `dataset` and `fields` objects makes sense - I feel it's more obvious what's happening.

Regarding additional fields - I've usually found they are quite awkward to use, but it's not a strong opinion. If you or someone else prefers it that way I can agree with it.

pawel-big-lebowski: > If you or someone else prefers it that way I can agree with it.

I've put same comment in the discussion above https://github.com/OpenLineage/OpenLineage/pull/1630#discussion_r1105847657

julienledem: I do think it is easier to read that way.

mobuchowski:
Will reopen when needed.

Issue #1625, opened 2023-02-14
Kafka transport needs more DEBUG-level logging


rossturk:
I believe that the OpenLineage Kafka transport should offer similar logging to the HTTP transport, as much as possible.

Right now, it does not provide any debugging messages. The result is that it is very hard to tell whether events are being emitted from openlineage-airflow - even with logging set to DEBUG, the task logs do not show any attempt to send a message.

The difference in DEBUG logging output can be seen by using the Python client to send messages to a local Kafka service, and then send messages to a local Marquez instance. The HTTP transport provides the payload of each message, while the Kafka transport shows a single message.

Ultimately this debugging information should be visible somewhere inside Airflow, when used as part of `openlineage-airflow`.

mobuchowski:
@rossturk I feel payload being logged on `DEBUG` should be a part of core client, while Transports should deal with logging transport-specific messages. This way any other transport will have this correctly implemented as well.

PR #1633, opened 2023-02-14
clients: debug logging of event is done only in base client class, not transports

mobuchowski:
This way we can ensure that `DEBUG` loglevel on properly configured loggers will always log the OL event, regardless of chosen transport.

This has the effect of double-logging when choosing `ConsoleTransport` with `DEBUG` loglevel, but I don't think it's a concern - `ConsoleTransport` is ment for debugging purposes only. 

Closes: https://github.com/OpenLineage/OpenLineage/issues/1625

### DIFF
--- a/client/python/openlineage/client/transport/kafka.py
+++ b/client/python/openlineage/client/transport/kafka.py
@@ -56,4 +56,4 @@ class KafkaTransport(Transport):
     def emit(self, event: RunEvent):
         self.producer.produce(topic=self.topic, value=Serde.to_json(event).encode('utf-8'))
         if self.flush:
-            self.producer.flush(timeout=2)
+            self.producer.flush(timeout=5)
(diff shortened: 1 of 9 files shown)


client/python/openlineage/client/transport/kafka.py
```diff
@@ -56,4 +56,4 @@ def __init__(self, config: KafkaConfig):
     def emit(self, event: RunEvent):
         self.producer.produce(topic=self.topic, value=Serde.to_json(event).encode('utf-8'))
         if self.flush:
-            self.producer.flush(timeout=2)
+            self.producer.flush(timeout=5)
```
pawel-big-lebowski, line 59: Wouldn't it be nice to let user override the timeout? 
If so, consider creating a separate issue for that.

pawel-big-lebowski (approved):

PR #1635, opened 2023-02-15
sql: add support for snowflake TRANSIENT keyword, bump sqlparser dependency

mobuchowski:
This PR mainly updates `sqlparser-rs` dependency - the keyword support happened there. 
The only changes happening here are due to internal `sqlparser-rs` changes. 

I will update `sqlp-release` branch to `snowflake/add-transient-keyword` and revert this change. This PR has this changed to not break build on other branches.

### DIFF
--- a/integration/common/tests/sql/test_parser.py
+++ b/integration/common/tests/sql/test_parser.py
@@ -266,7 +266,7 @@ def test_parse_bugged_cte():
             WHERE count > 1000 OR balance > 100000;
     """
     assert parse(sql).errors == [
-        ExtractionError(0, "Expected AS, found: (", sql)
+        ExtractionError(0, "Expected ), found: user_id", sql)
     ]
 
 
--- a/integration/sql/impl/src/visitor.rs
+++ b/integration/sql/impl/src/visitor.rs
@@ -8,11 +8,7 @@ use crate::context::Context;
 use crate::lineage::*;
 
 use anyhow::{anyhow, Result};
-use sqlparser::ast::{
-    Expr, Function, FunctionArg, FunctionArgExpr, Ident, ListAggOnOverflow, OrderByExpr, Query,
-    Select, SelectItem, SetExpr, Statement, TableAlias, TableFactor, TableWithJoins, WindowSpec,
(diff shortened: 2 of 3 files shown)


integration/sql/impl/Cargo.toml
```diff
@@ -13,4 +13,4 @@ crate-type = ["rlib"]
 
 [dependencies]
 anyhow = {workspace = true}
-sqlparser = {git = "https://github.com/mobuchowski/sqlparser-rs", branch = "sqlp-release"}
\ No newline at end of file
+sqlparser = {git = "https://github.com/mobuchowski/sqlparser-rs", branch = "snowflake/add-transient-keyword"}
```
pawel-big-lebowski, line 16: Is there a corresponding PR in `/
[sqlparser-rs](https://github.com/sqlparser-rs/sqlparser-rs)` ? 

What's the plan for future cooperation? Openlineage should not depend on personal forks and there is no guarantee the target PR won't change, not to mention the process of code reviews which does not exist

mobuchowski: >Is there a corresponding PR in / [sqlparser-rs](https://github.com/sqlparser-rs/sqlparser-rs) ?

It's already merged.

>What's the plan for future cooperation? Openlineage should not depend on personal forks and there is no guarantee the target PR won't change, not to mention the process of code reviews which does not exist

Then we should fork it in OL organization.

pawel-big-lebowski: > > Is there a corresponding PR in / [sqlparser-rs](https://github.com/sqlparser-rs/sqlparser-rs) ?
> 
> It's already merged.
> 
If it's merged, why do we point to fork? 
> > What's the plan for future cooperation? Openlineage should not depend on personal forks and there is no guarantee the target PR won't change, not to mention the process of code reviews which does not exist
> 
> Then we should fork it in OL organization.
It's a valid point.

pawel-big-lebowski (approved):
I would really appreciate follow-up issue to fork sqparser within OpenLineage repo.

Issue #1646, opened 2023-02-22
Missing events for super quick spark jobs


pawel-big-lebowski:
The following scenario is happening sometimes for really quick spark jobs: 

- `io.openlineage.spark.agent.OpenLineageSparkListener` is triggered when SQL job is started
-  It tries to get `queryExecution` from Spark's [SQLExectution](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/SQLExecution.scala)
- However, if Spark execution finished before triggering OpenLineage listener, `queryExecution` is removed when job finishes. The problem here is `SparkListener` does not have access to `queryExecution` on SQL start. 

We do have similar issue for that (https://github.com/OpenLineage/OpenLineage/issues/1096), which was initially created to fix flaky test. However, it is not the flaky test but non-deterministic bug.

PR #1650, opened 2023-02-23
improve collecting spark events on sql complete

pawel-big-lebowski:
### Problem

Spark integration can miss events in case of super quick spark jobs.

Closes: #1646

### Solution

Since Spark 3, `SparkListenerSQLExecutionEnd` contain `queryExecution` within itself, so can read it from there if we don't get from  on SQL start. We don't get rid of existing mechanism, bcz this would break Spark 2.x integrations. Although it's no longer supported, I tried to avoid removing it. 

### Checklist

----


integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
```diff
@@ -173,12 +171,37 @@ void testRDDName(SparkSession spark) {
...
     for (OpenLineage.RunEvent event : events) {
-      assertNotNull(
-          "Event can serialize",
-          OpenLineageClientUtils.newObjectMapper().writeValueAsString(event));
+      assertNotNull("Event can serialize", writeValueAsString(event));
     }
   }
+
+  // workaround to use OpenLineage client's ObjectMapper which is shaded
+  private Map<String, Object> readValue(File src)
```
tnazarew, line 183: i don't like the fact that those 3 methods are almost identical, but I don't see any good solution to extract some generic part

tnazarew (approved):

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/ContextFactory.java
```diff
@@ -48,4 +54,44 @@ public Optional<ExecutionContext> createSparkSQLExecutionContext(long executionI
...
+                  new OpenLineageRunEventBuilder(olContext, handlerFactory);
+              return new SparkSQLExecutionContext(
+                  event.executionId(), openLineageEventEmitter, olContext, runEventBuilder);
+            });
+  }
+
+  public static Optional<QueryExecution> executionFromCompleteEvent(
+      SparkListenerSQLExecutionEnd event) {
+    return Optional.ofNullable(
+            MethodUtils.getAccessibleMethod(event.getClass(), "qe", new Class[] {}))
```
mobuchowski, line 83: Why not just try-catch `MethodUtils.invokeMethod` with `forceAccess=true`?

integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
```diff
@@ -173,12 +171,37 @@ void testRDDName(SparkSession spark) {
...
       throws JsonProcessingException {
     for (OpenLineage.RunEvent event : events) {
-      assertNotNull(
-          "Event can serialize",
-          OpenLineageClientUtils.newObjectMapper().writeValueAsString(event));
+      assertNotNull("Event can serialize", writeValueAsString(event));
     }
   }
+
+  // workaround to use OpenLineage client's ObjectMapper which is shaded
```
mobuchowski, line 182: I don't think you need any shading workarounds here - `OpenLineageClientUtils` is a direct dependency that will be shaded later, and you can access any of it's shaded dependencies with fully-qualified class names too.

mobuchowski (changes requested):
I really like that we'll have more deterministic operation of Spark integration 🚀 

Added a few comments. Also, you might try to disable retry and run this code a few times on ci - to check if everything works.

integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
pawel-big-lebowski: Please look at my comment: https://github.com/OpenLineage/OpenLineage/pull/1650/files#r1118760545
This at least describes why is that.

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/ContextFactory.java
mobuchowski: You already catch `NoSuchMethodException` - no reason for doublechecking

integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
mobuchowski: >So the shading makes OpenLineageClientUtils return a shaded ObjectMapper instead.

That does not sound like it is a problem - the import of ObjectMapper should be for the shaded version.

>It is even more complex, because the shading is applied after compiling whole source code. So, if we use here shaded version, the code will not compile as shaded version is not available at the compile stage. But it is required during test runtime.

OK, but this is shaded in the Java client part, which is separate from shading here? So I don't really understand why does it matter.

integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
pawel-big-lebowski: I've managed to get rid of those.

integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/LibraryTest.java
pawel-big-lebowski: Well, it took me some time to realise that you're right! Thanks. 
I got rid of reflection here.

mobuchowski (approved):

Issue #1716, opened 2023-03-16
dbt: wrapper breaks with arguments before [run,test,build,seed]


rossturk:
If I run the following command, the wrapper fails to emit lineage:

```
OPENLINEAGE_URL=http://localhost:5000 dbt-ol --profiles-dir . run
```

This is because the wrapper looks for `sys.argv[1]` to be one of `['run', 'test', 'build', 'seed']`, but dbt allows users to provide arguments before the subcommand.

JDarDagran:
This requires a bit more of research but looks like parsing `run_results.json` should be a more viable option.
cc: @mobuchowski

PR #1724, opened 2023-03-19
Throw UnsupportedDbtCommand when finding unsupported entry in `args.which`.

JDarDagran:
Adjust `dbt-ol` script to scan all arguments and rely on `UnsupportedDbtCommand` exception.

### Problem

DBT integration checks for actual DBT command in two ways - by scanning `sys.argv[2:]` and from `run_results.json` file. First is more difficult to handle different combinations of parameter positions.

Closes: #1716 

### Solution

This PR changes detecting DBT command only to read from `run_results.json`.

### Checklist

----

### DIFF
--- a/integration/common/openlineage/common/provider/dbt.py
+++ b/integration/common/openlineage/common/provider/dbt.py
@@ -53,6 +53,8 @@ class SparkConnectionMethod(Enum):
     def methods():
         return [x.value for x in SparkConnectionMethod]
 
+class UnsupportedDbtCommand(Exception):
+    pass
 
 class SkipUndefined(Undefined):
     def __getattr__(self, name):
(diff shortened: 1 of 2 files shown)


mobuchowski:
I think we should always send COMPLETE (or FAILURE if failed) event regardless of what command was run, if we can't prevent it from sending START event. Thoughts @rossturk @JDarDagran ?

JDarDagran:
1. Now FAILURE is sent only based on JSON files produced by DBT. I believe the idea of supporting selected commands is to send such events only for those commands. Or maybe you were thinking of failed `dbt ...` command called from `dbt-ol` wrapper and then sending FAILURE event?
2. Does this mean we want to see a complete run that consists of no child runs?

mobuchowski:
>Does this mean we want to see a complete run that consists of no child runs?
I think if we send START event and there's no unrecoverable event (idk what, segmentation fault?) we should always send COMPLETE event.

>Or maybe you were thinking of failed dbt ... command called from dbt-ol wrapper and then sending FAILURE event?

I think whatever dbt does, it produces `run_results.json` right? If not, then we might just rely on subprocess return code, but that still needs to be checked.

rossturk:
I don't see a downside to sending `START` and `COMPLETE` events for dbt subcommands that do not perform operations on data, like `debug` and `compile`. They will result in orphaned nodes, which should not affect most users.

mobuchowski (approved):

PR #1795, opened 2023-04-24
Support custom transport types

nataliezeller1:
### Problem

Currently, OpenLineage supports only a limited set of predefined transport types. To add a new transport type, it is necessary to modify both the TransportFactory and TransportConfig to recognize the new type. This approach doesn't allow flexibility for users who want to extend the functionality with custom transport types without changing the core of OpenLineage.

### Solution

The proposed solution aims to address this issue by enabling customization and extension of transport types without the need to modify the core components of OpenLineage. 
This PR introduces a new interface - CustomTransportBuilder. To create a custom transport type, users should 
implement the CustomTransportBuilder interface, which defines the following methods:
String getType() - returns the transport type (to be set in spark.openlineage.transport.type)
TransportConfig getConfig() - return the custom TransportConfig instance
Transport build(TransportConfig config) - builds the custom Transport by its configuration

Using ServiceLoader pattern, custom implementations of CustomTransportBuilder will be discovered and loaded at run time. 
This will allow users to implement and register their custom transport types seamlessly.

### Checklist

----


client/java/src/main/java/io/openlineage/client/transports/CustomTransportResolver.java
```diff
@@ -0,0 +1,42 @@
...
+*/
+
+package io.openlineage.client.transports;
+
+import java.util.Optional;
+import java.util.ServiceLoader;
+import java.util.function.Predicate;
+import java.util.stream.StreamSupport;
+
+public class CustomTransportResolver {
```
pawel-big-lebowski, line 13: Just thinking loudly: wouldn't it make sense to have `TransportBuilder` (instead of `CustomTransportBuilder`) and load all transport types using it?

integration/spark/app/src/test/java/io/openlineage/spark/agent/ArgumentParserTest.java
```diff
@@ -52,13 +53,17 @@ void testTransportTypes() {
...
+    ArgumentParser argumentParserCustom =
+        ArgumentParser.parse(new SparkConf().set(ArgumentParser.SPARK_CONF_TRANSPORT_TYPE, "test"));
 
     assert (argumentParserConsole.getOpenLineageYaml().getTransportConfig()
         instanceof ConsoleConfig);
     assert (argumentParserHttp.getOpenLineageYaml().getTransportConfig() instanceof HttpConfig);
     assert (argumentParserKafka.getOpenLineageYaml().getTransportConfig() instanceof KafkaConfig);
     assert (argumentParserKinesis.getOpenLineageYaml().getTransportConfig()
         instanceof KinesisConfig);
+    assert (argumentParserCustom.getOpenLineageYaml().getTransportConfig()
```
pawel-big-lebowski, line 65: What extra scenario is verified when testing custom transport in Spark? Aren't the tests within client/java sufficient?

pawel-big-lebowski (commented):
This is great feature and I really enjoyed reading the created code. 
I've added two open questions to make sure proper code design decisions were made.

client/java/src/main/java/io/openlineage/client/transports/CustomTransportResolver.java
nataliezeller1: I initially thought it would be better to keep the original code without a change, although making everything work the same way makes sense. If we'll do that then I will need to add a dedicated builder for each existing transport (but it will replace the switch-case blocks and make the code cleaner). So if you think it makes sense to make this change, I'll go for it :)

integration/spark/app/src/test/java/io/openlineage/spark/agent/ArgumentParserTest.java
nataliezeller1: The test in spark (ArgumentParserTest) tests that the custom transportConfig was resolved correctly by the new TransportConfigTypeIdResolver (which replaces the JsonSubTypes that were previously defined in TransportConfig).
The test in client/java (TransportFactoryTest) tests that the custom transport was created successfully based on the custom config

client/java/src/main/java/io/openlineage/client/transports/CustomTransportResolver.java
pawel-big-lebowski: I think it makes sense because it's cleaner to build all the transports the same way. It can also save us writing extra test in ArgumentParserTest as `TransportConfigTypeIdResolver` will be already tested. 

@nataliezeller1 Please let me know if I am asking for too much refactoring for this PR. Otherwise I think it's worth making it consistent.

nataliezeller1: @pawel-big-lebowski I've changed the existing transports to be loaded the same way, let me know if you have any comment

pawel-big-lebowski (approved):

PR #1853, opened 2023-05-15
ci: remove releasing snapshots to Artifactory

mobuchowski:
Artifactory snapshots upload often break and usage does not justify continuously keeping up with it. This PR removes that uploads.


merobi-hub (approved):
LGTM, thank you

PR #1856, opened 2023-05-16
Use hatchling over setuptools and tox to orchestrate Python env setup

gaborbernat:
https://tox.wiki/en/latest/ is a project that automates Python environment setup for multiple Python versions and the de facto standard in Python world. hatchling is a modern replacement for setuptools.


client/python/README.md
```diff
@@ -3,22 +3,22 @@
 To install from source, run:
 
 ```bash
-$ python setup.py install
```
gaborbernat, line 6: As per https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html this is not supported anymore.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+      - id: check-merge-conflict
+      - id: check-toml
+      - id: debug-statements
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: "v0.0.265"
+    hooks:
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
```
JDarDagran, line 17: How about we add `insert-license` from `repo: https://github.com/Lucas-C/pre-commit-hooks`? We can do that in separate PR as well, just seems a nice addition while introducing pre-commit at all.

JDarDagran (approved):
@gaborbernat, that's a great PR. Thank you for your contribution. If you could just fix one missing header from the check.

@mobuchowski we should also follow same for `openlineage-airflow`

.pre-commit-config.yaml
gaborbernat: Would do it as follow-up as is orthogonal  👍

gaborbernat:
@JDarDagran done 👍

mobuchowski (approved):

integration/airflow/tests/integration/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN AIRFLOW_VERSION=`airflow version` && \
     cd ../airflow && \
     pip install --no-cache-dir --use-deprecated=legacy-resolver -r dev-requirements.txt \
     --constraint=https://raw.githubusercontent.com/apache/airflow/constraints-$AIRFLOW_VERSION/constraints-$PYTHON_VERSION.txt
-RUN python -m virtualenv dbt_venv && pwd && source dbt_venv/bin/activate &&  \
+RUN python -m virtualenv dbt_venv && pwd && source dbt_venv/bin/activate && pip install -U pip && \
     unset PIP_USER && \
     cd /app/openlineage/integration/sql && \
     pip install openlineage-sql --no-index --find-links target/wheels --force-reinstall && \
```
JDarDagran, line 26:
```suggestion
RUN python -m virtualenv dbt_venv && pwd && source dbt_venv/bin/activate && \
    unset PIP_USER && \
    python -m pip install pip --upgrade && \
    cd /app/openlineage/integration/sql && \
    pip install openlineage-sql --no-index --find-links target/wheels --force-reinstall && \
```

Seems to do the job for Airflow 2.1.4 integration test build

JDarDagran (approved):
@gaborbernat, thanks again for the PR! LGTM 👍

Issue #1868, opened 2023-05-19
Add runless event to spec


pawel-big-lebowski:
* create a spec for runless events,
 * adjust Java & Python client,

Follow the proposal: https://github.com/OpenLineage/OpenLineage/blob/main/proposals/1837/static_lineage.md

PR #1880, opened 2023-05-25
Emit job and dataset runless metadata

pawel-big-lebowski:
### Problem

Proposal: https://github.com/OpenLineage/OpenLineage/blob/main/proposals/1837/static_lineage.md

Closes: #1868

### Solution

 * Add proposed event types to spec,
 * Include client test,
 * Include support in python client, 
 * Verify event encoding in Marquez (to be done) 

### Checklist

----


spec/OpenLineage.json
```diff
@@ -64,6 +64,72 @@
         "schemaURL"
       ]
     },
+    "DatasetEvent": {
+      "type": "object",
+      "properties": {
+        "dataset": {
+          "allOf": [
+            { "$ref": "#/$defs/Dataset" }
```
pawel-big-lebowski, line 72: This is kind of workaround ;)

So, current code generation for Java works in a way such that base classes are generated as `interfaces` and only top level objects are classes. Dataset is only an interface while `InputDataset` and `OutputDataset` is a class. So, we cannot use `Dataset` directly here, instead we create an object with `allOf` containing `Dataset`. This advantage of this is that the existing code of generated client does not get modified. The disadvantage is a method name `newDatasetEventDataset` to create an instance of `DatasetEventDataset`  as in case of un-named objects, generated class name is a parent name concatenated with a property name.

client/java/src/main/java/io/openlineage/client/transports/Transport.java
```diff
@@ -23,4 +23,12 @@ enum Type {
   Transport(@NonNull final Type type) {}
 
   public abstract void emit(OpenLineage.RunEvent runEvent);
+
+  public void emit(OpenLineage.DatasetEvent datasetEvent) {
```
mobuchowski, line 27: Why not `abstract`? Do not want to enforce implementation in Transports made for previous versions?

client/java/src/main/java/io/openlineage/client/transports/ConsoleTransport.java
```diff
@@ -20,4 +20,16 @@ public void emit(OpenLineage.RunEvent runEvent) {
     // if DEBUG loglevel is enabled, this will double-log even due to OpenLineageClient also logging
     log.info(OpenLineageClientUtils.toJson(runEvent));
   }
+
+  @Override
+  public void emit(OpenLineage.DatasetEvent datasetEvent) {
```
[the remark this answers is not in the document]
pawel-big-lebowski: To make it work, we would need to introduce `Event` class/interface within a code that is generated from spec. To avoid hardcoding,  this should have been included within spec. So the json-schema would not only contain data definition, but also some hints on how to implement model classes.

client/java/src/main/java/io/openlineage/client/transports/Transport.java
pawel-big-lebowski: Users may have custom transport implementations. Making it `abstract` would enforce implementation on their side and would break their code even though they may have no interest in static metadata.

mobuchowski:
🚀 🚀 🚀

spec/OpenLineage.json
julienledem: Or create a named `StaticDataset` type to use here as a ref

spec/OpenLineage.json
```diff
@@ -64,6 +64,84 @@
...
+          "description": "URI identifying the producer of this metadata. For example this could be a git url with a given tag or sha",
+          "type": "string",
+          "format": "uri",
+          "example": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client"
+        },
+        "schemaURL": {
+          "description": "The JSON Pointer (https://tools.ietf.org/html/rfc6901) URL to the corresponding version of the schema definition for this RunEvent",
+          "type": "string",
+          "format": "uri",
+          "example": "https://openlineage.io/spec/2-0-0/OpenLineage.json#/$defs/DatasetEvent"
```
julienledem, line 90: I'm assuming this is how we decide what type of event this is.

spec/OpenLineage.json
```diff
@@ -64,6 +64,84 @@
...
+        }
+      },
+      "required": [
+        "eventTime",
+        "dataset",
+        "producer",
+        "schemaURL"
```
julienledem, line 97: maybe eventTime, producer and schemaURL are part of a base `Event` type added with a AllOf like we do for Input/OutputDatasets and facets.

julienledem (approved):
I added a few cosmetic comments (and a couple questions) but overall this looks great.
Please let me know which of those comments you would be considering adding.

julienledem (approved):
I added a few cosmetic comments (and a couple questions) but overall this looks great.
Please let me know which of those comments you would be considering adding.

client/java/src/test/java/io/openlineage/client/Events.java
```diff
@@ -8,7 +8,15 @@
 import java.net.URI;
 
 public class Events {
-  public static OpenLineage.RunEvent event() {
```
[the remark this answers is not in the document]
pawel-big-lebowski: sure, we can keep it.

client/java/generator/src/main/java/io/openlineage/client/TypeResolver.java
```diff
@@ -137,7 +138,15 @@ public ResolvedType visit(ArrayType arrayType) {
 
         @Override
         public ResolvedType visit(OneOfType oneOfType) {
-          throw new UnsupportedOperationException("oneOf is not supported yet " + oneOfType);
+          List<ResolvedType> resolvedTypes = oneOfType
+              .getTypes()
+              .stream()
+              .filter(type -> type instanceof RefType)
```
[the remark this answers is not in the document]
pawel-big-lebowski: That's kind of workaround for: 
```
"oneOf": [
    { "$ref": "#/$defs/RunEvent" },
    { "$ref": "#/$defs/DatasetEvent" },
    { "$ref": "#/$defs/JobEvent" }
  ]
```

We don't want to generate single Java class to handle that (may be undoable), but we want to generate code for each of types which is `RefType`. The method returns the first of the resolved types because it has to align with 
```
public ResolvedType visit(...) {
```
contract and changing contract to a list or optional would require changes across a whole code generator.

spec/OpenLineage.json
pawel-big-lebowski: I will try `StaticDataset` approach.

client/java/src/main/java/io/openlineage/client/transports/Transport.java
wslulciuc: What would be the usage of `emit(String)`? Our clients define models that make it clear the types allowed and enforces a structure. Allowing for `string` type allows for flexibility, but introduces the likelihood of invalid events being emitted.

spec/OpenLineage.json
pawel-big-lebowski: This is cool! I will try it out.

spec/OpenLineage.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/1-0-5/OpenLineage.json",
+  "$id": "https://openlineage.io/spec/2-0-0/OpenLineage.json",
```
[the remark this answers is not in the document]
wslulciuc: With `2-0-0`, can we also document how the static events will be handled by consumers? That is, without an `eventType`, it's not immediately obvious the `schemaURL` will be used to determine the type of event.

wslulciuc (approved):
Minor comments, but otherwise LGTM 💯 🥇

spec/OpenLineage.json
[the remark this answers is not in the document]
pawel-big-lebowski: https://github.com/OpenLineage/docs/issues/175

spec/OpenLineage.json
pawel-big-lebowski: yes, the approach has been tested within Marquez PR -> https://github.com/MarquezProject/marquez/pull/2495

pawel-big-lebowski:
@julienledem Superb comments!

- I've created `StaticDataset`
- I've introduced `BaseEvent` with schemaUrl, eventTime and producer fields
- I've reduced amount of `emit` methods by adding `emit(String)`

PR #1934, opened 2023-06-23
airflow: add possibility to force direct execution based on env variable

mobuchowski:
This adds possibility of using direct execution method on Airflow listener when the existence of non-sqlalchemy based Airflow event mechanism is confirmed. This happens by being on Airflow 2.6 or existence of env-variable `OPENLINEAGE_AIRFLOW_ENABLE_DIRECT_EXECUTION`


mobuchowski:
@tatiana I don't think it's something automatically testable without putting orders of magnitude of effort more in testing 🙂  or at least having very good idea how to simulate that. At least not unit testing, as this behavior is the result of some pretty complex fork and thread level issues... 

However, I've tested it manually on different versions... it works then.

JDarDagran (approved):

PR #1950, opened 2023-06-30
req: bump minimum supported python version to 3.8

mobuchowski:
Python 3.7 is EOL and that already [causes problems.](https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/7458/workflows/4d9071ba-fb19-4b2f-808a-8ec4b27c1379/jobs/121801) Let's bump this to 3.8 and keep aligned with Python EOL schedule.


merobi-hub (approved):
Thanks for turning this around so fast @mobuchowski

PR #1959, opened 2023-07-04
python, kafka: add workaround for airflow-sqlalchemy event mechanism bug

mobuchowski:
This PR adds workaround for another consequence of Airflow-sqlalchemy based event delivery mechanism. 

Due to known issues with fork and thread model there, Kafka producer left alone does not emit `COMPLETE` event.

Solution to that is to create a producer for each event when we detect that we're under Airflow 2.3 - 2.5. Unfortunately that means that Transport needs to be aware of Airflow, but that import is contained in a specific method.

### DIFF
--- a/client/python/openlineage/client/transport/kafka.py
+++ b/client/python/openlineage/client/transport/kafka.py
@@ -6,12 +6,15 @@ import logging
 from typing import TYPE_CHECKING, TypeVar, Union
 
 import attr
+from pkg_resources import parse_version
 
 from openlineage.client.serde import Serde
 from openlineage.client.transport.transport import Config, Transport
 from openlineage.client.utils import get_only_specified_fields
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


client/python/openlineage/client/transport/kafka.py
```diff
@@ -42,28 +45,69 @@ def from_dict(cls: type[_T], params: dict[str, str]) -> _T:
...
+
 class KafkaTransport(Transport):
     kind = "kafka"
     config = KafkaConfig
 
     def __init__(self, config: KafkaConfig) -> None:
+        self.topic = config.topic
+        self.flush = config.flush
+        self.kafka_config = config
+        self._is_airflow_sqlalchemy = _check_if_airflow_sqlalchemy_context()
```
JDarDagran, line 63: Let's make it cached_property maybe to avoid checking it on init?

JDarDagran (approved):

PR #1975, opened 2023-07-13
Add facet deletion

julienledem:
### Problem

The [last section of the static lineage proposal](https://github.com/OpenLineage/OpenLineage/blob/main/proposals/1837/static_lineage.md#semantics-of-facets-across-jobs-and-dataset-version) defines semantics of static facets attached to a job or dataset.

“job and dataset facets apply to that entity from that moment on until they are replaced by a new version of the same facet”

However this doesn’t allow deleting the facet altogether if the facet is no longer applicable.
For example: the dataset no longer has an owner.

This proposal adds a mechanism for deleting job and dataset facets.

### Solution

We propose to add a { _deleted: true } object that can take the place of any job or dataset facet (but not run or input/output facets which are valid only for a specific run)

### Checklist

----

### DIFF
--- a/client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
+++ b/client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
@@ -7,6 +7,7 @@ package io.openlineage.client;
 
 import static io.openlineage.client.TypeResolver.titleCase;
 import static javax.lang.model.element.Modifier.ABSTRACT;
+import static javax.lang.model.element.Modifier.DEFAULT;
 import static javax.lang.model.element.Modifier.FINAL;
 import static javax.lang.model.element.Modifier.PRIVATE;
 import static javax.lang.model.element.Modifier.PUBLIC;
@@ -24,9 +25,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+import javax.lang.model.element.Modifier;
+
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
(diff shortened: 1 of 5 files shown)


client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
```diff
@@ -440,26 +529,84 @@ private void generateDefaultImplementation(
...
+          }
+
+        });
+
+        factory.addJavadoc("@return a deleted $N", type.getName());
+        factory.addCode("return new $N(", "Default" + type.getName());
+        factory.addCode(CodeBlock.join(factoryParams, ", "));
+        factory.addCode(");\n");
+        containerTypeBuilder.addMethod(factory.build());
+      }
```
pawel-big-lebowski, line 572: isn't the `else` missing here?

client/java/src/test/java/io/openlineage/client/OpenLineageTest.java
```diff
@@ -109,6 +111,32 @@ void jsonDatasetEventSerialization() throws JsonProcessingException {
...
+
+    datasetEvent.getDataset().getFacets().getAdditionalProperties().put("documentation", ol.newDeletedDatasetFacet());
+
+    String json = mapper.writeValueAsString(datasetEvent);
+    DatasetEvent read = mapper.readValue(json, DatasetEvent.class);
+
+    assertEquals("ns", read.getDataset().getNamespace());
+    assertEquals("ds", read.getDataset().getName());
+    assertEquals(now, read.getEventTime());
+    assertEquals(Boolean.TRUE, read.getDataset().getFacets().getAdditionalProperties().get("documentation").get_deleted());
```
pawel-big-lebowski, line 139: Consider adding separate `isDeleted` instead of `get_deleted`

pawel-big-lebowski (approved):
Boffo code @julienledem 👍 💯 🥇 

Please look at my comments before merging. 
Also mind editing a `changelog`.

client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
julienledem: Ah I see how this is misleading. I'll move things around a bit to make the code clearer.
Short answer: `else` is not missing here.
The code bellow this comment is the existing code that I have modified to use my new `handleProperties` method that forces me to handle all cases of special properties (schemaURL, deleted, producer)
I have added a scoping `{ }` block. All this does is make sure the variable declared in the block are not visible outside. That's because I added the block above that creates `newDeletedDatasetFacet` exists only for those classes with the deleted field.
It is actually not necessary after i moved things around. I'll clean that up.

client/java/src/test/java/io/openlineage/client/OpenLineageTest.java
julienledem: great suggestion

julienledem (commented):
thank you

PR #1985, opened 2023-07-20
ci: change deploy key fingerprint for spec deploy

mobuchowski:


merobi-hub (approved):
Thank you for doing this, @mobuchowski

PR #2006, opened 2023-07-26
airflow: convert lineage from legacy file definition

mobuchowski:
To have more backwards compatibility with old Airflow legacy lineage description, coverage for File definition is added.

### DIFF
--- a/integration/airflow/openlineage/airflow/extractors/converters.py
+++ b/integration/airflow/openlineage/airflow/extractors/converters.py
@@ -22,7 +22,7 @@ def convert_from_object_storage_uri(uri: str) -> Optional[Dataset]:
             namespace=f"s3://{netloc}",
             name=path
         )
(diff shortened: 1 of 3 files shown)


integration/airflow/openlineage/airflow/extractors/converters.py
```diff
@@ -1,19 +1,49 @@
...
+            urlparse(uri)
+        )
+    except Exception:
+        return None
+    if scheme.startswith("s3"):
+        return Dataset(
+            namespace=f"s3://{netloc}",
+            name=path
+        )
+    elif scheme.startswith("gcs") or scheme.startswith("gs"):
```
JDarDagran, line 25:
```suggestion
    elif scheme.startswith(("gcs", "gs")):
```

JDarDagran (approved):

PR #2025, opened 2023-07-31
http, snowflake: stop using reusable session by default, do not send full event on snowflake complete

mobuchowski:
During testing 0.30.1 there's an issue of Snowflake connector clashing with HttpTransport. This PR fixes that by making two changes:

- Disable automatic `requests` session reusing - Snowflake is doing a lot of stuff connected to this. https://github.com/snowflakedb/snowflake-connector-python/blob/64fd9cc7a100d96ab5105c7e2ede7349c8b87d0f/src/snowflake/connector/connection.py#L1
- Do not run SnowflakeExtractor again on job complete. The data is already send in `START` event.


harels (approved):
lgtm

PR #2149, opened 2023-10-03
[Flink] expand iceberg source types

HuangZhenQiu:
#### One-line summary:

Currently, Flink Iceberg Source Wrapper only consider StreamingMonitorFunction for streaming use cases. For CDC and batch use cases, Flink users may directly use FlinkIcebergSource and FlinkIcebergTableSource Directly. This PR is to support 
FlinkIcebergSource and FlinkIcebergTableSource for the flink iceberg lineage.


pawel-big-lebowski:
@HuangZhenQiu  Great piece of code and I am super happy to go through this PR. 

Please modify the [changelog](https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md). 

Have you thought about writing flink example applications as we did in the past to serve as the integretion tests? Like [this one](https://github.com/OpenLineage/OpenLineage/blob/main/integration/flink/examples/stateful/src/main/java/io/openlineage/flink/FlinkIcebergApplication.java) which is then tested [here](https://github.com/OpenLineage/OpenLineage/blob/main/integration/flink/src/test/java/io/openlineage/flink/ContainerTest.java#L213).

Would be happy to hear your opinion about that. On one hand writing integration tests should not be required for all the issues. On the other hand, it is the integration tests which describe best what is supported within openlineage-flink integration at the moment. They also guarantee the feature will not get broken with future releases. This is extremely helpful in case of code that relies oftentimes on reflection methods and unit-tests are not able to cover all usecase scenarios.

pawel-big-lebowski (approved):

Issue #2166, opened 2023-10-06
[PROPOSAL] Dataset created from airflow athena integration must be named according to the location


dkt-sophie-ly:
**Purpose:**

Using the airflow athena extractor, the dataset created will be named as following:
- namespace: awsathena://host
- AwsDataCatalog.database.table

But according to the discussion here: https://github.com/OpenLineage/OpenLineage/issues/435
we should name the dataset according to its location and not the catalog.
Which also make sense here because other jobs from other integration can also read the very same data from it's location and now with the current behavior we are not able to link them with airflow athena jobs.

So the lineage is incomplete and we have a duplicate of the same dataset with different name (one with the location from s3 and another within the athena namespace).

Athena is only pointing to the aws data catalog so in order to complete the lineage and link athena job to other integration job using the same data we need to rename the dataset with its physical location.

<img width="1408" alt="image" src="https://github.com/OpenLineage/OpenLineage/assets/130998593/9bb9470b-0bbc-4af3-8eea-c0e926f5db3a">

For example here the dataset created is in a namespace called awsathena://athena.eu-west-1.amazonaws.com and and the dataset is named AwsDataCatalog.datalake_cds_parquet.d_business_unit. But it's actually the same dataset as this one:

<img width="1340" alt="image" src="https://github.com/OpenLineage/OpenLineage/assets/130998593/7cde8b72-1af9-483d-b87d-0306b34ce9ae">

We should be able to link the 2 lineages here because the way we read the data shouldn't define the name of the dataset.

**Proposed implementation**

The location of the data can be retrieve in athena get-meta-data function so we can replace the dataset name and namespace with the data physical location. We can also add a symlinks facets with the name of the catalog, the database and the table in Athena (like the discussion about the hive metastore/physical location linked).

I'll be happy to talk about it more and I'm currently working on a PR for this issue.

PR #2167, opened 2023-10-06
athena: change dataset name to its location

sophiely:
### Problem

According to the discussion here: https://github.com/OpenLineage/OpenLineage/issues/435
we should name the dataset according to its location and not the catalog.
Which also make sense here because other jobs from other integration can also read the very same data from it's location and now with the current behavior we are not able to link them with airflow athena jobs.

So the lineage is incomplete and we have a duplicate of the same dataset with different name (one with the location from s3 and another within the athena namespace).

Closes: [2166](https://github.com/OpenLineage/OpenLineage/issues/2166)

### Solution

The location of the data can be retrieve in athena get-meta-data function so we can replace the dataset name and namespace with the data physical location. We can also add a symlinks facets with the name of the catalog, the database and the table in Athena (like the discussion about the hive metastore/physical location linked).

### Checklist

----


pawel-big-lebowski:
@sophiely Please be aware, that the issue https://github.com/OpenLineage/OpenLineage/issues/435 you mentioned, has lead us to symlinks feature, which allows us registering dataset under multiple names. 

In case of spark integration, we decided to send physical location as a primary dataset identifier and include catalog identifier within the symlink facet. However, this approach is not obligatory. 

On the Marquez side, we implemented symlink feature in a way that, when a symlink for a dataset is sent, Marquez will know a dataset can be identified by different names. So, if you have a table in object storage that can be accessed by object storage physical location or by relational catalog on the top of it (like Athena), you just need to send a single OL event with the symlink facet. It does not matter if which idenfiier you put within dataset definition and which one will be contained in symlink facet. This will let Marquez know, that both identifiers correspond to the same dataset for all the other jobs, which can access data by object storage location or table name in catalog. 

To sum up: I think it's not required to replace existing identifier. Including a symlink should be sufficient.

pawel-big-lebowski:
@JDarDagran wdyt?

JDarDagran:
I think it's reasonable to do the same thing we do in Spark integration:

>  send physical location as a primary dataset identifier and include catalog identifier within the symlink facet.

This PR seems to do so.

dkt-sophie-ly:
Hi all ! 

Thanks for your reply.
So I understand that we just need to send the physical location information of the data in the symlink facets.

For now marquez seems to have issue with symlinks so I can't really say if symlink facet will solve my [issue](https://github.com/MarquezProject/marquez/issues/2645) or not.

Besides I don't think that athena extractor include s3 location in a symlink facets for the moment, should we change the PR in this direction ?

Also, for a dataset read by Athena the dataSource is indeed awsathena://athena.{region_name}.[amazonaws.com](http://amazonaws.com/) but the actual source of the data is s3, like you said @pawel-big-lebowski athena is just a relational catalog on top of it. Don't you think it can be misleading to call it dataSource then ?

JDarDagran:
> Besides I don't think that athena extractor include s3 location in a symlink facets for the moment, should we change the PR in this direction ?

You're right. We should add s3 location in a symlink facet.

> Also, for a dataset read by Athena the dataSource is indeed awsathena://athena.{region_name}.[amazonaws.com](http://amazonaws.com/) but the actual source of the data is s3, like you said @pawel-big-lebowski athena is just a relational catalog on top of it. Don't you think it can be misleading to call it dataSource then ?

That really depends. First (and I share it) perspective is that data comes from Athena and that is where dataSource is pointing at. We can also dig deeper and attach same dataset symlink to its actual blob location.

dkt-sophie-ly:
Hi all ! Just changed the PR according to what we have discussed.
Add s3 location in symlink facet

JDarDagran:
@dkt-sophie-ly thanks for the changes. We definitely should add some unit tests and mock some responses from API. This is a bigger task so I can take care of that in case it's too difficult.

dkt-sophie-ly:
Hi @JDarDagran 

Just added some tests for athena extractor :) let me know if it's ok for you

integration/airflow/tests/extractors/test_athena_extractor.py
```diff
@@ -0,0 +1,329 @@
...
+    SymlinksDatasetFacet,
+    SymlinksDatasetFacetIdentifiers,
+)
+from openlineage.common.dataset import Dataset, Field, Source
+from openlineage.common.models import DbColumn, DbTableSchema
+from openlineage.common.sql import DbTableMeta
+
+from airflow import DAG
+
+if is_airflow_version_enough("2.2.4"):
```
JDarDagran, line 20: Any reason this is 2.2.4 here but skips on 2.2.1?

JDarDagran (approved):
That's a first-class PR 🚀 ! Thank you for your contribution!
One small question though.

integration/airflow/tests/extractors/test_athena_extractor.py
dkt-sophie-ly: Hi @JDarDagran do you mean 2.1.4 ? 

Since AthenaHook is not compatible with airflow 2.1.4 I want to run these tests if airflow version >= 2.2.4

Tests for airflow 2.2.4:
image

Tests for airflow 2.1.4:
image

dkt-sophie-ly: Oh just a typo on my side, i'll fix this right now.

Thanks for noticing it 


```

JDarDagran: Thank you!

PR #2188, opened 2023-10-13
[SPARK] scala integration test with `rdd` and `toDf` support

pawel-big-lebowski:
### Problem

We don't have any test that covers Spark job written in Scala. Scala API provides smooth way to switch beetween RDD api and Dataset/dataframe api with `rdd` and `toDf` methods. Lineage is not tracked correctly in case of such operations. 

### Solution

* Write integration test for Scala job.
* Extract paths from `ParallelCollectionRDD` and `MapPartitionsRDD`.
* Refactor code to avoid extending if-else-if chain.
* Fix `ExternalRddVisitor` which is not turned on now and it's not present in any of the integration tests. 
* Avoid sending OL events for RDD operations when no outputs are detected. Events for RDD jobs, which have input only (like RDD creation without storing it), will not be emitted.

### Checklist

----

### DIFF
--- a/integration/spark/app/integrations/sparkrdd/1.json
+++ b/integration/spark/app/integrations/sparkrdd/1.json
@@ -17,11 +17,13 @@
   },
   "job" : {
     "namespace" : "ns_name",
-    "name" : "test_rdd.shuffled_map_partitions_hadoop"
+    "name" : "test_rdd.map_partitions_shuffled_map_partitions_hadoop"
   },
   "inputs" : [ {
     "namespace" : "gs.bucket",
     "name" : "gs://bucket/data.txt"
   } ],
-  "outputs" : [ ]
+  "outputs" : [ {
+    "namespace" : "file"
+  }]
 }
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java
@@ -99,9 +99,11 @@ class RddExecutionContext implements ExecutionContext {
   @Override
   @SuppressWarnings("PMD") //  f.setAccessible(true);
   public void setActiveJob(ActiveJob activeJob) {
+    log.debug("setActiveJob within RddExecutionContext {}", activeJob);
     RDD<?> finalRDD = activeJob.finalStage().rdd();
     this.jobSuffix = nameRDD(finalRDD);
     Set<RDD<?>> rdds = Rdds.flattenRDDs(finalRDD);
+    log.debug("flattenRDDs {}", rdds);
     this.inputs = findInputs(rdds);
     Configuration jc = new JobConf();
     if (activeJob.finalStage() instanceof ResultStage) {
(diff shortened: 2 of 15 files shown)


integration/spark/app/integrations/sparkrdd/1.json
```diff
@@ -16,12 +16,13 @@
     }
   },
   "job" : {
-    "namespace" : "ns_name",
-    "name" : "test_rdd.shuffled_map_partitions_hadoop"
```
julienledem, line 20: is this because the name will change every time? should that check a pattern of some sort?

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java
```diff
@@ -197,17 +199,20 @@ static String nameRDD(RDD<?> rdd) {
...
+    log.debug("start SparkListenerSQLExecutionEnd {}", sqlEnd);
   }
 
   @Override
   public void start(SparkListenerJobStart jobStart) {
-    if (inputs.isEmpty() && outputs.isEmpty()) {
-      log.info("RDDs are empty: skipping sending OpenLineage event");
+    log.debug("start SparkListenerJobStart {}", jobStart);
+    if (outputs.isEmpty()) {
+      log.info("Output RDDs are empty: skipping sending OpenLineage event");
```
julienledem, line 215: Is this a case where we would want to track that we don't understand the output or is it a case where we don't want to send an event? If the latter, please add a comment that explains this.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkScalaContainerTest.java
```diff
@@ -0,0 +1,162 @@
...
+      openLineageClientMockContainer.stop();
+    } catch (Exception e) {
+      logger.error("Unable to shut down openlineage client container", e);
+    }
+    network.close();
+  }
+
+  private GenericContainer createSparkContainer(String script) {
+    return new GenericContainer<>(
+            DockerImageName.parse("bitnami/spark:" + System.getProperty("spark.version")))
```
julienledem, line 90: maybe document the config in the javadoc at the top of the class for easy reference

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkScalaContainerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    addSparkConfig(command, "spark.jars.ivy=/tmp/.ivy2/");
+    addSparkConfig(command, "spark.openlineage.facets.disabled=");
+    addSparkConfig(
+        command, "spark.driver.extraJavaOptions=-Dlog4j.configuration=/opt/log4j.properties");
+
+    List<String> sparkShell =
+        new ArrayList(Arrays.asList("./bin/spark-shell", "--master", "local", "-i", script));
+    sparkShell.addAll(command);
+    sparkShell.addAll(
+        Arrays.asList("--jars", "/opt/libs/" + System.getProperty("openlineage.spark.jar")));
```
julienledem, line 127: same?

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkScalaContainerTest.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Test
+  void testScalaUnionRddToParquet() {
+    spark = createSparkContainer("rdd_union.scala");
+    spark.start();
+
+    await()
+        .atMost(Duration.ofSeconds(10))
+        .pollInterval(Duration.ofSeconds(2))
```
julienledem, line 141: isn't that a bit slow? Will this test take at a minimum 2 secs?

julienledem (approved):
This looks good to me. Just some cosmetic comments

integration/spark/app/integrations/sparkrdd/1.json
pawel-big-lebowski: the name just changed, it wasn't necessary to remove it.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkScalaContainerTest.java
pawel-big-lebowski: switched to 500ms

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java
pawel-big-lebowski: In case we don't understand the output, we use `log.warn` to notify it. However, there are cases when it does not make sense to send OL event as the event will not add any value to a lineage graph. This can be Spark action like `collect`/`count` or some internal trigger on `OpenlineageSparkListener`.

Issue #2161, opened 2023-10-05
[PROPOSAL] Add a Registry of Producers and Consumers in OpenLineage


julienledem:
# Purpose
This is the early stage of an idea to get community feedback on what an OpenLineage registry for producers, custom facets and consumers could be. Once this document is stable enough, I’ll create an official proposal on the OpenLineage repo.
## Goal
Allow third parties to register their implementations or custom extensions to make them easy to discover.
Shorten “Producer” and “schema url” values

# Proposed implementation
Current draft for discussion:

https://docs.google.com/document/d/1zIxKST59q3I6ws896M4GkUn7IsueLw8ejct5E-TR0vY/edit#heading=h.br8d2vy9wme9

Sheeri:
Proposed changes in PR #2237


Changes:
**02c1fc611a6005df3063d98475eb489f1e0b2d50** - Added a requirement

**b8f6e5b4adc11e5b910d89a08fed7a2585f61ec7** - Added an entry for "core" to propose changes to the spec with a "known" starting point.

This entry:
---
root_doc_URL: "https://openlineage.io/spec/facets/"
produced_facets: [
  "ol:core:1-0-0/ColumnLineageDatasetFacet.json",
  "ol:core:1-0-1/ColumnLineageDatasetFacet.json",
  "ol:core:1-0-0/DataQualityAssertionsDatasetFacet.json"
]
---

indicates that the documentation for the produced facets are at:
https://openlineage.io/spec/facets/1-0-0/ColumnLineageDatasetFacet.json
https://openlineage.io/spec/facets/1-0-1/ColumnLineageDatasetFacet.json
https://openlineage.io/spec/facets/1-0-0/DataQualityAssertionsDatasetFacet.json


**b099579e093e33461d4ba39c8f3de25c3fc518ef** - Proposing a new formatting of producer/consumer objects instead of an array of strings.

I have modified the "core" example, to show the translation between old and new format.
Added an "egeria" example which is both a producer and consumer.

- Producer and consumer doc root URLs may differ, so they are set inside the producer/consumer object
- It is assumed that the documentation link only applies to the facets owned by the same entity
  - e.g. egeria owns NewCustomFacet so the docs are at the Egeria doc URL
  - egeria does not own ColumnLineageDatasetFacet so there are no docs at the Egeria doc URL (unless it's extended by egeria?)
- "sample_URL" is where the examples/tests can be found (recommended but not required)
[...]

PR #2228, opened 2023-11-03
[PROPOSAL #2161] Add a Registry of Producers and Consumers in OpenLineage

julienledem:
### Problem

We need an OpenLineage registry for producers, custom facets and consumers.
- Allow third parties to register their implementations or custom extensions to make them easy to discover.
- Shorten “Producer” and “schema url” values

## Concept needing a registry:

Producers:
- Custom facet prefix to registry of facet schemas
- Producer uri to full URL of producer doc
- Facets produced
- Facet URI to full facet schema url

Consumers:
- URL to Documentation of facets understood.
- Facets consumed

## Requirements:
- Producers can create and evolve their custom facets without requiring approval from the OpenLineage project.
- Producers and Consumers can update the list of the facets they produce or consume without requiring approval from the OpenLineage project.
- URIs should be short
- A registered name can be both a producer and a consumer.

Closes: #2161

### Solution

A central registry that defines only names for producers and consumers but externalizes custom facets and what is consumed/produced.

### Checklist

----


mobuchowski:
Would be good to define what happens in producers that produce disjoint sets of facets - like Airflow does depending on operators, or Spark on different LogicalPlans.

mobuchowski:
@julienledem rebase should fix the CI failure.

proposals/2161/registry.md
```diff
@@ -110,6 +116,62 @@ airflow/
...
+	CODEOWNERS
+	registry.json
+	{
+		producer: {
+			root_doc_URL: "https://openlineage.io/spec/facets/",
+			sample_URL: "https://github.com/OpenLineage/OpenLineage/tree/main/spec/tests/",
+			facets: [
+				"ColumnLineageDatasetFacet.json": {
+					"owner": "core",
+					"spec_versions": [ "1-0-1", "1-0-0" ],
```
jenspfaug, line 134: What syntax will spec versions support? Is there one list entry per version or would there be support for =, <, <= prefixes? @Sheeri

proposals/2161/registry.md
```diff
@@ -0,0 +1,280 @@
...
+					"use_cases": [ "lineage", "catalog" ]
```
mobuchowski, line 135: I think we could have generic `tags` rather than `use_cases` - for example, we could have a tag for a dataset facets describing some physical properties of a dataset.

proposals/2161/registry.md
Sheeri: Great question! I see now that it probably should be using semantic versioning, this is already a solved problem.
https://github.com/npm/node-semver

I will change examples to:
"spec_versions": ">=1.0.0"
and
"spec_versions": "=1.0.0"

(I'm also not sure why I used the separator as - instead of . but I will change that too.)
We don't need to account for Advanced stuff like ~ or ^.

proposals/2161/registry.md
Sheeri: Sure, that works - a use case is one category of tags, and since they're not exclusive we can absolutely have a more generic "tag".

harels (approved):

pawel-big-lebowski (approved):
This proposal is great. Please conside change requests from other comments as they seem reasonable.

proposals/2161/registry.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Core facets
+As part of the creation of the registry, the core facets under "spec/facets" will be moved to the registry as well under the "core" name. They will follow all the same constraints as all the other facets in the registry. The "core" name is used to shorten the URIs. ex: "ol:core:{FacetName}"
+
+## Acceptance Guidelines
+To claim a name, an entity must have either documentation or a test/sample. "Reserving" a name prior to public functionality is discouraged.
+
+Corresponding values to be used:
+
+- Custom facet Prefix = `“${name}”`
```
ngorchakovaInG, line 41: do you mean here facets that won't be registered under newly create registry?

ngorchakova:
Would be good also to define a place for Facet java  classes and their builders? (Similar to that [shared ones  ](https://github.com/OpenLineage/OpenLineage/blob/f8eeeee0d5897b3e5811aabe6dcc0ad7f793df6d/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/LogicalPlanFacet.java))

proposals/2161/registry.md
julienledem: I think it's a great idea.
How about we create a separate github issue to define tags in the registry?
I'll remove the "use_cases" field from this proposal and @mobuchowski or @Sheeri, you create an issue to describe how we'd add tags to the registry. That way, I can unblock @harels getting the registry started and we give a proper description to the tag feature.

proposals/2161/registry.md
julienledem: That makes sense. In this case I meant `since` but I can see how we'd want to remove facets as well. 

Another thought I have is that we could derive this information from git history. So we could also not duplicate it in the file.

julienledem:
> Would be good to define what happens in producers that produce disjoint sets of facets - like Airflow does depending on operators, or Spark on different LogicalPlans.

Do we want to add a field for text documentation to explain cases like this?

julienledem:
> Would be good also to define a place for Facet java classes and their builders? (Similar to that [shared ones ](https://github.com/OpenLineage/OpenLineage/blob/f8eeeee0d5897b3e5811aabe6dcc0ad7f793df6d/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/LogicalPlanFacet.java))

I think yes. We could also include them in the generator if that works. As the schema moves to the main repo, I would put classes like this in the client. What do you think? @ngorchakova

julienledem:
I have removed "uses_cases" and "spec_versions" from the initial spec so we can get started while we finalize discussion around those items. I'm creating new issues to discuss them.

PR #2241, opened 2023-11-10
JobTypeJobFacet added and used in Flink

pawel-big-lebowski:
### Problem

Currently, a lineage backend has no way to distinguish if a `RunEvent` corresponds to `BATCH` or `STREAMING` job. We need to have the extra `jobType` property attached to have clear distinction about this. `BATCH` is assumed to be a default behaviour. 

Within Flink integration library, we need to distinguish which jobType a Flink job is, as Flink jobs can be both: batch and streaming. 

### Solution

 * Introduce new facet,
 * Include new facet within Flink integration. 
 
### Checklist

----

### DIFF
--- a/integration/flink/src/main/java/io/openlineage/flink/utils/JobTypeUtils.java
+++ b/integration/flink/src/main/java/io/openlineage/flink/utils/JobTypeUtils.java
@@ -1,0 +1,62 @@
+/*
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.flink.utils;
+
+import java.util.List;
+import org.apache.flink.api.common.RuntimeExecutionMode;
+import org.apache.flink.api.connector.source.Boundedness;
+import org.apache.flink.api.dag.Transformation;
+import org.apache.flink.streaming.api.transformations.WithBoundedness;
+
+/**
+ * Class to extract if a job is of streaming or batch type. It contains copy of code snippets from
+ * Flink internals to determine a job type.
+ */
+public class JobTypeUtils {
+
+  public static final String BATCH = "BATCH";
+  public static final String STREAMING = "STREAMING";
+
+  /**
+   * Modified version of
+   * https://github.com/apache/flink/blob/50ca1de29df7ca36c57e8524b5444a4d7529a5e1/flink-streaming-java/src/main/java/org/apache/flink/streaming/api/graph/StreamGraphGenerator.java#L494
+   *
+   * @param executionMode
+   * @param transformations
+   * @return
+   */
+  public static String extract(
+      RuntimeExecutionMode executionMode, List<Transformation<?>> transformations) {
+    if (executionMode != RuntimeExecutionMode.AUTOMATIC) {
+      return (executionMode == RuntimeExecutionMode.BATCH) ? BATCH : STREAMING;
+    }
+
+    return existsUnboundedSource(transformations) ? STREAMING : BATCH;
+  }
+
+  private static boolean existsUnboundedSource(List<Transformation<?>> transformations) {
+    return transformations.stream()
+        .anyMatch(
+            transformation ->
+                isUnboundedSource(transformation)
+                    || transformation.getTransitivePredecessors().stream()
+                        .anyMatch(t -> isUnboundedSource(t)));
+  }
+
+  /**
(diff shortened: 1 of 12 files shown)


spec/facets/ProcessingTypeJobFacet.json
```diff
@@ -0,0 +1,32 @@
...
+          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/JobFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "jobType": {
+              "description": "Job processing type.",
+              "type": "string",
+              "default": "BATCH",
+              "enum": ["BATCH", "STREAMING"]
```
mobuchowski, line 17: Can we do just `string` without enum here, and specify batch/streaming otherwise? Enums are cool but in schemas they introduce problems with forward compatibility.

integration/flink/src/main/java/io/openlineage/flink/utils/JobTypeUtils.java
```diff
@@ -0,0 +1,63 @@
...
+   * @param transformations
+   * @return
+   */
+  public static ProcessingTypeJobFacet.JobType extract(
+      RuntimeExecutionMode executionMode, List<Transformation<?>> transformations) {
+    if (executionMode != RuntimeExecutionMode.AUTOMATIC) {
+      return (executionMode == RuntimeExecutionMode.BATCH) ? BATCH : STREAMING;
+    }
+
+    return existsUnboundedSource(transformations) ? STREAMING : BATCH;
```
mobuchowski, line 38: 👍

spec/facets/ProcessingTypeJobFacet.json
pawel-big-lebowski: Aren't we able to extend enum in the future if we like?

pawel-big-lebowski: you're right. lets move this to strings.

spec/facets/ProcessingTypeJobFacet.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/2-0-2/ProcessingTypeJobFacet.json",
+  "$defs": {
+    "ProcessingTypeJobFacet": {
```
julienledem, line 5: I think, this is an opportunity to more closely define job types like we do Naming for Datasets.
I would propose to do the following:
```
jobType: {
  processingType: enum: "STREAM|BATCH"
  integration: STRING  ex: "SPARK|DBT|AIRFLOW|FLINK"
  jobType: STRING ex:"QUERY|COMMAND|DAG|TASK|JOB|MODEL|..."
}
```
Then the naming documentation for jobs would define:
- integration: SPARK
   - processingType: "STREAM|BATCH"
   - jobType: "JOB|COMMAND"
- integration: AIRFLOW
   - processingType: "BATCH"
   - jobType: "DAG|TASK"
- integration: DBT
   - processingType: "BATCH"
   - jobType: "PROJECT|MODEL"
- integration: FLINK
   - processingType: "STREAM|BATCH"
   - jobType: "JOB"

This doc would also clarify usage of parent jobs.
This will enable better visualization of jobs as well.

pawel-big-lebowski: Docs included in https://github.com/OpenLineage/docs/pull/257

mobuchowski: This duplicates `ProcessingEngineRunFacet`, but I like it. Generally I think this makes much more sense as job-level facet rather than run-based one, since the processing engine property wouldn't really change between job runs.

I think a very nice property of that facet is being able to distinguish whether the event is a "wrapper" job or a "leaf" job - something that we had to do "bottom-up" - only knowledge we had about it was either by treating job name in a semantically, or by looking at "children job" references. This solves the problem by making it explicite.

pawel-big-lebowski: Yes, I was also initially afraid of duplicating `ProcessingEngineRunFacet`. My current understanding is that it makes sense to have both. If airflow triggers snowflake SQL, then integration is `airflow` while processing engine should be `snowflake`, so integration has slightly different meaning: it tells more who is sending the event than what is processing data.

JDarDagran: I like an idea of having both of the facets and see need for them.

julienledem (approved):

julienledem (approved):

JDarDagran (approved):

PR #2259, opened 2023-11-14
sql: add quote information from sqlparser-rs

JDarDagran:
### Problem

Information if identifier was quoted in the SQL statement might be precious in some cases, e.g. when dealing with Snowflake.

### Solution

Add additional field `quotes` with optional values each corresponding for `database`, `schema` and `name`.
This change is backwards compatible. It was also tested with latest Airflow OpenLineage provider.

#### One-line summary:

Add quote information from sqlparser-rs

### Checklist

----

### DIFF
--- a/integration/sql/iface-java/src/test/java/io/openlineage/sql/OpenLineageSqlTest.java
+++ b/integration/sql/iface-java/src/test/java/io/openlineage/sql/OpenLineageSqlTest.java
@@ -39,7 +39,9 @@ class OpenLineageSqlTest {
     assertEquals(
         output,
         new SqlMeta(
-            Arrays.asList(new DbTableMeta("random-project", "dbt_test1", "source_table")),
+            Arrays.asList(
+                new DbTableMeta(
+                    "random-project", "dbt_test1", "source_table", new QuoteStyle("`", "`", "`"))),
             new ArrayList<DbTableMeta>(),
             Collections.emptyList(),
             Collections.emptyList()));
--- a/integration/common/tests/sql/test_parser.py
+++ b/integration/common/tests/sql/test_parser.py
@@ -355,7 +355,7 @@ def test_bigquery_escaping():
         dialect="bigquery",
         default_schema="public"
     )
-    assert sql_meta.in_tables == [DbTableMeta('random-project.dbt_test1.source_table')]
+    assert sql_meta.in_tables == [DbTableMeta('`random-project`.`dbt_test1`.`source_table`')]
 
 
 @pytest.mark.skipif(provider() == "python", reason="python does not understand DDL")
--- a/integration/sql/iface-java/src/lib.rs
+++ b/integration/sql/iface-java/src/lib.rs
@@ -82,13 +82,44 @@ impl AsJavaObject for rust_impl::SqlMeta {
     }
 }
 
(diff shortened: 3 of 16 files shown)


integration/sql/iface-java/src/test/java/io/openlineage/sql/OpenLineageSqlTest.java
```diff
@@ -39,7 +39,9 @@ void parseWithDialect() {
     assertEquals(
         output,
         new SqlMeta(
-            Arrays.asList(new DbTableMeta("random-project", "dbt_test1", "source_table")),
+            Arrays.asList(
+                new DbTableMeta(
+                    "random-project", "dbt_test1", "source_table", new Quotes("`", "`", "`"))),
```
pawel-big-lebowski, line 44: Why can't we follow the convention from `sqlparse-rs` and have quoting information included within the identifier?
In this case, in `DbTableMeta` each of the properties `schema`, `database` or `name` would be of `Ident` type instead of string, and each ident will contain information whether quote is used or not. 

I don't understand why `DbTableMeta` has all three properties repeated within `Quotes` class.

JDarDagran: We need both Java and Python interfaces to be backwards compatible. For that reason we can't replace `database`, `schema` and `name` properties from string to structure (which `Ident` really is).

We could have another constructor for `DbTableMeta` that takes `Ident`s as arguments. I can't see real example of someone using it though.

pawel-big-lebowski: Now I get it. Thanks!

integration/sql/iface-java/src/main/java/io/openlineage/sql/Quotes.java
```diff
@@ -0,0 +1,43 @@
...
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.sql;
+
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+public class Quotes {
+  private final String database;
```
pawel-big-lebowski, line 11: Could we rename this into `databaseQuoteStyle`, `schemaQuoteStyle`, `nameQuoteStyle` ?

mobuchowski: Is there any database where quote styles for those are different?

JDarDagran: Sure, let's hit the road 🚗

mobuchowski: I disagree with this rename, it's obvious in the context of the class. If anything, rename the class to `QuoteStyle`.

pawel-big-lebowski (approved):

JDarDagran:
@pawel-big-lebowski, @mobuchowski I've renamed `Quotes` to `QuoteStyle`. Please re-review. Thanks in advance!

mobuchowski (approved):

PR #2300, opened 2023-12-07
flink: try to fix release build

mobuchowski:
Flink release failed, try to bump gradle and make it more similar to Spark one.


merobi-hub (approved):
Thank you, Maciej

PR #2321, opened 2023-12-15
[FLINK] stick to dataset naming convention for kafka topics

pawel-big-lebowski:
### Problem

[Naming convention for Kafka](https://github.com/OpenLineage/OpenLineage/blob/main/spec/Naming.md#kafka) datasets assumes `kafka://` prefix which is currently missing within Flink integration. 

### Solution

* include prefix
* modify tests

### Checklist

----

### DIFF
--- a/integration/flink/src/main/java/io/openlineage/flink/utils/KafkaUtils.java
+++ b/integration/flink/src/main/java/io/openlineage/flink/utils/KafkaUtils.java
@@ -1,0 +1,26 @@
+/*
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+package io.openlineage.flink.utils;
+
+import io.openlineage.client.utils.DatasetIdentifier;
+import java.util.Properties;
+
+public class KafkaUtils {
+
+  private static final String KAFKA_DATASET_PREFIX = "kafka://";
+
+  /**
+   * Generate dataset identifier based on bootstrap servers in properties and topic name
+   *
+   * @param properties
+   * @param topic
+   * @return
+   */
+  public static DatasetIdentifier datasetIdentifierOf(Properties properties, String topic) {
+    String bootstrapServers = properties.getProperty("bootstrap.servers");
+
+    return new DatasetIdentifier(topic, String.format(KAFKA_DATASET_PREFIX + bootstrapServers));
+  }
+}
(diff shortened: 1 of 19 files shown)


integration/flink/src/main/java/io/openlineage/flink/utils/KafkaUtils.java
```diff
@@ -0,0 +1,26 @@
...
+
+  /**
+   * Generate dataset identifier based on bootstrap servers in properties and topic name
+   *
+   * @param properties
+   * @param topic
+   * @return
+   */
+  public static DatasetIdentifier datasetIdentifierOf(Properties properties, String topic) {
+    String bootstrapServers = properties.getProperty("bootstrap.servers");
```
harels, line 22: thinking if we should normalize the bootstrap servers property?
for example, strip out `https://` prefixes and trailing `/`

integration/flink/src/main/java/io/openlineage/flink/visitor/KafkaSinkVisitor.java
```diff
@@ -46,10 +46,10 @@ public List<OpenLineage.OutputDataset> apply(Object kafkaSink) {
                   datasetFacetsBuilder.schema(
                       AvroSchemaUtils.convert(context.getOpenLineage(), schema)));
 
-      log.debug("Kafka output topic: {}", topic);
+      log.debug("Kafka output topic: {}", di.getName());
 
       return Collections.singletonList(
-          outputDataset().getDataset(topic, bootstrapServers, datasetFacetsBuilder));
+          outputDataset().getDataset(di.getName(), di.getNamespace(), datasetFacetsBuilder));
     } catch (IllegalAccessException e) {
```
harels, line 53: proposal, add a `getDataset` method that takes in `DatasetIdentifier` instead of passing name and namespace?

harels (approved):
lgtm, made a couple of suggestions but can be merged as is

integration/flink/src/main/java/io/openlineage/flink/utils/KafkaUtils.java
pawel-big-lebowski: I thought bootstrap servers is just a comma separated list of hosts and cannot contain `http://`.

The bigger problem is that the list of bootstrap servers can change over time, while we still would like to have the same dataset, like a topic on a Kafka cluster. We'll need to revisit this some day.

harels: oh, I think you're right about that

harels: and yes, broker servers might not be the best identifier for a kafka namespace. but to your point, out of scope for this PR.

PR #2379, opened 2024-01-19
fix/spark-integration-catalog-symlink-without-warehouse

algorithmy1:
👋 Hi there,

**Issue 1**

We've observed an issue with symlinks generated by Spark integration, specifically when dealing with a **Glue Catalog Table**. The dataset namespace ends up being the same as the dataset name. This issue arises from the method used to parse the dataset's URI.

For example, in this Unit Test:
https://github.com/OpenLineage/OpenLineage/blob/aa382998c48ab2d7488052fe27b972efcf63f7f9/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/PathUtilsTest.java#L165

The URI `hdfs://namenode:8020/warehouse/table` leads to:
- dataset.name=/warehouse/table
- dataset=hdfs://namenode:8020

And for symlinks:
- symlinks[0].name=db.table
- symlinks[0].namespace=/warehouse

However, in our scenario:
https://github.com/OpenLineage/OpenLineage/blob/aa382998c48ab2d7488052fe27b972efcf63f7f9/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java#L54

After logging the function, 

```
catalogTable: Map(Catalog -> spark_catalog, Database -> dev_db, Table -> feat_count, Created Time -> Fri Jan 19 13:01:04 UTC 2024, Last Access -> UNKNOWN, Created By -> Spark , Type -> EXTERNAL, Provider -> parquet, Location -> s3://dev-bucket/feat_count)
catalogTable.storage(): Storage(Location: s3://dev-bucket/feat_count)
metastoreUri: Optional.empty
catalogTable.identifier(): `spark_catalog`.`dev_db`.`feat_count`
nameFromTableIdentifier(catalogTable.identifier(): dev_db.feat_count
di.getName(): feat_count
StringUtils.substringBeforeLast(di.getName(), File.separator): feat_count (which is the namespace calculated but the name of a table in our case)
```

we found the following for `catalogTable`:
- Location: s3://dev-bucket/feat_count

The issue stems from `StringUtils.substringBeforeLast(di.getName(), File.separator)`, which, if the separator is not in `di.getName()`, simply returns `di.getName()`.

Consequently, we end up with:
- URI = s3://dev-bucket/feat_count
- dataset.name=s3://dev-bucket
- dataset=feat_count
- symlinks[0].name=dev_db.feat_count
- symlinks[0].namespace=feat_count

**Issue 2**

Creating a namespace from the table path prefix doesn't seem logical in our context. Each team has its own database and places tables freely within the bucket. A more sensible approach would be:
- symlinks[0].name=feat_count
- symlinks[0].namespace=spark_catalog.dev_db

What are your thoughts on this? I'm eager to contribute and resolve this issue.

Apache Spark integration covers `S3` and `Glue Catalog` filesystem operations, and it's tested with AWS EMR.

### Checklist

- [x] Sign off on your work.
- [x] Ensure your pull request title is according to our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests).
- [x] Include tests where relevant.
- [x] Keep your changes small and self-contained.
- [ ] Update any relevant documentation.
- [ ] Add a one-liner to the changelog describing the change's purpose, if necessary.
- [ ] Version the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver), if relevant.
- [ ] Add a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files, if relevant.


integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/PathUtilsTest.java
```diff
@@ -180,6 +180,26 @@ void testFromCatalogWithDefaultStorage() throws URISyntaxException {
...
+    when(catalogTable.identifier()).thenReturn(tableIdentifier);
+    when(tableIdentifier.database()).thenReturn(Option.apply("db"));
+    when(tableIdentifier.table()).thenReturn("table");
+
+    DatasetIdentifier di = PathUtils.fromCatalogTable(catalogTable, Optional.of(sparkConf));
+    assertThat(di.getName()).isEqualTo("table");
+    assertThat(di.getNamespace()).isEqualTo("s3://s3-db");
+    assertThat(di.getSymlinks()).hasSize(1);
+    assertThat(di.getSymlinks().get(0).getName()).isEqualTo("db.table");
+    assertThat(di.getSymlinks().get(0).getNamespace()).isEqualTo("");
```
pawel-big-lebowski, line 200: A reference backend implementation creates lineage graph based on symlinks as well. Thus it requires uniqueness of the symlinks. In this case, which is a spark table without megastore URI, we want to use send table name within the symlink, although we must still enforce its uniqueness and cannot leave empty `namespace`. That's why `s3://s3-db` as it provides the wanted guarantees.

algorithmy1: Yes I agree with you.
But with the current code we get
symlinks[0].namespace=feat_count
where feat_count is the name of the table.

pawel-big-lebowski: Yes, I think `name` should be a table name. `namespace` should point to metastore URI if it is present. Otherwise this can be warehouse location within the bucket. In the example you provide warehouse location is just a bucket root.

algorithmy1: Do you mean by bucket root just a "/" ?

algorithmy1: For now we have a namespace table in the database full of table names and it is due to `StringUtils.substringBeforeLast(di.getName(), File.separator)`.

In our case each team has its own bucket and its own database.

Some chose to write at the root of the bucket and some chose to write in some path. Both use the Glue Catalog.

For us there is no notion of warehouse. 

What do you think could be a good solution ?

algorithmy1: If I follow what you said, in these examples :

1. db_team1.table1 -> (s3://db_team1_bucket/table1)
2. db_team2.table2 -> (s3://db_team2_bucket/folder2/table2)

Currently 
- in the 1st example, the namespace of the symlink is `table1`
- in the 2nd example, the namespace of the symlink is `/folder2`

Do you think we should have ?
- `symlinks[0].namespace="/"` instead of `table1`
- `symlinks[0].namespace="/folder2"`

If it is the case, I can implement it.

(Knowing that in our case, the number of namespace such as `/folder2` can be very high. Almost as much as the number of tables).

pawel-big-lebowski: For `when(catalogStorageFormat.locationUri()).thenReturn(Option.apply(new URI("s3://s3-db/table")));`
it should be:
```
assertThat(di.getName()).isEqualTo("table");
assertThat(di.getNamespace()).isEqualTo("s3://s3-db");
assertThat(di.getSymlinks().get(0).getName()).isEqualTo("db.table");
assertThat(di.getSymlinks().get(0).getNamespace()).isEqualTo("s3://s3-db");
```

`when(catalogStorageFormat.locationUri()).thenReturn(Option.apply(new URI("s3://s3-db/some-path/table")));`
it should be:
```
assertThat(di.getName()).isEqualTo("/some-path/table");
assertThat(di.getNamespace()).isEqualTo("s3://s3-db");
assertThat(di.getSymlinks().get(0).getName()).isEqualTo("db.table");
assertThat(di.getSymlinks().get(0).getNamespace()).isEqualTo("s3://s3-db/some-path/");
```

does it make sense?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -79,7 +79,8 @@ public static DatasetIdentifier fromCatalogTable(
     } else {
       return di.withSymlink(
           nameFromTableIdentifier(catalogTable.identifier()),
-          StringUtils.substringBeforeLast(di.getName(), File.separator),
+          StringUtils.substringBeforeLast(
+              catalogTable.storage().locationUri().get().toString(), File.separator),
```
pawel-big-lebowski, line 83: Please make sure execution chain does not cause null pointer exception, especially for `locationUri()` which returns an Optional.

algorithmy1:
Can you tell me please what should I do so that the tests pass ?


```
#!/bin/bash -eo pipefail
1 ERROR: Missing environment variable {i}
2
3 Exited with code exit status 1
4
CircleCI received exit code 1
```

pawel-big-lebowski (approved):

Issue #2375, opened 2024-01-18
[PROPOSAL] Support timeout for openlineage spark listener


athityakumar:
**Purpose:**

Currently, we have seen 1-2% of our spark workloads, whose spark jobs complete within the normal duration; but the openlineage spark listener takes hours to complete processing all spark events - which keeps the infrastructure up, resulting in SLA impacts as well as infrastructure cost.

While we have been seeing different root-causes for the spark listener taking a long time to finish processing events, we'd like to have a way to reduce the blast radius. That way, even if openlineage does end up going to the long-running scenario, we have a way to configure a hard timeout limit - so that we can be sure that the jobs don't end up going out of SLA, even if lineage doesn't get captured in such cases.

**Proposed implementation**
This section describes *how* you propose to model it. 

We can add support for a new spark conf, that can be configured by users like: spark.openlineage.listener.timeout.seconds=120 (say, 2 mins)

Internally, if a timeout is configured by the user, we can have an internal `Executor + Future.get()` with timeout / use Guava `SimpleTimeLimiter` to ensure all the spark events handled by OL spark listener have a max timeout rather than running for hours.

Relevant slack thread: https://openlineage.slack.com/archives/C01CK9T7HKR/p1705161285825369

mobuchowski:
Agreed with this approach - just would want to add that if the timeout is not configured, the execution path should be the same as it already is.

mobuchowski:
Solved by https://github.com/OpenLineage/OpenLineage/pull/2609

PR #2392, opened 2024-01-23
[SPARK] adds timeout for event processing in spark listener

athityakumar:
### Problem

Currently, we have seen 1-2% of our spark workloads, whose spark jobs complete within the normal duration; but the openlineage spark listener takes hours to complete processing all spark events - which keeps the infrastructure up, resulting in SLA impacts as well as infrastructure cost.

While we have been seeing different root-causes for the spark listener taking a long time to finish processing events, we'd like to have a way to reduce the blast radius. That way, even if openlineage does end up going to the long-running scenario, we have a way to configure a hard timeout limit - so that we can be sure that the jobs don't end up going out of SLA, even if lineage doesn't get captured in such cases.

Closes: #2375 

### Solution

Add wrapper functions for all the methods that have the business logic of handling/processing the spark events and invoke them with/without timeouts depending on whether the `spark.openlineage.event.processing.timeout.seconds` spark-conf is provided or not.

If the conf is provided and can be parsed into a valid Long, then timeout will be applicable. Else, timeout will not be applied and the behavior of the spark listener will remain unchanged.

The logic of timeout is added natively using Executors, Future - and on timeout, the Futures are cancelled so as to respect the timeout duration specified. We're not using external dependencies that have timeout implementations like SimpleTimeLimiter - as adding new dependencies could cause conflicts with the dependencies of the spark job's code itself & we want to minimise external dependencies for spark listeners as much as possible.

The timeout changes in the openlineage-spark integration have been already tested in Spark jobs on Databricks; we'll also test it on EMR soon.

### Checklist

----


pawel-big-lebowski:
Perhaps in the future we could extend `DebugFacet` we have to contain some metrics about OL event generation. For example, it would be nice to know amount of time spent on event serializing and amount of time spent on each visitor and dataset builder.

athityakumar:
> Perhaps in the future we could extend DebugFacet we have to contain some metrics about OL event generation. For example, it would be nice to know amount of time spent on event serializing and amount of time spent on each visitor and dataset builder.

That makes sense - created https://github.com/OpenLineage/OpenLineage/issues/2396 for this.

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
```diff
@@ -53,11 +53,14 @@ public class ArgumentParser {
           Arrays.asList("transport.properties.", "transport.urlParams.", "transport.headers."));
   public static final String SPARK_CONF_CUSTOM_ENVIRONMENT_VARIABLES =
       "spark.openlineage.facets.custom_environment_variables";
+  public static final String SPARK_CONF_EVENT_PROCESSING_TIMEOUT_SECONDS =
+      "spark.openlineage.event.processing.timeout.seconds";
```
pawel-big-lebowski, line 57: please update changelog and readme. 
you can also prepare additional PR to update docs. [docs](https://github.com/OpenLineage/docs/blob/main/docs/integrations/spark/spark.md) and readme contain same information, copied into two locations. Sorry for inconvenience.

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
```diff
@@ -75,6 +78,16 @@ public static ArgumentParser parse(SparkConf conf) {
     findSparkConfigKey(conf, SPARK_CONF_NAMESPACE).ifPresent(builder::namespace);
     findSparkConfigKey(conf, SPARK_CONF_JOB_NAME).ifPresent(builder::jobName);
     findSparkConfigKey(conf, SPARK_CONF_PARENT_RUN_ID).ifPresent(builder::parentRunId);
+    findSparkConfigKey(conf, SPARK_CONF_EVENT_PROCESSING_TIMEOUT_SECONDS)
+        .ifPresent(
+            timeout -> {
+              try {
+                builder.eventProcessingTimeoutSeconds(Optional.of(Long.parseLong(timeout)));
+              } catch (Exception e) {
```
pawel-big-lebowski, line 86: would it make sense to log an error in this case?

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -74,6 +81,33 @@ public static void init(ContextFactory contextFactory) {
...
+                "Processing onOtherEventFn with timeout: {} seconds",
+                eventProcessingTimeoutSeconds.get());
+            onOtherEventFn(event);
+            return null;
+          },
+          eventProcessingTimeoutSeconds.get(),
+          TimeUnit.SECONDS,
+          true);
+    } catch (Exception e) {
+      log.debug("onOtherEvent Timeout Exception: {}", e.getMessage(), e);
```
pawel-big-lebowski, line 106: shouldn't this be `warn` level if not an `error`

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -286,8 +401,41 @@ private void initializeContextFactoryIfNotInitialized() {
...
+    }
+    return result;
+  }
+
   private static boolean checkIfDisabled() {
     String isDisabled = Environment.getEnvironmentVariable("OPENLINEAGE_DISABLED");
     return Boolean.parseBoolean(isDisabled);
   }
+
+  public static Optional<Long> getEventProcessingTimeoutSeconds() {
```
pawel-big-lebowski, line 438: can we make it `private`?

integration/spark/app/src/test/java/io/openlineage/spark/agent/OpenLineageSparkListenerTest.java
```diff
@@ -131,15 +131,25 @@ void testOpenlineageDisableDisablesExecution() throws URISyntaxException {
...
+        OpenLineageSparkListener.init(contextFactory);
+        OpenLineageSparkListener listener = new OpenLineageSparkListener();
 
-      verify(contextFactory, never()).createSparkSQLExecutionContext(anyLong());
+        listener.onJobStart(
+            new SparkListenerJobStart(
+                0, 2L, ScalaConversionUtils.<StageInfo>asScalaSeqEmpty(), new Properties()));
+
+        verify(contextFactory, never()).createSparkSQLExecutionContext(anyLong());
+      }
```
pawel-big-lebowski, line 152: Is it doable to write a unit test that verifies the behaviour? 

Possible test scenario:
 - create a context that for methods like `context.end(jobEnd);` that sleeps `100ms` 
 - configure timeout to `10ms`
 - verify execution lasting less than 100ms

pawel-big-lebowski (commented):
This feature is really cool. 
Please let me know what do you think about the comments?.

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -104,6 +138,33 @@ private static void sparkSQLExecEnd(SparkListenerSQLExecutionEnd endEvent) {
   /** called by the SparkListener when a job starts */
   @Override
   public void onJobStart(SparkListenerJobStart jobStart) {
+    Optional<Long> eventProcessingTimeoutSeconds =
+        OpenLineageSparkListener.getEventProcessingTimeoutSeconds();
+
+    if (!eventProcessingTimeoutSeconds.isPresent()) {
```
mobuchowski, line 144: Can we move this to wrapper method rather than calling it in every event processing method? As in, modify `callWithTimeout` to do this check

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
athityakumar: We had some instances where the timeout logs / error stack trace became huge with warn/errors (timeout for multiple events), as teams usually hide just the debug logs in production. This kinda became an issue for folks trying to debug their actual spark job logs - that's where we had added it as debug

pawel-big-lebowski:
Hi @athityakumar, I think you are working on something similar to my PR #2407, which is a circuit breaker which prevent generating, serialising and emitting OL event, when a job runs out of memory. I was thinking on what is the best way to make those two in a consistent way: I am working on memory circuit breaker while you're working on timeout circuit breaker. 

My idea is to: 
 * include timeout into circuit breaker config I am working on 
 * extend your `callWithTimeout` in a way so that it checks memory circuit breaker every few secs and verifies if we are not running out of memory. 

Would you mind if I did some Proof of Concept of that in my PR and reuse some of your code? I'll mention you in changelog of course if this approach is going to work. What do you say?

PR #2436, opened 2024-02-15
[FLINK] support lineage for Flink JDBC connector

HuangZhenQiu:
#### One-line summary:
Support Flink lineage for use cases that use JDBC connector.

----


integration/flink/src/main/java/io/openlineage/flink/visitor/wrapper/JdbcSinkWrapper.java
```diff
@@ -0,0 +1,75 @@
...
+package io.openlineage.flink.visitor.wrapper;
+
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.flink.connector.jdbc.JdbcConnectionOptions;
+import org.apache.flink.connector.jdbc.internal.GenericJdbcSinkFunction;
+import org.apache.flink.connector.jdbc.internal.JdbcOutputFormat;
+import org.apache.flink.connector.jdbc.internal.connection.JdbcConnectionProvider;
+import org.apache.flink.connector.jdbc.internal.connection.SimpleJdbcConnectionProvider;
+import org.apache.flink.connector.jdbc.internal.options.InternalJdbcConnectionOptions;
```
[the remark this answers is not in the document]
mobuchowski: Reflection magic :)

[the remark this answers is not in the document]
HuangZhenQiu: I combine the two solution together. But somehow, I find integration test is not triggered, even through it runs successfully in local.

integration/flink/app/src/main/java/io/openlineage/flink/visitor/VisitorFactoryProvider.java
```diff
@@ -0,0 +1,51 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.flink.visitor;
+
+import org.apache.flink.FlinkVersion;
+
+public class VisitorFactoryProvider {
```
HuangZhenQiu, line 10: This class can be removed for now.

integration/flink/flink117/build.gradle
```diff
@@ -0,0 +1,139 @@
...
+    assertjVersion = '3.25.3'
+    junit5Version = '5.10.2'
+    mockitoVersion = '5.2.0'
+    jdbcVersion = '3.1.1-1.17'
+}
+
+dependencies {
+    implementation(project(path: ":shared"))
+    api "io.openlineage:openlineage-java:${project.version}"
+    api "io.openlineage:openlineage-sql-java:1.8.0"
```
HuangZhenQiu, line 102: If I change to ${project.version}, integration test will fail due to fail to load the native libs.

HuangZhenQiu:
@pawel-big-lebowski @mobuchowski 
I have used reflection in the jdbc wrapper. But used separate module for test cases of different flink versions. Please review your PR at your most convenient time.

integration/flink/flink117/build.gradle
mobuchowski: This is working when running on CI, please move it back to `${project.version}` once it passes locally.

integration/flink/shared/src/test/resources/events/expected_cassandra.json
```diff
@@ -0,0 +1,15 @@
+{
```
mobuchowski, line 1: Looks like those files were copied, not moved - is that intentional?

integration/flink/app/src/main/java/io/openlineage/flink/visitor/VisitorFactoryProvider.java
mobuchowski: If unused yet, I'd remove it and add when needed

integration/flink/src/main/java/io/openlineage/flink/visitor/wrapper/JdbcSinkWrapper.java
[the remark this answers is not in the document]
mobuchowski: @HuangZhenQiu which integration test? Can you point to specific one, and version you're testing on?

integration/flink/app/build.gradle
```diff
@@ -0,0 +1,368 @@
+/*
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+import org.apache.tools.ant.filters.ReplaceTokens
+import io.franzbecker.gradle.lombok.task.DelombokTask
+
+plugins {
+    id 'java'
```
d-m-h, line 10: Applying the `java` plugin is unnecessary, as the `java-library` plugin already does that. This can be safely removed.

integration/flink/app/build.gradle
```diff
@@ -0,0 +1,368 @@
...
+    testImplementation "org.mockito:mockito-junit-jupiter:${mockitoVersion}"
+    testImplementation "org.projectlombok:lombok:${lombokVersion}"
+
+    testAnnotationProcessor "org.projectlombok:lombok:${lombokVersion}"
+}
+
+sourceSets {
+    main.java.srcDirs = ["src/main/java"]
+    test.java.srcDirs = ["src/test/java"]
+}
```
d-m-h, line 136: This is unnecessary. The `java-library` plugin will already creates the `main` and `test` source sets, and it configures them. This can be safely removed.

integration/flink/app/build.gradle
```diff
@@ -0,0 +1,368 @@
...
+    mavenLocal()
+    mavenCentral()
+    maven {
+        url = "https://packages.confluent.io/maven/"
+    }
+}
+
+configurations {
+    lombok
+}
```
d-m-h, line 48: You declare this configuration, but I don't see it being used. Can you point me to the spot where you use this configuration?

integration/flink/app/build.gradle
```diff
@@ -0,0 +1,368 @@
...
+    required { isReleaseVersion }
+    def signingKey = findProperty("signingKey")
+    def signingPassword = findProperty("signingPassword")
+    useInMemoryPgpKeys(signingKey, signingPassword)
+    sign publishing.publications.mavenJava
+}
+
+shadowJar {
+    dependsOn(':examples:stateful:shadowJar')
+    dependsOn(':examples:stateful:pmdMain')
```
d-m-h, line 313: Please, do not depend on the tasks of other modules. This can cause Gradle to perform unnecessary configuration work. If you need the outputs of the ":examples:stateful:shadowJar" task, then you should declare a dependency on the `:examples:stateful` module like so:

```groovy
runtimeOnly(project(path: ':examples:stateful', configuration: 'shadow'))
```

The dependency on `:examples:stateful:pmdMain` is confusing to me though. Could you explain why you depend on another modules `pmdMain` task?

integration/flink/examples/stateful/build.gradle
```diff
@@ -58,17 +58,18 @@ ext {
...
+            "1.17": ["cassandra": "3.1.0-1.17", "kafka": flinkVersion, "jdbc": "3.1.1-1.17", "alternativeShort": "1.17"],
+            "1.18": ["cassandra": "3.1.0-1.17", "kafka": "3.0.2-1.18", "jdbc": "3.1.1-1.17", "alternativeShort": "1.17"]
     ]
 
     versions = versionsMap[flinkVersionShort]
 }
 
 dependencies {
-    implementation project(':')
+    implementation(project(path: ":app"))
```
d-m-h, line 71: I'm concerned that we're going to run into a cyclic dependency here. In the `app/build.gradle` you're asking for the output of ':extamples:stateful:shadowJar', but over here, you're asking for the JAR of `app`.

integration/flink/flink115/build.gradle
```diff
@@ -0,0 +1,139 @@
+/*
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+plugins {
+    id 'java'
```
d-m-h, line 7: See my previous note about the necessity of `java` when you're applying `java-library`.

integration/flink/flink115/build.gradle
```diff
@@ -0,0 +1,139 @@
...
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
+}
+
+group 'io.openlineage.flink'
+
+sourceSets {
+    main.java.srcDirs = ["src/main/java"]
+    test.java.srcDirs = ["src/test/java"]
+}
```
d-m-h, line 57: See my previous comments about source set configuration

integration/flink/flink117/build.gradle
```diff
@@ -0,0 +1,139 @@
+/*
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+plugins {
+    id 'java'
```
d-m-h, line 7: See my previous comment about applying the `java` plugin when the `java-library` plugin is already present

integration/flink/flink117/build.gradle
```diff
@@ -0,0 +1,139 @@
...
+pmdMain {
+    dependsOn shadowJar
+    reports {
+        html.required = true
+    }
+}
+
+configurations {
+    lombok
+}
```
d-m-h, line 45: Lombok configuration. Is it used at all in this build?

integration/flink/flink117/build.gradle
```diff
@@ -0,0 +1,139 @@
...
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
+}
+
+group 'io.openlineage.flink'
+
+sourceSets {
+    main.java.srcDirs = ["src/main/java"]
+    test.java.srcDirs = ["src/test/java"]
+}
```
d-m-h, line 57: See my commentary about default source set configuration when the `java-library` plugin is present

integration/flink/flink115/build.gradle
```diff
@@ -0,0 +1,139 @@
...
+pmdMain {
+    dependsOn shadowJar
+    reports {
+        html.required = true
+    }
+}
+
+configurations {
+    lombok
+}
```
d-m-h, line 45: See my previous commentary about the lombok configuration. If you aren't using it, please remove it.

integration/flink/shared/build.gradle
```diff
@@ -0,0 +1,182 @@
+/*
+/* Copyright 2018-2023 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+import io.franzbecker.gradle.lombok.task.DelombokTask
+
+plugins {
+    id 'java'
```
d-m-h, line 9: `java` plugin is applied unncessarily in the presence of the `java-library` plugin

integration/flink/shared/build.gradle
```diff
@@ -0,0 +1,182 @@
...
+pmdMain {
+    dependsOn shadowJar
+    reports {
+        html.required = true
+    }
+}
+
+configurations {
+    lombok
+}
```
d-m-h, line 47: Please review the creation and usage of this configuration

integration/flink/shared/build.gradle
```diff
@@ -0,0 +1,182 @@
...
+    sourceCompatibility = JavaVersion.VERSION_11
+    targetCompatibility = JavaVersion.VERSION_11
+}
+
+group 'io.openlineage.flink'
+
+sourceSets {
+    main.java.srcDirs = ["src/main/java"]
+    test.java.srcDirs = ["src/test/java"]
+}
```
d-m-h, line 59: Please remove this.

d-m-h (changes requested):
Overall, there is a lot of duplication across the build.gradle files. This duplication can be isolated into a, so-called, convention plugin, and uniformly applied to each build.gradle, thereby removing the duplication.

integration/flink/app/build.gradle
HuangZhenQiu: I moved the original build.gradle from root to the app module. I think it just to make integration test run after all of the shadow jar are created. 

https://github.com/OpenLineage/OpenLineage/blob/main/integration/flink/build.gradle#L306

After changing the to runtimeOnly(project(path: ':examples:stateful', configuration: 'shadow')). Error comes from build.

Could not find method project() for arguments [{path=:examples:stateful, configuration=shadow}] on task ':app:shadowJar' of type com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar

integration/flink/examples/stateful/build.gradle
HuangZhenQiu: It is actually not cyclic dependency. Example depends on app in terms of lib. App's shadow jar depends on example's shadow jar.

integration/flink/shared/src/test/resources/events/expected_cassandra.json
HuangZhenQiu: It is not intentional. I moved whole resources folder. Only json files lose the git history. Do you have any suggestion on this?

integration/flink/app/build.gradle
HuangZhenQiu: Removed

integration/flink/app/build.gradle
HuangZhenQiu: Removed

HuangZhenQiu:
@d-m-h 
Thanks for the suggestions. I cleaned useless configs and java plugin as you suggested. I think we can use the similar configuration plugin for Spark to remove the duplicated gradle sections for spotless, Lombok and etc. I will work on it in a separate PR.

integration/flink/src/main/java/io/openlineage/flink/visitor/wrapper/JdbcSinkWrapper.java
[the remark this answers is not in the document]
HuangZhenQiu: @mobuchowski 
The JDBC integration test is added here. It tests on Flink version from 1.16 to 1.18.
https://github.com/OpenLineage/OpenLineage/pull/2436/files#diff-a720abad723520bf34db6903161fc07623dcab146baf7dcd0cb7c3b8b0a05f19R311

integration/flink/app/build.gradle
d-m-h: Oh, my apologies, I should have been more clearer.

The dependency should be added to the module's `dependencies { }` block.

Concretely,

```groovy
dependencies {
	// other dependencies

	runtimeOnly(project(path: ':examples:stateful', configuration: 'shadow'))

	// test dependencies
}
```

You can then verify that the project appears on the runtime classpath by doing the following:

```groovy
tasks.register("printSourceSetDependencyInformation") {
    group = "debug"
    description = "Prints the dependencies per configuration for a given source set, use '-Psource.set.name=<value>'"

    doLast {
        final String sourceSetName = project.findProperty('source.set.name').toString()
        if (sourceSetName == null) {
            throw new GradleException("Please provide a source set name using '-Psource.set.name=<value>'")
        }

        final SourceSet sourceSet = sourceSets.getByName(sourceSetName)
        if (sourceSet == null) {
            throw new GradleException("Source set with name $sourceSetName does not exist.")
        }

        final List<String> configurationNames = Arrays.stream(SourceSet.class.getDeclaredMethods())
                .filter { Method m -> m.name.startsWith("get") && m.name.endsWith("ConfigurationName") }
                .map { Method m -> m.invoke(sourceSet).toString() }
                .sorted(Comparator.naturalOrder())
                .collect(Collectors.toList())

        Consumer<String> printer = { String configurationName ->
            println("Dependencies for configuration: ${configurationName}")

            try {
                final def deps = configurations.getByName(configurationName).dependencies
                if (deps.isEmpty()) {
                    println("No dependencies have been added to this configuration.")
                } else {
                    deps.forEach { Dependency dependency ->
                        println("- ${dependency.group}:${dependency.name}:${dependency.version}")
                    }
                }
            } catch (UnknownConfigurationException e) {
                println("Configuration $configurationName does not exist.")
                return
            } finally {
                println()
            }
        }

        configurationNames.forEach { String configurationName ->
            printer.accept(configurationName)
        }
    }
}
```

You can run the above task like so:

```bash
./gradlew app:printSourceSetDependencyInformation -Psource.set.name=main
```

You want to make sure that the `runtimeClasspath` has the `example stateful` entry there. (The axact name can very, based on how the projects are configured)

HuangZhenQiu: @d-m-h 
Thanks for the suggestion. Using the fixtures jar in dependencies is a better solution. It does't need to build shadow jar for app module. Please review the change.

integration/flink/shared/src/test/resources/events/expected_cassandra.json
HuangZhenQiu: @mobuchowski 
Double checked in IntelliJ. The git history is kept for these files, as git blame still shows original author. I guess it is a bug in GitHub for this json files.

integration/flink/app/build.gradle
```diff
@@ -0,0 +1,249 @@
...
+        }
+        html {
+            enabled true
+            destination = file(coverageDir)
+        }
+    }
+}
+
+spotlessJava {
+    dependsOn("compileJava", "compileTestJava", ":examples:stateful:generateAvroJava")
```
d-m-h, line 232: Now that 'app' doesn't depend on the :examples:stateful project for anything other than fixtures, you can remove the dependency on ':examples:stateful:generateAvroJava'.

I must say, it looks much better now.

HuangZhenQiu: @d-m-h Thanks for the suggestion on the improvement.

mobuchowski (approved):
Thanks for the contribution @HuangZhenQiu !

PR #2455, opened 2024-02-22
spark: use prebuild Spark images from quay.io

mobuchowski:
**PROBLEM**

The CI/CD pipeline sometimes times out when building the docker images needed for the integration test, due to throttling that the Apache archive applies to the downloads of the Spark binaries.

**SOLUTION**

We built and pushed Spark images to `quay.io` under the `openlineage/spark` repository. We now use those images in the integration tests.


mobuchowski:
Will move this gradle plugin to a separate place in the codebase in following PR.

d-m-h (approved):
Looks good to me

PR #2457, opened 2024-02-23
spark: add runtime property facet

Ruihua98:
#### Summary:
Add a new facet called RuntimePropertyFacet to capture RuntimeConfig of Spark session, because the existing SparkPropertyFacet can only capture static config of spark context.
This Facet will be added in both RDD-related runs and SQL-related runs. Users can use this feature by setting spark.openlineage.capturedRuntimeProperties.

#### One-line summary:
Add a new facet called RuntimePropertyFacet to capture RuntimeConfig of Spark session.

### Checklist

----


d-m-h:
Hi @Ruihua98 - thanks for your contribution. Going over your code, it seems your justification for creating this is only because the existing `integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/SparkPropertyFacetBuilder.java` file uses 'SparkContext' and in your approach you are using `SparkSession`?

I'm not sure I fully understand what this is attempting to do then, because the `SparkConf` object is shared between the `SparkSession` and `SparkContext` objects.

Could you elaborate a bit more on where the current approach falls short?

mobuchowski:
@Ruihua98 to be honest, I was thinking more of modifying existing `SparkPropertyFacetBuilder` to additionally capture runtime properties and adding it to RDDExecutionContext. Are there some problems with doing it that way? I don't think it breaks anything, although I might be wrong.

Ruihua98:
> @Ruihua98 to be honest, I was thinking more of modifying existing `SparkPropertyFacetBuilder` to additionally capture runtime properties and adding it to RDDExecutionContext. Are there some problems with doing it that way? I don't think it breaks anything, although I might be wrong.

Hi~ We agree to modify the existing facet to achieve this. In fact, I have tried to write a [draft](https://github.com/OpenLineage/OpenLineage/compare/main...Ruihua98:OpenLineage:1.3.1-mod1) before I submit this PR. I found the code in that draft branch a little bit redundant. 
1.  I don't know how to use SparkPropertyFacetBuilder in RddExecutionContext because I didn't find  OpenLineageContext object in RddExecutionContext and this object is a parameter in SparkPropertyFacetBuilder's constructor method. 
2.  I don't know how to deal with these lines of code in SparkPropertyFacetBuilder because I didn't find event.properties() in SparkListenerEvent object. (maybe I was wrong here)
` event.properties().entrySet().stream().filter(e -> allowerProperties.contains(e.getKey())).forEach(e -> m.putIfAbsent(e.getKey().toString(), e.getValue()))`

The two reasons above make the code in my draft branch complex, so I finally submit a new facet. I fully agree to change my code to modify the old facet if you can help me find out how to solve the above two problems.

mobuchowski:
>I don't know how to use SparkPropertyFacetBuilder in RddExecutionContext because I didn't find OpenLineageContext object in RddExecutionContext and this object is a parameter in SparkPropertyFacetBuilder's constructor method.

Feel free to add, for example `SparkPropertyRddExecutionFacetBuilder` that would accept objects that you need to construct the facet, or modify `SparkPropertyFacetBuilder` to be able to use it from both contexts.
For example, `OpenLineageContext` is used only to get `SparkConf` which could be passed directly:
```
    conf = context.getSparkContext().getConf();
```

>I don't know how to deal with these lines of code in SparkPropertyFacetBuilder because I didn't find event.properties() in SparkListenerEvent object. (maybe I was wrong here)
 event.properties().entrySet().stream().filter(e -> allowerProperties.contains(e.getKey())).forEach(e -> m.putIfAbsent(e.getKey().toString(), e.getValue()))
 
This code assumes that it's called with particular `SparkListenerJobStart` event, not generic `SparkListenerEvent`. If you need to emit this on other events, then two options would be to either add different `FacetBuilder` without that assumption, or modify code to check if event is `SparkListenerJobStart` using `instanceof`, and only adding those properties in that case.

I would go with modifying existing `SparkPropertyFacetBuilder`, because then there is no worry what happens if both try to add the same facet.

Ruihua98:
> > I don't know how to use SparkPropertyFacetBuilder in RddExecutionContext because I didn't find OpenLineageContext object in RddExecutionContext and this object is a parameter in SparkPropertyFacetBuilder's constructor method.
> 
> Feel free to add, for example `SparkPropertyRddExecutionFacetBuilder` that would accept objects that you need to construct the facet, or modify `SparkPropertyFacetBuilder` to be able to use it from both contexts. For example, `OpenLineageContext` is used only to get `SparkConf` which could be passed directly:
> 
> ```
>     conf = context.getSparkContext().getConf();
> ```
> 
> > I don't know how to deal with these lines of code in SparkPropertyFacetBuilder because I didn't find event.properties() in SparkListenerEvent object. (maybe I was wrong here)
> > event.properties().entrySet().stream().filter(e -> allowerProperties.contains(e.getKey())).forEach(e -> m.putIfAbsent(e.getKey().toString(), e.getValue()))
> 
> This code assumes that it's called with particular `SparkListenerJobStart` event, not generic `SparkListenerEvent`. If you need to emit this on other events, then two options would be to either add different `FacetBuilder` without that assumption, or modify code to check if event is `SparkListenerJobStart` using `instanceof`, and only adding those properties in that case.
> 
> I would go with modifying existing `SparkPropertyFacetBuilder`, because then there is no worry what happens if both try to add the same facet.

Thank you very much. I have finished another version by only modifying existing `SparkPropertyFacetBuilder` with your help. [New Branch](https://github.com/Ruihua98/OpenLineage/tree/feature/modify-spark-property-facet) Should I end this PR and create a new PR on the new branch?

mobuchowski:
@Ruihua98 either open a new PR or push changes to this one, I'm okay with both :)

Ruihua98:
> Add a new facet called RuntimePropertyFacet to capture RuntimeConfig of Spark session, because the existing SparkPropertyFacet can only capture static config of spark context.
> This Facet will be added in both RDD-related runs and SQL-related runs.

I have submitted a New PR https://github.com/OpenLineage/OpenLineage/pull/2480 , and this old one can be closed.

PR #2475, opened 2024-02-28
[SPARK][FLINK][JAVA] http timeout fix

### LINKED ISSUE #2474: `HttpTransport` timeout unit unclear
In the `HttpTransport`, the docs specify `millis` as the unit, see [`HttpTransport.withTimeout(Double)`](https://github.com/OpenLineage/OpenLineage/blob/main/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java#L53), but if the timeout is overwritten, the client multiples the timeout * `1000`:

```
private static CloseableHttpClient withTimeout(Double timeout) {
    int timeoutMs;
    if (timeout == null) {
      timeoutMs = 5000;
    } else {
      timeoutMs = (int) (timeout * 1000);
    }
...
```

* Why do we use `double`, not `int`?
* If a user passes `2000`, the timeout would be (based on the above) `2000` * `1000` = `2000000`

pawel-big-lebowski:
### Problem

Documentation says timeout should be provided in millis
Screenshot 2024-02-28 at 09 56 37

But the code treats it as if it was provided in seconds 

Fixes #2474

### Solution

 * Write failing test
 * Modify the code 

### Checklist

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 # Changelog
 
 ## [Unreleased](https://github.com/OpenLineage/OpenLineage/compare/1.9.1...HEAD)
+### Added
 * **Flink: improve Cassandra lineage metadata** (https://github.com/OpenLineage/OpenLineage/pull/2479) [@HuangZhenQiu](https://github.com/HuangZhenQiu)
   *Use Cassandra cluster info as dataset namespace, and combine keyspace with table name as dataset name.*
 * **Flink: bump Flink JDBC connector version to 3.1.2-1.18 for Flink 1.18** (https://github.com/OpenLineage/OpenLineage/pull/2472) [@HuangZhenQiu](https://github.com/HuangZhenQiu)  
   *Bump Flink JDBC connector version to 3.1.2-1.18 for Flink 1.18.*
+
+### Fixed
+* **Spark: fix `HttpTransport` timeout.** [`#2475`](https://github.com/OpenLineage/OpenLineage/pull/2475) [@pawel-big-lebowski](https://github.com/pawel-big-lebowski)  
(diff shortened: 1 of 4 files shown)


mobuchowski:
@pawel-big-lebowski I think we should do this other way around and make docs fit the reality rather than breaking the timeout for people using floats

pawel-big-lebowski:
> @pawel-big-lebowski I think we should do this other way around and make docs fit the reality rather than breaking the timeout for people using floats

@mobuchowski  I was thinking about this as well and I think we shouldn't. Suppose there's a user who configured timeout to 10000 and lives unaware of the fact timeout settings will not work. Once we fix it, timeout will work again. If we change the behaviour, timeout settings for such users will get broken even more. 

So the tradeoff is: shall we break the contract for those who follow the docs and are unaware of the problem or shall we break the contract for those (probably a single user) who found out the cause and hacked the system?

Additionally, circuit breaker timeout also uses milliseconds.

d-m-h:
My 2 cents:

1. Add an explicit configuration to specify the time unit (minutes, seconds, milliseconds)
2. Default the above time unit to milliseconds
3. Allow users to configure accordingly.

This way, you remove the ambiguity of "is this 5 milliseconds, or 5 seconds?"

pawel-big-lebowski:
Nice idea from offline discussions to make as many people happy as possible
 * We leave `timeout` as is in http config (treat it in seconds, remove it from docs), we mark config entry deprecated in the code
 * We include `timeoutMillis`  config entry and take its value for timeout if it's present

@mobuchowski would it make sense to you?

tatiana:
> Nice idea from offline discussions to make as many people happy as possible
> 
> * We leave `timeout` as is in http config (treat it in seconds, remove it from docs), we mark config entry deprecated in the code
> * We include `timeoutMillis`  config entry and take its value for timeout if it's present
> 
> @mobuchowski would it make sense to you?

I really like this strategy, @pawel-big-lebowski , since it is backwards compatible and also supports users specifying milliseconds. Credits to @julienledem from our chat yesterday :D

PR #2484, opened 2024-03-02
upgrade spotless

julienledem:
### Problem

the current version of spotless we use does not work with jdk 21
https://github.com/diffplug/spotless/issues/1819

### Solution

upgrade spotless plugin


client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
```diff
@@ -13,8 +13,10 @@
 public class DatasetIdentifierUtils {
 
   private static final String DEFAULT_SCHEME = "file";
+
   /** The directory separator, a slash, as a character. */
   public static final char SEPARATOR_CHAR = '/';
+
```
julienledem, line 19: an I ran `./gradlew :spotlessApply`

pawel-big-lebowski (approved):
looks good to me

PR #2486, opened 2024-03-03
client: add info from http response to client exception

davidjgoss:
### Problem

When using the `HttpTransport` with the Java client, if the receiving service (Marquez or whatever) responds with a non-2xx status then an `OpenLineageClientException` is thrown with a descriptive message. The consuming code may want to do something different with that exception depending on the response - e.g. we might retry the operation for a 502 but not for a 400. Currently this information has to be parsed out of the exception message.

### Solution

This PR adds `HttpTransportResponseException`, a subclass of `OpenLineageClientException` that includes fields for status code and body - this is what is now thrown in the event of a non-2xx response. Note that other exception-raising scenarios like timeouts and bad config will behave as previously. Also, the `message` on the exception remains the same for backwards compatibility.

#### One-line summary:

Adds the status code and body as properties on the thrown exception when a non-success response is encountered in the HTTP transport.

### Checklist

----

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
@@ -147,11 +147,8 @@ public final class HttpTransport extends Transport implements Closeable {
   private void throwOnHttpError(@NonNull HttpResponse response) throws IOException {
     final int code = response.getStatusLine().getStatusCode();
     if (code >= 400 && code < 600) { // non-2xx
-      String message =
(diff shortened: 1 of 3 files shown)


pawel-big-lebowski (approved):
Looks good to me. 

Minor: wouldn't `HttpTransportResponseException` constructor be a better place to create `message`?

davidjgoss:
> Minor: wouldn't HttpTransportResponseException constructor be a better place to create message?

Good point, let's do that.

PR #2491, opened 2024-03-05
[FLINK] support lineage for hybrid source

HuangZhenQiu:
#### One-line summary:
Support lineage for hybrid source users who combine KafkaSource and IcebergSource.

----

### DIFF
--- a/integration/flink/shared/src/main/java/io/openlineage/flink/visitor/HybridSourceVisitor.java
+++ b/integration/flink/shared/src/main/java/io/openlineage/flink/visitor/HybridSourceVisitor.java
@@ -1,0 +1,74 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.flink.visitor;
+
+import io.openlineage.client.OpenLineage;
+import io.openlineage.flink.api.OpenLineageContext;
+import io.openlineage.flink.visitor.wrapper.WrapperUtils;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.flink.api.connector.source.Source;
+import org.apache.flink.connector.base.source.hybrid.HybridSource;
+import org.apache.flink.connector.base.source.hybrid.HybridSource.SourceFactory;
+
+@Slf4j
+public class HybridSourceVisitor extends Visitor<OpenLineage.InputDataset> {
+  public static final String SOURCE_LIST_ENTRY_CLASS =
+      "org.apache.flink.connector.base.source.hybrid.HybridSource$SourceListEntry";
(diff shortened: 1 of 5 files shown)


integration/flink/shared/src/main/java/io/openlineage/flink/visitor/HybridSourceVisitor.java
```diff
@@ -0,0 +1,82 @@
...
+                WrapperUtils.<SourceFactory>getFieldValue(
+                    sourceListEntryClass, sourceEntry, "factory");
+            if (sourceListOpt.isPresent()) {
+              Class passThroughSourceFactory = Class.forName(PASS_THROUGH_SOURCE_FACTORY_CLASS);
+              Optional<Source> sourceOpt =
+                  WrapperUtils.<Source>getFieldValue(
+                      passThroughSourceFactory, sourceFactoryOpt.get(), "source");
+              if (sourceOpt.isPresent()) {
+                Source source = sourceOpt.get();
+                if (source instanceof KafkaSource) {
```
pawel-big-lebowski, line 61: What if we want to add more sources and visitors here in the future? I think this should be run on all `getInputVisitors`  from `VisitorFactoryImpl.java`. Would we be able to reuse `io.openlineage.flink.visitor.lifecycle.FlinkExecutionContext#getInputDatasets` for this? This is looking really similar, at least for the first glance.

HuangZhenQiu: Make sense. I think it is a good suggtion. I will do the change.

integration/flink/shared/src/main/java/io/openlineage/flink/visitor/HybridSourceVisitor.java
```diff
@@ -0,0 +1,78 @@
...
+              Optional<Source> sourceOpt =
+                  WrapperUtils.<Source>getFieldValue(
+                      passThroughSourceFactory, sourceFactoryOpt.get(), "source");
+              if (sourceOpt.isPresent()) {
+                Source source = sourceOpt.get();
+                VisitorFactoryImpl visitorFactory = new VisitorFactoryImpl();
+                visitorFactory.getInputVisitors(context).stream()
+                    .forEach(
+                        visitor -> {
+                          if (visitor.isDefinedAt(source)) {
```
pawel-big-lebowski, line 63: this can be extracted to filter on the stream two lines above

pawel-big-lebowski (approved):
Looks good to me 👍 🚀  Minor comment to look at.
It is incredible to see @HuangZhenQiu continuously contributing to the project. Love that.

integration/flink/shared/src/main/java/io/openlineage/flink/visitor/HybridSourceVisitor.java
HuangZhenQiu: Thanks for the suggestion.. Simplified accordingly.

PR #2496, opened 2024-03-07
metrics, java: add support for Micrometer metrics

mobuchowski:
This PR adds support for Micrometer-based telemetry. This comprises of: 

- adding `MeterRegistryFactory` mechanism that allows to construct Micrometer's `MeterRegistry` based on passed configuration
- adding `MicrometerProvider` mechanism that allows you to configure metrics from any OpenLineage integration and get configured `MeterRegistry` instance
- adding proper metrics config to OpenLineage config
- adding `StatsDMetricsBuilder`
- implementing instrumentation into Java client. 

The metrics mechanism can forward metrics to any Micrometer compatible implementation - list [here](https://docs.micrometer.io/micrometer/reference/implementations.html) - however compatible wrapper `MeterRegistryFactory` needs to be added. This PR adds only support for StatsD based one, as well as in-memory `SimpleMetricsBuilder` and `CompositeMetricsBuilder` that allows you to configure multiple backends. However, global registry provided by `MicrometerProvider` is a `CompositeMetricsBuilder`, so that using it manually is generally unadvisable.

PR #2528, opened 2024-03-18
metrics, spark: add support for telemetry mechanism in Spark integration 

mobuchowski:
This PR follows previous metrics PR https://github.com/OpenLineage/OpenLineage/pull/2496 and adds instrumentation to the Spark integration. This comprises of 

- setting global Spark version, OpenLineage-Spark version, disabled facets tags
- emitting event received counters (Application, Job, SQL Start/End) 
- adding timers around RunFacets
- adding serialized metrics to DebugFacet


client/java/build.gradle
```diff
@@ -63,10 +64,13 @@ dependencies {
     implementation 'org.apache.httpcomponents:httpclient:4.5.14'
     implementation 'commons-logging:commons-logging:1.3.0'
     implementation 'org.slf4j:slf4j-api:1.7.36'
+    implementation "io.micrometer:micrometer-core:${micrometerVersion}"
```
pawel-big-lebowski, line 67: This has some extra dependencies. Moreover this can clash with users having micrometer already installed on their class path. Could we assume users provide this manually? Anyway they need to add jar specific to a micrometer backend they chose.

client/java/src/main/java/io/openlineage/client/metrics/MetricsResolver.java
```diff
@@ -0,0 +1,114 @@
...
+   */
+  public static MeterRegistry clear() {
+    registry.close();
+    registry = new CompositeMeterRegistry();
+    return registry;
+  }
+
+  private static List<MetricsBuilder> getMetricsBuilders() {
+    List<MetricsBuilder> builders = new ArrayList<>();
+    if (ReflectionUtils.hasClass("io.micrometer.statsd.StatsdMeterRegistry")) {
```
pawel-big-lebowski, line 103: I am assuming this is done like this bcz there is no way to automatically inject config entries to micrometer. In other words, for a user wanting to send metrics to backend `X`, we need to implement `XMetricsBuilder`. If this is necessary, please keep in mind to state this in docs.

client/java/src/test/resources/config/metrics-complex.yaml
```diff
@@ -0,0 +1,6 @@
+transport:
```
pawel-big-lebowski, line 1: minor: filename should be `composite` as well, right?

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -96,17 +106,20 @@ public void onOtherEvent(SparkListenerEvent event) {
...
     getSparkSQLExecutionContext(startEvent.executionId())
         .ifPresent(
-            context ->
-                circuitBreaker.run(
-                    () -> {
-                      context.start(startEvent);
-                      return null;
-                    }));
+            context -> {
+              meterRegistry.counter("openlineage.spark.event.sql.start").increment();
```
pawel-big-lebowski, line 110: ❤️

client/java/src/main/java/io/openlineage/client/metrics/MetricsResolver.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Adds a MeterRegistry to the common OpenLineage meter registry based on the provided
+   * configuration.
+   *
+   * @param config The configuration for the MeterRegistry.
+   * @return Common registry configured with the added MeterRegistry.
+   */
+  public static MeterRegistry addMeterRegistryFromConfig(Map<String, Object> config) {
```
pawel-big-lebowski, line 64: Would love to find later on in the docs short definition of: meter, metric, meter registry and metric builder. 
I found that non-straightforward: metric resolver class reading meter registry from config. 
If a class resolves MeterRegistry from config, shouldn't it be `MeterRegistryResolver` instead of `MetricResolver`?

pawel-big-lebowski (commented):
Left few single comments' feedback above. Overall I am really happy with the progress you do on the PR and the shape it becomes. Amazing job @mobuchowski 🥇

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -257,6 +272,11 @@ private static void clear() {
 
   @Override
   public void onApplicationEnd(SparkListenerApplicationEnd applicationEnd) {
+    meterRegistry.counter("openlineage.spark.event.app.end").increment();
+    meterRegistry
+        .counter("openlineage.spark.event.app.end.memoryusage")
+        .increment(
+            (100.0 * (double) RuntimeUtils.totalMemory()) / (double) RuntimeUtils.maxMemory());
```
pawel-big-lebowski, line 279: how about adding `getMemoryUsage` into RuntimeUtils ?

pawel-big-lebowski (commented):
This code looks really well and I appreciate the solution to extract timers for facet builders. 

Did you experience any problems with testing? I couldn't find any new tests added. Would it make sense to see a debug facet filled with timers for each visitor and builder?

client/java/src/main/java/io/openlineage/client/metrics/MetricsResolver.java
mobuchowski: Comment is obsoleted, it's now `MeterRegistryFactory`.

client/java/src/test/resources/config/metrics-complex.yaml
mobuchowski: Fixed.

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
mobuchowski: Done.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkGenericIntegrationTest.java
```diff
@@ -115,6 +116,43 @@ void sparkEmitsApplicationLevelEvents() {
...
   }
 
+  @Test
+  @SneakyThrows
+  void sparkGathersMetrics() {
+    Dataset<Row> df = createTempDataset();
+
+    Dataset<Row> agg = df.groupBy("a").count();
+    agg.write().mode("overwrite").csv("/tmp/test_data/test_output/");
+    Thread.sleep(1000);
```
pawel-big-lebowski, line 126: It's preferred to use await library to avoid `Thread.sleep`. Is there a reason for doing it this way?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/JobMetricsHolder.java
```diff
@@ -60,7 +60,7 @@ public Map<Metric, Number> pollMetrics(int jobId) {
   public void cleanUp(int jobId) {
     Set<Integer> stages = jobStages.remove(jobId);
     stages = stages == null ? Collections.emptySet() : stages;
-    stages.forEach(s -> jobStages.remove(s));
+    stages.forEach(s -> stageMetrics.remove(s));
```
pawel-big-lebowski, line 63: still valid?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
```diff
@@ -503,15 +516,35 @@ private OpenLineage.JobFacets buildJobFacets(
...
       OpenLineage.JobFacetsBuilder jobFacetsBuilder) {
-    events.forEach(event -> builders.forEach(fn -> fn.accept(event, jobFacetsBuilder::put)));
+    events.forEach(
+        event ->
+            builders.forEach(
+                fn ->
+                    openLineageContext
+                        .getMeterRegistry()
+                        .timer(
+                            "openlineage.spark.facets.job.execution.time",
```
pawel-big-lebowski, line 526: would it make sense to have a single class to store all the metric names as constants?

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkContainerIntegrationTest.java
```diff
@@ -312,4 +312,16 @@ void testRddWithParquet() {
 
     verifyEvents(mockServerClient, "pysparkRDDWithParquetComplete.json");
   }
+
+  @Test
+  void testMetricsEmit() {
```
pawel-big-lebowski, line 317: Would it make sense to rename it to `testDebugFacet`? 
It could be useful to add other assertions related to debug Facet within the same test in the future.

pawel-big-lebowski (commented):
Minor cosmetic comments + a question about `thread.sleep`  which I would love to avoid

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkGenericIntegrationTest.java
mobuchowski: Removed

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
mobuchowski: Done

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkGenericIntegrationTest.java
mobuchowski: Removed

integration/spark/shared/src/main/java/io/openlineage/spark/agent/JobMetricsHolder.java
mobuchowski: Rebased

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
mobuchowski: No, but I think particular classes could hold that - however most often the name is used one time so I don't see very much use of it.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkContainerIntegrationTest.java
mobuchowski: Done

Issue #2499, opened 2024-03-08
[Bug] openlineage-spark crashes spark driver on Databricks


efthymiosh:
We are in the process of configuring the `io.openlineage.spark.agent.OpenLineageSparkListener` for our databricks compute.

Running workloads crashes the spark driver with the following stacktrace logged:

```
24/03/07 14:14:05 ERROR Utils: throw uncaught fatal error in thread spark-listener-group-shared
java.lang.NoClassDefFoundError: com/databricks/sdk/scala/dbutils/DbfsUtils
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.getDbfsUtils(DatabricksEnvironmentFacetBuilder.java:124)
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.getDatabricksEnvironmentalAttributes(DatabricksEnvironmentFacetBuilder.java:92)
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.build(DatabricksEnvironmentFacetBuilder.java:58)
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.build(DatabricksEnvironmentFacetBuilder.java:32)
	at io.openlineage.spark.api.CustomFacetBuilder.accept(CustomFacetBuilder.java:40)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.lambda$null$27(OpenLineageRunEventBuilder.java:508)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.lambda$buildRunFacets$28(OpenLineageRunEventBuilder.java:508)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.buildRunFacets(OpenLineageRunEventBuilder.java:508)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.populateRun(OpenLineageRunEventBuilder.java:328)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.buildRun(OpenLineageRunEventBuilder.java:304)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.buildRun(OpenLineageRunEventBuilder.java:265)
[...]

PR #2530, opened 2024-03-20
fix: missing libs in final jar

algorithmy1:
### Problem

Since OpenLineage 1.9.1, we started getting this error in Databricks environments

```
ERROR Utils: throw uncaught fatal error in thread spark-listener-group-shared
java.lang.NoClassDefFoundError: com/databricks/sdk/scala/dbutils/DbfsUtils
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.getDbfsUtils(DatabricksEnvironmentFacetBuilder.java:124)
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.getDatabricksEnvironmentalAttributes(DatabricksEnvironmentFacetBuilder.java:92)
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.build(DatabricksEnvironmentFacetBuilder.java:58)
	at io.openlineage.spark.agent.facets.builder.DatabricksEnvironmentFacetBuilder.build(DatabricksEnvironmentFacetBuilder.java:32)
	at io.openlineage.spark.api.CustomFacetBuilder.accept(CustomFacetBuilder.java:40)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.lambda$null$27(OpenLineageRunEventBuilder.java:508)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.lambda$buildRunFacets$28(OpenLineageRunEventBuilder.java:508)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.buildRunFacets(OpenLineageRunEventBuilder.java:508)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.populateRun(OpenLineageRunEventBuilder.java:328)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.buildRun(OpenLineageRunEventBuilder.java:304)
	at io.openlineage.spark.agent.lifecycle.OpenLineageRunEventBuilder.buildRun(OpenLineageRunEventBuilder.java:265)
	at io.openlineage.spark.agent.lifecycle.SparkSQLExecutionContext.start(SparkSQLExecutionContext.java:221)
	at io.openlineage.spark.agent.OpenLineageSparkListener.lambda$null$13(OpenLineageSparkListener.java:178)
	at io.openlineage.client.circuitBreaker.NoOpCircuitBreaker.run(NoOpCircuitBreaker.java:27)
	at io.openlineage.spark.agent.OpenLineageSparkListener.lambda$onJobStart$14(OpenLineageSparkListener.java:176)
	at java.util.Optional.ifPresent(Optional.java:159)
	at io.openlineage.spark.agent.OpenLineageSparkListener.onJobStart(OpenLineageSparkListener.java:172)
	at org.apache.spark.scheduler.SparkListenerBus.doPostEvent(SparkListenerBus.scala:37)
	at org.apache.spark.scheduler.SparkListenerBus.doPostEvent$(SparkListenerBus.scala:28)
	at org.apache.spark.scheduler.AsyncEventQueue.doPostEvent(AsyncEventQueue.scala:42)
	at org.apache.spark.scheduler.AsyncEventQueue.doPostEvent(AsyncEventQueue.scala:42)
	at org.apache.spark.util.ListenerBus.postToAll(ListenerBus.scala:118)
	at org.apache.spark.util.ListenerBus.postToAll$(ListenerBus.scala:102)
	at org.apache.spark.scheduler.AsyncEventQueue.super$postToAll(AsyncEventQueue.scala:114)
	at org.apache.spark.scheduler.AsyncEventQueue.$anonfun$dispatch$1(AsyncEventQueue.scala:114)
	at scala.runtime.java8.JFunction0$mcJ$sp.apply(JFunction0$mcJ$sp.java:23)
	at scala.util.DynamicVariable.withValue(DynamicVariable.scala:62)
	at org.apache.spark.scheduler.AsyncEventQueue.org$apache$spark$scheduler$AsyncEventQueue$$dispatch(AsyncEventQueue.scala:109)
	at org.apache.spark.scheduler.AsyncEventQueue$$anon$2.$anonfun$run$1(AsyncEventQueue.scala:105)
	at org.apache.spark.util.Utils$.tryOrStopSparkContext(Utils.scala:1493)
	at org.apache.spark.scheduler.AsyncEventQueue$$anon$2.run(AsyncEventQueue.scala:105)
Caused by: java.lang.ClassNotFoundException: com.databricks.sdk.scala.dbutils.DbfsUtils
	at java.net.URLClassLoader.findClass(URLClassLoader.java:387)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:419)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:352)
	... 33 more
```

Closes: https://github.com/OpenLineage/OpenLineage/issues/2499

### Solution

The package `com.databricks:dbutils-api_2.11:0.0.4` and `com.databricks:databricks-sdk-java:0.14.0` exist bu default in the databricks clusters while the package `com.databricks:databricks-dbutils-scala_2.12:0.1.4` need to be added to the jar.

#### One-line summary:

Add the `com.databricks:databricks-dbutils-scala_2.12:0.1.4` in the `gradle.build`.

### Checklist

----

### DIFF
--- a/integration/spark/shared/build.gradle
+++ b/integration/spark/shared/build.gradle
@@ -47,7 +47,7 @@ dependencies {
     compileOnly("org.apache.spark:spark-sql-kafka-0-10_${scalaBinaryVersion}:${sparkVersion}")
     compileOnly("org.apache.spark:spark-sql_${scalaBinaryVersion}:${sparkVersion}")
 
-    compileOnly("com.databricks:databricks-dbutils-scala_${scalaBinaryVersion}:${databricksVersion}") {
+    implementation("com.databricks:databricks-dbutils-scala_${scalaBinaryVersion}:${databricksVersion}") {
         exclude(group: "com.fasterxml.jackson.core")
     }
     compileOnly("com.google.cloud.spark:spark-bigquery-with-dependencies_${scalaBinaryVersion}:${bigqueryVersion}") {
@@ -94,7 +94,7 @@ dependencies {
     scala212CompileOnly("org.apache.spark:spark-sql-kafka-0-10_2.12:${sparkVersion}")
     scala212CompileOnly("org.apache.spark:spark-sql_2.12:${sparkVersion}")
 
-    scala212CompileOnly("com.databricks:databricks-dbutils-scala_2.12:${databricksVersion}") {
+    scala212Implementation("com.databricks:databricks-dbutils-scala_2.12:${databricksVersion}") {
         exclude(group: "com.fasterxml.jackson.core")
     }
     scala212CompileOnly("com.google.cloud.spark:spark-bigquery-with-dependencies_2.12:${bigqueryVersion}") {
@@ -140,7 +140,7 @@ dependencies {
     scala213CompileOnly("org.apache.spark:spark-sql-kafka-0-10_2.13:${sparkVersion}")
     scala213CompileOnly("org.apache.spark:spark-sql_2.13:${sparkVersion}")
 
-    scala213CompileOnly("com.databricks:databricks-dbutils-scala_2.13:${databricksVersion}") {
+    scala213Implementation("com.databricks:databricks-dbutils-scala_2.13:${databricksVersion}") {
         exclude(group: "com.fasterxml.jackson.core")
     }
     scala213CompileOnly("com.google.cloud.spark:spark-bigquery-with-dependencies_2.13:${bigqueryVersion}") {
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
@@ -110,7 +110,7 @@ public class DatabricksEnvironmentFacetBuilder
   private static Optional<DbfsUtils> getDbfsUtils()
       throws ClassNotFoundException, InstantiationException, IllegalAccessException,
           IllegalArgumentException, InvocationTargetException {
-    Class dbutilsClass = Class.forName("com.databricks.dbutils_v1.impl.DbfsUtilsImpl");
+    Class dbutilsClass = Class.forName("com.databricks.sdk.scala.dbutils.SdkDBUtilsImpl");
     Constructor[] dbutilsConstructors = dbutilsClass.getDeclaredConstructors();
     if (dbutilsConstructors.length == 0) {
       log.warn(
--- a/integration/spark/app/build.gradle
+++ b/integration/spark/app/build.gradle
@@ -109,7 +109,7 @@ dependencies {
     testImplementation("org.mockito:mockito-core:${mockitoVersion}")
     testImplementation("org.mockito:mockito-inline:${mockitoVersion}")
     testImplementation("org.mockito:mockito-junit-jupiter:${mockitoVersion}")
-    testImplementation("com.databricks:databricks-sdk-java:0.4.0") {
+    testImplementation("com.databricks:databricks-sdk-java:0.7.0") {
         exclude group: 'com.fasterxml.jackson.core'
         exclude group: 'com.fasterxml.jackson.module'
     }
--- a/integration/spark/app/src/test/java/io/openlineage/spark/agent/DatabricksUtils.java
+++ b/integration/spark/app/src/test/java/io/openlineage/spark/agent/DatabricksUtils.java
@@ -189,10 +189,11 @@ public class DatabricksUtils {
             .build();
 
     log.info("Creating cluster");
+    Map<String, String> emptyHeaders = new HashMap<>();
     CreateClusterResponse response =
         workspace
             .apiClient()
-            .POST("/api/2.0/clusters/create", createCluster, CreateClusterResponse.class);
+            .POST("/api/2.0/clusters/create", createCluster, CreateClusterResponse.class, emptyHeaders);
 
     return response.getClusterId();
   }
--- a/integration/spark/app/src/test/java/io/openlineage/spark/agent/DatabricksUtils.java
+++ b/integration/spark/app/src/test/java/io/openlineage/spark/agent/DatabricksUtils.java
@@ -193,7 +193,11 @@ public class DatabricksUtils {
     CreateClusterResponse response =
         workspace
             .apiClient()
-            .POST("/api/2.0/clusters/create", createCluster, CreateClusterResponse.class, emptyHeaders);
+            .POST(
+                "/api/2.0/clusters/create",
+                createCluster,
+                CreateClusterResponse.class,
+                emptyHeaders);
 
     return response.getClusterId();
   }
@@ -258,7 +262,9 @@ public class DatabricksUtils {
 
   @SneakyThrows
   private static List<RunEvent> fetchEventsEmitted(WorkspaceClient workspace) {
-    return workspace.dbfs().readAllLines(Paths.get(DBFS_EVENTS_FILE), StandardCharsets.UTF_8)
+    return workspace
+        .dbfs()
+        .readAllLines(Paths.get(DBFS_EVENTS_FILE), StandardCharsets.UTF_8)
         .stream()
         .map(event -> OpenLineageClientUtils.runEventFromJson(event))
         .collect(Collectors.toList());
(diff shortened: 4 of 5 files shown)


integration/spark/shared/build.gradle
```diff
@@ -47,7 +47,7 @@ dependencies {
     compileOnly("org.apache.spark:spark-sql-kafka-0-10_${scalaBinaryVersion}:${sparkVersion}")
     compileOnly("org.apache.spark:spark-sql_${scalaBinaryVersion}:${sparkVersion}")
 
-    compileOnly("com.databricks:databricks-dbutils-scala_${scalaBinaryVersion}:${databricksVersion}") {
+    implementation("com.databricks:databricks-dbutils-scala_${scalaBinaryVersion}:${databricksVersion}") {
```
pawel-big-lebowski, line 50: ohh, I thought we don't want to add this dependency. We just want to make use its classes if it is on the classpath. Otherwise we end up accumulating all the dependencies from all the spark runtimes, extensions and libs.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
```diff
@@ -110,7 +110,7 @@ private Map<String, Object> getDatabricksEnvironmentalAttributes(SparkListenerJo
   private static Optional<DbfsUtils> getDbfsUtils()
       throws ClassNotFoundException, InstantiationException, IllegalAccessException,
           IllegalArgumentException, InvocationTargetException {
-    Class dbutilsClass = Class.forName("com.databricks.dbutils_v1.impl.DbfsUtilsImpl");
+    Class dbutilsClass = Class.forName("com.databricks.sdk.scala.dbutils.SdkDBUtilsImpl");
```
pawel-big-lebowski, line 113: What it looks to me, we should remove `throws ClassNotFoundException` from method signature, catch the exception and return `Optional.empty` if so. Moreover, it may have sense to support both class names and choose the one that is present.

pawel-big-lebowski (commented):
Thanks @algorithmy1 for investigating and preparing a solution for this. Please mind my comments and tell what do you think about this. Super happy to see your contribution again 💯

integration/spark/shared/build.gradle
algorithmy1: I think we should not include any extra packages in the shadow jar, this way we don't end up with a really fat jar 🤣
For this, I suggest that we roll back this imports in the [commit](https://github.com/OpenLineage/OpenLineage/commit/89beaadb9a608244e5375d0f76a124f78c5167d8#diff-5542f4a02d45d03e2fdc704d8bf3aab869f919d72a4785a077c52acf86c3aa6a) since `com.databricks.dbutils_v1` that has `com.databricks.dbutils_v1.DbfsUtils` already exists on the `classpath`.
If we don't do so people will be struggling installing OpenLineage in Databricks.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
algorithmy1: It's good idea to remove the `throws ClassNotFoundException` but we should log some warning that the Databricks Facet is not built because the package is not present in the classpath.
For me, I think we can keep `com.databricks.dbutils_v1.impl.DbfsUtilsImpl` which already exists by default and simply remove `com.databricks.sdk.scala.dbutils.SdkDBUtilsImpl`.

Tell me that do you think.

d-m-h:
Yes, it's as I remember. The issue with `com.databricks:dbutils-api_$scalaBinVersion:$dbutilsVersion` is that it lacks a variant for Scala 2.13. If _any_ of the methods that we use from those classes expect or return Scala Sequences we will run into errors when running in a Scala 2.13 environment.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
d-m-h: I have one question, would the warning be logged in non-Databricks environments? If the answer is yes, then it doesn't make sense to log the warning.

algorithmy1:
> Yes, it's as I remember. The issue with `com.databricks:dbutils-api_$scalaBinVersion:$dbutilsVersion` is that it lacks a variant for Scala 2.13. If _any_ of the methods that we use from those classes expect or return Scala Sequences we will run into errors when running in a Scala 2.13 environment.

Should we then keep `compileOnly` instead of `implementation` for `com.databricks:databricks-dbutils-scala_${scalaBinaryVersion}:${databricksVersion}`. The final user will be responsible for adding this package alongside with OpenLineage's one. And finally, we remove `throws ClassNotFoundException` and replace it with some warning message in order to keep the ETL safe from OpenLineage related failures.

What do you think ?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/DatabricksEnvironmentFacetBuilder.java
algorithmy1: Something similar ti what we see [here](https://github.com/OpenLineage/OpenLineage/blob/c82c208f886ab57e0299f904d665bac74b413856/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java#L81) It goes in the Log4j output alongside with spark's logs.

algorithmy1:
> > Yes, it's as I remember. The issue with `com.databricks:dbutils-api_$scalaBinVersion:$dbutilsVersion` is that it lacks a variant for Scala 2.13. If _any_ of the methods that we use from those classes expect or return Scala Sequences we will run into errors when running in a Scala 2.13 environment.
> 
> Should we then keep `compileOnly` instead of `implementation` for `com.databricks:databricks-dbutils-scala_${scalaBinaryVersion}:${databricksVersion}`. The final user will be responsible for adding this package alongside with OpenLineage's one. And finally, we remove `throws ClassNotFoundException` and replace it with some warning message in order to keep the ETL safe from OpenLineage related failures.
> 
> What do you think ?

If I'm not mistaken, Databricks only supports clusters that run on Scala 2.12.

image

d-m-h:
The biggest concern then becomes that when the Scala 2.13 variant is being built, the databricks things need to be excluded / not considered, otherwise the build will fail.

mobuchowski:
@d-m-h @algorithmy1 
It seems to me that the actual solution here would be
- keep `compileOnly` in the gradle definition
- provide two versions of `DatabricksEnvironmentFacetBuilder`, one that uses `com.databricks.sdk.scala` approach, and one that uses `com.databricks.dbutils_v1` approach
- gate them behind existence of the particular classes at runtime, using `ReflectionUtils.hasClasses`

what do you think?

d-m-h:
> @d-m-h @algorithmy1 It seems to me that the actual solution here would be
> 
> * keep `compileOnly` in the gradle definition
> * provide two versions of `DatabricksEnvironmentFacetBuilder`, one that uses `com.databricks.sdk.scala` approach, and one that uses `com.databricks.dbutils_v1` approach
> * gate them behind existence of the particular classes at runtime, using `ReflectionUtils.hasClasses`
> 
> what do you think?

It _could_ work. It's not 100% full-proof, as we can't prevent users from using the 2.13 variant with the 2.12 runtime. Overall, I am happy to with that approach.

mobuchowski:
>It could work. It's not 100% full-proof, as we can't prevent users from using the 2.13 variant with the 2.12 runtime.

Ah, the check could also check that - if we think it's necessary.

pawel-big-lebowski:
Hi everyone, I think that this one https://github.com/OpenLineage/OpenLineage/pull/2537 should make it work.

mobuchowski:
I will close that as #2537 should fix it - please reopen if that's not true or you disagree with me closing it in any way.

PR #2538, opened 2024-03-27
spark: add missing relocate snakeyaml incantation

mobuchowski:


pawel-big-lebowski (approved):
Looks like this one -> https://github.com/OpenLineage/OpenLineage/blob/9079c15c73f169370cdec2090d52873fa845746e/integration/spark/app/build.gradle#L237 

is no longer valid. Please verify that other jars are relocated.

mobuchowski:
@pawel-big-lebowski removed no-op relocates from `app/build.gradle`. Now, the relocation is being done only at proper, top level jar.

PR #2546, opened 2024-03-29
[SQL] Fix parsing JDBCOptions(table=...) containing subquery

dolfinus:
### Problem

👋 Thanks for opening a pull request! Please include a brief summary of the problem your change is trying to solve, or bug fix. If your change fixes a bug or you'd like to provide context on why you're making the change, please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) as follows:

In the following case:
```python
df = spark.read.jdbc(table="(select * from table where filter='value') T")
```

`openlineage-spark` returns datasets with names like `database.(select * from table where filter='value') T`, not `database.table`. This is because `JDBCOptions.table` is expected to always return a table name, not subquery.

### Solution

If `JDBCOptions.table` contains a subquery, fallback to branch with SQL parsing to extract table names from this query.

I've also added a test that `openlineage-spark` can parse query of unknown JDBC dialect.

#### One-line summary:

Fix `openlineage-spark` producing datasets with names like `database.(select * from table)` for JDBC sources.

### Checklist

----

### DIFF
--- a/integration/spark/shared/src/test/java/io/openlineage/spark/agent/lifecycle/plan/JdbcRelationHandlerTest.java
+++ b/integration/spark/shared/src/test/java/io/openlineage/spark/agent/lifecycle/plan/JdbcRelationHandlerTest.java
@@ -45,6 +45,7 @@ public class JdbcRelationHandlerTest {
   String invalidJdbc = "(test) SPARK_GEN_SUBQ_0";
   String url = "postgresql://localhost:5432/test";
   String mysqlUrl = "mysql://localhost:3306/test";
+  String unknownUrl = "unknown://localhost:1234/test";
   StructType schema =
       new StructType().add("k", DataTypes.IntegerType).add("j", DataTypes.StringType);
 
@@ -91,6 +92,12 @@ public class JdbcRelationHandlerTest {
 
   @Test
   void testHandlingJdbcDbTableAsSubQuery() {
+    CaseInsensitiveMap params =
+        CaseInsensitiveMap$.MODULE$.apply(
+            ScalaConversionUtils.fromJavaMap(
+                Collections.singletonMap(
+                    JDBCOptions$.MODULE$.JDBC_TABLE_NAME(), jdbcDbTableAsSubQuery)));
+    when(jdbcOptions.parameters()).thenReturn(params);
     when(jdbcOptions.tableOrQuery()).thenReturn(jdbcDbTableAsSubQuery);
     StructType schema1 =
         new StructType().add("k", DataTypes.IntegerType).add("j1", DataTypes.StringType);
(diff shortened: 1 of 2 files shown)


integration/spark/shared/src/test/java/io/openlineage/spark/agent/lifecycle/plan/JdbcRelationHandlerTest.java
```diff
@@ -91,6 +92,12 @@ void testHandlingJdbcTable() {
 
   @Test
   void testHandlingJdbcDbTableAsSubQuery() {
```
pawel-big-lebowski, line 94: Would it make sense to create a test `testHandlingJdbcDbTableAsSubQueryWithUnknownDialect`? I am not following how `testUnknownDialect` verifies a behaviour described within issue description.

dolfinus: `testUnknownDialect` is just checking that if here `dialect` is `null`, query still can be parsed using generic dialect:
https://github.com/OpenLineage/OpenLineage/blob/62276dc99da5f51a112b9877c9fd8fe545db8fcc/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/JdbcSparkUtils.java#L105

dolfinus: Redarding `testHandlingJdbcDbTableAsSubQueryWithUnknownDialect` test, I don't think it will make much sense. Both cases `table is not set, query=SELECT ...` and `table=(SELECT...)` are now handler in the same way, and covered by tests above

pawel-big-lebowski: Common practice about fixing issues and unit tests is to: (1) write a failing unit test and (2) fix a code to make a failing test pass. 

Issue description is 
```
openlineage-spark returns datasets with names like database.(select * from table where filter='value') T, not database.table
```
Is there any test verifying this? 
How does the introduced test `testUnknownDialect` relate to issue description and code that has been modified?

dolfinus: > (1) write a failing unit test
> (2) fix a code to make a failing test pass

I've updated `testHandlingJdbcDbTableAsSubQuery` test to cover this. Previously it covered just the same case as `testHandlingJdbcQuery`

pawel-big-lebowski (approved):

PR #2564, opened 2024-04-03
[COMMON] feat: Add support for SCRIPT type jobs in BigQuery

kacpermuda:
### Problem

When using SCRIPT type jobs in BigQuery, no lineage is extracted, because SCRIPT job has no lineage information - it only spawns child jobs that have that information.

### Solution

Extract lineage information from child jobs when dealing with SCRIPT type job. 

I removed query string from `BigQueryJobRunFacet` - it can increase event size a lot and it's already included in SqlJobFacet so it's not necessary here. 

I also added deduplication of input and output datasets to avoid duplicates in case the script job writes to / read from a table multiple times.

### Checklist

----

### DIFF
--- a/integration/common/openlineage/common/provider/bigquery.py
+++ b/integration/common/openlineage/common/provider/bigquery.py
@@ -1,9 +1,11 @@
 # Copyright 2018-2024 contributors to the OpenLineage project
 # SPDX-License-Identifier: Apache-2.0
 
+import copy
 import json
 import logging
 import traceback
+import warnings
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 
 import attr
@@ -89,7 +91,21 @@ class BigQueryStatisticsDatasetFacet(BaseFacet):
 class BigQueryFacets:
     run_facets: Dict[str, BaseFacet] = attr.ib()
     inputs: List[Dataset] = attr.ib()
-    output: Optional[Dataset] = attr.ib(default=None)
+    outputs: List[Dataset] = attr.ib()
+    _output: Optional[Dataset] = attr.ib(default=None)
+
+    @property
+    def output(self) -> Optional[Dataset]:
+        warnings.warn(
+            "The 'output' attribute will be removed in a future version. Use 'outputs' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._output
+
+    @output.setter
+    def output(self, value: Optional[Dataset]) -> None:
+        self._output = value
 
 
 class BigQueryDatasetsProvider:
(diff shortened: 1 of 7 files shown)


integration/common/openlineage/common/provider/bigquery.py
```diff
@@ -1,9 +1,10 @@
 # Copyright 2018-2024 contributors to the OpenLineage project
 # SPDX-License-Identifier: Apache-2.0
-
+import copy
```
JDarDagran, line 3:
```suggestion

import copy
```

integration/common/openlineage/common/provider/bigquery.py
```diff
@@ -141,40 +162,70 @@ def get_facets(self, job_id: str) -> BigQueryFacets:
...
-            raise ValueError(f"BigQuery properties did not have required data: queryPlan - {json_props}")
-
-        out_stage = stages[-1]
-        out_rows = out_stage.get("recordsWritten", None)
-        out_bytes = out_stage.get("shuffleOutputBytes", None)
+        return BigQueryFacets(
+            run_facets=run_facets,
+            inputs=list({i.name: i for i in inputs if i}.values()),
+            outputs=self._deduplicate_outputs(outputs),
+            output=outputs[0] if len(outputs) == 1 else None,
```
JDarDagran, line 169: Couldn't this default to first element if exists when integration uses older `output` attribute because of no update there?

JDarDagran (commented):
I love you've made it backwards compatible with notice of deprecation :rocket:
Maybe it's time we should really get rid of this dependency and keep logic only in Airflow integration? Same goes for `RedshiftDataDatasetsProvider`. It's of course not the subject of this PR, just raising this. cc @mobuchowski 

Would it be also possible to modify existing BQ integration test?

integration/common/openlineage/common/provider/bigquery.py
kacpermuda: I think that we should somehow indicate to people that there are more than one output. If there are multiple outputs and somebody is using `output` - they will still receive something and may not be aware that there are more. My idea was - if they see no output at all, they will look into it, probably discover `outputs` and update the code. I'm open to changes, no hard arguments here.

JDarDagran: I don't really like the idea of limiting users' experience just to force them to use newer libraries. It wouldn't be the case if the change was only for `SCRIPT` type jobs which don't work so far. This on the other hand breaks output on purpose for any other job.

kacpermuda: I'm not sure i understand. At the moment, you can only receive a single output dataset from a single BQ job (one job == one statement). After my change in this PR, for all jobs other than `SCRIPT` you will still receive a single output dataset (as a dataset under `output` and as a list with one dataset under `outputs`). For `SCRIPT` however, if there are multiple outputs, you have to use `outputs` to get all of them and the output will be empty (just like right now, the script job returns no input/outputs). So as far as i understand, the behaviour won't change (apart from `output` being deprecated).

JDarDagran: Sorry, you're right!
I don't have strong opinion though now. Maybe let's not leave first output for `SCRIPT` jobs as this may lead to some strange behaviours and questioned results. Could you please leave some comment at this place to explain the logic?

kacpermuda: Sure. Multiple outputs are there for more complex scripts that create or insert data into multiple tables (multiple INSERT INTO, CREATE statements etc.) - then we will receive multiple output datasets and we need to use `outputs` because the `output` will be empty. However, the script can also be a single statement (f.e. CREATE TABLE AS) with some variables declared at the top (f.e. DECLARE var1 ...) - in this case, we will receive a single output dataset. Even though the statement type is `SCRIPT` we can treat it as a regular statement and also make this single output available under `output`. This way, even if somebody doesn't know that they are using a `SCRIPT` job, because they are simply running a single statement with variables, will still be able to use `output` until removed from deprecation. I think it should not lead to any unexpected behaviours, but maybe i am missing something.

kacpermuda: Added a [comment](https://github.com/OpenLineage/OpenLineage/pull/2564/files#diff-3be24ccd3c1a25a965537f442716bdf033abcd9346056db006b1922466f81487R167-R170) in the file as well, let me know if it requires some changes.

JDarDagran: Looks great, thanks!

JDarDagran (approved):

PR #2572, opened 2024-04-04
[CLIENT] Fix missing pkg_resources module on Python 3.12

dolfinus:
### Problem

After installing `openlineage-python` on Python 3.12 virtual environment, I try to import it:
```python
from openlineage.client import OpenLineageClient
```

I got an exception:
```
ImportError occurred when trying to import numpy module.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/maxim/Repo/airflow-cd/venv/lib/python3.12/site-packages/openlineage/client/__init__.py", line 5, in <module>
    from openlineage.client.client import OpenLineageClient, OpenLineageClientOptions
  File "/home/maxim/Repo/airflow-cd/venv/lib/python3.12/site-packages/openlineage/client/client.py", line 19, in <module>
    from openlineage.client.transport import Transport, TransportFactory, get_default_factory
  File "/home/maxim/Repo/airflow-cd/venv/lib/python3.12/site-packages/openlineage/client/transport/__init__.py", line 9, in <module>
    from openlineage.client.transport.kafka import KafkaConfig, KafkaTransport
  File "/home/maxim/Repo/airflow-cd/venv/lib/python3.12/site-packages/openlineage/client/transport/kafka.py", line 12, in <module>
    from pkg_resources import parse_version
ModuleNotFoundError: No module named 'pkg_resources'
```

This is because `pkg_resources` is not a part of stdlib, but actually provided by `setuptools`. And setuptools is not automatically installed to virtual environments since Python 3.12.

For example, Python 3.11:
```bash
$ python -V
Python 3.11.8
$ python -m venv venv
$ source venv/bin/activate
$ pip freeze --all
pip==24.0
setuptools==65.5.0
```

Python 3.12:
```bash
$ python -V
Python 3.12.2
$ python -m venv venv
$ source venv/bin/activate
$ pip freeze --all
pip==24.0
```

#### One-line summary:

Remove `pkg_resources` dependency and replace it with [packaging](https://packaging.pypa.io/en/latest/version.html) lib.

### Checklist

----

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
   "python-dateutil>=2.8.2",
   "pyyaml>=5.4",
   "requests>=2.20.0",
+  "packaging>=21.0",
 ]
 optional-dependencies.kafka = [
   "confluent-kafka>=2.1.1",
--- a/integration/airflow/openlineage/airflow/__init__.py
+++ b/integration/airflow/openlineage/airflow/__init__.py
@@ -6,13 +6,13 @@
 import logging
 
 from openlineage.airflow.version import __version__
-from pkg_resources import parse_version
+from packaging.version import Version
 
 from airflow.version import version as AIRFLOW_VERSION
 
 __author__ = """OpenLineage"""
 
-if parse_version(AIRFLOW_VERSION) < parse_version("2.1.0"):  # type: ignore
+if Version(AIRFLOW_VERSION) < Version("2.1.0"):  # type: ignore
     logging.warning(
         f"""
         OpenLineage support for Airflow version {AIRFLOW_VERSION} is REMOVED.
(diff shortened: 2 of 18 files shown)


integration/airflow/openlineage/airflow/__init__.py
```diff
@@ -6,21 +6,21 @@
 import logging
 
 from openlineage.airflow.version import __version__
-from pkg_resources import parse_version
+from packaging.version import parse as Version
```
kacpermuda, line 9:
```suggestion
from packaging.version import Version
```

client/python/pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
   "python-dateutil>=2.8.2",
   "pyyaml>=5.4",
   "requests>=2.20.0",
+  "packaging",
```
kacpermuda, line 32: Not sure which one, but we probably shouldn't leave it without some min version ?

kacpermuda (commented):
LGTM

client/python/pyproject.toml
dolfinus: I don't know either, `Version` class is present in this package since v14.0.0 which was a first released version

kacpermuda: @JDarDagran any thoughts ?

dolfinus: Why there should be a minimal version set in the first place?

mobuchowski: I would rather pin it to some higher version - less chance that something breaks if we use it in some other way and someone else decides to use very low version.

mobuchowski (approved):

PR #2575, opened 2024-04-04
release, quickfix: make spark release job use scala interfaces build in their own release job

mobuchowski:
It's a quickfix for release failures.

After that, a refactor of CI is needed to get rid of all dependencies for release jobs: 
- they should reuse artifact generated by build job
- they should not depend on any other job but a build job for a given artifact


merobi-hub (approved):
LG, thank you

JDarDagran (approved):

PR #2583, opened 2024-04-05
[SPARK][FLINK][JAVA] Support yaml config files together with SparkConf/FlinkConf

pawel-big-lebowski:
### Problem

Client-java documentation says it can read configs directly from `openlineage.yaml` file, while Spark integration reads config entries from SparkConf only. Flink integration can read it from file and Flink conf exclusively. This is confusing for the user.

Problems: 
 * There is no way to add flink specific config to Flink integration.
 * Some config entries in Flink conf will not work if passed through yaml file.
 * Spark integration does not support reading conf from yaml file at all.

### Solution

 * rename `OpenlineageYaml` -> `OpenlineageConfig`
 * create `SparkOpenlineageConfig`, `FlinkOpenlineageConfig` extending the above to keep entries specific to integration
 * modify code to use only `OpenLineageConfig` classes. 
 * update doc to mention that both ways can be used interchangeably and final documentation will be a merge of all values provided. 

### Checklist

----


client/java/src/main/java/io/openlineage/client/transports/HttpConfig.java
```diff
@@ -28,7 +29,29 @@ public final class HttpConfig implements TransportConfig {
       timeout; // deprecated, will be removed in 1.13, assumes timeout is in seconds
 
   @Getter @Setter private @Nullable Integer timeoutInMillis;
-  @Getter @Setter private @Nullable TokenProvider auth;
-  @Getter @Setter private @Nullable Map<String, String> urlParams;
-  @Getter @Setter private @Nullable Map<String, String> headers;
+
+  @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
```
pawel-big-lebowski, line 33: added to prevent serialising the field within debug facet

mobuchowski: 👍

client/java/src/main/java/io/openlineage/client/OverwriteConfig.java
```diff
@@ -0,0 +1,75 @@
...
+package io.openlineage.client;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Contains methods which allows overwriting values of a config object with values from another
+ * object
+ */
+public interface OverwriteConfig<T> {
```
mobuchowski, line 15: I see you've added this mechanism, but what's the actual purpose of it? Can't we just create new config instance rather than making config mutable if we need to merge config entries from multiple sources?

pawel-big-lebowski: I think we can go with the approach you describe: change the interface to return a new value instead
will prepare a change for this.

d-m-h (commented):
I'm on a phone right now, so it's a bit difficult to see the changes, so forgive me if I ask something that is obvious.

When converting from flat properties to a config object, how is this done?

To we try and infer a hierarchy from using dot separated properties? For example, if we do `spark.openlineage.transport.foo.bar`, do we assume that the object has the hierarchy of:

```yaml
transport:
\tfoo:
\t\tbar: $value
```

If so, this could pose a problem in my use case, as I have letting the custom transport act as a pass through for the values the spark integration provides. For example, I have the following property:

`spark.openlineage.transport.properties.cluster.federation`

The `cluster.federation` is a property that the library my transport wraps expects. So if this code tries to infer hierarchy, this would break my integration.

pawel-big-lebowski:
@d-m-h Actually this is already done within this piece of code -> https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java#L88

I don't think this PR changes this behaviour anyway.

mobuchowski (approved):
Amazing job @pawel-big-lebowski. That change was very much needed. I'm approving but please wait with merging till we hear if @d-m-h's concerns are answered. 

One thing you can improve in the meantime is improving comments on MergeConfig class - I believe we can explain the semantics better and in simple language.

dolfinus:
I see that changelog entry was added, but for 1.12.0 instead of Unreleased. Fix: #2636

pawel-big-lebowski:
@dolfinus thank you, sry

PR #2584, opened 2024-04-05
spark: use hashset in column level lineage instead of iterating through linkedlist

mobuchowski:
Original implementation used `LinkedList` which iteration performance is terrible. There's no real need to use list in that place - `HashSet` will be more performant collection.


harels (approved):
lgtm

Issue #2595, opened 2024-04-10
[PROPOSAL] Using gzip for HTTP request body


dolfinus:
**Purpose:**

Currently HTTP request body is just plain JSON values, send to API without any compression. For example, this is one of events send by Spark integration: [event.json](https://github.com/OpenLineage/OpenLineage/files/14934779/event.json)

This is 3.8Kb of raw data. But using gzip, it can be compressed to just 861 bytes [event.json.gz](https://github.com/OpenLineage/OpenLineage/files/14934778/event.json.gz)

**Proposed implementation**

For Java client, wrap `HTTPEntity` with [GzipCompressingEntity](https://www.javadoc.io/doc/org.apache.httpcomponents/httpclient/4.5.2/org/apache/http/client/entity/GzipCompressingEntity.html) to automatically compress body to gzip while sending a request.
For Python client this is as simple as using [builtin zlib module](https://stackoverflow.com/a/38866506/23601543)

To avoid breaking backward compatibility, this can be enabled using HTTPTransport configuration, like:
```yml
transport:
  type: http
  url: http://localhost:5000
  compression: gzip  # deflate, null, etc
```

mobuchowski:
In past jobs I've also seen https://github.com/facebook/zstd used very successfully for similar tasks. Advantage for zlib/gzip is just that it's buildin for Java and Python.

I think adding compression as opt-in mechanism is definitely a good idea - especially we see really some repeated data in the events.

dolfinus:
Just tried zstd, with default compression level 3 it produces file with size 894 bytes, with level compression level 19 it is 860 bytes. So not a much difference, although I've tested file with quite a small file (Airflow task events contains much more data).

mobuchowski:
Yeah, might be not worth the effort for small jsons.

dolfinus:
Here are Python #2604 and Java implementations #2603. Checked on Marquez, it can handle requests with gzipped content.

I haven't added compression to Proxy because it is implemented in third-party package [Jersey](https://javadoc.io/static/org.glassfish.jersey.bundles/jaxrs-ri/2.35/org/glassfish/jersey/message/GZipEncoder.html).

dolfinus:
Documentation: https://github.com/OpenLineage/docs/pull/318

PR #2603, opened 2024-04-12
[Java] Add GZIP compression to HttpTransport

dolfinus:
### Problem

Implements: #2595

#### One-line summary:

Add `compression` option to `HttpTransport` config of Java client, with `gzip` implementation.

### Checklist

----

### DIFF
--- a/client/java/src/test/java/io/openlineage/client/transports/HttpTransportTest.java
+++ b/client/java/src/test/java/io/openlineage/client/transports/HttpTransportTest.java
@@ -25,6 +25,8 @@ import static org.mockito.Mockito.when;
 
 import io.openlineage.client.OpenLineageClient;
 import io.openlineage.client.OpenLineageClientException;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.URI;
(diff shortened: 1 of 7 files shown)


client/java/src/test/java/io/openlineage/client/transports/HttpTransportTest.java
```diff
@@ -269,6 +275,52 @@ void customHeaders() throws IOException {
...
+    client.emit(runEvent());
+
+    HttpEntity resultEntity = map.get("test").getEntity();
+    assertThat(resultEntity).isInstanceOf(GzipCompressingEntity.class);
+
+    GzipCompressingEntity gzippedEntity = (GzipCompressingEntity) resultEntity;
+    ByteArrayOutputStream compressedData = new ByteArrayOutputStream();
+    gzippedEntity.writeTo(compressedData);
+
+    GZIPInputStream uncompressedStream =
```
pawel-big-lebowski, line 308: I like this test 👍

pawel-big-lebowski (approved):

pawel-big-lebowski:
@dolfinus would you be so kind to modify changelog file? You can within this PR some entry about #2598 as well.

dolfinus:
> would you be so kind to modify changelog file?

I thought that changelog entries are generated using commit messages or PR titles, just before the release.

client/java/src/test/java/io/openlineage/client/transports/HttpTransportTest.java
dolfinus: Well, I'm not really. The only thing is tested here is ApacheHTTP client classes

rahul-madaan:
I have a few questions, 

1. backward compatibility - Do existing OL consumers need to update anything on their backend to accept compressed events?
2. will the events be compressed by default even if we don't set `compression: gzip` in transport?

kacpermuda:
@rahul-madaan If i understand correctly, both in Python and Java this was added as an optional feature and will not change the default behaviour - events will still be sent uncompressed by default, so there is no problem with being backward compatible. I would, however, recommend the OL consumers to be able to accept compressed events as it will probably be included in the next release.

mobuchowski (approved):

PR #2633, opened 2024-04-22
flink: add metrics support

mobuchowski:
This adds support for https://github.com/OpenLineage/OpenLineage/pull/2496 metrics for Flink integration.

### DIFF
--- a/integration/flink/shared/src/main/java/io/openlineage/flink/client/Versions.java
+++ b/integration/flink/shared/src/main/java/io/openlineage/flink/client/Versions.java
@@ -1,0 +1,34 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.flink.client;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.util.Properties;
+
+public class Versions {
+
+  public static final URI OPEN_LINEAGE_PRODUCER_URI = getProducerUri();
+
+  private static URI getProducerUri() {
+    return URI.create(
+        String.format(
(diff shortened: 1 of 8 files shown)


integration/flink/shared/src/main/java/io/openlineage/flink/client/Versions.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public static final URI OPEN_LINEAGE_PRODUCER_URI = getProducerUri();
+
+  private static URI getProducerUri() {
+    return URI.create(
+        String.format(
+            "https://github.com/OpenLineage/OpenLineage/tree/%s/integration/flink", getVersion()));
+  }
+
+  @SuppressWarnings("PMD")
```
pawel-big-lebowski, line 23: Please mention particular warning you want to suppress.

pawel-big-lebowski (commented):
Please provide description of this PR of what it adds and specify what metrics will be produced. This super helpful when trying to figure what was changed in the past by a PR.  

Documentation PR is important as well:
 - There is metrics section in Java client, however it is not linked from Spark integration. I think it should be included in both Spark and Flink integrations (include md partial)
 - Additionally, I think it's important to provide info specific to Spark and Flink. As a Spark/Flink user I would like to know which metrics are emitted just to know if this is going to solve the problem I have or not.

integration/flink/shared/src/main/java/io/openlineage/flink/client/Versions.java
mobuchowski: @pawel-big-lebowski to be honest, I've copied this fragment from Spark integration. Should I change this there as well?

pawel-big-lebowski (approved):

PR #2646, opened 2024-04-25
spark, glue: add Glue symlink if reading from Glue catalog table 

mobuchowski:
If we're using Glue catalog table, the dataset symlink should point to Glue catalog table name. 

For now, the Glue dataset naming is not standarized - ideally this PR would start a discussion around that naming.


integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/PathUtilsTest.java
```diff
@@ -162,10 +162,34 @@ void testFromCatalogTableWithStorage() throws URISyntaxException {
...
+    when(catalogTable.identifier()).thenReturn(TableIdentifier.apply(TABLE));
+    when(catalogStorageFormat.locationUri())
+        .thenReturn(Option.apply(new URI("s3://bucket/directory")));
+
+    DatasetIdentifier di = PathUtils.fromCatalogTable(catalogTable, Optional.of(sparkConf));
+    assertThat(di.getName()).isEqualTo("directory");
+    assertThat(di.getNamespace()).isEqualTo("s3://bucket");
+    assertThat(di.getSymlinks()).hasSize(1);
+    assertThat(di.getSymlinks().get(0).getName()).isEqualTo(TABLE);
+    assertThat(di.getSymlinks().get(0).getNamespace()).isEqualTo("glue");
```
pawel-big-lebowski, line 185: The notion of symlink is used to provide an information that two datasets are the same if any of the symlinks are equal. That's why symlinks should provide the same level of uniqueness like primary dataset names. Does this still hold for the solution proposed here? 

Is it possible to have two different glues (projects, namespaces, whatever naming they use) and table with the same name in each of them? If it is possible, we end up with different datasets being marked as the same dataset.

kacpermuda: The ARN(Amazon Resource Name) for Glue is documented [here](https://docs.aws.amazon.com/glue/latest/dg/glue-specifying-resource-arns.html#data-catalog-resource-arns), according to this [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html) the `Amazon Resource Names (ARNs) uniquely identify AWS resources`.

So, probably, for namespace we could go for `aws:glue:<region>:<account-id>` or something similar?

mobuchowski: Using ARNs for purely-AWS resources would probably be a good idea.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -75,14 +74,27 @@ public static DatasetIdentifier fromCatalogTable(
...
           symlink.getName(), symlink.getNamespace(), DatasetIdentifier.SymlinkType.TABLE);
+    } else if (catalogTable.provider().isDefined()
+        && sparkConf.isPresent()
+        && "hive".equals(catalogTable.provider().get())
+        && "com.amazonaws.glue.catalog.metastore.AWSGlueDataCatalogHiveClientFactory"
+            .equals(sparkConf.get().get("spark.hadoop.hive.metastore.client.factory.class", ""))) {
+      String region = System.getenv().get("AWS_DEFAULT_REGION");
+      if (region == null || region.isEmpty()) {
+        region = System.getenv("AWS_REGION");
+      }
```
kacpermuda, line 91: Can `region` still can be null in the end or the AWS_REGION is always set?

mobuchowski: @kacpermuda I've seen that `AWS_DEFAULT_REGION` is supported in all currently supported Glue versions, the `AWS_REGION` seems to be a good fallback given that there is confusion in AWS whether to use one var or another.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -75,14 +74,27 @@ public static DatasetIdentifier fromCatalogTable(
...
     Optional<URI> metastoreUri = extractMetastoreUri(sparkConf);
     // TODO: Is the call to "metastoreUri.get()" really needed?
-    //   Java's Optional should prevent the null in the first place.
+    // Java's Optional should prevent the null in the first place.
     if (metastoreUri.isPresent() && metastoreUri.get() != null) {
       // dealing with Hive tables
       DatasetIdentifier symlink = prepareHiveDatasetIdentifier(catalogTable, metastoreUri.get());
       return di.withSymlink(
           symlink.getName(), symlink.getNamespace(), DatasetIdentifier.SymlinkType.TABLE);
+    } else if (catalogTable.provider().isDefined()
```
pawel-big-lebowski, line 83: Minor: consider extracting this to separate method like `withAwsGlueSymlink` in case more of such conditions are added in future.

integration/spark/shared/build.gradle
```diff
@@ -62,6 +62,7 @@ dependencies {
     testFixturesApi(platform("org.junit:junit-bom:${junit5Version}"))
     testFixturesApi("org.junit.jupiter:junit-jupiter")
     testFixturesApi("org.junit.jupiter:junit-jupiter-params")
+    testFixturesApi("org.junit-pioneer:junit-pioneer:1.9.1")
```
pawel-big-lebowski, line 65: is there anything in tests that uses it?

pawel-big-lebowski (approved):
Happy to onboard this feature. 

I am hoping to have an integration test for Glue in future. 
I find code pieces like:
```
 "com.amazonaws.glue.catalog.metastore.AWSGlueDataCatalogHiveClientFactory"
            .equals(sparkConf.get().get("spark.hadoop.hive.metastore.client.factory.class", ""))
```
hard to maintain and likely to break in the future.

integration/spark/shared/build.gradle
mobuchowski:
```
  @SetEnvironmentVariable(key = "AWS_DEFAULT_REGION", value = "us-west-2")
```

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
mobuchowski: This code generally calls for greater refactor :)

PR #2650, opened 2024-04-30
Add log info when emitting lineage from spark

algorithmy1:
## Overview
This pull request introduces enhancements to logging within the Spark integration of OpenLineage. The goal is to enable clear and accessible logs that indicate whether an event has been successfully sent to the backend API.

## Importance
Accurate logging is crucial for monitoring and verifying the reliability of event dispatching.

## Changes Proposed
- Implement additional logging statements that confirm the dispatch of events.
- These logs will provide immediate feedback in the Spark execution logs, aiding in quick diagnostics and status checks by administrators and developers.

Your feedback and suggestions on this enhancement are welcome.

### Checklist

----

### DIFF
--- a/client/java/gradle.properties
+++ b/client/java/gradle.properties
@@ -1,1 +1,1 @@
-version=1.10.2
+version=1.11.0-SNAPSHOT
--- a/client/python/openlineage/client/constants.py
+++ b/client/python/openlineage/client/constants.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
-__version__ = "1.10.2"
+__version__ = "1.11.0"
 
 DEFAULT_TIMEOUT_MS = 5000
 DEFAULT_NAMESPACE_NAME = "default"
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -6,7 +6,7 @@ requires = [
 
 [project]
 name = "openlineage-python"
-version = "1.10.2"
+version = "1.11.0"
 description = "OpenLineage Python Client"
 readme = "README.md"
 keywords = [
--- a/client/python/setup.cfg
+++ b/client/python/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.10.2
+current_version = 1.11.0
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?P<rc>.*)
--- a/integration/airflow/openlineage/airflow/version.py
+++ b/integration/airflow/openlineage/airflow/version.py
@@ -1,4 +1,4 @@
 # Copyright 2018-2024 contributors to the OpenLineage project
 # SPDX-License-Identifier: Apache-2.0
 
-__version__ = "1.10.2"
+__version__ = "1.11.0"
--- a/integration/airflow/setup.cfg
+++ b/integration/airflow/setup.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.10.2
+current_version = 1.11.0
 commit = False
 tag = False
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?P<rc>.*)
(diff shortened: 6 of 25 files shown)


integration/spark/app/src/main/java/io/openlineage/spark/agent/EventEmitter.java
```diff
@@ -60,6 +60,8 @@ public void emit(OpenLineage.RunEvent event) {
       this.client.emit(event);
       log.debug(
           "Emitting lineage completed successfully: {}", OpenLineageClientUtils.toJson(event));
+      log.info(
+          "Emitting lineage completed successfully with run id: {}", event.getRun().getRunId());
```
d-m-h, line 64: Hi @algorithmy1. I don't understand the purpose of this change. You're almost duplicating the log messages from the line above it. If you want to change the log content of the already emitted log message, I think that there are better ways of achieving this. 

Is there any particular reason why you want only the run id to be emitted?

algorithmy1: Hi @d-m-h, thank you for your feedback. Let me address your concerns:

- Distinctiveness from Existing Logs: The addition of this log line, which records only the run ID upon successful event dispatch, does not duplicate but rather complements existing logs. This specific line is crucial for quickly verifying event dispatch without sifting through more verbose log details.
- Necessity in Complex Systems: In large organizations, it can be challenging to confirm whether OpenLineage has been correctly deployed across all clusters. This logging change simplifies identifying which clusters are properly integrated with OpenLineage, ensuring comprehensive lineage tracking.
- Event Tracking: Sometimes, a start event is logged without a corresponding end event, often due to job failures or network issue. By logging the run ID at each crucial step, we can better trace these issues back to their source, improving reliability and maintenance responses.
- Cluster and Event Linkage: This change aids in correlating specific logs with their respective clusters.
- Architecture Complexity: Given our architecture, where events pass through multiple layers (API proxies, Kafka, Marquez/DataDog), tracking can be challenging. Logging run IDs directly helps trace events that might get lost due to networking issues or other disruptions.
- Elasticsearch and Datadog Integration: Linking these concise log entries with detailed event data in Datadog via Elasticsearch streamlines event tracking and troubleshooting, enhancing our overall data governance framework.

(OpenLineage events are dispatched to Marquez and DataDog while Spark logs are sent to elasticsearch in our case)

I hope this clarifies the intent and utility of the proposed log enhancements. I look forward to your thoughts and any further suggestions you might have.

algorithmy1: Hi, @d-m-h Do you have any updates concerning this thread please ?
I am available for any further question.

pawel-big-lebowski: Some of the concerns mentioned can be solved with Spark metrics mechanism -> https://openlineage.io/docs/integrations/spark/metrics

Looking from the users' point, the ability to match OpenLineage events with application logs is something huge. 

I think it would make sense to add extra if statement:
```
if (log.isDebugEnabled()) {
   log.debug(
          "Emitting lineage completed successfully  with run id: {}: {}", event.getRun().getRunId(), OpenLineageClientUtils.toJson(event));
} else {
  log.info(
          "Emitting lineage completed successfully with run id: {}", event.getRun().getRunId());
}
```
But this is still a system which relies on log message structure which is not tested nor guaranteed to remain like this in future. 

To wrap up: I wouldn't recommend mapping `Logger <~> OL events` based on log entries (instead, one can add extra OL facet to let identify the logs related to the OL event), but I don't mind adding extra log if it is not duplicating the existing one.

mobuchowski: I like Paweł's idea.

algorithmy1: I like it as well.
I think it is good to have the information in the logs.
I love the metrics, we should update the Openlineage version used.
I will put it as a task in my backlog to add a facet that leads to the job's execution page + logs physical location.

Issue #2657, opened 2024-05-03
[BUG] [dbt] Overwritten job name when a project is launched with different models or profiles


blacklight:
Consider [the way the `dbt-ol` wrapper](https://github.com/OpenLineage/OpenLineage/blob/c9b06a429c374c71749c75e7c94e82ed6f0a3f22/integration/dbt/scripts/dbt-ol#L144) initializes the name of the underlying OpenLineage job:

```python
start_event = dbt_run_event_start(
    job_name=f"dbt-run-{processor.project_name}",
    job_namespace=job_namespace,
    parent_run_metadata=parent_run_metadata,
)
```

What this means is that two invocations of the same dbt project but from different jobs (e.g. in two different tasks of the same Airflow DAG), with different profiles or different models/model selectors will still result in OpenLineage events that have the same job name, thus making it harder to disambiguate.

The generated OpenLineage events should probably include at least the dbt profile and model(s) as part of the job name, along the lines of what is done to disambiguate Spark jobs.

PR #2658, opened 2024-05-03
[dbt] Support a less ambiguous logic to generate job names.

blacklight:
### Problem

The name of the `dbt-ol run` wrapper job generated by the connector isn't sufficiently unique, as it only depends on the dbt project name.

Multiple tasks that launch the same project, but with different profiles and/or models, will experience name clashes.

Closes: #2657

### Solution

If a new environment variable is set (`OPENLINEAGE_DBT_USE_EXTENDED_JOB_NAME`, used in order to prevent breaking existing flows, but it should ideally be the default), then the name of the job published on the event will change from:

`dbt-run-{dbt_project_name}`

To:

`dbt-run-{dbt_project_name}-{dbt_profile_name}-{dbt_model}`

#### One-line summary:

Include profile and models in the dbt job name to make it more unique.

### Checklist

----


integration/common/openlineage/common/provider/dbt/local.py
```diff
@@ -101,10 +104,45 @@ def __init__(
...
         self.profile_name = profile_name or dbt_project.get("profile")
         if not self.profile_name:
             raise KeyError(f"profile not found in {dbt_project}")
 
+    @property
+    def _use_extended_job_name(self):
+        # The name of the job published on the OpenLineage events for the dbt-ol run wrapper
+        # should include extended attributes (profile and model) only if this environment
+        # variable is set, in order to prevent breaking back-compatibility.
+        var = os.getenv("OPENLINEAGE_DBT_USE_EXTENDED_JOB_NAME", "false").lower()
```
blacklight, line 118: @mobuchowski @JDarDagran do you think it makes sense to control the job name structure using this environment variable (I introduced it to avoid the case where a job is named `A` before the update and `B` afterwards), or is it ok to just apply the new logic to all the workflows?

integration/common/openlineage/common/provider/dbt/local.py
```diff
@@ -101,10 +104,45 @@ def __init__(
...
+        if not self._use_extended_job_name:
+            return job_name
+
+        if self.profile_name:
+            job_name = f"{job_name}-{self.profile_name}"
+        if self.models:
+            job_name = f"{job_name}-" + "-".join(
+                [self._sanitize_job_name_component(model) for model in self.models]
+            )
+        elif self.selector:
```
mobuchowski, line 141: Do you think that if chain covers all the potential cases when we want to have extended job name? 

I'm not that up to date with `dbt` lately so I don't know - we don't have to cover everything here, but if there's something else I'd leave a comment there.

integration/common/openlineage/common/provider/dbt/local.py
mobuchowski: @blacklight any idea whether we might want to use enum here rather than boolean? Like, explicitely choosing between job names depending on `model` or `selector` rather than forcing the hierarchy in `job_name` function?

mobuchowski (commented):
I like the logic more than introducing templates.

As for testing this, there are both tests in `common` [here](https://github.com/OpenLineage/OpenLineage/tree/main/integration/common/tests/dbt) for parsing manifests, and Airflow-based tests.

I think adding Airflow-based test would be a good idea here: to do that you need to add DAG here: https://github.com/OpenLineage/OpenLineage/tree/main/integration/airflow/tests/integration/tests/airflow/dags and then write expected OL event here: https://github.com/OpenLineage/OpenLineage/tree/main/integration/airflow/tests/integration/requests

Current dbt DAG: https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/tests/integration/tests/airflow/dags/dbt_dag.py

integration/common/openlineage/common/provider/dbt/local.py
```diff
@@ -84,6 +85,8 @@ def __init__(
         profile_name: Optional[str] = None,
         target: Optional[str] = None,
         target_path: Optional[str] = None,
+        models: Optional[Sequence[str]] = None,
```
JDarDagran, line 88: I think it should not be limited to local processor only. I'm not entirely sure if and how this can be done with dbt cloud but additional logic seems not to interfere with base processor class.

integration/common/openlineage/common/provider/dbt/local.py
blacklight: That would make sense if dbt supports a mix of models list + model selector, but I haven't toyed enough with dbt to know if that is the case.

My understanding is that dbt supports either a list of models, or a model selector, but not both (at least not in the cases I've seen).

If that is the case then I don't see much value in an enum approach - either the user wants an extended job name or not, and if so then we already know which components to expect.

@JDarDagran do you know if a model+selector approach is supported?

integration/common/openlineage/common/provider/dbt/local.py
blacklight: Talking from the experience of how we use dbt in Booking, I think that a combination of profile + model list/selector should be sufficient for us to uniquely identify a task. But I'm not sure if this may be the case for everyone. I've just pushed a comment for this property that explains the `job_name` construction strategy and invites anyone with more clue than me to contribute if they feel like more identifiers are needed.

integration/common/openlineage/common/provider/dbt/local.py
JDarDagran: `dbt run --target dev --select tag:nightly --selector nightly` this simple example seems to be skipping `--select tag:nightly` if passing `--selector`

JDarDagran (approved):

integration/common/openlineage/common/provider/dbt/local.py
mobuchowski: >If that is the case then I don't see much value in an enum approach - either the user wants an extended job name or not, and if so then we already know which components to expect.

I'm okay with that

mobuchowski (approved):

PR #2675, opened 2024-05-07
[SPARK] Spark extension Java interfaces

ddebowczyk92:
### Problem

This pull request introduces Java interfaces that fulfill the same role as traits in the `spark-extension-scala` module. The `spark-extension` module addresses the needs of Spark extension developers who want to keep their repositories 'Scala version agnostic'

### Solution

PR contains new artifact `spark-extension` with interfaces as well as Spark integration internals

### Checklist

----


integration/spark-interfaces/src/main/java/io/openlineage/spark/extension/v1/OlExprId.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Objects;
+
+/**
+ * Class to contain reference to Spark's ExprId without adding dependency to Spark library
+ *
+ * @see <a
+ *     href="https://github.com/apache/spark/blob/ce5ddad990373636e94071e7cef2f31021add07b/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/namedExpressions.scala#L48">namedExpressions.scala</a>
+ */
+public final class OlExprId {
```
pawel-big-lebowski, line 15: Just thinking loudly. If we were to do that again I would start with Java version first. 
However, scala version is still marked as experimental. 

What if we made Scala interfaces depend on Java package and use Java classes instead of Scala case classes in both packages? After this change we would leave Scala interfaces `ColumnLevelLineageNode`, `InputLineageNode` and `OutputLineageNode` but their return type would be Java classes. 

I think it's the last moment to have it doable. This will simplify on the openlineage-spark connector side and reduce ricks of interface packages grow inconsistently. 

@mobuchowski does it make sense?

pawel-big-lebowski (commented):
Code looks good to me. 
There's one thought in the comments which I would like to discuss offline to make sure we're heading good direction.

integration/spark-interfaces/src/main/java/io/openlineage/spark/extension/v1/OlExprId.java
mobuchowski: I agree - we don't have any known implementation yet. Better to do it right then way overly cautious.

dolfinus:
Is that the same as #2809?

PR #2693, opened 2024-05-14
python: use v2 Python facets

JDarDagran:
### Problem

#2520 introduced v2 facets in Python client. Rest of Python code requires migration to use them.

### Solution

Remove usage of v1 facets. In https://github.com/apache/airflow/pull/39614 there is removed `openlineage.common` dependency on BigQuery in Google provider so there should be no discrepancy between versions.

#### One-line summary:

Migrate integrations to use v2 python facets.

### Checklist

----

### DIFF
--- a/integration/airflow/openlineage/airflow/adapter.py
+++ b/integration/airflow/openlineage/airflow/adapter.py
@@ -9,20 +9,21 @@ from typing import TYPE_CHECKING, Dict, List, Optional
 from openlineage.airflow.extractors import TaskMetadata
 from openlineage.airflow.utils import DagUtils, redact_with_exclusions
 from openlineage.airflow.version import __version__ as OPENLINEAGE_AIRFLOW_VERSION
-from openlineage.client import OpenLineageClient, OpenLineageClientOptions, set_producer
-from openlineage.client.facet import (
-    BaseFacet,
-    DocumentationJobFacet,
-    ErrorMessageRunFacet,
-    JobTypeJobFacet,
-    NominalTimeRunFacet,
-    OwnershipJobFacet,
-    OwnershipJobFacetOwners,
-    ParentRunFacet,
-    ProcessingEngineRunFacet,
-    SourceCodeLocationJobFacet,
+from openlineage.client import OpenLineageClient, OpenLineageClientOptions
+from openlineage.client import set_producer_v2 as set_producer
+from openlineage.client.event_v2 import Job, Run, RunEvent, RunState
+from openlineage.client.facet_v2 import (
+    JobFacet,
+    RunFacet,
+    documentation_job,
+    error_message_run,
+    job_type_job,
+    nominal_time_run,
+    ownership_job,
+    parent_run,
+    processing_engine_run,
+    source_code_location_job,
 )
-from openlineage.client.run import Job, Run, RunEvent, RunState
 from openlineage.client.uuid import generate_static_uuid
 
 from airflow.stats import Stats
@@ -45,8 +46,8 @@ set_producer(_PRODUCER)
 # https://openlineage.io/docs/spec/facets/job-facets/job-type
 # They must be set after the `set_producer(_PRODUCER)`
 # otherwise the `JobTypeJobFacet._producer` will be set with the default value
-_JOB_TYPE_DAG = JobTypeJobFacet(jobType="DAG", integration="AIRFLOW", processingType="BATCH")
-_JOB_TYPE_TASK = JobTypeJobFacet(jobType="TASK", integration="AIRFLOW", processingType="BATCH")
+_JOB_TYPE_DAG = job_type_job.JobTypeJobFacet(jobType="DAG", integration="AIRFLOW", processingType="BATCH")
+_JOB_TYPE_TASK = job_type_job.JobTypeJobFacet(jobType="TASK", integration="AIRFLOW", processingType="BATCH")
 
 log = logging.getLogger(__name__)
 
@@ -129,7 +130,7 @@ class OpenLineageAdapter:
         nominal_end_time: str,
         owners: List[str],
         task: Optional[TaskMetadata],
-        run_facets: Optional[Dict[str, BaseFacet]] = None,  # Custom run facets
+        run_facets: Optional[Dict[str, RunFacet]] = None,  # Custom run facets
     ) -> str:
         """
         Emits openlineage event of type START
@@ -150,7 +151,7 @@ class OpenLineageAdapter:
         """
         from airflow.version import version as AIRFLOW_VERSION
 
-        processing_engine_version_facet = ProcessingEngineRunFacet(
+        processing_engine_version_facet = processing_engine_run.ProcessingEngineRunFacet(
             version=AIRFLOW_VERSION,
             name="Airflow",
             openlineageAdapterVersion=OPENLINEAGE_AIRFLOW_VERSION,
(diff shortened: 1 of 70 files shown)


integration/airflow/openlineage/airflow/adapter.py
```diff
@@ -44,8 +44,8 @@
 # https://openlineage.io/docs/spec/facets/job-facets/job-type
 # They must be set after the `set_producer(_PRODUCER)`
 # otherwise the `JobTypeJobFacet._producer` will be set with the default value
-_JOB_TYPE_DAG = JobTypeJobFacet(jobType="DAG", integration="AIRFLOW", processingType="BATCH")
-_JOB_TYPE_TASK = JobTypeJobFacet(jobType="TASK", integration="AIRFLOW", processingType="BATCH")
+_JOB_TYPE_DAG = job_type_job.JobTypeJobFacet(jobType="DAG", integration="AIRFLOW", processingType="BATCH")
+_JOB_TYPE_TASK = job_type_job.JobTypeJobFacet(jobType="TASK", integration="AIRFLOW", processingType="BATCH")
```
kacpermuda, line 48: nit: we could possibly just pass the producer as an argument now, it should be simpler.

kacpermuda: nit: the _PRODUCER on line 39 is two strings now, i think we can clean that

integration/airflow/openlineage/airflow/utils.py
```diff
@@ -510,6 +515,7 @@ def is_json_serializable(item):
 
 
 def _is_name_redactable(name, redacted):
+    print("checking if redactabled", name, redacted)
```
kacpermuda, line 518: nit: remove the print?

integration/airflow/setup.cfg
```diff
@@ -46,7 +46,7 @@ deps = -r dev-requirements.txt
 	airflow-2.6.1: apache-airflow==2.6.1
 whitelist_externals = bash
 commands = bash -ec "if [[ ! -f $0/airflow.db ]]; then airflow db reset -y; fi" {envdir}
-	python -m pytest --cov=openlineage --junitxml=test-results/junit.xml
+	python -m pytest --cov=openlineage --junitxml=test-results/junit.xml -vv
```
kacpermuda, line 49: Is this a desired change or a leftover ?

integration/airflow/tests/integration/test_integration.py
```diff
@@ -518,7 +518,7 @@ def test_airflow_mapped_task_facet(airflow_db_conn):
     started_events = [event for event in mapped_events if event["eventType"] == "START"]
     assert len(started_events) == 3
     assert started_events[0]["job"]["name"] == f"{dag_id}.{task_id}"
-    assert set([event["job"]["facets"]["sourceCode"]["source"] for event in started_events]) == set(
+    assert set([event["job"]["facets"]["sourceCode"]["sourceCode"] for event in started_events]) == set(
```
kacpermuda, line 521: Should we mention all the changes of facet keys and attributes in a changelog or somewhere else? Just thinking out loud what would be best - maybe it's more important for the OL airflow provider, but still should be mentioned somewhere.

integration/airflow/tests/test_macros.py
```diff
@@ -51,6 +51,7 @@ def test_lineage_parent_id():
         task_id="task_id",
         execution_date="execution_date",
         try_number=1,
+        dag_run=mock.MagicMock(run_id="run_id"),
```
kacpermuda, line 54: I think it works without that line, am i wrong?

integration/airflow/tests/test_openlineage_adapter.py
```diff
@@ -128,11 +129,12 @@ def test_emit_start_event():
...
                 facets={
-                    "documentation": DocumentationJobFacet(description="description"),
-                    "jobType": JobTypeJobFacet(processingType="BATCH", integration="AIRFLOW", jobType="TASK"),
+                    "documentation": documentation_job.DocumentationJobFacet(description="description"),
+                    "jobType": job_type_job.JobTypeJobFacet(
+                        processingType="BATCH", integration="AIRFLOW", jobType="TASK"
+                    ),
                 },
             ),
-            producer=_PRODUCER,
```
kacpermuda, line 135: If we're removing this producer line from tests, i would remove it also from the adapter so it's not passed to RunEvent. Right now the producer is still passed explicitly, so it can stay in the tests as well.

integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -123,8 +125,10 @@ class ParentRunMetadata:
     job_name: str = attr.ib()
     job_namespace: str = attr.ib()
 
-    def to_openlineage(self) -> ParentRunFacet:
-        return ParentRunFacet.create(runId=self.run_id, name=self.job_name, namespace=self.job_namespace)
+    def to_openlineage(self) -> parent_run.ParentRunFacet:
+        return parent_run.ParentRunFacet.create(
```
kacpermuda, line 129: nit: I think `parent_run.ParentRunFacet.create()` is deprecated? I saw it in multiple places, just assuming we are ignoring this deprecation for now.

integration/common/tests/dbt/spark/thrift/result.json
```diff
@@ -605,7 +605,7 @@
         "eventType": "START",
         "inputs": [
             {
-                "facets": {
+                "inputFacets": {
```
kacpermuda, line 608: question: it looks like quite a big change, why has it change and should we communicate it somehow?

kacpermuda (approved):
Nice 🚀 Leaving some comments and questions

integration/airflow/openlineage/airflow/adapter.py
JDarDagran: > nit: we could possibly just pass the producer as an argument now, it should be simpler.

I don't think we're going away with `set_producer` approach? For me it's still main way to set producer, maybe I'm wrong 🤔 

> nit: the _PRODUCER on line 39 is two strings now, i think we can clean that
It's not two string but the concatenation might look better, agreed.

integration/airflow/setup.cfg
JDarDagran: Leftover.

integration/airflow/tests/integration/test_integration.py
JDarDagran: In this case it's different. This is something that is fixed and now is correct with JSON Schema spec. Besides, this is OL standalone package, I wouldn't go strongly with deprecations/notices here.

integration/airflow/tests/test_macros.py
JDarDagran: Yes, I think the logic of the macro changed in between, this wasn't fixed.

integration/airflow/openlineage/airflow/adapter.py
kacpermuda: > It's not two string but the concatenation might look better, agreed.

That's exactly what i meant :)

> I don't think we're going away with set_producer approach? For me it's still main way to set producer, maybe I'm wrong 🤔

We are not, I don't liked the idea that we have to rely on defining something (facets) after calling something (set_producer) if there is another, more reliable way of making sure it works as expected. Just a preference, so no pressure here.

integration/airflow/tests/test_openlineage_adapter.py
JDarDagran: I removed it from the adapter.

integration/common/openlineage/common/provider/dbt/processor.py
JDarDagran: Yes, I should have remove the usage of it. Just did.

integration/common/tests/dbt/spark/thrift/result.json
JDarDagran: If you track what OpenLineage spec defines, you might see that this is correct key for the `InputDatasetFacet`. Generally speaking we need some strategy what to do with changes that are actually fixes.

mobuchowski: >Generally speaking we need some strategy what to do with changes that are actually fixes.

If something was a bug, we fix it. This is different than something that wasn't the best solution, but it was intended as it.

JDarDagran:
@kacpermuda  @mobuchowski I've added `facets` section in dbt integration with TODO comment to remove it next release.

mobuchowski (approved):

PR #2698, opened 2024-05-16
[SPARK] verify jar content after build

pawel-big-lebowski:
### Problem

Spark integration contains several subproject and it's important to guarantee that released jar contains  desired classes only. Common pitfalls we would like to avoid: 
 * releasing unshaded dependency to external library that could clash with classes user has on classpath, 
 * not-including openlineage-spark classes from some subproject we have.

In this PR, a tool is added to verify `shadowJar` content and prevent issues described from happening. Such issues are hard to prevent currently and require manual verification of manually unpacked jar content. 

PR contains gradle extension which runs the validation. Together with this PR, there is another PR #2715  which introduces possible jar issues in order to show the validation is working and proper error messages are produced. 

I was unable to verify if Scala version used in compilation matches scala version of the release artifact, although such check would be useful too.

### DIFF
--- a/integration/spark/app/build.gradle
+++ b/integration/spark/app/build.gradle
@@ -83,7 +83,6 @@ dependencies {
     implementation(project(path: ":spark34", configuration: activeRuntimeElementsConfiguration))
     implementation(project(path: ":spark35", configuration: activeRuntimeElementsConfiguration))
     implementation("org.apache.httpcomponents.client5:httpclient5:5.3")
-    implementation("io.micrometer:micrometer-core:${micrometerVersion}")
 
     compileOnly("org.apache.spark:spark-core_${scala}:${spark}")
     compileOnly("org.apache.spark:spark-sql_${scala}:${spark}")
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/ContextFactory.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/ContextFactory.java
@@ -125,7 +125,7 @@ public class ContextFactory {
   public static Optional<QueryExecution> executionFromCompleteEvent(
       SparkListenerSQLExecutionEnd event) {
     try {
-      return Optional.of((QueryExecution) MethodUtils.invokeMethod(event, "qe", null));
+      return Optional.of((QueryExecution) MethodUtils.invokeMethod(event, "qe", (Object[]) null));
     } catch (NoSuchMethodException e) {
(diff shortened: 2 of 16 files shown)


integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/ContextFactory.java
```diff
@@ -125,7 +125,7 @@ public Optional<ExecutionContext> createSparkSQLExecutionContext(
   public static Optional<QueryExecution> executionFromCompleteEvent(
       SparkListenerSQLExecutionEnd event) {
     try {
-      return Optional.of((QueryExecution) MethodUtils.invokeMethod(event, "qe", null));
+      return Optional.of((QueryExecution) MethodUtils.invokeMethod(event, "qe", (Object[]) null));
```
pawel-big-lebowski, line 128: get rid of irritating warning during build & shadowJar

integration/spark/shared/build.gradle
```diff
@@ -44,7 +43,6 @@ dependencies {
     api("io.openlineage:openlineage-java:${project.version}")
     api("io.openlineage:openlineage-sql-java:${project.version}")
     api("io.openlineage:spark-interfaces-scala_${scalaBinaryVersion}:${project.version}")
-    api("io.micrometer:micrometer-core:${micrometerVersion}")
```
pawel-big-lebowski, line 47: no need to add micrometer in each subproject as it's already in openlineage-java.

integration/spark/app/build.gradle
```diff
@@ -27,7 +27,7 @@ ext {
     mockitoVersion = '4.11.0'
     postgresqlVersion = '42.7.1'
     testcontainersVersion = '1.19.3'
-    micrometerVersion = '1.12.4'
+    micrometerVersion = '1.13.0'
```
mobuchowski, line 30: This needs to be in sync with Java client - maybe just get the dependency transitively from it?

mobuchowski (changes requested):

integration/spark/app/build.gradle
pawel-big-lebowski: Within the PR I got rid of `micrometer-core` dependencies from Spark integration. It's not needed as this is already present within `openlineage-java`. This setting has only effect for `testImplementation`, so this doesn't need to be in sync with Java client and our tests will fail in case of breaking changes in version.

mobuchowski (approved):

Issue #560, opened 2022-02-23
[INTEGRATION][Kafka] Standarize kafka dataset naming


mobuchowski:
Natural representation of Kafka cluster on clients is a list of bootstrap servers. This would be dataset namespace. 

This list can change over time. Additionally, same cluster can have different representation on different clients - whether by shuffling the list, or having different parts of the same bootstrap.server list.

We can do "normalization" like sorting bootstrap.server IPs to have it more easily match between different clients.

The question is whether we should do this, and whether OpenLineage consumers should do something to connect different representations of same cluster. An example would be to treat all bootstrap.server list that have one matching IP as one cluster.

pawel-big-lebowski:
> An example would be to treat all bootstrap.server list that have one matching IP as one cluster.
If only there was some neat data structure for that ;-)

Perhaps the same problem applies not only to Kafka but also hive servers. I think this should be solved on the Marquez side by introducing there a concept and a separate table to store each bootstrap or Hive server in a separate row. We could assume separators like `,` or `;` and do not care if it is Kafka or anything else.

FranMorilloAWS:
Any news on this?

dolfinus:
Same issues can be faced on with all clustered databases, like Clickhouse, Oracle, MySQL, etc.

There actually 2 ways this could be implemented:
1. Proposed example from the issue description, e.g. having a namespace name like `kafka://first.of.nodes`, and also send some custom facet with list of all known nodes (e.g. `bootstrap.servers` for Kafka). So it's up to backend how to group events matching all of the nodes in this list to one namespace.
2. Use some unique internal name, like `cluster_id` in Kafka or `DBID` for Oracle.

But there are issues.

First case (one host + list of all hosts):
1. It is not always possible to get list of these nodes. For example, `bootstrap.servers` in Kafka may contain only address of one Kafka broker, but then it may return address of all other brokers to the client (using `ADVERTISED_LISTENERS`). So connection address specified in client or baked into Spark execution plan (e.g. `JDBCRelation`) may contain only part of the real list.
2. User may set either IP address or DNS name, both of them are valid connection address. This is may be the case even on non-cluster installations. Also user may pass both `domain` and `domain.company.com` which may resolve to the same IP.
  Someone can say that it will be better to resolve this DNS to IP first, and then use it, but there are things like geoDNS. And reverse DNS lookup is not always possible.

Second case (unique internal name):
[...]

PR #2720, opened 2024-05-23
Dataset namespace resolver feature

pawel-big-lebowski:
### Problem

Oftentimes a dataset's namespace is identified by host used to access it. For example, for Kafka clusters, a first element of bootstrap servers is used to create the namespace. This was the initial approach that deserves to improved and enhanced. 

This PR introduces dataset namespace resolving mechanism, which resolves dataset namespaces based on the resolvers configured. The core mechanism is implemented in `openlineage-java` and can be used within Flink and Spark integrations. 

It contains out-of-the-box resolvers like `hostList`, `pattern` or `patternGroup` type presented below:
```
dataset:
  namespaceResolvers:
      kafka-prod:
        type: hostList
        hosts: ['kafka-prod13.company.com', 'kafka-prod15.company.com']
      cassandra-prod:
        type: pattern
        # 'cassandra-prod7.company.com', 'cassandra-prod8.company.com'
        regex: 'cassandra-prod(\d)+\.company\.com'
      test-pattern:
        type: patternGroup
        # 'cassandra-test-7.company.com', 'cassandra-test-8.company.com', 'kafka-test-7.company.com', 'kafka-test-8.company.com'
        regex: '(?<cluster>[a-zA-Z-]+)-(\d)+\.company\.com:[\d]*'
        matchingGroup: "cluster"
```

and the ability to add custom namespace resolver through the Service Loader. More info will be put in the docs.

Related to: #560

Corresponding docs PR: https://github.com/OpenLineage/docs/pull/329

----

### DIFF
--- a/client/java/src/test/resources/config/datasetNamespaceResolver.yaml
+++ b/client/java/src/test/resources/config/datasetNamespaceResolver.yaml
@@ -1,0 +1,19 @@
+transport:
+  type: console
+dataset:
+  namespaceResolvers:
+      kafka-prod:
+        type: hostList
+        hosts: ['kafka-prod13.company.com', 'kafka-prod15.company.com']
+        schema: kafka
+      cassandra-prod:
+        type: pattern
+        # 'cassandra-prod7.company.com', 'cassandra-prod8.company.com'
+        regex: 'cassandra-prod(\d)+\.company\.com'
+        schema: cassandra
+      test-pattern:
+        type: patternGroup
+        # 'cassandra-test-7.company.com', 'cassandra-test-8.company.com', 'kafka-test-7.company.com', 'kafka-test-8.company.com'
+        regex: '(?<cluster>[a-zA-Z-]+)-(\d)+\.company\.com:[\d]*'
+        matchingGroup: "cluster"
+        schema: cassandra
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
@@ -197,8 +197,6 @@ public class ArgumentParser {
   }
 
   private static boolean isArrayType(String value) {
-    return value.startsWith(ARRAY_PREFIX_CHAR)
-        && value.endsWith(ARRAY_SUFFIX_CHAR)
-        && value.contains(DISABLED_FACETS_SEPARATOR);
+    return value.startsWith(ARRAY_PREFIX_CHAR) && value.endsWith(ARRAY_SUFFIX_CHAR);
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 ### Added
 * **Spark: add `jobType` facet to Spark application events** [`#2719`](https://github.com/OpenLineage/OpenLineage/pull/2719) [@dolfinus](https://github.com/dolfinus)  
     *Add `jobType` facet to `runEvent`s emitted by `SparkListenerApplicationStart`.*
+* **Spark & Flink: Introduce dataset namespace resolver.** [`#2720`](https://github.com/OpenLineage/OpenLineage/pull/2720) [@pawel-big-lebowski](https://github.com/pawel-big-lebowski)  
+  *Enable resolving dataset namespace with predefined resolvers like: `HostListNamespaceResolver`, `PatternNamespaceResolver`, `PatternMatchingGroupNamespaceResolver` or custom implementation loaded with ServiceLoader. Feature is useful to resolve hostnames into cluster identifiers.*
 
 ### Fixed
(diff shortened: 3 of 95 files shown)


client/java/src/main/java/io/openlineage/client/dataset/hostResolver/DatasetHostResolverBuilder.java
```diff
@@ -0,0 +1,15 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.dataset.hostResolver;
+
+public interface DatasetHostResolverBuilder {
```
d-m-h, line 8: Maybe a better name would be "Dataset**Namespace**Resolver" to keep up with the already established vocabulary used within OpenLineage?

dolfinus:
Is it possible to update list of resolvers in already running Spark session?

pawel-big-lebowski:
> Is it possible to update list of resolvers in already running Spark session?

Not now. The list is initiated based on the configuration, that is needed to initiate `OpenLineageContext`

client/java/src/main/java/io/openlineage/client/utils/DatasetNamespaceCombinedResolver.java
```diff
@@ -0,0 +1,91 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.utils;
```
mobuchowski, line 6: Why not have this class in the package together with rest of the relevant classes?

client/java/src/test/resources/config/datasetNamespaceResolver.yaml
```diff
@@ -0,0 +1,16 @@
+transport:
+  type: console
+dataset:
+  namespaceResolvers:
```
mobuchowski, line 4: Looking at those examples, I wonder if we should not add optional field indicating `schema` of the namespace that the resolver would be operating on. 

For example, I would like to specify that `kafka-prod` resolver would work only on `kafka` namespaces, while `cassandra-prod` would work on `cassandra` namespaces.

I think there exists a risk of mixing those up, especially with too wide regex.

client/java/src/main/java/io/openlineage/client/dataset/namespaceResolver/DatasetNamespaceResolver.java
```diff
@@ -0,0 +1,23 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.dataset.namespaceResolver;
```
mobuchowski, line 6: Everywhere in this PR where we use dot case: maybe `namespace.resolver` instead of `namespaceResolver`?

I don't like mixing up the case styles - let's go with regular Java `dot.separated.case` in packages/directory names and Spark/Flink conf.

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
```diff
@@ -197,8 +197,6 @@ private static List<String> getJsonPath(String keyPath) {
   }
 
   private static boolean isArrayType(String value) {
-    return value.startsWith(ARRAY_PREFIX_CHAR)
-        && value.endsWith(ARRAY_SUFFIX_CHAR)
-        && value.contains(DISABLED_FACETS_SEPARATOR);
```
mobuchowski, line 202: Why was this removed?

client/java/src/main/java/io/openlineage/client/dataset/namespaceResolver/HostListNamespaceResolverBuilder.java
```diff
@@ -0,0 +1,24 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.dataset.namespaceResolver;
+
+public class HostListNamespaceResolverBuilder implements DatasetNamespaceResolverBuilder {
```
mobuchowski, line 8: Would be awesome to get proper documentation of those in OpenLineage/docs, but I'm sure you've already thought of that.

mobuchowski (commented):
@pawel-big-lebowski astounding work. Would love to hear some clarifications though.

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
pawel-big-lebowski: problem with this are single element arrays which do ton contain `;` separator

mobuchowski (approved):

PR #2727, opened 2024-05-24
http: override debug level

mobuchowski:
When a DAG author in Airflow enables DEBUG logging for HTTPConnection with:
```
http_client.HTTPConnection.debuglevel = 1
```
the OpenLineage logs: headers, body, etc. We don't want these requests logged!

Here is an example of a reproduction:
```
import requests
import http.client as http_client

http_client.HTTPConnection.debuglevel = 1
resp = requests.get("http://localhost:8080/login")

send: b'GET /login/ HTTP/1.1\r\nHost: localhost:8080\r\nUser-Agent: python-requests/2.31.0\r\nAccept-Encoding: gzip, deflate, zstd\r\nAccept: */*\r\nConnection: keep-alive\r\nCookie: session=e9194549-d5d1-440e-883c-c0f4608f0def.fDuMSL2im-gbcnJP03WRbJc0KqI\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server: Werkzeug/2.2.3 Python/3.11.9
header: Date: Thu, 23 May 2024 17:20:48 GMT
header: Content-Type: text/html; charset=utf-8
header: Content-Length: 18079
header: Cache-Control: no-store
header: X-Robots-Tag: noindex, nofollow
header: Set-Cookie: session=e9194549-d5d1-440e-883c-c0f4608f0def.fDuMSL2im-gbcnJP03WRbJc0KqI; Expires=Sat, 22 Jun 2024 17:20:48 GMT; HttpOnly; Path=/; SameSite=Lax
header: Connection: close
```

This fixes it so there's no debug logging of HTTP request when uploading OpenLineage events, even if the DAG author has turned it on for their task.


client/python/openlineage/client/transport/http.py
```diff
@@ -160,6 +162,9 @@ def set_adapter(self, adapter: HTTPAdapter) -> None:
             self.session.mount(self.url, adapter)
 
     def emit(self, event: Event) -> Response:
+        # If anyone overrides debuglevel manually, we can potentially leak any secrets to logs.
```
kacpermuda, line 165:
```suggestion
        # If anyone overrides debuglevel manually, we can potentially leak secrets to logs.
```

client/python/openlineage/client/transport/http.py
```diff
@@ -160,6 +162,9 @@ def set_adapter(self, adapter: HTTPAdapter) -> None:
...
+        # Override this setting to make sure it does not happen.
+        http_client.HTTPConnection.debuglevel = 0
```
kacpermuda, line 167: Is it a global change? Maybe we should set it back to 1 after we emit the event?

dolfinus:
> the OpenLineage logs: headers, body, etc. We don't want these requests logged!

Why not? User explicitly asked for debug logs, why this should be ignored by OpenLineage?

client/python/openlineage/client/transport/http.py
mobuchowski: Changed.

mobuchowski:
>Why not? User explicitly asked for debug logs, why this should be ignored by OpenLineage?

Giant majority of times they set this up because they want to see debug logs for their own code using requests/http.client - not us, and they don't realize they could leak secrets from OL. 

The settings are not isolated - it's better to not log them rather than to assume users want to increase loglevel of OL specifically. 

In contrast, for Java we'd not have this problem because HTTP library is shaded.

kacpermuda (approved):

jedcunningham (approved):

JDarDagran (approved):

PR #2743, opened 2024-05-30
[SPARK] upgrade supported versions to 3.4.2->3.4.3 and 3.5.0->3.5.1

pawel-big-lebowski:
### Problem

Upgrade to latest Spark versions used by the integration.

Issue encountered: 
 * For a job writing to JDBC table, although the job is uses `SparkSqlExecutionContext`, it is also using rdd job to when writing to jdbc. As a result, some of the listener calls run through `SparkSqlExecutionContext` while other run with `RddExecutionContext` which is makes the test fail because `setActiveJob` method that is necessary for output statistics to be up an running. 
 * As a solution to this `OpenLineageSparkListener` will store `Integer activeJobId` property to share jobId among SparkSQL and RDD execution contexts.

### DIFF
--- a/integration/spark/vendor/snowflake/build.gradle
+++ b/integration/spark/vendor/snowflake/build.gradle
@@ -17,12 +17,12 @@ ext {
             '3.1.3': '2.11.0',
             '3.2.4': '2.11.0',
             '3.3.4': '2.11.0',
-            '3.4.2': '2.15.0'
+            '3.4.3': '2.15.0'
     ]
 
     sparkProp = project.findProperty('spark.version').toString()
     // because snowflake doesn't have a 3.5.0 connector yet
-    spark = sparkProp == "3.5.0" ? "3.4.2" : sparkProp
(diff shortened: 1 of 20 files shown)


integration/spark/vendor/snowflake/build.gradle
```diff
@@ -17,12 +17,12 @@ ext {
...
             '3.2.4': '2.11.0',
             '3.3.4': '2.11.0',
-            '3.4.2': '2.15.0'
+            '3.4.3': '2.15.0'
     ]
 
     sparkProp = project.findProperty('spark.version').toString()
     // because snowflake doesn't have a 3.5.0 connector yet
-    spark = sparkProp == "3.5.0" ? "3.4.2" : sparkProp
+    spark = sparkProp == "3.5.1" ? "3.4.3" : sparkProp
```
pawel-big-lebowski, line 25: `sparkProp >= "3.5.0"`

mobuchowski (approved):

PR #2744, opened 2024-05-30
spark: make sure debug logging is guarded when it can cause function call

mobuchowski:
If loglevel is higher than debug, then there's no point in having to call methods result of which won't be used - especially if they are rather heavy, like `OpenLineageClientUtils.toJson(event)`

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
@@ -71,7 +71,9 @@ class SparkSQLExecutionContext implements ExecutionContext {
 
   @Override
   public void start(SparkListenerSQLExecutionStart startEvent) {
-    log.debug("SparkListenerSQLExecutionStart - executionId: {}", startEvent.executionId());
+    if (log.isDebugEnabled()) {
+      log.debug("SparkListenerSQLExecutionStart - executionId: {}", startEvent.executionId());
+    }
     if (!olContext.getQueryExecution().isPresent()) {
       log.info(NO_EXECUTION_INFO, olContext);
       return;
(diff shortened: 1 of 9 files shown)


integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -71,7 +71,9 @@ public SparkSQLExecutionContext(
 
   @Override
   public void start(SparkListenerSQLExecutionStart startEvent) {
-    log.debug("SparkListenerSQLExecutionStart - executionId: {}", startEvent.executionId());
+    if (log.isDebugEnabled()) {
```
pawel-big-lebowski, line 74: I don't like the redundant approach of checking logging level in our code. This should happen within logger library. 
Extra condition makes only sense when resolving `debug` method parameters can be expensive like json. Otherwise I don't think it's necessary. 
I wouldn't do that for `getClass` and `getContainerName` as well.

pawel-big-lebowski (approved):

integration/spark/spark34/src/main/java/io/openlineage/spark34/agent/lifecycle/plan/column/CreateReplaceInputDatasetBuilder.java
```diff
@@ -40,7 +40,10 @@ public boolean isDefinedAt(SparkListenerEvent event) {
 
   @Override
   public boolean isDefinedAtLogicalPlan(LogicalPlan x) {
-    log.debug("Calling isDefinedAtLogicalPlan on {} with children {}", x.getClass(), x.children());
+    if (log.isDebugEnabled()) {
+      log.debug(
+          "Calling isDefinedAtLogicalPlan on {} with children {}", x.getClass(), x.children());
```
tatiana, line 45: @mobuchowski are there any circumstances where `x` could be null?

integration/spark/spark34/src/main/java/io/openlineage/spark34/agent/lifecycle/plan/column/CreateReplaceInputDatasetBuilder.java
```diff
@@ -50,7 +53,9 @@ public boolean isDefinedAtLogicalPlan(LogicalPlan x) {
 
   @Override
   public List<OpenLineage.InputDataset> apply(SparkListenerEvent event, LogicalPlan x) {
-    log.debug("Calling apply on {}", x.getClass());
+    if (log.isDebugEnabled()) {
+      log.debug("Calling apply on {}", x.getClass());
```
tatiana, line 57: Same question here

PR #2755, opened 2024-06-04
[SPARK] spark lineage CLI verifier

pawel-big-lebowski:
### Problem

It's difficult to verify correctness of Openlineage Spark integration from command line which makes it difficult to build custom integrations tests to be run before upgrading `openlineage-spark` jar. 

### Solution

Implement tool that allows running configurable integration tests through command line:
```
./integration/spark/cli/configurable-test.sh --spark ./integration/spark/cli/spark-conf.yml --test ./integration/spark/cli/tests 

```

where spark config file is of a form:
```
appName: "CLI test application"
sparkVersion: 3.4.3
scalaBinaryVersion: 2.12
enableHiveSupport: true
packages:
  - org.apache.iceberg:iceberg-spark-runtime-3.2_2.13:1.4.0
sparkConf:
  key.subkey.subsubkey1: value1
  key.subkey.subsubkey2: value2
```
and `--test` param specifies directory containing:
 * sql script file
 * expected events in form of json per each file

Relevant documentation PR: https://github.com/OpenLineage/docs/pull/340

This PR also adds nightly-builds which run:
 * CLI verifier,
 * Databricks integration tests.

Once PR is merged, nightly-build in circle ci needs to be pointed into main branch. 
----


pawel-big-lebowski:
passing CI ru with `nightly-run` mode enabled: https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/11042/workflows/70d18bc7-67b2-489a-ab63-89763aad5477/jobs/234606

integration/spark/app/src/test/java/io/openlineage/spark/agent/util/RunEventVerifier.java
```diff
@@ -0,0 +1,283 @@
...
+            new StringBuilder()
+                .append("Couldn't match event for ")
+                .append(expectationFile.getName())
+                .append(System.lineSeparator())
+                .append(bestEffortDiffDescription(events, expectedJson))
+                .append(System.lineSeparator())
+                .toString();
+      }
+
+      matchesAll = matchesAll && matches;
```
tnazarew, line 93: you can also look into `Arrays.stream(expectationFiles).allMatch()`

integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableIntegrationTest.java
```diff
@@ -0,0 +1,210 @@
...
+            .findFirst();
+
+    if (config.getPackages() != null) {
+      command.add("--packages");
+      command.add(String.join(",", config.getPackages()));
+    }
+
+    command.add("--jars");
+    command.add(olJar.get().toAbsolutePath().toString());
+    command.add(getPySparkFile(runDir).getPath());
```
tnazarew, line 163: in theory you could have just one python file called something like `runner.py` which would have only two lines
```
import sys
eval(sys.argv[1])
```
and you could supply the the python code that way, this saves you creating temp files and later reading from them. It could also add layer of complexity with possible escaping some characters or supplying arguments in general so idk if it's better or not

tnazarew (commented):
overall great contribution, I petition to move all our integration tests to this structure in near future

.circleci/continue_config.yml
```diff
@@ -671,6 +676,29 @@ jobs:
...
+  configurable-integration-test-spark:
+    working_directory: ~/openlineage/
+    machine:
+      image: ubuntu-2004:current
+    resource_class: large
+    steps:
+      - *checkout_project_root
+      - when:
+          condition:
+            equal: [ active, << pipeline.parameters.nightly-run >> ]
```
mobuchowski, line 688: Not sure why this needs to be true?

.circleci/workflows/openlineage-spark.yml
```diff
@@ -45,6 +45,8 @@ workflows:
               spark-version: [ '3.2.4', '3.3.4', '3.4.2', '3.5.0' ]
           requires:
             - approval-integration-spark
+      - configurable-integration-test-spark:
+          context: integration-tests
```
mobuchowski, line 49: Do we need to run this as integration tests?

client/java/src/main/java/io/openlineage/client/transports/FileTransport.java
```diff
@@ -51,7 +51,7 @@ private void emit(String eventAsJson) {
           eventAsJson.replace(System.lineSeparator(), "") + System.lineSeparator(),
           StandardCharsets.UTF_8,
           true);
-      log.info("emitted event: " + eventAsJson);
+      log.debug("emitted event: " + eventAsJson);
```
mobuchowski, line 54: 👍

integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableIntegrationTest.java
mobuchowski: Might work but passing Python code via command line parameter sounds like too evil hack compared to just adding a file :)

integration/spark/app/src/test/java/io/openlineage/spark/agent/util/RunEventVerifier.java
```diff
@@ -0,0 +1,283 @@
...
+import shaded_package.com.fasterxml.jackson.core.JsonProcessingException;
+import shaded_package.com.fasterxml.jackson.databind.JsonNode;
+import shaded_package.com.fasterxml.jackson.databind.JsonSerializer;
+import shaded_package.com.fasterxml.jackson.databind.ObjectWriter;
+import shaded_package.net.javacrumbs.jsonunit.core.Configuration;
+import shaded_package.net.javacrumbs.jsonunit.core.Option;
+import shaded_package.net.javacrumbs.jsonunit.core.internal.Diff;
+import shaded_package.net.javacrumbs.jsonunit.core.internal.Options;
+import shaded_package.net.javacrumbs.jsonunit.core.listener.DifferenceContext;
+import shaded_package.net.javacrumbs.jsonunit.core.listener.DifferenceListener;
```
mobuchowski, line 32: Why `shaded_package`?

mobuchowski (commented):
Great PR, just a bit of nits.

.circleci/continue_config.yml
pawel-big-lebowski: There are some tests which I consider heavy enough so that they shouldn't be run with each commit. There are also tests, like databricks tests, which use external resources and should not be run each commit because of race condition possible when accessing external resources. Nightly builds can be a remedy to those problems. 

Currently, it is difficult to modify CI because there is a configuration flow which uses complex logic to create continuation workflow out of other workflow configuration files. CircleCI (up to my knowledge) does not accept any more workflows in main `config.yml` apart from configuration step + continuation workflow. 

Based on that, I decided to run the same workflow with extra param `nightly-run` = `active`. Then each each job, that is expected to run nightly and not on each commit, has to contain such running condition. 

To sum up: a nightly build is a regular build + bunch of jobs that normally don't get triggered bcz of such condition.

integration/spark/app/src/test/java/io/openlineage/spark/agent/util/RunEventVerifier.java
pawel-big-lebowski: looks sweet but docs says `This is a short-circuiting terminal operation.`, which will not collect full `diffDescription` in case of multiple files not being matched.

integration/spark/app/src/test/java/io/openlineage/spark/agent/util/RunEventVerifier.java
pawel-big-lebowski: yeah, adding proper source dependency. Thanks.

.circleci/continue_config.yml
tnazarew: to the last point, wouldn't it be better to have nightly build only run tests that are not run during regular builds? The whole point is that it would run on `main` code that has regular build already ran on it.

tnazarew (approved):
great as always!

Issue #2708, opened 2024-05-21
[BUG][SPARK] Wrong symlink location for Hive metastore tables


dolfinus:
### OpenLineage integration/client

Spark integration

### OpenLineage version

io.openlineage:openlineage-spark_2.12:1.14.0

### Technology and package versions

pyspark==3.4.3

### Problem details

If I'm using SparkSession with `enableHiveSupport()`. While creating table (see example below), I got event like this:
```json
{
  "eventType": "COMPLETE",
  "eventTime": "2024-05-21T09:37:57.768Z",
  "run": {
    "runId": "998c044c-2677-4856-93cc-02f34ca02cd3",
    "facets": {
      "nominalTime": null,
      "parent": {
        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.14.0/integration/spark",
        "_schemaURL": "https://openlineage.io/spec/facets/1-0-1/ParentRunFacet.json#/$defs/ParentRunFacet",
        "run": {
          "runId": "0983eeb7-9095-40e2-b664-f0de866a2baa"
        },
        "job": {
          "namespace": "default",
          "name": "pyspark_shell"
        }
      },
      "spark_properties": {
        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.14.0/integration/spark",
        "_schemaURL": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/RunFacet",
        "properties": {
          "spark.master": "local[*]",
          "spark.app.name": "pyspark-shell"
        }
      },
      "processing_engine": {
        "name": "spark",
        "version": "3.4.3",
        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.14.0/integration/spark",
        "_schemaURL": "https://openlineage.io/spec/facets/1-1-1/ProcessingEngineRunFacet.json#/$defs/ProcessingEngineRunFacet",
        "openlineageAdapterVersion": "1.14.0"
      },
      "environment-properties": {
        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.14.0/integration/spark",
        "_schemaURL": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/RunFacet",
        "environment-properties": {}
      }
    }
  },
  "job": {
    "namespace": "default",
    "name": "myapp.execute_create_table_command.mydb_db_some_table",
    "facets": {
      "documentation": null,
      "sourceCodeLocation": null,
[...]

PR #2761, opened 2024-06-06
Spark: Fix Hive metastore namespace

dolfinus:
### Problem

Reading keys like `spark.hadoop.*` from SparkConf returns the value only if some Hadoop configuration option was set via Spark config. But if option was set using Hadoop config files, e.g. `$SPARK_CONF_DIR/core-site.xml` or `$SPARK_CONF_DIR/hive-site.xml`, Spark config will not contain such keys at all. This may result to producing wrong Hive metastore URL.

Closes: #2708

### Solution

Instead of reading `spark.hadoop.*` keys from SparkConf, read them from Hadoop's Configuration class (without prefix).

Before:
изображение

After:
изображение

#### One-line summary:

Fix dataset namespace for case then Hive metastore URL is set using `$SPARK_CONF_DIR/hive-site.xml` file.

### Checklist

----

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/HiveTableRelationVisitor.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/HiveTableRelationVisitor.java
@@ -35,8 +35,13 @@ public class HiveTableRelationVisitor<D extends Dataset>
 
   @Override
   public List<D> apply(LogicalPlan x) {
+    if (!context.getSparkSession().isPresent()) {
+      return Collections.emptyList();
+    }
+
     HiveTableRelation hiveTable = (HiveTableRelation) x;
-    DatasetIdentifier datasetId = PathUtils.fromCatalogTable(hiveTable.tableMeta());
+    DatasetIdentifier datasetId =
+        PathUtils.fromCatalogTable(hiveTable.tableMeta(), context.getSparkSession().get());
     return Collections.singletonList(factory.getDataset(datasetId, x.schema()));
   }
 }
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
@@ -13,8 +13,10 @@ import java.util.Optional;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.spark.SparkConf;
+import org.apache.spark.SparkContext;
 import org.apache.spark.sql.SparkSession;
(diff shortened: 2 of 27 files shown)


pawel-big-lebowski (commented):
Is there any way to avoid static SparkContext field? More in comments.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -26,7 +28,7 @@ public class PathUtils {
 
   private static final String DEFAULT_SCHEME = "file";
   private static final String DEFAULT_SEPARATOR = "/";
-  private static Optional<SparkConf> sparkConf = Optional.empty();
+  private static Optional<SparkContext> sparkContext = Optional.empty();
```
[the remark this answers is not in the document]
dolfinus: Removed sparkContext and sparkConf from PathUtils state. Now methods tries to get active SparkSession on each call.

I could replace occurrences of `PathUtils.fromCatalogTable(table)` with `PathUtils.fromCatalogTable(table, openLineageContext.getSparkSession())` to pass session explicitly, but I'm not sure that should be returned from this method if Optional.empty() is passed. Returning `null` may lead to NullPointerExceptions, returning Optional require rewriting DatasetFactory to handle not add dataset to OL event.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/HiveTableRelationVisitor.java
```diff
@@ -35,8 +35,13 @@ public HiveTableRelationVisitor(@NonNull OpenLineageContext context, DatasetFact
 
   @Override
   public List<D> apply(LogicalPlan x) {
+    if (!context.getSparkSession().isPresent()) {
```
pawel-big-lebowski, line 38: Would it make sense to make `PathUtils.fromCatalogTable` get `OpenLineageContext` as parameter and return `Optional<DatasetIdentifier>`? This would allow having a single session existence check.

dolfinus: IMHO there is no difference between having method like this:
```java
DatasetIdentifier fromCatalogTable(CatalogTable, SparkSession) {
  ...
}

////

if (!table.isPresent() || !sparkSession.isPresent()) {
  return Collections.emptyList();
}

di = fromCatalogTable(table.get(), sparkSession.get());
return Collections.singletonList(outputDataset().getDataset(di))
```
and like this:
```java
Optional<DatasetIdentifier> fromCatalogTable(Optional<CatalogTable>, Optional<SparkSession>) {
   ...
}

////

di = fromCatalogTable(table, sparkSession);
if (di.isPresent()) {
  return Collections.singletonList(outputDataset().getDataset(di.get()));
} else {
  return Collections.emptyList();
}
```

Passing the entire OpenLineageContext instead of SparkSession is just the same as last method.

pawel-big-lebowski (approved):

PR #2630, opened 2024-04-22
fix: enable python generated classes deserialization using cattrs

mgorsk1:
### Problem

Closes: https://github.com/OpenLineage/OpenLineage/issues/2629

### Solution

Please describe your change as it relates to the problem, or bug fix, as well as any dependencies. If your change requires a schema change, please describe the schema modification(s) and whether it's a _backwards-incompatible_ or _backwards-compatible_ change, then select one of the following:

> **Note:** All schema changes require discussion. Please [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) for context.

- [ ] Your change modifies the [core](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json) OpenLineage model
- [ ] Your change modifies one or more OpenLineage [facets](https://github.com/OpenLineage/OpenLineage/tree/main/spec/facets)

If you're contributing a new integration, please specify the scope of the integration and how/where it has been tested (e.g., Apache Spark integration supports `S3` and `GCS` filesystem operations, tested with AWS EMR).

### Checklist

- [x] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [ ] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [ ] Your changes are accompanied by tests (_if relevant_)
- [ ] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] Your comment includes a one-liner for the changelog about the specific purpose of the change (_if necessary_)
- [ ] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_if relevant_)
- [ ] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----
SPDX-License-Identifier: Apache-2.0\
Copyright 2018-2023 contributors to the OpenLineage project

PR #2772, opened 2024-06-14
Python: enable deserialization to Python classes

JDarDagran:
### Problem

There's no out-of-the-box solution that would allow deserialize OL events into Python classes. As the continuation of the effort made to generate Python facets out of JSON schemas this PR aims to enable such serialization.

### Solution

1. Use old type annotations to enable Python 3.8 compatibility.
2. Add global variable that registers and holds keys under which facets should be stored in the structure.
3. Add logic to properly deserialize nested types into proper classes.

#### Few additional points
* attrs are limited to accept attributes names that Python can assign into variables. Therefore, e.g. for `my-facet.great` key in event payload this is going to be sanitized into `my_facet_great` and original named would be stored in attr.field(metadata={"name":"my-facet.great"})
* above lacks tests
* for unresolved types there are basic types created with additional attributes (e.g. `run`->`facets`->`custom` would be `RunFacet` with some additional attributes)
* one can register their own custom facets using `utils.register_facet_key` to enable richer deserialization
* after couple approaches Pydantic didn't turn out to solve issues with deserializing nested fields into proper classes. e.g. anything under `run`->`facets` goes into generic `RunFacet`. After some discussion with @mgorsk1 I still don't see clear way how Pydantic could be significantly better for deserialization task. https://github.com/OpenLineage/OpenLineage/pull/2630#discussion_r1605762463 might work but is complicated and does not leverage to use Pydantic instead

### Checklist

----


mgorsk1:
True, pydantic approach, cleaner as it is, hit the same issue attrs approach had and after working on it for quite a while I also did not found quick solution. However, what I think is the root origin of all these issues (both attrs/pydantic) is [OpenLineage.json](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json#L117) model. For example, the way events are modeled now, there is no connection between any of the facet (like ParentRunFacet, ErrorMessageFacet) and Run.facets object. That's why any approach to turn this model into attrs/pydantic/dataclasses models will require custom workarounds.

dolfinus:
What's next? Hope this PR can be reviewed soon.

mobuchowski:
@JDarDagran does this PR works with Airflow without any particular changes?

JDarDagran:
In what way it could work/not work with Airflow?
It sure needs extra `deserialize` - I don't think we want to install cattrs by default.

dolfinus:
Airflow integration does not use deserialization, IIRC.

Regarding `deserialize` extra, this looks great. Could you remove Draft mark, please?

PR #2773, opened 2024-06-14
Java: Improve namespace format for SQLServer

dolfinus:
### Problem

See #2762

### Solution

* Add `JdbcExtractor` interface which can extract components of JDBC url from a string input.
* Add generic implementation `GenericJdbcExtractor` for simple URL formats (like Postgres) and specialized `SqlServerJdbcExtractor` for Microsoft SQL Server.
* Split `JdbcUtils` to `JdbcUrlSanitizer` & `JdbcDatasetUtils`.
* Add a lot of tests for `JdbcDatasetUtils.getDatasetIdentifier` with different JDBC url formats (previously only Postgres URLs were tested). I've reused `PlanUtilsTest.provideJdbcUrls` for that, it looks like it was not used for some time.

#### One-line summary:

Improve namespace format for SQLServer.

### Checklist

- [X] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [X] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [X] Your changes are accompanied by tests (_if relevant_)
- [ ] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [X] Your comment includes a one-liner for the changelog about the specific purpose of the change (_not required for changes to tests, docs, or CI config_)
- [ ] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_if relevant_)
- [ ] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----
SPDX-License-Identifier: Apache-2.0\
Copyright 2018-2024 contributors to the OpenLineage project

PR #2767, opened 2024-06-12
Spark: fallback to spark.sql.warehouse.dir as table namespace

dolfinus:
### Problem

Before swapping dataset (name+namespace=location) and symlink (name=table, namespace=metastoreUri or warehouseDir) in #2718, make value of warehouseDir consistent.

See https://github.com/OpenLineage/OpenLineage/issues/2718#issuecomment-2152697956.

### Solution

* If there is no metastore, use `spark.sql.warehouse.dir` or `hive.metastore.warehouse.dir` as symlink namespace. But only if table location matches the warehouse.
* If there is no metastore, and table has a custom location, do not create symlink at all - other Spark sessions can access such a table only using location, as it is not stored in warehouse. Previously symlink was just the same as location, but in different format.
* If table is created within custom database, symlink still points to warehouse location. Previously it included path to a database dir.
* Rewrite `PathUtilsTests`

Before:
```yaml
datasets:
  - namespace: hdfs://namenode:8020
    name: /warehouse/with/metastore/database.db/table
    facets:
      symlinks:
        - namespace: hive://metastore:port
          name: database.table
          type: TABLE

  - namespace: hdfs://namenode:8020
    name: /warehouse/without/metastore/database.db/table
    facets:
      symlinks:
        - namespace: hdfs://namehode:8020/warehouse/without/metastore/database.db
          name: database.table
          type: TABLE

  - namespace: hdfs://namenode:8020
    name: /some/custom/location
    facets:
      symlinks:
        - namespace: hdfs://namenode:8020/some/custom/location
          name: database.table
          type: TABLE
```

After:
```yaml
datasets:
  - namespace: hdfs://namenode:8020
    name: /warehouse/with/metastore/database.db/table
    facets:
      symlinks:
        - namespace: hive://metastore:port
          name: database.table
          type: TABLE

  - namespace: hdfs://namenode:8020
    name: /warehouse/without/metastore/database.db/table
    facets:
      symlinks:
        - namespace: hdfs://namehode:8020/warehouse/without/metastore  # no /database.db in the path
          name: database.table
          type: TABLE

  - namespace: hdfs://namenode:8020
    name: /some/custom/location
    # no symlinks for locations not matching warehouse
```

#### One-line summary:

In case then metastore is not used, fallback to `spark.sql.warehouse.dir` or `hive.metastore.warehouse.dir` as table namespace, instead of duplicating table's location.

### Checklist

[...]

PR #2782, opened 2024-06-18
Spark, Flink: Fix S3 dataset names

dolfinus:
### Problem

S3 paths like `s3://bucket/some` are converted to DatasetIdentifier with `name=some`, but `s3://bucket/some/path` is converted to `/some/path`. Object storages like S3 are effectively key-value storages, keys are never start with `/`.

Also HadoopFileSystem can use different S3 implementations (`s3://`, `s3a://`, `s3n://`) which may lead to producing different datasets for the same location, and is also different from [naming convention](https://openlineage.io/docs/spec/naming/).

### Solution

* Introduce new `FilesystemDatasetUtils` class with methods `fromLocation(URI)` and `fromLocationAndName(URI,String)`.
* Introduce `FilesystemDatasetExtractor` interface and implementations for local filesystem (`/path` or `file:/path`), object storages (`scheme://bucket/path`) and generic filesystem (`scheme:/authority/path`).
* Replace `s3a://` and `s3n://` schemes with `s3://`
* Drop `DatasetIdentifierUtils`, and rewrite its tests for new classes.
* Update `PathUtils.fromPath()` and `PathUtils.fromURI()` to use new classes. As local filesystem extractor is specially designed to handle paths without any schema, `.fromPath(Path, String)` and `.fromURI(URI,String)` methods are deleted.

#### One-line summary:

Drop leading slash from object storage dataset name. Convert `s3a://` and `s3n://` schemes to `s3://`.

### Checklist

----


client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
```diff
@@ -5,88 +5,94 @@
...
+    return normalizedName;
+  }
 
-    boolean hasWindowsDrive = (WINDOWS && HAS_DRIVE_LETTER_SPECIFIER.matcher(uri.getPath()).find());
-
-    int startPositionWithoutWindowsDrive = 0;
-    if (hasWindowsDrive) {
-      startPositionWithoutWindowsDrive = (uri.getPath().charAt(0) == SEPARATOR_CHAR ? 3 : 2);
+  private static String removeFirstSlashIfSingleSlashInString(String name) {
+    // But why?
```
dolfinus, line 88: Could someone give an explanation why `/some/path` should start with `/` but `some-path` should not? I can't find any reference for this.

pawel-big-lebowski: What happens when you remove this? 
I think the answer can be found in `PathUtilsTest`: 
```
 di = PathUtils.fromURI(new URI("s3://data-bucket/path"), FILE);
 assertThat(di.getName()).isEqualTo("path");
 assertThat(di.getNamespace()).isEqualTo("s3://data-bucket");
```

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
```diff
@@ -5,88 +5,94 @@
...
+    return new DatasetIdentifier(normalizedName, normalizedNamespace);
   }
 
-  private static String removeLastSlash(String name) {
-    if (!name.isEmpty() && name.charAt(name.length() - 1) == File.separatorChar) {
-      return name.substring(0, name.length() - 1);
+  /* Convert namespace from URI to normalized string.
+   * Final slashes are removed, `/path` and `file:///path` are converted to `file:/path`
+   */
+  public static String normalizeNamespace(URI namespaceUri, String defaultScheme) {
```
pawel-big-lebowski, line 56: Any reason for making this method public ?

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
pawel-big-lebowski: @dolfinus pls let me know if you want to do something about it within this PR?

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
dolfinus: It is required by #2767. But I can make it private in this PR, and convert to public in #2767.

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
dolfinus: Let me rephrase my question.

Currently path like `s3://bucket/path` is converted to `namespace=s3://bucket, name=path`, but `s3://bucket/some/path` is converted to `namespace=s3://bucket, name=/some/path`. Is this required?

dolfinus: > @dolfinus pls let me know if you want to do something about it within this PR?

Not now, probably in a separated PR.

kacpermuda: > Currently path like s3://bucket/path is converted to namespace=s3://bucket, name=path, but s3://bucket/some/path is converted to namespace=s3://bucket, name=/some/path. Is this required?

I think the leading slash should always be removed, unless it's the only character in name (f.e. `s3://bucket/` should be converted to namespace=s3://bucket and name=/)

dolfinus: > I think the leading slash should always be removed

It is removed, but in some cases starting slash is removed too

dolfinus: Is this applied to s3 only?

kacpermuda: Yes, that's what i meant - the starting / leading slash, the one at the beginning. Not sure why it's NOT removed sometimes, imho it should always be removed.

pawel-big-lebowski:
```
 DatasetIdentifier di = PathUtils.fromURI(new URI("file:///home/test"), null);
    assertThat(di.getName()).isEqualTo("/home/test");
    assertThat(di.getNamespace()).isEqualTo(FILE);
```
If a name is path, then it makes sense for path to start with slash. The same applies for HDFS location:
```
di = PathUtils.fromPath(new Path("hdfs://namenode:8020/home/test"));
    assertThat(di.getName()).isEqualTo(HOME_TEST);
    assertThat(di.getNamespace()).isEqualTo("hdfs://namenode:8020");
```

I think the problem is that single method is used for multiple contexts (same as it used to be within jdbc url parsing). The logic may have sense for local files and hdfs, but makes less sense for object storages. Unfortunately, everything if implemented currently within the same code-flow.

dolfinus: Do you mean that there should be multiple different classes implementing dataset namespace+name normalization, instead of one `DatasetIdentifierUtils`?

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
pawel-big-lebowski: it's ok, no need to do that. thanks for answering.

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifierUtils.java
mobuchowski: I think yes - per-scheme one would be great? Similar to your contribution in https://github.com/OpenLineage/OpenLineage/pull/2773 :)

dolfinus: Hm, ok, I'll try to implement this

dolfinus: Done. Updated PR title and description to track the changes.

pawel-big-lebowski (approved):
Thank you @dolfinus for your efforts in improving quality of the OpenLineage codebase. This PR brings a huge improvement and makes existing logic a lot clearer. 
This will let the community implement future changes in dataset naming logic with a limited risk of breaking what's been implemented so far.

pawel-big-lebowski:
some failing integration tests:
```
SparkContainerIntegrationTest > testPysparkSQLOverwriteDirHiveTest() FAILED
    java.lang.AssertionError at MockServerClient.java:864

SparkReadWriteIntegTest > testCreateDataSourceTableAsSelect(Path, SparkSession) FAILED
```

dolfinus:
> SparkReadWriteIntegTest > testCreateDataSourceTableAsSelect(Path, SparkSession) FAILED

It is failed at first runs because of some internal exception raised by Spark (according to stack trace), but next run was successful.

> SparkContainerIntegrationTest > testPysparkSQLOverwriteDirHiveTest() FAILED

For some reason, running the individual test does not show any errors, but running the entire `SparkContainerIntegrationTest` suite sometimes leads to such error. Trying to investigate.

dolfinus:
Failing test `testPysparkSQLOverwriteDirHiveTest` is correlated with this message in logs:
```
ERROR io.openlineage.spark.agent.lifecycle.ContextFactory - Query execution is null: can't emit event for executionId 1
```
https://github.com/OpenLineage/OpenLineage/blob/ca5142cbf0715a995352a9b959725bd9f3264f9a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/ContextFactory.java#L71-L75

Spark sometimes does not return `queryExecution` information for `START` event for a job with `lifecycleStateChange: OVERWRITE`, but always return it for `COMPLETE` event. Test expects both `START` and `COMPLETE` events to be emitted, and thus failing.

Changes in `InsertIntoHiveDirVisitor` does not seem related to this error, as this class is used after `queryExecution` nullability check. Is it possible to mark some test to retry if FAILED?

dolfinus:
https://github.com/apache/spark/blob/ce5ddad990373636e94071e7cef2f31021add07b/sql/core/src/main/scala/org/apache/spark/sql/execution/SQLExecution.scala#L40-L44

Probably failing `testPysparkSQLOverwriteDirHiveTest` and `testCreateDataSourceTableAsSelect` are both related to updates of `ConcurrentHashMap` from the code above.

dolfinus:
Failing MavOS CI job is preventing test from being run

mobuchowski:
Let's run the tests after the blackout and merge it then.

mgorsk1:
hey @dolfinus they seem to be still inconsistencies re s3/s3a/s3n in different parts of OL events: 

image

dolfinus:
Please create a dedicated issue for that, with more details regarding this example

PR #2792, opened 2024-06-21
CI: release: move to OSSRH token

mobuchowski:
Maven central did a change: https://community.sonatype.com/t/401-content-access-is-protected-by-token-authentication-failure-while-performing-maven-release/12741/6

This makes release use generated token from Sonatype Nexus.


merobi-hub (approved):
Thanks!

PR #2822, opened 2024-07-03
[SPARK] configurable test with docker image provided

pawel-big-lebowski:
Extend configurable integration test feature to be able to get Docker image name as a a name. 

Example config entries:]
```
docker:
  image: "apache/spark:3.3.3-scala2.12-java11-python3-ubuntu"
  sparkSubmit: /opt/spark/bin/spark-submit
  waitForLogMessage: ".*ShutdownHookManager: Shutdown hook called.*"
```

Related docs PR: https://github.com/OpenLineage/docs/pull/344/files

### DIFF
--- a/.circleci/workflows/openlineage-spark.yml
+++ b/.circleci/workflows/openlineage-spark.yml
@@ -46,6 +46,7 @@ workflows:
           requires:
             - approval-integration-spark
       - configurable-integration-test-spark:
+          context: integration-tests
       - workflow_complete:
           requires:
             - integration-test-integration-spark-scala-2_12
--- a/integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableTestConfig.java
+++ b/integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableTestConfig.java
@@ -3,7 +3,9 @@
 /* SPDX-License-Identifier: Apache-2.0
 */
 
-package io.openlineage.spark.agent.util;
+package io.openlineage.spark.agent;
+
+import static io.openlineage.spark.agent.SparkContainerProperties.SPARK_DOCKER_IMAGE;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -14,11 +16,11 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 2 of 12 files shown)


.circleci/workflows/openlineage-spark.yml
```diff
@@ -46,6 +46,7 @@ workflows:
           requires:
             - approval-integration-spark
       - configurable-integration-test-spark:
+          context: integration-tests
```
pawel-big-lebowski, line 49: otherwise we end up with: https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/11214/workflows/63ffa876-b03a-4da0-9b83-d7a2d6999545

integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableTestConfig.java
```diff
@@ -29,17 +31,22 @@
...
 import org.testcontainers.shaded.org.yaml.snakeyaml.constructor.Constructor;
+import org.testcontainers.utility.DockerImageName;
 
 @NoArgsConstructor
 @Getter
 @Setter
 @Slf4j
 public class ConfigurableTestConfig {
+
+  public static final Path IVY_DIR = Paths.get("/usr/lib/openlineage/integration/spark/cli/.ivy2");
```
mobuchowski, line 45: Why do we need that?

integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableTestConfig.java
```diff
@@ -83,31 +100,22 @@ public Path generateRunDir(Path testDir) {
...
-
-    perms.add(PosixFilePermission.OTHERS_READ);
-    perms.add(PosixFilePermission.OTHERS_WRITE);
-    perms.add(PosixFilePermission.OTHERS_EXECUTE);
-
-    perms.add(PosixFilePermission.GROUP_READ);
-    perms.add(PosixFilePermission.GROUP_WRITE);
-    perms.add(PosixFilePermission.GROUP_EXECUTE);
-
+    Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rwxrwxrwx");
```
mobuchowski, line 105: 👍

mobuchowski (approved):
Looks good. However, would be nice to explain why do we need that Ivy dir.

integration/spark/app/src/test/java/io/openlineage/spark/agent/ConfigurableTestConfig.java
pawel-big-lebowski: So, `.ivy2` points to a location where packages added with `--packages` are stored. We wan't to have a Docker volume for this so that running multiple tests will lead do downloading package only once. 

However, when working with different Docker images, this is causing problems. A first container writes some xml files to `.ivy/cache` dir. Then a second container tries to read all the files from that directory and fails bcz of permission exception (this is happening on linux only where docker constrains are more strict). The solution to this is to clear `.ivy` cache directory `@BeforeAll`. 

The line you refer is just a constant variable used three times.

PR #2823, opened 2024-07-03
fix flaky test

pawel-big-lebowski:
Recent upgrade of supported Spark version introduced flaking test for Spark 3.4.3.
The issue is related to some Spark 3.4.3 throwing `ConcurrentModificationException` when run within `SparkAgentTestExtension`.


pawel-big-lebowski:
Screenshot 2024-07-03 at 12 58 13
two consecutive successful runs

tnazarew (approved):

Issue #2762, opened 2024-06-06
[FEATURE] Improve JDBC urls format for Oracle, MSSQL, Teradata


dolfinus:
### Motivation

Currently OpenLineage integration for Spark supports JDBC urls with following formats:
* `jdbc:type://host:port` - dataset has hame `schema.table`
* `jdbc:type://host:port/database` - database is extracted and used as prefix for dataset name, e.g. `database.schema.table`

Query params (everything after `?`) are trimmed off, as well as user & password parts of URL.
[events_mysql.json](https://github.com/user-attachments/files/15618079/events_mysql.json)

But some databases have different JDBC url formats:
* MSSQL - `jdbc:sqlserver://host:port;database=database` or `jdbc:sqlserver://host:port;databaseName=database` (see [documentation](https://learn.microsoft.com/en-us/sql/connect/jdbc/building-the-connection-url?view=sql-server-ver16))
* Teradata - `jdbc:teradata://host/DBS_PORT=port,DATABASE=database` (see [documentation](https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_2.html#URL_DATABASE))
* Oracle - `jdbc:oracle:thin:@host:port:sid` or `jdbc:oracle:thin:@//host:port/service` (not mentioning [TNS notation](https://docs.oracle.com/en/database/oracle/oracle-database/21/jjdbc/data-sources-and-URLs.html#GUID-6F729E4D-064B-4FD9-AE92-1BD44B8BE5EF), it is hard to parse).

In this case, OpenLineage integration returns datasets like:
* `namespace=sqlserver://localhost:1433;databaseName=database`, `name=schema.table` [events_mssql.json](https://github.com/user-attachments/files/15618131/events_mssql.json)
* `namespace=teradata://host/DBS_PORT=port,DATABASE=database`, `name=schema.table` (sorry, no example of events for now)
* `namespace=oracle:thin:@host:port:sid`, `name=schema.table` [events_oracle.json](https://github.com/user-attachments/files/15618140/events_oracle.json)

This leads to producing different dataset namespaces for the different JDBC URLs, although some of them are pointing to the same database, just in different connection formats.
More than that, because other connection params are not stripped away, they are also included to dataset namespace. So if user passed custom connection option, like timeout, this will lead to adding one more dataset with a new namespace.

### Description

Proposal - handle those URLs like:
[...]

PR #2826, opened 2024-07-03
Spark,Flink: Add TeradataJdbcExtractor

dolfinus:
### Problem

See #2762.

#### One-line summary:

Convert JDBC URLs like `jdbc:teradata/host/DBS_PORT=1024,DATABASE=somedb` to datasets with namespace `teradata://host:1024` and name `somedb.table`.

### Checklist

----

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/utils/jdbc/TeradataJdbcExtractor.java
+++ b/client/java/src/main/java/io/openlineage/client/utils/jdbc/TeradataJdbcExtractor.java
@@ -1,0 +1,61 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.utils.jdbc;
+
+import java.net.URISyntaxException;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+
+public class TeradataJdbcExtractor implements JdbcExtractor {
+  // https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_2.html#URL_DATABASE
+
+  private static final String SCHEME = "teradata";
+  private static final String PORT_PROPERTY = "DBS_PORT";
+  private static final String DEFAULT_PORT = "1025";
+  private static final String DATABASE_PROPERTY = "DATABASE";
+
+  private static final Pattern URL =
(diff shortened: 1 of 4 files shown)


client/java/src/main/java/io/openlineage/client/utils/jdbc/TeradataJdbcExtractor.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Optional;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+
+public class TeradataJdbcExtractor implements JdbcExtractor {
+  // https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_2.html#URL_DATABASE
+
+  private static final String SCHEME = "teradata";
```
pawel-big-lebowski, line 18: By the way, according to the standard: https://www.rfc-editor.org/rfc/rfc3986#section-3.1:
```
 Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names
```
which means one should do `jdbcUri.toLowerCase().startsWith(SCHEME.toLowerCase());`

pawel-big-lebowski (commented):
Thank you for working on this. Single comment added. Would you mind adding `teradata` to the docs https://openlineage.io/docs/spec/naming#dataset-naming as well?

client/java/src/main/java/io/openlineage/client/utils/jdbc/TeradataJdbcExtractor.java
dolfinus: This raises a question - should URLs be normalized (lowercase scheme, lowercase authority) before constructing dataset namespace? Currently passing URLs in different cases leads to producing different datasets.

Also? for most RDBMS object names are case insensitive, so users may pass both `SCHEMA.TABLE` and `schema.table` with the same result. Should these be normalized as well?

dolfinus: Moved this discussion to https://github.com/OpenLineage/OpenLineage/issues/2829.

Regarding behavior of `isDefined` method, I'd prefer to implement this in a separated PR, because currently URLs like `JDBC:TERADATA://...` are not properly handled by `JdbcDatasetUtils` - it trims prefix `jdbc:`, but not `JDBC:`

pawel-big-lebowski (approved):

PR #2863, opened 2024-07-19
Spark: Make Delta dataset symlink consistent with non-Delta tables

dolfinus:
### Problem

Currently DeltaHandler has its own logic for generating Dataset symlink, and it does this by stripping the last component from table location. There are 2 issues with that:
* Table metadata can be stored in the metastore, like Hive, but symlink does not reflect that.
* Table could have custom location, which may produce symlink which has no meaning.

### Solution

Reuse existing logic from `PathUtils.fromCatalogTable`.

#### One-line summary:

Make dataset symlinks for Delta and non-Delta tables consistent.

### Checklist

----

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
@@ -9,30 +9,23 @@ import io.openlineage.client.OpenLineage;
 import io.openlineage.client.utils.DatasetIdentifier;
 import io.openlineage.spark.agent.util.PathUtils;
 import io.openlineage.spark.api.OpenLineageContext;
-import java.io.File;
-import java.net.URI;
-import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.Path;
 import org.apache.spark.sql.SparkSession;
-import org.apache.spark.sql.catalyst.TableIdentifier;
+import org.apache.spark.sql.catalyst.catalog.CatalogTable;
 import org.apache.spark.sql.connector.catalog.Identifier;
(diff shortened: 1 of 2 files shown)


dolfinus:
The only piece of code left which generates symlink namespace by stripping up last part of table's location is [AbstractDatabricksHandler](https://github.com/OpenLineage/OpenLineage/blob/caecf5f39a4b2bb25513a3a0ef7a5346b4dd4a0d/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java). It just copies the previous implementation of `DeltaHandler`, but I have no Databricks account to check whether the same change could be applied here as well.

pawel-big-lebowski:
> The only piece of code left which generates symlink namespace by stripping up last part of table's location is [AbstractDatabricksHandler](https://github.com/OpenLineage/OpenLineage/blob/caecf5f39a4b2bb25513a3a0ef7a5346b4dd4a0d/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java). It just copies the previous implementation of `DeltaHandler`, but I have no Databricks account to check whether the same change could be applied here as well.

In case you wanted to verify behaviour for Databricks, there is a databricks integration test: https://github.com/OpenLineage/OpenLineage/blob/aa6ba51c5019609b6db160eac1986b0607219232/integration/spark/app/src/test/java/io/openlineage/spark/agent/DatabricksIntegrationTest.java#L44
It can be run from circle ci when extra parameter `databricks-test`=`enabled` is added. I can start it when needed. I don' think it's convenient for development purpose but should be sufficient to guarantee the change did not break anything.

pawel-big-lebowski (approved):

pawel-big-lebowski (approved):

dolfinus:
> It can be run from circle ci when extra parameter databricks-test=enabled is added. I can start it when needed. I don' think it's convenient for development purpose but should be sufficient to guarantee the change did not break anything.

I currently don't have information about the result type of `com.databricks.sql.transaction.tahoe.catalog.DeltaCatalog.loadTable(identifier)` call. For open source Delta at least I have sources to get possible types, and I can extract `catalogTable` object from its attributes. For Databricks I don't have any information about that.

PR #2890, opened 2024-07-27
Proxy config

jonathanlbt1:
Adding configuration to run openlineage-proxy directly from within its folder with docker-compose.yml file. 

I also edited the README file to add a description of how to use it. It will make it a lot quicker for anybody to have a go at it and see the application running.


proxy/backend/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN ./gradlew --no-daemon shadowJar
 FROM eclipse-temurin:11
 WORKDIR /usr/src/app
 COPY --from=build /usr/src/app/build/libs/openlineage-proxy-*.jar /usr/src/app
-COPY proxy.dev.yml proxy.dev.yml
```
pawel-big-lebowski, line 18: Can we keep it `proxy.dev.yml` to indicate it is an example setup to be run locally?

proxy/backend/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN ./gradlew --no-daemon shadowJar
...
+COPY proxy.yml proxy.yml
 COPY docker/entrypoint.sh entrypoint.sh
-EXPOSE 5000 5001
+EXPOSE 6000 6001
```
pawel-big-lebowski, line 20: any particular reason for this change?

pawel-big-lebowski (commented):
Thank you for refreshing proxy setup. 
It's been a while since modifying this part. 
I left few minor comments. Let me know what do you think.

proxy/backend/Dockerfile
jonathanlbt1: Sure thing! I'll include the changes in this file but I'll leave them commented with some instructions so people know how to proceed.

pawel-big-lebowski (approved):

proxy/backend/Dockerfile
jonathanlbt1: Yes, Marquez is configured to work on port 5000, so I changed this port so one can use Marquez to give it a try.

PR #2899, opened 2024-07-31
feature, refactor: Refactor Spark facets configuration

arturowczarek:
* Spark facets are configured using prefix spark.openlineage.facets.<facet name>.enabled property.

### Problem

Disabling one facet implicitly enables facets which were disabled by default (logical plan and unknown entitites). Enabling debug facet doesn't follow the convention used in Spark properties.

### Solution

Debug, logical plan and unknown entity facets have their own configuration classes with field `enabled`. They have to be explicitly enabled if someone needs them.

### Checklist

----


integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
```diff
@@ -53,8 +53,6 @@ public class ArgumentParser {
       "spark.openlineage.jobName.replaceDotWithUnderscore";
   private static final String SPARK_CONF_FACETS_DISABLED = "spark.openlineage.facets.disabled";
 
-  private static final String SPARK_CONF_DEBUG_FACET = "spark.openlineage.debugFacet";
```
arturowczarek, line 56: We replace `spark.openlineage.debugFacet` `String` property with `spark.openlineage.facets.debug.enabled` boolean property

d-m-h: Whilst I am all for unifying how we configure facets, there are several things that are missing for me:

1. Deprecation notices before removal
2. Retaining backwards compatibility whilst the deprecation notice is still in place.
3. Documentation changea about these changes

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkOpenLineageConfig.java
```diff
@@ -26,51 +26,53 @@
...
     this.job = job;
+    this.facetsConfig = facets;
   }
 
   @Override
-  public FacetsConfig getFacetsConfig() {
+  public SparkFacetsConfig getFacetsConfig() {
     if (facetsConfig == null) {
-      facetsConfig = new FacetsConfig();
+      // the configuration creates the default values in the getters, when they are accessed for the
```
arturowczarek, line 73: I spent some time trying to understand where the facets config came from when I deserialised the configuration without these properties and had nulls there, but when accesses I saw values. I continue the same pattern for the facets configuration class.
We are ok with the nulls there. When someone accesses them, the values are going to be created with some default values.

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
arturowczarek: Ad 3. Documentation will be changed
Ad 2 and 3. I'm not sure if maintaining backward compatibility and adding deprecation warnings is worth it. For `spark_unknown` and `spark.logicalPlan` we can't maintain it, because this pull request changes the how they are activated. We could do something about `debug` facet but is it really worth it? It would complicate the logic quite significantly (parsing the field, but hiding it; using it when present; deciding what takes precedence; how to handle merges; writing extra tests). If I add the warning, how long should it stay there? Maybe we could add warning (or even error) when we see the `spark.openlineage.debugFacet` configuration but don't use the value? But when should we remove it? These three facets are pretty advanced and used only when something is not working. Why don't we assume that when someone needs to debug the application they copy the value from the documentation anyway?

mobuchowski: I think for `debugFacet` specifically we could make the change outright - as the name suggests, it's the debug facet. There's no guarantee of stability of it and I really hope no one is using it in production.

Other setting

```
spark.openlineage.facets.disabled
```

we have to keep  - but the logic should not override 

```
spark.openlineage.facets.XXX.disabled
```

By that I mean 
```
spark.openlineage.facets.spark_unknown.disabled
spark.openlineage.facets.spark.logicalPlan.disabled
```
should be disabled, and lack of them in `spark.openlineage.facets.disabled` should not cause those two to be enabled unless someone adds the new-form config.

However, if someone added something additional to `spark.openlineage.facets.disabled` - it should be disabled.

So, basically, both configs, but new form having priority over the others, with sensible defaults - `spark_unknown`, `spark.logicalPlan` and `debug` disabled by default in the new form.

mobuchowski: As for documentation, deprecation warnings - definitely agree with Damien.

arturowczarek:
I close this one and implement https://github.com/OpenLineage/OpenLineage/pull/2912 instead.

PR #2908, opened 2024-08-02
Detect incorrect java version when compiling

pawel-big-lebowski:
Provide user with meaningful error when compiling Spark integration with incorrect Java version.


tnazarew (approved):
should be helpful, nice!

Issue #2907, opened 2024-08-01
[FEATURE] - Support custom headers for Python http client


kr-igor:
### Motivation

OpenLineage support `Authorization` http header, but some destinations may require a custom header. Specifically, Datadog requires `DD_API_KEY` to be set. It's not currently possible to set custom headers for the Python client.

### Description

Proposal is to support `OPENLINEAGE_HTTP_HEADERS: key=value;key1=value1` environment variable to set custom headers.
[ol.patch](https://github.com/user-attachments/files/16461991/ol.patch)

### Do you plan to make this contribution yourself?

- [X] I am interested in doing this work

kacpermuda:
Hey, thanks for the interest in OpenLineage. FYI, you do not need to create an issue if you have a change ready to submit! You can open a [pull request](https://github.com/OpenLineage/OpenLineage/pulls) immediately instead. Even if you are not sure if it's something that will end up being added to the code, it's always easier to discuss over actual implementation 😄

mobuchowski:
@kr-igor I've tried to find a solution to the problem more generically in https://github.com/OpenLineage/OpenLineage/issues/2916

mobuchowski:
@kr-igor Also - really thanks for starting the discussion and opening PR. You just hit a place where we have larger issue we need to solve :)

Issue #2667, opened 2024-05-06
[FEATURE] Add OPENLINEAGE_TRANSPORT variable to python client


kacpermuda:
Like in [Airflow](https://airflow.apache.org/docs/apache-airflow-providers-openlineage/stable/configurations-ref.html#transport) we should let the users setup any transport with json-string as env variable, not only .yml file.

PR #2910, opened 2024-08-02
Python client: support OPENLINEAGE_HTTP_HEADERS

kr-igor:
### Motivation

OpenLineage support `Authorization` http header, but some destinations may require a custom header. Specifically, Datadog requires `DD_API_KEY` to be set. It's not currently possible to set custom headers for the Python client.

### Description

Proposal is to support `OPENLINEAGE_HTTP_HEADERS: key=value;key1=value1` environment variable to set custom headers.

https://github.com/OpenLineage/OpenLineage/issues/2907

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -228,6 +228,13 @@ class OpenLineageClient:
         if endpoint is not None:
             config.endpoint = endpoint
 
+        headers_str = os.environ.get("OPENLINEAGE_HTTP_HEADERS")
+        if headers_str is not None:
+            for val in headers_str.split(","):
+                if "=" in val:
+                    key, value = val.split("=")
+                    config.headers[key] = value
+
         return HttpTransport(config)
 
     @staticmethod
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -230,7 +230,7 @@ class OpenLineageClient:
 
         headers_str = os.environ.get("OPENLINEAGE_HTTP_HEADERS")
         if headers_str is not None:
-            for val in headers_str.split(","):
+            for val in headers_str.split(";"):
                 if "=" in val:
                     key, value = val.split("=")
                     config.headers[key] = value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
   *Spark integration is always compiled with Java 17, while tests are running on both Java 8 and Java 17 according to the configuration.*
 * **Spark: Support preview release of Spark 4.0 .** [`#2854`](https://github.com/OpenLineage/OpenLineage/pull/2854) [@pawel-big-lebowski](https://github.com/pawel-big-lebowski)  
   *Include Spark 4.0 preview release in the integration tests.*
-
+* **Python client**: support custom http headers for http transport [`#2910`](https://github.com/OpenLineage/OpenLineage/pull/2910) [@kr-igor](https://github.com/kr-igor)
+`OPENLINEAGE_HTTP_HEADERS` environment variable can be used to set custom headers.
 
 ### Fixed
 * **Spark: fix issue with kafka source when saving with for each batch method** [`#2868`](https://github.com/OpenLineage/OpenLineage/pull/2868) [@imbruced](https://github.com/Imbruced)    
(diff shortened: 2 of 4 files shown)


harels:
thanks @kr-igor, changes LGTM! can you please add this addition to the changelog? https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md

Would be great to also follow up with a PR to update the docs here https://github.com/OpenLineage/docs/blob/main/docs/client/python.md

client/python/openlineage/client/client.py
```diff
@@ -228,6 +228,13 @@ def _http_transport_from_env_variables() -> HttpTransport:
         if endpoint is not None:
             config.endpoint = endpoint
 
+        headers_str = os.environ.get("OPENLINEAGE_HTTP_HEADERS")
+        if headers_str is not None:
+            for val in headers_str.split(","):
```
mobuchowski, line 233: I would use `;`, as the commit description suggests

client/python/openlineage/client/client.py
```diff
@@ -228,6 +228,13 @@ def _http_transport_from_env_variables() -> HttpTransport:
         if endpoint is not None:
             config.endpoint = endpoint
 
+        headers_str = os.environ.get("OPENLINEAGE_HTTP_HEADERS")
```
kacpermuda, line 231: I'm just wondering, if we are later translating this to a dict, is it necessary to introduce another logic here with some splitting and equal signs? Can't we just ask the user to provide json dict in `OPENLINEAGE_HTTP_HEADERS` ? 

Overall, maybe it would be better, instead of adding this feature, to simply implement [#2667](https://github.com/OpenLineage/OpenLineage/issues/2667) - let the user provide transport dict to openlineage client, and in there he could add some extra http headers. I just think that introducing another config variable, that is limited to http transport may not be the best way of handling it in the long run. Let me know your thoughts.

kr-igor:
> thanks @kr-igor, changes LGTM! can you please add this addition to the changelog? https://github.com/OpenLineage/OpenLineage/blob/main/CHANGELOG.md
> 
> Would be great to also follow up with a PR to update the docs here https://github.com/OpenLineage/docs/blob/main/docs/client/python.md

@harels I've updated the change log, please, review.

client/python/openlineage/client/client.py
kr-igor: Thanks for catching this! I've updated to `;`

client/python/openlineage/client/client.py
kr-igor: Thank you for the review! I think using a dictionary env variable format is less user-friendly than setting a single variable (way easier to make a mistake, you have to set multiple values etc). My goal was to make the onboarding to OpenLineage as simple as possible. 
I'd support both in the long run. Dictionary for consistency / advanced users and single-variable-per-value as a simpler alternative.

kacpermuda: I understand your perspective. I believe using a dictionary environment variable format can actually enhance user-friendliness, but maybe let's hear some more opinions on this - maybe I'm in a minority 😄 

Also correct me if I'm wrong, but if I tried to provide this custom http headers in a .yml file (and not via env variable) with your current implementation, I'd have to provide it as a dict anyway right? `HttpConfig.from_dict` simply reads the provided parameters but it does not perform the conversion from `key=value` to `{"key": "value"}`. So I think it's not working properly at the moment.

kacpermuda: Also, one more concern about this format `key=value;key1=value1`. I think both `;` and `=` are valid characters that can be included in HTTP header values (I could not find any confirmation stating otherwise), so providing headers in this way can be erroneous and hard to debug, when it breaks.

mobuchowski: @kr-igor can we do the one thing without the other? Add http headers to http transport, but not add env variable now, and fix it in https://github.com/OpenLineage/OpenLineage/issues/2916?

mobuchowski:
@kr-igor can we add the HTTP headers to the config, without adding env variable now - and fix it in https://github.com/OpenLineage/OpenLineage/issues/2916 ?

JDarDagran:
@kr-igor I'm closing this one, I'll post a PR for configuring Python client with env vars similar to #3094

Issue #2808, opened 2024-06-28
[BUG]  wrong table taken added to inputs from SQL queries incl WITH statements in openlineage-spark integration


sfeese:
### OpenLineage integration/client

Spark integration

### OpenLineage version

openlineage-spark_2.12:1:16:0

### Problem details

Openlineage event created by spark-integration does not contain correct input table when query contains multiple tables. For both queries below the input table will be `authors`.

Excerpt of produced OL event
````json
    "inputs": [
        {
            "namespace": "postgres://localhost:5432",
            "name": "demo.authors",
            "facets": {
                "dataSource": {
                    "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.16.0/integration/spark",
                    "_schemaURL": "https://openlineage.io/spec/facets/1-0-1/DatasourceDatasetFacet.json#/$defs/DatasourceDatasetFacet",
                    "name": "postgres://localhost:5432",
                    "uri": "postgres://localhost:5432"
                },
                "schema": {
                    "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.16.0/integration/spark",
                    "_schemaURL": "https://openlineage.io/spec/facets/1-1-1/SchemaDatasetFacet.json#/$defs/SchemaDatasetFacet",
                    "fields": [
                        {
                            "name": "author_id",
                            "type": "integer"
                        },
                        {
                            "name": "name",
                            "type": "string"
                        },
                        {
                            "name": "birth",
                            "type": "date"
                        },
                        {
                            "name": "death",
                            "type": "date"
                        }
                    ]
                }
            },
            "inputFacets": {}
        }
    ],
````

Query 1
```python
    query = """
        WITH ns AS (
            SELECT author_id AS id
            FROM books
            WHERE book_id <= 20
        )
        SELECT authors.*
        FROM authors
        INNER JOIN ns ON authors.author_id = ns.id
    """
```

Query 2
```python
    query = """
        WITH ns AS (
            SELECT author_id AS id
            FROM authors
            WHERE author_id <= 20
        )
        SELECT books.*
        FROM books
        INNER JOIN ns ON books.author_id = ns.id
    """
```

Pyspark code snippet
```python
    df = (
        spark.read.format("jdbc")
        .option("url", "jdbc:postgresql://localhost:5432/demo")
        .option("query", query)
        .option("user", "demo")
        .option("password", "demo")
        .option("driver", "org.postgresql.Driver")
        .load()
    )

[...]

PR #2918, opened 2024-08-07
Fix issue with only one table in inputs from sql query while reading from jdbc.

Imbruced:
### Problem
Closes: #2808

### Solution

Take more elements from array than the first one.

### Checklist

----


integration/spark/app/src/test/java/io/openlineage/spark/agent/InvalidTablesInJdbcInputsTest.java
```diff
@@ -0,0 +1,261 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.utility.DockerImageName;
+
+@Tag("integration-test")
+@Testcontainers
+@Slf4j
+class InvalidTablesInJdbcInputsTest {
```
pawel-big-lebowski, line 36: Does this test really verify invalid tables?

integration/spark/app/src/test/java/io/openlineage/spark/agent/InvalidTablesInJdbcInputsTest.java
```diff
@@ -0,0 +1,261 @@
...
+        handler.getEvents("test_postgres_query_when_multiple_tables_are_in_the_query").stream()
+            .filter(event -> !event.getInputs().isEmpty())
+            .collect(Collectors.toList());
+
+    assertThat(runEvents).hasSizeGreaterThanOrEqualTo(4);
+
+    runEvents.stream()
+        .map(OpenLineage.RunEvent::getInputs)
+        .forEach(
+            inputs -> {
```
pawel-big-lebowski, line 152: Are we extracting schemas in this case? If not, then it's at least worth writing a comment.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/JdbcSparkUtils.java
```diff
@@ -39,13 +39,26 @@ public static <D extends OpenLineage.Dataset> List<D> getDatasets(
     String jdbcUrl = relation.jdbcOptions().url();
     Properties jdbcProperties = relation.jdbcOptions().asConnectionProperties();
 
+    if (meta.columnLineage().isEmpty() && meta.inTables().size() > 1) {
```
pawel-big-lebowski, line 42: how come empty column level lineage makes here a difference? Could you add some comment?

integration/spark/app/src/test/java/io/openlineage/spark/agent/InvalidTablesInJdbcInputsTest.java
Imbruced: ++ not good name for test :)

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/JdbcSparkUtils.java
Imbruced: I ve tried to not make any regression. It might be that the condition is not necessary in this form.

integration/spark/app/src/test/java/io/openlineage/spark/agent/InvalidTablesInJdbcInputsTest.java
Imbruced: fixed

integration/spark/app/src/test/java/io/openlineage/spark/agent/InvalidTablesInJdbcInputsTest.java
Imbruced: its interesting how we can handle schema when we have more than one table, what we can do is to somehow parse the query and look for column references but I am not sure if its the case here and for simplicity is better to keep the schema to be empty StructType ?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/JdbcSparkUtils.java
Imbruced: I ve changed that a little in newest commit, the issue is with how we handle the schema when we have more than one table and we dont have column lineage. In that case we need for each column to set schema to empty StructType or parse plan (but I think thats what column lineage will at the end ll do)

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/JdbcSparkUtils.java
```diff
@@ -40,11 +40,20 @@ public static <D extends OpenLineage.Dataset> List<D> getDatasets(
...
+      return meta.inTables().stream()
+          .map(
+              dbtm -> {
+                DatasetIdentifier di =
+                    JdbcDatasetUtils.getDatasetIdentifier(
+                        jdbcUrl, dbtm.qualifiedName(), jdbcProperties);
+                return datasetFactory.getDataset(
+                    di.getName(),
+                    di.getNamespace(),
+                    numberOfTables == 1 ? schema : new StructType());
```
pawel-big-lebowski, line 54: Won't this result in schema facet with empty fields? If so, this is something that should be avoided.
 
If any other OL event emits dataset with schema facet, the schema will be attached to a dataset up until the moment a new schema facet is sent. Sending empty schema facet will clear existing information.

Imbruced: That example is only possible to happen when we have no column lineage + number of tables in the jdbc query is more than one. Maybe instead this solution we should take one step backward and add column level lineage for jdbc queries with with statement ? I don't see any better solution without stepping back than putting the empty Schema field. Do you have any ideas ?

pawel-big-lebowski: How about adding new method `io.openlineage.spark.api.DatasetFactory#getDataset(java.lang.String, java.lang.String)` without `StructType schema` parameter?

Imbruced: fixed

PR #2925, opened 2024-08-09
Python: Add CompositeTransport.

JDarDagran:
### Problem

Currently, to send OL event to multiple locations there is separate proxy component needed. This PR aims to reduce this need and let users send events using multiple transports in Python Client.

### Solution

Add `CompositeTransport` that can accept other transport configs to instantiate transports and use them to emit events.

#### One-line summary:

Add CompositeTransport.

### Checklist

----

### DIFF
--- a/client/python/openlineage/client/transport/composite.py
+++ b/client/python/openlineage/client/transport/composite.py
@@ -1,0 +1,81 @@
+# Copyright 2018-2024 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import logging
+from functools import cached_property
+from typing import TYPE_CHECKING, Any
+
+import attr
+from openlineage.client.transport.transport import Config, Transport
+from openlineage.client.utils import get_only_specified_fields
+
+if TYPE_CHECKING:
+    from openlineage.client.client import Event
+
+log = logging.getLogger(__name__)
+
+
+@attr.s
+class CompositeConfig(Config):
+    """
+    CompositeConfig is a configuration class for CompositeTransport.
+
+    Attributes:
+        transports:
+            A list of dictionaries, where each dictionary represents the configuration
+            for a child transport. Each dictionary should contain the necessary parameters
+            to initialize a specific transport instance.
+
+        continue_on_failure:
+            If set to True, the CompositeTransport will attempt to emit the event using
+            all configured transports, regardless of whether any previous transport
+            in the list failed to emit the event. If set to False, an error in one
+            transport will halt the emission process for subsequent transports.
+    """
+
+    transports: list[dict[str, Any]] = attr.ib()
+    continue_on_failure: bool = attr.ib(default=True)
+
+    @classmethod
+    def from_dict(cls, params: dict[str, Any]) -> CompositeConfig:
+        """Create a CompositeConfig object from a dictionary."""
+        if "transports" not in params:
+            msg = "composite `transports` not passed to CompositeConfig"
+            raise RuntimeError(msg)
+        return cls(**get_only_specified_fields(cls, params))
+
+
+class CompositeTransport(Transport):
+    """CompositeTransport is a transport class that emits events using multiple transports."""
+
+    kind = "composite"
+    config_class = CompositeConfig
+
+    def __init__(self, config: CompositeConfig) -> None:
+        """Initialize a CompositeTransport object."""
+        self.config = config
+
+    @cached_property
+    def transports(self) -> list[Transport]:
+        """Create and return a list of transports based on the config."""
+        from openlineage.client.transport import get_default_factory
+
+        transports = []
+        for transport_config in self.config.transports:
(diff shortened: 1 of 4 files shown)


client/python/tests/test_composite.py
```diff
@@ -0,0 +1,96 @@
...
+from openlineage.client.transport.composite import CompositeConfig, CompositeTransport
+from openlineage.client.transport.transport import Transport
+
+
+def test_composite_loads_full_config() -> None:
+    config = CompositeConfig.from_dict(
+        {
+            "type": "composite",
+            "transports": {
+                "kafka": {
```
mobuchowski, line 17: Can we move transport name to key inside transport, rather than just being a key tied to `CompositeTransport`?

client/python/openlineage/client/transport/composite.py
```diff
@@ -0,0 +1,62 @@
...
+if TYPE_CHECKING:
+    from openlineage.client.client import Event
+
+log = logging.getLogger(__name__)
+
+
+@attr.s
+class CompositeConfig(Config):
+    transports: dict[str, dict[str, Any]] = attr.ib()
+    continue_on_failure: bool = attr.ib(default=True)
```
kacpermuda, line 23: Shouldn't we set this to False by default, so that the user has to explicitly allow failures to pass silently ?

client/python/openlineage/client/transport/composite.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def emit(self, event: Event) -> None:
+        for transport_name, transport in self.transports.items():
+            try:
+                log.debug("Emitting event using %s", transport_name)
+                transport.emit(event)
+            except Exception as e:  # noqa: BLE001
+                msg = f"Transport {transport_name} failed to emit event: {e}"
+                if self.config.continue_on_failure:
+                    log.debug(msg)
```
kacpermuda, line 60: Imho, we may not want to raise an error but we probably want to see it in the logs, without debug logs turned on. Maybe we should provide some more details on debug level, but i would consider either also logging on a warning / exception level or replace the debug log entirely.

client/python/openlineage/client/transport/composite.py
```diff
@@ -0,0 +1,62 @@
...
+            transports[transport_name] = get_default_factory().create(transport_config)
+        return transports
+
+    def emit(self, event: Event) -> None:
+        for transport_name, transport in self.transports.items():
+            try:
+                log.debug("Emitting event using %s", transport_name)
+                transport.emit(event)
+            except Exception as e:  # noqa: BLE001
+                msg = f"Transport {transport_name} failed to emit event: {e}"
```
dolfinus, line 58: Please use % syntax for message strings instead of f-strings

client/python/openlineage/client/transport/composite.py
JDarDagran: I would say it's better than other transports emit successfully even if some fail. Having OL event delivered is imho more desired than getting info in logs / throwing exception.

client/python/openlineage/client/transport/composite.py
JDarDagran: How would I do that both for log and exception? One requires `"msg %s" % var`, another `"msg %s", var`. f-string just seemed easier solution here.

client/python/tests/test_composite.py
JDarDagran: I know it's for easier configuration through env vars but maybe let's add it in the coming PR related to this? (I'm assuming this PR merged -> next PR with env vars).

client/python/openlineage/client/transport/composite.py
mobuchowski: Yeah, I'd add warning level. This - CompositeTransport - is a clear case where user had to configure something themselves to get that result.

client/python/openlineage/client/transport/composite.py
dolfinus: Instruments like Sentry use message as a key for grouping errors. Using f-string leads to generating unique messages, and grouping will not be possible.
Also logging methods are lazy, and if debug log level is disabled, then message will not be templated at all.

JDarDagran: Got it, I hope it works now.

client/python/openlineage/client/transport/composite.py
JDarDagran: Changed to warning level.

client/python/openlineage/client/transport/composite.py
kacpermuda: I understand your point and partially agree - sending OL events should be the the most desired action. I'm still not convinced that we should do this by default, but maybe mentioning that behaviour in the docs is enough, so that the users are aware of it and can adjust this on their own. Feel free to resolve this one if that's okay.

JDarDagran: Added this to docstring. Also I checked how it renders to generated API reference:
image

PR #2949, opened 2024-08-21
fix, ci: non-committers should be able to run build jobs

mobuchowski:
https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/12077/workflows/cee92a54-1335-461d-b002-bcec346a6f79

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ setup: true
 # the continuation orb is required to continue a pipeline based on
 # the path of an updated fileset
 orbs:
-  continuation: circleci/continuation@0.2.0
+  continuation: circleci/continuation@1.0.0
 
 # optional parameter when triggering to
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -138,42 +138,51 @@ jobs:
...
+      - when:
+          condition:
+            not:
+              or:
+                - equal: [ 'main',  << pipeline.git.branch >> ]
+                - << pipeline.git.tag >>
+          steps:
+            - continuation/continue:
+                configuration_path: complete_config.yml
+                parameters: '{ "build-context": "pr" }'
```
pawel-big-lebowski, line 183: Assuming there is no easy way to assign parameter value based on condition and that's why different conditions are necessary.

pawel-big-lebowski (approved):
Seems to be a valid solution to the problem encountered.

PR #2961, opened 2024-08-22
CI and deploy steps for docs

pawel-big-lebowski:
*Problem*: we want to have a single repo where developers push code and documentation changes within a single PR. 

As a result of the above, we moved `Openlineage/docs` repo into the main repository. In future we would like to allow versioning of the whole website but we don't want to store whole history in the main repo. Because this will get too big, because this should contain javadoc and we don't want to store javadoc within the main repo. 

As a solution to above, we decided to store  current version of the docs in the main repo while having a separate `OpenLineage-site` repo for the versioned documentation. Website will be served as GitHub page of `Openlineage-site` repo. In future, a separate GitHub action will create a new version of the site within `openlineage-site` while keeping the all the other versions. This will be done automatically. Users should only push changes to the main repo. Synchronisation with additional repo should be hidden.  This PR adds a logic to synchronise current state to additional repo with a javadoc and openapi page being generated and pushed into the repo. 

Moreover, documentation is also used to server facet and Openlineage spec definitions. Whenever, a version of a facet is bumped `1.0.5` -> `1.0.6`, a process should create a new copy of facet definition and store it under `/1.0.6/` path so that it is reachable via URL.  This is done in precommit and verified on CI site. Precommit generates a copy of facets changes and puts it under proper location. Additionally, a CI step verifies if the content is generated.

The PR in its current shape is pushing changes into my fork `pawel-big-lebowski/openlineage-site`. This is happening (bcz I couldn't obtain fine-grained personal access token to `OpenLineage/openlineage-site`. Anyway, this is good for testing. 

To verify a correctness of the code created: 
 * Visit [`pawel-big-lebowski.github.io/openlineage-site`](https://pawel-big-lebowski.github.io/openlineage-site/)
 * Verify openapi is generated  -> https://pawel-big-lebowski.github.io/openlineage-site/apidocs/openapi/
 * Verify javadoc is generated -> https://pawel-big-lebowski.github.io/openlineage-site/apidocs/javadoc/
 * Modify a spec (bump facet version or version in OpenLineage.json) to see recommit works as expected
 * See https://github.com/pawel-big-lebowski/OpenLineage-site to view last commits. For example this is an initial commit -> https://github.com/pawel-big-lebowski/openlineage-site/commit/d818a1333f4902d770d5d796ac76d9f6a7b98cc0


.github/workflows/website-snapshot.yml
```diff
@@ -0,0 +1,86 @@
...
+          java-version: '17'
+      - uses: shogo82148/actions-setup-perl@v1
+        with:
+          perl-version: "5.38"
+      - name: Regenerate Javadoc to website
+        run: |
+          cd source/client/java
+          ./gradlew javadoc
+          mkdir -p ../../website/static/apidocs/javadoc
+          cp -r build/docs/javadoc ../../website/static/apidocs
```
JDarDagran, line 42: I can see you've created script to regenerate docs. Why not use it here?

.circleci/continue_config.yml
```diff
@@ -1182,6 +1165,16 @@ jobs:
...
 
+  verify-spec-snapshot:
+    working_directory: ~/openlineage/spec
+    docker:
+      - image: cimg/python:3.8
+    steps:
+      - *checkout_project_root
+      - run:
+          name: Verify spec snapshot
+          command: ./verify-spec-snapshot-up-to-date.sh
```
JDarDagran, line 1176: There's already pre-commit hook for that. What additional verification does this job do?

.github/workflows/website-snapshot.yml
```diff
@@ -0,0 +1,86 @@
...
+          for i in $(ls -d ./facets/* | sort); do cp $i/*.json ${LATEST_VERSION}/facets; done;
+
+          echo "LATEST_VERSION=${LATEST_VERSION}" >> "$GITHUB_ENV"
+
+          pushd $LATEST_VERSION
+          ln -sf ../OpenLineage.yml .
+          popd
+      - name: Generate redoc
+        run: |
+          npm install redoc-cli
```
JDarDagran, line 67: While we change the code we could update deprecated `redoc-cli` to `@redocly/cli` package.

.github/workflows/website-snapshot.yml
```diff
@@ -0,0 +1,86 @@
...
+          yarn run redoc-cli build -o source/website/static/apidocs/openapi/index.html source/website/static/spec/$LATEST_VERSION/OpenLineage.yml --title "OpenLineage API Docs"
+      - name: Clean temp files
```
JDarDagran, line 69: There's also Python API reference generation: https://github.com/OpenLineage/OpenLineage/blob/main/client/python/docs/release-api-reference.sh

I think we should add it as well.

.github/workflows/website-snapshot.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Push target repo
+        run: |
+          rm -r target/* 
+          cp -r source/website/* target
+          cd target
+          git config user.name "GitHub Action Website Snapshot"
+          git config user.email "<>"
+          git add .
+          if ! git diff-index --quiet HEAD; then
+            git commit -m "refreshing website content from main repo"
```
JDarDagran, line 84: Could this message be more informative? Maybe commit hash or number of PR that triggered the workflow?

JDarDagran (changes requested):

website/docusaurus.config.js
```diff
@@ -26,8 +26,8 @@ const linksSocial = [
 const config = {
   title: 'OpenLineage',
   tagline: 'OpenLineage',
-  url: 'https://openlineage.github.io',
-  baseUrl: '/',
+  url: 'https://pawel-big-lebowski.github.io', // TODO: replace with openlineage.io in the next step
+  baseUrl: '/openlineage-site', // TODO: replace with openlineage.io in the next PR
```
JDarDagran, line 30: I don't know much about configuration here but main page https://pawel-big-lebowski.github.io/openlineage-site/ does not work. Can this be adjusted so we can see all works correctly?

JDarDagran: According to https://docusaurus.io/docs/api/docusaurus-config#baseUrl `Always has both leading and trailing slash.`: probably it should be `/openlineage-site/`

JDarDagran: or: `baseUrl: process.env.NODE_ENV === 'production' ? '/openlineage-site/' : '/',` and run `NODE_ENV=production yarn build` in `openlineage-site` repo.

.github/workflows/website-snapshot.yml
pawel-big-lebowski: The script was added when javadoc was generated with pre-commit which is no longer used. I will remove the script instead. Thanks for spotting this.

.circleci/continue_config.yml
pawel-big-lebowski: you're right. precommit script is no longer run in docker.

website/docusaurus.config.js
pawel-big-lebowski: I think the main page issue can be fixed with this -> https://github.com/OpenLineage/OpenLineage/pull/2961/files#diff-ca969f89131bf8aecb9ccf3a81ba10fea069104412c9733943fcbb8a0ee8275aR43
which will work both when server from domain without baseUrl and domain with `openlineage-site` as baseUrl. 

Production env is not necessary I believe. Once we switch prod into `openlineage.io`, `baseUril` will be removed and this should work the same way on localhost and on prod.

.github/workflows/website-snapshot.yml
pawel-big-lebowski: added committed like this -> https://github.com/pawel-big-lebowski/openlineage-site/commit/19beda359d2ad31f14ed701eb1dcf540ae5890d0

.github/workflows/website-snapshot.yml
pawel-big-lebowski: good point Jakub. I fixed this.

.github/workflows/website-snapshot.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Push target repo
+        run: |
+          rm -r target/* 
+          cp -r source/website/* target
+          cd target
+          git config user.name "GitHub Action Website Snapshot"
+          git config user.email "<>"
+          git add .
+          if ! git diff-index --quiet HEAD; then
+            git commit -m "Refreshing website content from main repo with commit ${GITHUB_SHA}"
```
JDarDagran, line 93: I guess this might be more readable?
```suggestion
            git commit -m "Refreshing website content from main repo.
    
            Source commit: https://github.com/${GITHUB_REPOSITORY}/commit/${GITHUB_SHA}"
```

website/docusaurus.config.js
JDarDagran: This changed works for me locally and in your forked GH Pages.

JDarDagran (approved):
Outstanding work @pawel-big-lebowski  🚀 
Let's make neccessary changes once `OpenLineage/openlineage-site` is enabled to publish to GitHub Pages.

PR #2968, opened 2024-08-23
ci: automatically close sonatype staging repo

mobuchowski:
This should automate closing of repository.

Next step - after this succeeds - is using `closeAndReleaseSonatypeStagingRepository` instead of just closing it - the author plugins recommend running just close when adding it for the first time.


pawel-big-lebowski (approved):
Option seems to do exactly what's needed. Not sure why it isn't already present. Let's give it a try during next release.

PR #2995, opened 2024-08-27
ci: exclude dependabot from static analysis

mobuchowski:
Dependabot can't access secrets, so the workflow always fails.


kacpermuda (approved):

JDarDagran (approved):

pawel-big-lebowski:
I am afraid it doesn't work still -> https://github.com/OpenLineage/OpenLineage/pull/2956
@mobuchowski  How do you know this is a proper GitHub actor?

PR #2998, opened 2024-08-28
fix: release sql should depend on compile- jobs, rebuild the shadowJar to depend on native libs

mobuchowski:


pawel-big-lebowski (approved):
could you comment it with TODO to point that this should be done differently?

mobuchowski:
@pawel-big-lebowski done.

PR #2999, opened 2024-08-28
fix: Support IDENTIFIER function instead of treating it like table name

kacpermuda:
### Problem

IDENTIFIER keyword, valid in certain SQL dialects (snowflake, databricks, mssql), is misplaced as table name in some cases (e.g. `SELECT col1 FROM IDENTIFIER(table)` would produce table named `IDENTIFIER` and not `table`)

### Solution

Whenever we have full information about the Table, we can try to repair the lineage. This PR adds support for this identifier in SELECT, MERGE, UPDATE, DELETE statements. For now only static identifiers are supported (e.g. IDENTIFIER("table_name") => table_name), when variable is used, we remove this table from lineage to avoid wrong lineage (with IDENTIFIER table).

### Future work
1. We don't always (not in all statements) have the full information about the table, sometimes we just receive the already parsed ObjectName - I will raise a separate PR to the sqlparser-rs source repo and try to add this feature there.
2. IDENTIFIER also support variables, but we currently do not have Visitor for `Statement::SetVariable`. As a followup PR i may try to add support for variables and keep them in our context so that we can support them f.e. in `IDENTIFIER`.

#### One-line summary:
Support IDENTIFIER function in some statements instead of treating it like table name

### Checklist

----

### DIFF
--- a/integration/sql/impl/src/visitor.rs
+++ b/integration/sql/impl/src/visitor.rs
@@ -8,9 +8,9 @@ use anyhow::{anyhow, Result};
 use sqlparser::ast::{
     AlterTableOperation, Expr, FromTable, Function, FunctionArg, FunctionArgExpr,
     FunctionArguments, Ident, ObjectName, Query, Select, SelectItem, SetExpr, Statement, Table,
-    TableFactor, Use, WindowSpec, WindowType, With,
+    TableFactor, TableFunctionArgs, Use, WindowSpec, WindowType, With,
 };
-use sqlparser::dialect::DatabricksDialect;
+use sqlparser::dialect::{DatabricksDialect, MsSqlDialect, SnowflakeDialect};
 
 pub trait Visit {
     fn visit(&self, context: &mut Context) -> Result<()>;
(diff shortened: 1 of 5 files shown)


integration/sql/impl/src/visitor.rs
```diff
@@ -690,12 +716,49 @@ impl Visit for Table {
 
 // --- Utils ---
 
-fn get_table_name_from_table_factor(table: &TableFactor) -> Result<Vec<Ident>> {
-    if let TableFactor::Table { name, .. } = table {
-        Ok(name.clone().0)
+fn get_table_name_from_table_factor(
+    table: &TableFactor,
+    context: &Context,
+) -> Result<Option<Vec<Ident>>> {
```
mobuchowski, line 722: Either Result or Option - both do not make sense, unless you want to express some three-value logic, and there are better tools for that like Enums

kacpermuda: Updated.

mobuchowski (approved):

PR #3002, opened 2024-08-29
CI: Group Dependabot MRs

dolfinus:
### Problem

Currently Dependabot creates a new PR for each dependency it updates:
изображение

### Solution

My proposal is to create one PR for each dependency group, containing all the updates since previous week:
изображение

Here is an example:
https://github.com/MobileTeleSystems/horizon/blob/8b21b805254d1d0b118660f973d2f177e5368201/.github/dependabot.yml#L22-L26
https://github.com/MobileTeleSystems/horizon/pull/75
изображение

### Checklist

----

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -32,6 +32,11 @@ updates:
     ignore:
       - dependency-name: "org.slf4j.slf4j-api"
       - dependency-name: "org.openapi.generator"
+    # https://til.simonwillison.net/github/dependabot-python-setup
+    groups:
+      client/java:
+        patterns:
+          - '*'
 
   - package-ecosystem: "gradle"
     directory: "integration/flink"
@@ -42,6 +47,10 @@ updates:
       - "dependabot"
     ignore:
       - dependency-name: "org.slf4j.slf4j-api"
+    groups:
+      integration/flink:
+        patterns:
+          - '*'
 
   - package-ecosystem: "gradle"
     directory: "integration/spark"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -32,6 +32,11 @@ updates:
     ignore:
       - dependency-name: "org.slf4j.slf4j-api"
       - dependency-name: "org.openapi.generator"
+    # https://til.simonwillison.net/github/dependabot-python-setup
+    groups:
+      python-packages:
+        patterns:
+          - '*'
```
kacpermuda, line 39: I like the idea, but I rarely merge those updates so I'm not sure how often we do some selective merges, where we update some and packages and leave others. Let's hear from @mobuchowski or @pawel-big-lebowski here.

Also, shouldn't we at least split those PRs into multiple groups f.e. java / python / rust ? Or if they are to be grouped together, maybe we should not name the group `python-packages`?

dolfinus: Fixed name

pawel-big-lebowski: My only concern are versions we don't want to bump as CI is failing on new version. Like Hadoop map reduce on the screenshot attached from 2.10.2 -> 3.4.0. But I assume this should be solved telling dependabot to ignore certain packages.

dolfinus: If this change introduces more issues than is solves, then I'm okay with closing this PR

pawel-big-lebowski: I am happy to test this in action.

pawel-big-lebowski (approved):

.github/dependabot.yml
pawel-big-lebowski: Let's wait a day or two in case someone comes with strong `no`. Otherwise, we merge it.

mobuchowski (approved):

.github/dependabot.yml
mobuchowski: Very easy to reverse that change if we don't like - let's do it.

Issue #2916, opened 2024-08-06
Allow whole OL configuration to be specified by ENV variables


mobuchowski:
For now, our main model of OL config is an `openlineage.yml` file with specified nested fields. Over the years, we've had a special set of some configuration variables that accumulated either due to history (like `OPENLINEAGE_URL` to configure HTTP transport before we've had the concept of transport) or operational ease (like `OPENLINEAGE_DISABLED` to stop OL action without need to physically remove all OL code and config from relevant system).

Additional complication is that OL is multi-language project - so Python and Java (and potentially Go, Rust, C++...) clients and configs would necessarily have to differ - and integration specific config - what makes sense in Airflow, is not really relevant in Spark.

So, not all was possible to configure via env variables - we've been adding new configurations with `openlineage.yml` or integration-specific context, as with Airflow or Spark conf. 

However, environment variables are the easiest to use operationally in modern data context, especially running on Kubernetes or so. It would be great if we could use those consistently.

There are few problems with that mechanism.

1. Transports. We want to also configure transports that way. However, they have different types. We want to have some hierarchy mechanism to decide what variables go where. Also, we want to introduce "composite" transport mechanism to allow people to emit events to different destinations, without needing to set up proxy. 
    - Naive way would be to add list of transports to OL transport mechanism. However, we'd prefer that logic to be handled with specialized "composite" transport that would allow us to attach behavior to different transport execution, with added benefit of not requiring change of current configuration. This will be described separately by @JDarDagran 
    - That still leaves us with problem of knowing which env configuration to attach to different mechanism - since env variables badly map to unordered lists. The solution might be to add optional `name` field to transport config - and allow people to configure it this way, with potential hierarchy being resolved by name hierarchy. An example, which would create HTTP transport with configured additional HTTP headers :
     ```env
     OPENLINEAGE_TRANSPORT_BACKEND_TYPE: http
     OPENLINEAGE_TRANSPORT_BACKEND_URL: http://something.datadog.com
     OPENLINEAGE_TRANSPORT_BACKEND_HTTP_HEADERS: key=value;key1=value1
     ``` 
[...]

PR #3029, opened 2024-09-02
Spark/add parsing dynamic info from env var

algorithmy1:
### Problem

Currently, there is a limitation when supplying parentJobName or parentRunId for jobs scheduled on Databricks clusters. These jobs, which can be Python jobs or notebook jobs, do not allow parameters to be supplied via Spark Submit configuration arguments. This limitation affects our ability to create new clusters for each job while passing the necessary parameters.

Closes: #ISSUE-NUMBER

### Solution

Inspired from [Issue 2916](https://github.com/OpenLineage/OpenLineage/issues/2916), I propose adding the capability to supply parentJobName and parentRunId (and any relevant argument) via environment variables. This solution ensures that the configuration can be overridden in the following order of precedence:

1- Yaml file
2- Environment Variables
3- Spark Configuration

Changes Proposal:

Introduce a method to extract configuration from environment variables.
Update the parse method to incorporate configuration from environment variables.

#### One-line summary:

Add support for supplying parent job details from environment variables in Databricks clusters.

### Checklist

- [ 
/] You've updated any relevant documentation (_if relevant_)

----

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
@@ -26,6 +26,8 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.Map;
+import java.util.HashMap; 
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
```diff
@@ -37,26 +39,40 @@
...
       "spark.openlineage.jobName.replaceDotWithUnderscore";
+  
   private static final String SPARK_CONF_FACETS_DISABLED = "spark.openlineage.facets.disabled";
 
   private static final String SPARK_CONF_DEBUG_FACET = "spark.openlineage.debugFacet";
 
   private static final String SPARK_TEST_EXTENSION_PROVIDER =
       "spark.openlineage.testExtensionProvider";
+  
+  private static final String VAR_ENV_OPENLINEAGE_APP_NAME = "OPENLINEAGE__APP_NAME";
```
mobuchowski, line 73: For regular `OPENLINEAGE_` env vars I'd keep single `_` for consistency with rest of env vars, unless we have some good reason to not do that.

pawel-big-lebowski, line 73: What is the reason for having those properties named in a non standard way?

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
```diff
@@ -134,6 +160,83 @@ private static void extractSparkSpecificConfigEntriesFromSparkConf(
         .ifPresent(a -> config.getFacetsConfig().setDisabledFacets(a));
   }
 
+  private static Optional<SparkOpenLineageConfig> extractOpenLineageConfFromEnvVar() {
```
pawel-big-lebowski, line 163: Would it make sense to create separate class `EnvPropertiesArgumentParser` with newly created methods?

pawel-big-lebowski (commented):
Nice feature, thank you for working on this. 
Please fix pmd errors making CI red.

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
algorithmy1: Because appName should be overriddenAppName in the targetConfig.
Since it is a special case, I defined the string to use it later

integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java
arturowczarek: I would take a slightly different approach. I would use `io.openlineage.spark.agent.ArgumentParser#extractOpenLineageConfFromSparkConf` as a base (with changed name, for example `configFromProps`). Instead of `SparkConf`, I would accept `List<Pair> props`. Then, in `io.openlineage.spark.agent.ArgumentParser#extractOpenLineageConfFromSparkConf` I would call `filterProperties(conf)` and pass it to the new method. Similarly, I would create extractLineageConfFromEnvVariables(), where I would use environmental variables to create a list of pairs ready to pass to the new extracted method. What do you think?

arturowczarek (changes requested):
I'm glad to see somebody has started working on it.

There is one tricky thing though we must address. When you update your branch to include the newest changes, you'll see the way we disable facets has changed. And reading it from environmental variables won't work, because the names of the facets (at least one) are not consistent. There is a `spark_unknown` facet which won't be properly handled. I think it is the only exception. Maybe we should rename it to `sparkUnknown`, since it is an internal facet anyway?

JDarDagran:
@algorithmy1 I just merged #3094. Thank you for your effort, I was just working on the configuration mechanism as a whole, both for Java and Python client.

PR #3043, opened 2024-09-04
Add GCP Dataplex transport as separate module

ddebowczyk92:
This PR adds `GCP` transport as separate module.

### Checklist

----

### DIFF
--- a/.circleci/workflows/openlineage-java.yml
+++ b/.circleci/workflows/openlineage-java.yml
@@ -6,6 +6,13 @@ workflows:
             tags:
               only: /^[0-9]+(\.[0-9]+){2}(-rc\.[0-9]+)?$/
           context: << pipeline.parameters.build-context >>
+      - build-transports-dataplex:
+          filters:
+            tags:
+              only: /^[0-9]+(\.[0-9]+){2}(-rc\.[0-9]+)?$/
+          context: << pipeline.parameters.build-context >>
+          requires:
+            - build-client-java
       - compile-integration-sql-java-linux:
           filters:
             tags:
@@ -51,6 +58,15 @@ workflows:
           requires:
             - build-integration-sql-java
             - build-client-java
+      - release-transports-dataplex:
+          filters:
+            tags:
+              only: /^[0-9]+(\.[0-9]+){2}(-rc\.[0-9]+)?$/
+            branches:
+              only: main
+          context: release
+          requires:
+            - build-transports-dataplex
       - release-integration-sql-java:
           filters:
             tags:
@@ -69,3 +85,4 @@ workflows:
           requires:
             - release-client-java
             - release-integration-sql-java
+            - release-transports-dataplex
--- a/client/java/transports-dataplex/README.md
+++ b/client/java/transports-dataplex/README.md
@@ -40,7 +40,7 @@ be emitted correctly.
   provided [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials)
   are used
 - `mode` - enum that specifies the type of client used for publishing OpenLineage events to Dataplex. Possible values:
-  `sync` (synchronous) or `async` (asynchronous). Optional, default: `sync`.
+  `sync` (synchronous) or `async` (asynchronous). Optional, default: `async`.
 
 ### Behavior
 
--- a/client/java/transports-dataplex/README.md
+++ b/client/java/transports-dataplex/README.md
@@ -1,0 +1,45 @@
+# Google Cloud Platform Dataplex Transport
+
+This library provides a transport layer that integrates OpenLineage with Google Cloud Platform's Dataplex service.
+It wraps the `com.google.cloud.datalineage:producerclient-java8` library into an OpenLineage transport, allowing you to
+emit lineage events directly to Dataplex using `gRPC` channel.
+
+## Getting Started
+
+### Adding the Dependency
+
+To use this transport in your project, you need to include the following dependency in your build configuration. This is
+particularly important for environments like `Spark`, where this transport must be on the classpath for lineage events
+to
+be emitted correctly.
+
+**Maven:**
+
+```xml
+
+<dependency>
+    <groupId>io.openlineage</groupId>
+    <artifactId>transports-dataplex</artifactId>
+    <version>YOUR_VERSION_HERE</version>
+</dependency>
+```
+
+### Configuration
+
+- `type` - string, must be `"dataplex"`. Required.
+- `endpoint` - string, specifies the endpoint to which events are sent. Optional.
+- `projectId` - string, the project quota identifier. If not provided, it is determined based on user credentials. Optional.
+- `locations` - string, [Dataplex locations](https://cloud.google.com/dataplex/docs/locations). Optional, default:
+  `"us"`.
+- `credentialsFile` - string, path
+  to the [Service Account credentials JSON file](https://developers.google.com/workspace/guides/create-credentials#create_credentials_for_a_service_account).
+  Optional, if not
+  provided [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials)
+  are used
+- `headers` - dictionary specifying request headers. Optional.
+
+### Behavior
+
+- Events are serialized to JSON, included as a part of `gRPC` request, and then dispatched to the `Dataplex` endpoint.
+- Two constructors are available: one accepting both `SyncLineageClient` and `DataplexConfig` and another solely accepting
+  `DataplexConfig`.
--- a/client/java/transports-dataplex/README.md
+++ b/client/java/transports-dataplex/README.md
@@ -36,7 +36,6 @@ be emitted correctly.
   Optional, if not
   provided [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials)
   are used
-- `headers` - dictionary specifying request headers. Optional.
 
 ### Behavior
 
--- a/client/java/transports-dataplex/README.md
+++ b/client/java/transports-dataplex/README.md
@@ -27,18 +27,22 @@ be emitted correctly.
 ### Configuration
 
 - `type` - string, must be `"dataplex"`. Required.
-- `endpoint` - string, specifies the endpoint to which events are sent, default value is `datalineage.googleapis.com:443`. Optional.
-- `projectId` - string, the project quota identifier. If not provided, it is determined based on user credentials. Optional.
+- `endpoint` - string, specifies the endpoint to which events are sent, default value is
+  `datalineage.googleapis.com:443`. Optional.
+- `projectId` - string, the project quota identifier. If not provided, it is determined based on user credentials.
+  Optional.
 - `location` - string, [Dataplex location](https://cloud.google.com/dataplex/docs/locations). Optional, default:
   `"us"`.
 - `credentialsFile` - string, path
-  to the [Service Account credentials JSON file](https://developers.google.com/workspace/guides/create-credentials#create_credentials_for_a_service_account).
+  to
+  the [Service Account credentials JSON file](https://developers.google.com/workspace/guides/create-credentials#create_credentials_for_a_service_account).
   Optional, if not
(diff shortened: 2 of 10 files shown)


mobuchowski:
@ddebowczyk92 `GCP` -> `Dataplex`?

.circleci/workflows/openlineage-java.yml
```diff
@@ -68,4 +84,5 @@ workflows:
               only: /^[0-9]+(\.[0-9]+){2}(-rc\.[0-9]+)?$/
           requires:
             - release-client-java
-            - release-integration-sql-java
\ No newline at end of file
+            - release-integration-sql-java
```
pawel-big-lebowski, line 87: ?

pawel-big-lebowski (commented):
My concern is: is this code specific to OpenLineage or certain lineage backend? In case of the latter: are we happy to build and release transports for all the other OL backends in case someone wants it? What are the advantages of openlineage community taking care of this? 

The disadvantage of this approach is that: we take responsibility for the code without being able to verify its correctness. One can easily make a change within `DataplexTransport` which breaks the integration while passing all the tests. It's difficult to maintain such package, as even upgrading dependencies like `com.google.cloud:google-cloud-datalineage' is risky. 

If we want to maintain such code, I think we need an integration test with real DataPlex credentials, same as we have for BigQuery.

mobuchowski:
>My concern is: is this code specific to OpenLineage or certain lineage backend? In case of the latter: are we happy to build and release transports for all the other OL backends in case someone wants it? What are the advantages of openlineage community taking care of this?

I believe there is no problem with us maintaining wrappers (transports are wrappers for specific libraries) for sending events to significant OpenLineage backends. I don't see a difference between DataplexTransport and KafkaTransport here for example, given that they both produce consumable JAR that the transport uses.

Adding this as a separate releasable Maven package is actually a step forward for modularity: even if something breaks inside, it affects only users that use it - it's not the case with `Kafka` or `Kinesis` transports now for example.

>The disadvantage of this approach is that: we take responsibility for the code without being able to verify its correctness. One can easily make a change within DataplexTransport which breaks the integration while passing all the tests. It's difficult to maintain such package, as even upgrading dependencies like `com.google.cloud:google-cloud-datalineage' is risky.

I don't agree with this, the analogy is comparable to KafkaTransport. We don't test if `'org.apache.kafka:kafka-clients:3.8.0'` can emit data to Kafka. If Google breaks their dependency, it will be on Google to fix it. We don't have to do any workarounds for this, we can purely point to the older version of a dependency or disable release for particular transport.

>If we want to maintain such code, I think we need an integration test with real DataPlex credentials, same as we have for BigQuery.

@tnazarew is writing integration consumer tests that would be able to cover whether particular transport produces information that's relevant to the particular consumer.

client/java/transports-dataplex/README.md
```diff
@@ -0,0 +1,45 @@
...
+- `endpoint` - string, specifies the endpoint to which events are sent. Optional.
+- `projectId` - string, the project quota identifier. If not provided, it is determined based on user credentials. Optional.
+- `locations` - string, [Dataplex locations](https://cloud.google.com/dataplex/docs/locations). Optional, default:
+  `"us"`.
+- `credentialsFile` - string, path
+  to the [Service Account credentials JSON file](https://developers.google.com/workspace/guides/create-credentials#create_credentials_for_a_service_account).
+  Optional, if not
+  provided [Application Default Credentials](https://cloud.google.com/docs/authentication/application-default-credentials)
+  are used
+- `headers` - dictionary specifying request headers. Optional.
```
jphalip, line 39: Can you indicate the types of headers one would want to pass here, and their function?

client/java/transports-dataplex/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Behavior
+
+- Events are serialized to JSON, included as a part of `gRPC` request, and then dispatched to the `Dataplex` endpoint.
```
jphalip, line 43: Small typo: "as part of _a_ `gRPC` request"

client/java/transports-dataplex/README.md
```diff
@@ -0,0 +1,45 @@
...
+    <groupId>io.openlineage</groupId>
+    <artifactId>transports-dataplex</artifactId>
+    <version>YOUR_VERSION_HERE</version>
+</dependency>
+```
+
+### Configuration
+
+- `type` - string, must be `"dataplex"`. Required.
+- `endpoint` - string, specifies the endpoint to which events are sent. Optional.
```
jphalip, line 30: Provide the format for the endpoint

client/java/transports-dataplex/README.md
```diff
@@ -0,0 +1,45 @@
...
+- `projectId` - string, the project quota identifier. If not provided, it is determined based on user credentials. Optional.
+- `locations` - string, [Dataplex locations](https://cloud.google.com/dataplex/docs/locations). Optional, default:
```
jphalip, line 32: It should be `location` (singular).

client/java/transports-dataplex/README.md
jphalip: @midamkina Is the endpoint config redundant with the ProjectId+Location combination?

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
```diff
@@ -0,0 +1,132 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class DataplexTransport extends Transport implements Closeable {
+
+  private final SyncLineageClient producerClient;
```
jphalip, line 34: @midamkina Should we also give the option to use the Async client?

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
```diff
@@ -0,0 +1,132 @@
...
+  private <T extends OpenLineage.BaseEvent> void emitEvent(T event) {
+    try {
+      String eventJson = OpenLineageClientUtils.toJson(event);
+      Struct openLineageStruct = OpenLineageHelper.jsonToStruct(eventJson);
+      ProcessOpenLineageRunEventRequest request =
+          ProcessOpenLineageRunEventRequest.newBuilder()
+              .setParent(parent)
+              .setOpenLineage(openLineageStruct)
+              .build();
+      producerClient.processOpenLineageRunEvent(request);
```
jphalip, line 76: @midamkina Should we also inspect the contents of the response to establish whether the request was successful or not, or is it ok to just rely on an exception to be raised in case of failure?

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
```diff
@@ -0,0 +1,132 @@
...
+    } catch (Exception e) {
+      throw new OpenLineageClientException(e);
```
jphalip, line 78: It might be good to log the error but not throw an exception. Otherwise this would cause the job to fail.
@midamkina What to you think?

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexConfig.java
```diff
@@ -0,0 +1,45 @@
...
+@ToString
+public class DataplexConfig implements TransportConfig, MergeConfig<DataplexConfig> {
+
+  @Getter @Setter private @Nullable String endpoint;
+
+  @Getter @Setter private @Nullable String projectId;
+
+  @Getter @Setter private @Nullable String credentialsFile;
+
+  @Getter @Setter private @Nullable String locations;
```
pulkit-jain-G, line 29: It should be location.

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexConfig.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@NoArgsConstructor
+@AllArgsConstructor
+@ToString
+public class DataplexConfig implements TransportConfig, MergeConfig<DataplexConfig> {
```
pulkit-jain-G, line 21: Suggestion: we can have one more property name `mode`. which could have sync or async value. Based on these user provided value we can decide to use SyncLineageProducerClient or AsyncLineageProducerClient. This way we are giving more freedom to our user to decide how should they wanna publish lineage into Dataplex.

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
midamkina: I don't think so. Now if the user wants the job to proceed, they can just try-catch this exception. But if we change it to logging and the user wants the job to fail, there will be no straightforward way to do that.

client/java/transports-dataplex/README.md
jphalip: @midamkina Is the endpoint of the form `projects/{project}/locations/{location}/`? If so, is the distinction here only about the service's project vs the billing project?

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
jphalip: From an implementation point of view, one potential challenge about the Async client is that we would need a way to track the success/failure of the API call asynchronously as well.

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
jphalip: Ok I see. So the thing is to make sure that the client integration (e.g. Spark) must properly catch that exception, and then log the error without failing the Spark job.

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexConfig.java
ddebowczyk92: @pulkit-jain-G adressed your comment, please re-review

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
ddebowczyk92: @jphalip done

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexConfig.java
pulkit-jain-G: Thanks @ddebowczyk92!

client/java/transports-dataplex/src/main/java/io/openlineage/client/transports/dataplex/DataplexTransport.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  static class ProducerClientWrapper implements Closeable {
+    private final SyncLineageClient syncLineageClient;
+    private final AsyncLineageClient asyncLineageClient;
+    private final String parent;
+
+    protected ProducerClientWrapper(DataplexConfig config) throws IOException {
+      LineageSettings settings;
+      if (DataplexConfig.Mode.async == config.getMode()) {
```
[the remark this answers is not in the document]
ddebowczyk92: switched default mode to async

pulkit-jain-G (approved):
LGTM

PR #3045, opened 2024-09-04
ci: get labels from GitHub, filter jobs based on them 

mobuchowski:
We are running full tests on each PR. This causes our CI to be severly throttled. This reduces matrix of Spark tests on PRs to three chosen ones - Spark 2.4.8 with Java 8 and Scala 2.12 as lowest dependency, Spark 3.5.1 with Scala 2.13 as highest dependency, and Spark 4. If you want to run all tests, you need to add label `full-tests` and rerun the CI.

This is done by getting GH labels from inside the setup CircleCI job, and using it in Python filtering script.

### DIFF
--- a/dev/filter_matrix.py
+++ b/dev/filter_matrix.py
@@ -1,0 +1,26 @@
+# Copyright 2018-2024 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+
+import yaml
+
+with open("complete_config.yml") as f:
+    d = yaml.safe_load(f)
+
+for _, workflow_definition in d["workflows"].items():
+    jobs = workflow_definition.get("jobs") if isinstance(workflow_definition, dict) else None
+    if not jobs:
+        continue
+
+    for job in jobs:
+        if "test-integration-spark" in job:
+            test_job = job["test-integration-spark"]
+        elif "integration-test-integration-spark" in job:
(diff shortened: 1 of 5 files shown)


dev/filter_matrix.py
```diff
@@ -0,0 +1,29 @@
...
+    for job in jobs:
+        if "test-integration-spark" in job:
+            test_job = job["test-integration-spark"]
+        elif "integration-test-integration-spark" in job:
+            integration_test_job = job["integration-test-integration-spark"]
+
+    approvals = list(
+        filter(lambda x: isinstance(x, dict) and next(iter(x.values())).get("type") == "approval", jobs)
+    )
+    for job in filter(lambda x: x is not None, [test_job, integration_test_job]):
```
JDarDagran, line 23:
```suggestion
    for job in filter(None, [test_job, integration_test_job]):
```

mobuchowski: Fixed.

JDarDagran:
Why put scripts into `dev` directory? I'd rather put it into `.circleci`

JDarDagran (approved):
I left small comment, other than that looks good.

It's a pity though that circleCI does not have out-of-the-box capability to read GH labels.

pawel-big-lebowski:
@mobuchowski Please mind nightly build which should trigger full tests as well

PR #3076, opened 2024-09-11
fix nightly run

pawel-big-lebowski:
* Fix configurable test to run on architecture other than `linux-gnu` (`amazoncorretto:17-alpine-jdk` image runs `linux-musl`)]
* Fix databricks integration test for Spark 3.5.2 to select proper runtime
* Modify slack notifications as post-step of spark integration tests

Successful run: https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/12630/workflows/b5138d8e-0f90-4f0d-9003-f7fa076c033a

### DIFF
--- a/.circleci/workflows/openlineage-spark.yml
+++ b/.circleci/workflows/openlineage-spark.yml
@@ -95,6 +95,14 @@ workflows:
               spark-version: [ '3.4.2', '3.5.2' ]
           requires:
             - approval-integration-spark
+          post-steps:
+            - when:
+                condition:
+                  equal: [ active, << pipeline.parameters.nightly-run >> ]
+                steps:
+                  - slack/notify:
+                      event: fail
+                      template: basic_fail_1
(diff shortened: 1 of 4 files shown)


.circleci/workflows/openlineage-spark.yml
```diff
@@ -120,16 +128,32 @@ workflows:
               ]
           requires:
             - approval-integration-spark
+          post-steps:
```
JDarDagran, line 131: Is this neccessary? Why not stay with previous implementation but add `when: always`?

pawel-big-lebowski: I don't an easy way to solve this. 

The problem is:
 * we do have many jobs
 * if a job does fails, next job is not starting
 * slack notify needs to depend on all the tests' jobs.

JDarDagran (approved):

PR #3094, opened 2024-09-16
Configure Java client with environment variables.

JDarDagran:
### Problem

Partially closes: #2916 

### Solution

Add the option in Java client and adjust code in Spark. Seems Flink handles this without code changes.

### Checklist

----

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/Clients.java
+++ b/client/java/src/main/java/io/openlineage/client/Clients.java
@@ -29,9 +29,16 @@ public final class Clients {
     if (isDisabled()) {
       return OpenLineageClient.builder().transport(new NoopTransport()).build();
     }
-    final OpenLineageConfig openLineageConfig =
-        OpenLineageClientUtils.loadOpenLineageConfigYaml(
-            configPathProvider, new TypeReference<OpenLineageConfig>() {});
+    OpenLineageConfig openLineageConfig;
+    try {
+      openLineageConfig =
+          OpenLineageClientUtils.loadOpenLineageConfigYaml(
+              configPathProvider, new TypeReference<OpenLineageConfig>() {});
+    } catch (OpenLineageClientException e) {
+      openLineageConfig =
+          OpenLineageClientUtils.loadOpenLineageConfigFromEnvVars(
+              new TypeReference<OpenLineageConfig>() {});
+    }
     return newClient(openLineageConfig);
   }
 
--- a/client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
+++ b/client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
@@ -1,0 +1,160 @@
+/*
+/* Copyright 2018-2024 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.utils;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.openlineage.client.Environment;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class OpenLineageEnvParser {
+
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+  private static final String PREFIX = "OPENLINEAGE__";
+
+  /**
+   * Parses all environment variables starting with OPENLINEAGE__ into a single nested JSON
+   * structure.
+   *
+   * @return A single nested JSON structure with all matching environment variables.
+   * @throws JsonProcessingException
+   */
+  public static String parseAllOpenLineageEnvVars() throws JsonProcessingException {
+    ObjectNode root = objectMapper.createObjectNode();
+
+    // Filter all environment variables starting with 'OPENLINEAGE__'
+    Map<String, String> env = Environment.getAllEnvironmentVariables();
+    env.entrySet().stream()
+        .filter(entry -> entry.getKey().startsWith(PREFIX))
+        .sorted(Map.Entry.comparingByKey(Comparator.reverseOrder()))
+        .map(
+            entry -> {
+              String envKey = entry.getKey();
+              String envValue = entry.getValue();
+              try {
+                // Parse the individual environment variable into a nested structure
+                return parseEnvToNestedJson(envKey, envValue);
+              } catch (Exception e) {
+                log.warn("Failed to parse environment variable: {}", envKey);
+                return null;
+              }
+            })
+        .filter(Objects::nonNull) // Filter out any failed parsing attempts
+        .forEach(parsedJson -> merge(root, parsedJson));
+    return objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(root);
+  }
+
+  /**
+   * Parses a single environment variable key and value into a nested JSON structure.
+   *
+   * @param envKey The environment variable key.
+   * @param envValue The environment variable value.
+   * @return A nested JSON structure based on the key.
+   */
+  private static ObjectNode parseEnvToNestedJson(String envKey, String envValue) throws Exception {
+    // Remove the 'OPENLINEAGE__' prefix
+    String strippedKey = envKey.substring(PREFIX.length());
+
+    // Split the remaining key on '__' into a List<String>
+    List<String> keyParts =
+        Stream.of(strippedKey.split("__"))
+            .map(String::toLowerCase) // Lowercase the keys if needed
+            .map(OpenLineageEnvParser::convertSnakeCaseToCamelCase)
+            .collect(Collectors.toList());
+
+    // Create the final value node. If it's valid JSON, parse it; otherwise, treat
+    // it as a string.
+    JsonNode valueNode;
+    try {
+      valueNode = objectMapper.readTree(envValue);
+    } catch (JsonProcessingException e) {
+      // If it's not valid JSON, treat it as a plain string
+      valueNode = objectMapper.convertValue(envValue, JsonNode.class);
+    }
+
+    // Create nested structure based on keyParts
+    ObjectNode root = objectMapper.createObjectNode();
+    ObjectNode currentNode = root;
+
+    // Create nested objects for each key part except the last
+    for (int i = 0; i < keyParts.size() - 1; i++) {
+      ObjectNode nextNode = objectMapper.createObjectNode();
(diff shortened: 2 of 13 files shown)


client/java/src/main/java/io/openlineage/client/Clients.java
```diff
@@ -29,9 +29,16 @@ public static OpenLineageClient newClient(ConfigPathProvider configPathProvider)
...
-    final OpenLineageConfig openLineageConfig =
-        OpenLineageClientUtils.loadOpenLineageConfigYaml(
-            configPathProvider, new TypeReference<OpenLineageConfig>() {});
+    OpenLineageConfig openLineageConfig;
+    try {
+      openLineageConfig =
+          OpenLineageClientUtils.loadOpenLineageConfigYaml(
+              configPathProvider, new TypeReference<OpenLineageConfig>() {});
+    } catch (OpenLineageClientException e) {
+      openLineageConfig =
```
pawel-big-lebowski, line 38: Why are we reading conf from env var when exception is thrown?
Shouldn't we read from both places and merge entries, as we already have a mechanism to merge configs.

integration/spark/app/src/test/java/io/openlineage/spark/agent/ArgumentParserTest.java
```diff
@@ -159,6 +164,33 @@ void testConfToKinesisConfig() {
...
 
+  @Test
+  void testLoadConfigFromEnvVars() {
+    try (MockedStatic mocked = mockStatic(Environment.class)) {
+      when(Environment.getAllEnvironmentVariables())
+          .thenReturn(Collections.singletonMap("OPENLINEAGE__TRANSPORT__TYPE", "console"));
+
+      SparkConf sparkConf = new SparkConf();
+      SparkOpenLineageConfig config = ArgumentParser.parse(sparkConf);
+      assertThat(config.getTransportConfig()).isInstanceOf(ConsoleConfig.class);
```
pawel-big-lebowski, line 175: isn't a console already a default transport? Would this test fail if environment variable not set?

integration/flink/shared/src/main/java/io/openlineage/flink/client/FlinkConfigParser.java
```diff
@@ -59,7 +59,7 @@ private static Optional<FlinkOpenLineageConfig> extractOpenLineageConfFromFile()
               OpenLineageClientUtils.loadOpenLineageConfigYaml(
                   new DefaultConfigPathProvider(), new TypeReference<FlinkOpenLineageConfig>() {}));
     } catch (OpenLineageClientException e) {
-      log.info("Couldn't log config from file, will read it from SparkConf");
+      log.info("Couldn't log config from file, will read it from FlinkConf");
```
pawel-big-lebowski, line 62: thank you for fixing

client/java/src/main/java/io/openlineage/client/Clients.java
JDarDagran: I don't think we want to merge configs from yaml and env vars. The precedence is YAML -> env vars, not a mix of those.

integration/spark/app/src/test/java/io/openlineage/spark/agent/ArgumentParserTest.java
JDarDagran: You're right, I'll change the test.

client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
```diff
@@ -0,0 +1,154 @@
...
+   */
+  private static void merge(ObjectNode mainNode, ObjectNode updateNode) {
+    updateNode
+        .fields()
+        .forEachRemaining(
+            entry -> {
+              String fieldName = entry.getKey();
+              JsonNode value = entry.getValue();
+
+              if (mainNode.has(fieldName)) {
```
Imbruced, line 134: What do you think about keeping code (happy path) left + early return 
```java
    if (!mainNode.has(fieldName)) {
      mainNode.set(fieldName, value);
      return;
    }

    // If the field already exists and both are objects, recursively merge
    JsonNode existingValue = mainNode.get(fieldName);
    if (existingValue.isObject() && value.isObject()) {
      merge((ObjectNode) existingValue, (ObjectNode) value);
      return;
    }

    mainNode.set(fieldName, value);
```


I think that way code is simpler to read

client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * Converts a string from snake_case to camelCase.
+   *
+   * @param value The input string in snake_case.
+   * @return The input string converted to camelCase.
+   */
+  private static String convertSnakeCaseToCamelCase(String value) {
+    return value.indexOf('_') != -1
```
Imbruced, line 112: I think thats nit but for me is more readable 
```java
    if (value.indexOf('_') != -1) {
      return value;
    }
    
    String initialPart = value.substring(0, value.indexOf('_'));
    
    String secondPart = Arrays.stream(
            value
              .substring(value.indexOf('_') + 1)
              .split("_"))
              .map(s -> Character.toUpperCase(s.charAt(0)) + s.substring(1))
              .collect(Collectors.joining()
    );
    
    return initialPart + secondPart;
```

client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
```diff
@@ -0,0 +1,154 @@
...
+              String envValue = entry.getValue();
+              try {
+                // Parse the individual environment variable into a nested structure
+                return parseEnvToNestedJson(envKey, envValue);
+              } catch (Exception e) {
+                log.warn("Failed to parse environment variable: {}", envKey);
+                return null;
+              }
+            })
+        .filter(parsedJson -> parsedJson != null) // Filter out any failed parsing attempts
```
Imbruced, line 54: NIT: Objects::nonNull ?

client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
```diff
@@ -0,0 +1,154 @@
...
+        .forEach(parsedJson -> merge(root, parsedJson));
+    String jsonString = objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(root);
+    return jsonString;
```
Imbruced, line 57: NIT:
```java
return objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(root)
```

client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
```diff
@@ -0,0 +1,154 @@
...
+            .map(String::toLowerCase) // Lowercase the keys if needed
+            .map(OpenLineageEnvParser::convertSnakeCaseToCamelCase)
+            .collect(Collectors.toList());
+
+    // Create the final value node. If it's valid JSON, parse it; otherwise, treat
+    // it as a string.
+    JsonNode valueNode;
+    try {
+      valueNode = objectMapper.readTree(envValue);
+    } catch (Exception e) {
```
Imbruced, line 83: I am not sure if I understand that fully.
I think readTree can throw
```java
JsonProcessingException, JsonMappingException
```

Can we lower the error scope ?

client/java/src/test/java/io/openlineage/client/utils/OpenLineageEnvParserTest.java
```diff
@@ -0,0 +1,86 @@
...
+            "{\n"
+                + "  \"api\": {\n"
+                + "    \"url\": \"http://example.com/api\",\n"
+                + "    \"timeout\": \"invalid_json\"\n"
+                + "  }\n"
+                + "}"),
+        // Unmatching variable
+        Arguments.of(
+            Map.of(
+                "OPENLINEAGE__API__URL", "http://example.com/api",
```
Imbruced, line 70: great that you have complex json covered also :)

client/java/src/main/java/io/openlineage/client/utils/OpenLineageEnvParser.java
JDarDagran: Lowered to `JsonProcessingException`.

Imbruced (approved):

website/docs/client/java/partials/java_dynamic_env_vars.md
```diff
@@ -0,0 +1,143 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+:::info
+This feature is available in OpenLineage versions >= 1.23.0
```
pawel-big-lebowski, line 5: we version the docs ;-)

website/docs/client/java/partials/java_dynamic_env_vars.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+is equivalent to passing following YAML configuration:
+```yaml
+transport:
+  type: http
+  url: http://localhost:5050
+  endpoint: api/v1/lineage
+  auth:
+  type: api_key
```
pawel-big-lebowski, line 50: two types at the same indent level?

website/docs/client/java/partials/java_dynamic_env_vars.md
```diff
@@ -0,0 +1,143 @@
...
+```yaml
+transport:
+  properties:
+    key:
+      serializer: org.apache.kafka.common.serialization.StringSerializer
+```
+which is not a valid configuration for Spark.
+
+</TabItem>
+
```
pawel-big-lebowski, line 142: My understanding is that env vars do not allow providing array params.
Perhaps this may be documented.

pawel-big-lebowski (approved):
Thank You @JDarDagran for working on this 🥇 👍 

Please revise some minor comments added in the docs. Feel free to merge after.

website/docs/client/java/partials/java_dynamic_env_vars.md
JDarDagran: I thought so too.

website/docs/client/java/partials/java_dynamic_env_vars.md
JDarDagran: Thanks for noticing.

website/docs/client/java/partials/java_dynamic_env_vars.md
pawel-big-lebowski: For example https://openlineage.io/docs/client/java/configuration#host-list-resolver -> has an ability to provide a list of hosts. Is this doable to express though env vars? I thought no. If no. it may be worth to put this into docs.

JDarDagran: I've added another example in docs as well as unit test that covers this.

One thing that could be mentioned explicitly is that one can't use `-` in env vars. This means e.g. for namespaceResolvers you may name them as follows in the test: `resolvedName` but not `resolved-name`. Not sure if those names matter at all but maybe it's worth to mention.

Issue #3084, opened 2024-09-13
[PROPOSAL] Column lineage with dataset dependencies separate from fields dependencies


arturowczarek:
**Purpose:**
In column lineage, we keep dependencies for fields and the whole datasets in one property `fields`. Field dependencies, which are used to produce the values of the fields, should stay there. The dataset dependencies (present because of the operations: sort, distinct, group by, filtering, etc), should be extracted to a separate property: `dataset`.
In the current form, representing dataset dependencies as field dependencies is realised by adding dependency to every column in the dataset. So if we have a table with columns `a`, `b`, `c`, `d` and we sort it with `d`, we receive four dependencies (every column depends indirectly on `d`). In some scenarios (for example the operation `DISTINCT`, see: https://github.com/OpenLineage/OpenLineage/issues/2874), this can be really problematic.

**Proposed implementation**
Depending on the logical plan, we should produce the dependency either to the `fields` property or `dataset` property.

Example

We have the job where we read the data from the table, remove duplicates and save it in different location.

```python
(
    spark.read.parquet(source_table_location)
    .distinct()
    .write.mode("overwrite")
    .format("parquet")
    .save("s3://aowczarek-glue-test/iceberg_warehouse/olt56.db/s3_table_2_v2/")
)
```

We use every column from the dataset. It is creating direct dependencies.
The operation `distinct` is creating dataset dependency to every column.

This is how column lineage looks like

```json
{
  "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.23.0-SNAPSHOT/integration/spark",
  "_schemaURL": "https://openlineage.io/spec/facets/1-1-1/ColumnLineageDatasetFacet.json#/$defs/ColumnLineageDatasetFacet",
  "fields": {
    "id": {
      "inputFields": [
        {
          "namespace": "s3://aowczarek-glue-test",
          "name": "/iceberg_warehouse/olt56.db/s3_table_v2",
          "field": "last_name",
          "transformations": [
            {
              "type": "INDIRECT",
              "subtype": "GROUP_BY",
              "description": "",
              "masking": false
            }
          ]
        },
        {
          "namespace": "s3://aowczarek-glue-test",
          "name": "/iceberg_warehouse/olt56.db/s3_table_v2",
          "field": "first_name",
          "transformations": [
            {
              "type": "INDIRECT",
              "subtype": "GROUP_BY",
              "description": "",
              "masking": false
            }
          ]
        },
        {
[...]

PR #3097, opened 2024-09-17
feature: Dataset dependencies in separate fields

arturowczarek:
* When flag columnLineage.deprecatedMechanismEnabled, then the dataset dependencies are extracted from property field into separate field (dataset)

### Problem

The dataset dependencies are mixed with field dependencies

Relates to: #3084
This is a continuation of: https://github.com/OpenLineage/OpenLineage/pull/3098
The next change is #3100

### Solution

We should have the dataset dependencies separate

- [X] Your change modifies one or more OpenLineage [facets](https://github.com/OpenLineage/OpenLineage/tree/main/spec/facets)

### Checklist

- [X] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_if relevant_)
- [X] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----


spec/facets/ColumnLineageDatasetFacet.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/facets/1-1-0/ColumnLineageDatasetFacet.json",
+  "$id": "https://openlineage.io/spec/facets/1-2-0/ColumnLineageDatasetFacet.json",
```
pawel-big-lebowski, line 3: Spec changes should be done in separate PRs

arturowczarek: https://github.com/OpenLineage/OpenLineage/pull/3098

integration/spark/shared/src/test/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilderDeprecatedMechanismTest.java
tnazarew, line 1: same here

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/ColumnLevelLineageUtils.java
```diff
@@ -55,7 +55,21 @@ public static Optional<OpenLineage.ColumnLineageDatasetFacet> buildColumnLineage
...
-    facetBuilder.fields(context.getBuilder().build());
+    boolean datasetLineageEnabled =
+        context
+            .getOlContext()
+            .getOpenLineageConfig()
+            .getColumnLineageConfig()
+            .isDatasetLineageEnabled();
+    if (!datasetLineageEnabled) {
+      log.warn(
+          "DEPRECATION WARNING: The columnLineage.datasetLineageEnabled configuration is set to false. This flag will default to false in the future versions and soon will be removed (defaulting to true).");
```
tnazarew, line 66: good deprecation message, should it be `warn` though? It's just information so maybe `info`, `WARNING` in the logs should suggest that something could fail here

tnazarew (changes requested):

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/ColumnLevelLineageUtils.java
arturowczarek: It is going to fail in just a couple of releases.

tnazarew: so in couple of releases it will be a good warning :)

arturowczarek: Now it warns that the default is going to change and your events will be affected

tnazarew: I mean, it's not that big of a deal since it will be removed relatively soon. 

Nonetheless warning should tell you about something that may cause failure, so I wouldn't log warnings about something concerning future versions (which current users may not upgrade to anytime soon)

tnazarew (approved):

website/docs/integrations/spark/configuration/spark_conf.md
```diff
@@ -6,20 +6,21 @@ title: Spark Config Parameters
...
+| spark.openlineage.job.owners.<ownership-type\>        | Specifies ownership of the job. Multiple entries with different types are allowed. Config key name and value are used to create job ownership type and name (available since 1.13).                                                                                                                                                                                                                 | spark.openlineage.job.owners.team="Some Team" |
+| spark.openlineage.columnLineage.datasetLineageEnabled | **Since 1.24.0** Makes the dataset dependencies to be included in their own property `dataset` in the column lineage pattern. If this flag is set to `false`, then the dataset dependencies are merged into `fields` property. The default value is `false` but soon the default will change to `true` and soon after this configuration will be removed. **It is recommended to set it to `true`** | true                                          |
```
pawel-big-lebowski, line 26: There is no need to mention the docs this way anymore.

integration/spark/app/src/test/java/io/openlineage/spark/agent/column/ColumnLevelLineageIcebergTest.java
```diff
@@ -378,7 +380,9 @@ void testCTEQuery() {
     assertColumnDependsOn(facet, "c", FILE, T1_EXPECTED_NAME, "a");
     assertColumnDependsOn(facet, "d", FILE, T1_EXPECTED_NAME, "b");
     assertColumnDependsOnInputs(facet, "c", 1);
-    assertColumnDependsOnInputs(facet, "d", 2);
+    assertColumnDependsOnInputs(facet, "d", 1);
+    assertDatasetDependsOnType(
```
pawel-big-lebowski, line 384: Is this the only test affected in the class by dataset depenendency feature? If so, why do we want to create whole `ColumnLevelLineageIcebergOnlyFieldDependenciesTest` file?

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkOpenLineageConfig.java
```diff
@@ -32,15 +34,18 @@ public class SparkOpenLineageConfig extends OpenLineageConfig<SparkOpenLineageCo
...
+  private String parentJobNamespace;
+  private String parentRunId;
+  private String overriddenAppName;
+  @NonNull private String debugFacet;
+  private String testExtensionProvider;
+  private JobNameConfig jobName;
+  private JobConfig job;
+
+  @JsonProperty("columnLineage")
+  private ColumnLineageConfig columnLineageConfig;
```
pawel-big-lebowski, line 48: Would it make sense to have it as a part of `DatasetConfig`?

pawel-big-lebowski (commented):
I think the original idea behind PR was: let's copy existing tests, bcz there will be a lot of tests changes when implementing dataset lineage feature. The problem with this reasoning is that it turned out only 4 existing tests are affected by the new feature, but going this way requires copying 900 lines of code. 

Is my understanding correct?

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkOpenLineageConfig.java
arturowczarek: It is often said that nothing is more permanent than s temporary change, but I hope it won't be this case. Moving it to `DatasetConfig` would require extending this class for Spark integration (because we don't have this facet in Java) and it carries the risk introducing some regression. Additionally, it would take extra 30-60 minutes of extra work that could be better spent elsewhere. Is it worth it?

integration/spark/app/src/test/java/io/openlineage/spark/agent/column/ColumnLevelLineageIcebergTest.java
arturowczarek: I chose to copy the files for two reasons. First, to finally remove it, we just delete the entire files without thinking. Second, and most importantly, there is only one place in the tests where the new dataset dependency mode flag is set. Switching it would affect every test. Without hard-to-review to review refactor, it would be impossible to control the flag's value. Additionally, at some point, we'd need to reverse the refactor to avoid causing confusion. The `OnlyFieldDependenciesTest` suites are just copies of what was there so far and are intended for removal.

pawel-big-lebowski (approved):

PR #3121, opened 2024-09-27
Require Java 17 only for Spark 4.0

jphalip:
Without this change, I get this kind of error whenever I run any gradle task:

```
Execution failed for task ':shared:compileScala212Java'.
> This project will not compile with Java version below 17.
```

My understanding is that Java 17 is only required for Spark 4.0. Is that correct? If so, then this change should hopefully more appropriately scope this requirement.

### DIFF
--- a/integration/spark/buildSrc/src/main/kotlin/io/openlineage/gradle/plugin/CommonConfigPlugin.kt
+++ b/integration/spark/buildSrc/src/main/kotlin/io/openlineage/gradle/plugin/CommonConfigPlugin.kt
@@ -87,10 +87,6 @@ class CommonConfigPlugin : Plugin<Project> {
                     // never run compile on CI without property being set
                     throw RuntimeException("java.compile.home should be always set on CI env")
                 }
-
-                if (!target.hasProperty("java.compile.home") && JavaVersion.current() < JavaVersion.VERSION_17) {
-                    throw RuntimeException("This project will not compile with Java version below 17.")
-                }
             }
         }
 
--- a/integration/spark/spark40/build.gradle
+++ b/integration/spark/spark40/build.gradle
@@ -13,6 +13,12 @@ scalaVariants {
     create("2.13")
 }
 
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(17)
+    }
+}
+
 idea {
     module {
         testSources.from(sourceSets.testScala213.java.srcDirs)


pawel-big-lebowski:
@jphalip Although Java 17 is required for Spark 4 only, keep in mind we do release a single jar that should work will all the Spark versions. Spark openlineage library is built with target compatibility set to Java 8. 

However, in order to compile with Spark 4 dependencies, it has to be compiled with Java 17. 

Fortunately, we are able to run tests which compile code with Java 17 and run gradle tests on Java 8, as some Spark versions that we still support don't allow Java 17. 

Let me know if this makes sense to you.

jphalip:
Hi @pawel-big-lebowski -- My issue is that I can't run any tests for Spark, even older versions, because I keep getting the error:
```
Execution failed for task ':shared:compileScala212Java'.
> This project will not compile with Java version below 17.
```
Am I missing something?

pawel-big-lebowski:
@jphalip you need to compile project with Java 17. If you want to run tests on older Java, you still need to compile code with Java 17. This is done in CI with: 
```
./gradlew --no-daemon --console=plain --stacktrace build check -Pspark.version=${SPARK} -Pscala.binary.version=${SCALA} -Pjava.compile.home=${JAVA17_HOME}
```

jphalip:
> @jphalip you need to compile project with Java 17. If you want to run tests on older Java, you still need to compile code with Java 17. This is done in CI with:
> 
> ```
> ./gradlew --no-daemon --console=plain --stacktrace build check -Pspark.version=${SPARK} -Pscala.binary.version=${SCALA} -Pjava.compile.home=${JAVA17_HOME}
> ```

Couldn't that be set in the Gradle conf itself, so that the user doesn't need to specify it explicitly every time?

pawel-big-lebowski:
hmyy, if you're using Java 17, there should be no need to specify it.

pawel-big-lebowski (dismissed):

jphalip:
@pawel-big-lebowski The compilation errors are related to some classes (e.g. `ColumnLineageDatasetFacetFieldsAdditionalInputFields`) that have been removed recently. Maybe the error is due to some old code being cached somewhere on the CI server?

pawel-big-lebowski:
@jphalip there was such issue this week. could you please try rebasing? sry for inconvenience.

PR #3142, opened 2024-10-08
test: Add integration tests for EMR

arturowczarek:
* Spark integration has integration tests for EMR

### Problem

We are missing the integration tests for EMR. Without them some types of bugs remain undetected and we may introduce regressions without knowing it.

### Solution

We should have integration tests. We don't have a sponsor for the AWS infrastructure, so every developer should be able to run them manually using their infrastructure.

**For reviewers**: Start reading from `EmrIntegrationTest` class.

### Checklist

----

### DIFF
--- a/integration/spark/app/build.gradle
+++ b/integration/spark/app/build.gradle
@@ -123,6 +123,12 @@ dependencies {
         exclude group: 'com.fasterxml.jackson.core'
         exclude group: 'com.fasterxml.jackson.module'
     }
+
+    testImplementation(platform("software.amazon.awssdk:bom:2.28.11"))
+    testImplementation("software.amazon.awssdk:auth")
+    testImplementation("software.amazon.awssdk:emr")
+    testImplementation("software.amazon.awssdk:s3")
+
     testImplementation("org.mock-server:mockserver-netty:5.14.0:shaded") {
         exclude group: 'com.google.guava', module: 'guava'
         exclude group: 'com.fasterxml.jackson.core'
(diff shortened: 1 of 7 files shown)


mobuchowski:
@arturowczarek can we run "regular" integration tests on EMR, or only special tests otherwise located in the `emr_test_jobs` directory?

arturowczarek:
@mobuchowski The loading class locates the files in the specific directory for simplicity. We can extend it if necessary, but passing just the name of the file is convenient, because the name is used to create temporary directory where everything for this test is kept. Otherwise we could have multiple scripts with the same name and they would keep the events under the same location. Alternatively we could have a better strategy for determining the location for the events.

mobuchowski:
Makes sense, I just think we'd also potentially benefit with running a bunch of "regular" integration tests on EMR regularly.

arturowczarek:
Yes, it is one of the methods we could further develop these tests.
There is another challenge we have to find a good solution to - managing resources and determining which tests to run. In some tests I'd like to add, there is a need to have two AWS accounts (to have two separate Glue catalogs). It is not common that developers have access to multiple accounts. In such scenarios they have to skip these tests. I'd also like to create some (possibly) Terraform file to create the infrastructure. I'll do it when we have someone to sponsor it.

pawel-big-lebowski (commented):
I think this is huge step forward in improving OL event collection on AWS. 
The code looks good to me.

What system properties are required to be set to run the test?

arturowczarek:
The parameters are listed in `DynamicParameter` enum. Most of them have some reasonable defaults (instance size, EMR version, roles which should be present in every account). Only one is required (`openlineage.tests.bucketName`). Apart from that you have to be authenticated (have ~/.aws/credentials file or use any method described in [Default credentials provider chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html))

PR #3147, opened 2024-10-10
Fix/emr implicit glue catalog

arturowczarek:
### Problem

The mechanism for determining Glue catalog ID is broken when we don't pass it explicitly.

Closes: #ISSUE-NUMBER

### Solution

We should use account id when the user doesn't pass the catalog ID explicitly.

### Checklist

----

### DIFF
--- a/.github/workflows/integration-tests-spark-aws.yml
+++ b/.github/workflows/integration-tests-spark-aws.yml
@@ -1,0 +1,52 @@
+name: Integration tests for Spark AWS
+
+# Runs the AWS integration tests
+#
+# The workflow is triggered manually. To run, you have to install GitHub CLI, setup secrets and run:
+# gh workflow run "Integration tests for Spark AWS" --ref <name of your branch>
+#
+# The required secrets are:
+# - EMR_TESTS_AWS_ACCESS_KEY_ID
+# - EMR_TESTS_AWS_SECRET_ACCESS_KEY
+# - EMR_TESTS_AWS_REGION
+# - EMR_TESTS_BUCKET_NAME
+# - EMR_TESTS_EC2_SUBNET_ID
+#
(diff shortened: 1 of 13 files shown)


pawel-big-lebowski (commented):
I was able once to see your live demo and these features are great. 
The problem with this code is that, as a reviewer, one cannot see if it working. 

Would it be possible to provide a CI step to run the tests which takes AWS credentials as run parameters? In this case, you would be able to prove this is working by providing your creds for a single run. 

The advantage for the community is that we can turn it on easily later, once get a sponsor for those kind of tests.

arturowczarek:
> I was able once to see your live demo and these features are great. The problem with this code is that, as a reviewer, one cannot see if it working.
> 
> Would it be possible to provide a CI step to run the tests which takes AWS credentials as run parameters? In this case, you would be able to prove this is working by providing your creds for a single run.
> 
> The advantage for the community is that we can turn it on easily later, once get a sponsor for those kind of tests.

I've added a GitHub workflow to run the tests. The details how to do it are in the workflow file.
Example execution: https://github.com/arturowczarek/OpenLineage/actions/runs/11347676564
There is a problem I've submitted to AWS security where credentials are always printed out. Before they fix it, we have to use the key and revoke it once used.

mobuchowski:
Hey @arturowczarek - I believe we could fix the errorenous secrets printing by utilizing add-mask GitHub Actions functionality. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#masking-a-value-in-a-log

Not leaking secrets is pretty important when using public CI, I'd be more comfortable merging this change after it's fixed.

mobuchowski (approved):

PR #3159, opened 2024-10-15
local file system should not point to IP, added remote file system doc

mobuchowski:
Namespace for a `local file system`: `file` dataset mentions `host`, or `IP` and `port` when still talking about "local file system". It seems clear that this is not "local" if we need to mention IP or port.

This also adds `remote file system` dataset - which is for example used by `ftp` and `sftp` assets.

Draft to facilitate discussion.

### DIFF
--- a/spec/Naming.md
+++ b/spec/Naming.md
@@ -1,5 +1,7 @@
 # Naming
 
+## This document is obsolete, please look at the [OpenLineage Naming Conventions](https://openlineage.io/docs/spec/naming)
+
 We define the unique name strategy per resource to ensure it is followed uniformly independently of who is producing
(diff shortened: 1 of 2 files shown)


JDarDagran:
You're right that mentioning a host, IP, or port usually implies some sort of networked or remote resource, which conflicts with the idea of a truly "local" file system. For **true** local file systems, you'd typically reference file paths directly (e.g., `/local/path/to/file`), without needing to specify network-related information.

pawel-big-lebowski:
I don't see ideal solution for this. This proposal brings clear improvement by distinguishing datasets that can be globally identified from those which only make sense in the context of local machine running.

pawel-big-lebowski (approved):

PR #3215, opened 2024-11-03
docs: fix outdated Spark-related docs

mobuchowski:


arturowczarek (commented):
It is going to be outdated next month. Maybe we should replace it with a placeholder or a variable which would update with every release instead?

mobuchowski:
That's a good idea, want to contribute that? 
Main issue was the deprecated way HTTP transport was configured though.

JDarDagran (approved):

PR #3240, opened 2024-11-13
collect input/output statistics from Spark jobs

pawel-big-lebowski:
Improve Spark integration to collect statistics about input and output datasets. 
Prepare test to check if the stats are correct. 

For outputs, Spark integration always extracts all the statistics: amount of files written, bytes size and records written. 
For inputs, bytes size and file count is always returned. Amount of records read is fetched only for DataSourceV2 inputs. 

Another PR is planned to extract even more detailed statistics for Iceberg datasets.

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2ScanRelationOnEndInputDatasetBuilder.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2ScanRelationOnEndInputDatasetBuilder.java
@@ -6,26 +6,14 @@
 package io.openlineage.spark3.agent.lifecycle.plan;
 
 import io.openlineage.client.OpenLineage.InputDataset;
-import io.openlineage.client.OpenLineage.InputStatisticsInputDatasetFacet;
 import io.openlineage.client.dataset.DatasetCompositeFacetsBuilder;
 import io.openlineage.spark.api.AbstractQueryPlanInputDatasetBuilder;
 import io.openlineage.spark.api.DatasetFactory;
 import io.openlineage.spark.api.OpenLineageContext;
 import io.openlineage.spark3.agent.utils.DataSourceV2RelationDatasetExtractor;
-import java.lang.reflect.InvocationTargetException;
-import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.reflect.MethodUtils;
-import org.apache.iceberg.BaseCombinedScanTask;
-import org.apache.iceberg.ContentFile;
-import org.apache.iceberg.DataFile;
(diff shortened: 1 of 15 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2ScanRelationOnEndInputDatasetBuilder.java
```diff
@@ -50,10 +62,51 @@ public List<InputDataset> apply(DataSourceV2ScanRelation plan) {
...
         factory, context, relation, datasetFacetsBuilder);
   }
 
+  private Optional<InputStatisticsInputDatasetFacet> getInputStatisticsInputDatasetFacet(
+      DataSourceV2ScanRelation plan) {
+    if (plan.scan() == null) {
+      return Optional.empty();
+    }
+    try {
+      List<ScanTask> tasks = (List<ScanTask>) MethodUtils.invokeMethod(plan.scan(), true, "tasks");
```
mobuchowski, line 79: Where is this coming from? I don't see it implemented for Scan interface: https://github.com/apache/spark/blob/381f3691bd481abc8f621ca3f282e06db32bea31/sql/catalyst/src/main/java/org/apache/spark/sql/connector/read/Scan.java#L46

mobuchowski: If it's Iceberg, shouldn't it be separated somehow? Or, not at least, have some isolated code path.

mobuchowski (approved):
One point about iceberg support, besides that looking good.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DataSourceV2ScanRelationOnEndInputDatasetBuilder.java
pawel-big-lebowski: For Iceberg, I am planning a separate PR -> https://github.com/OpenLineage/OpenLineage/pull/3256/files which will introduce a separate facet. This PR is more generic but gets limited amount of metrics. For iceberg we can obtain much more information, but I think it's worth putting it into a separate iceberg specific facet. In some cases, we get both facets and I don't think this duplication is a problem at all.

mobuchowski: My point is about separating the specific Iceberg implementation of `getInputStatisticsInputDatasetFacet` in `DataSourceV2ScanRelation`, as it implements some Iceberg specific functionality. Maybe to some `IcebergUtils` if we don't have `vendor` code for it yet.

pawel-big-lebowski: Ohh,  you're right, now I get it. I will remove feature for iceberg and bring it back in the separate PR.

PR #3256, opened 2024-11-19
Spark & Iceberg - include CommitReport and ScanReports facets 

pawel-big-lebowski:
Collect iceberg Scan and Commit reports.

Apache Iceberg provides a MetricsReporter (https://iceberg.apache.org/docs/1.4.1/metrics-reporting/) which allows collecting information `ScanReport` like
```
INFO org.apache.iceberg.metrics.LoggingMetricsReporter - Received metrics report: 
ScanReport{
    tableName=scan-planning-with-eq-and-pos-delete-files, 
    snapshotId=2, 
    filter=ref(name="data") == "(hash-27fa7cc0)", 
    schemaId=0, 
    projectedFieldIds=[1, 2], 
    projectedFieldNames=[id, data], 
    scanMetrics=ScanMetricsResult{
        totalPlanningDuration=TimerResult{timeUnit=NANOSECONDS, totalDuration=PT0.026569404S, count=1}, 
        resultDataFiles=CounterResult{unit=COUNT, value=1}, 
        resultDeleteFiles=CounterResult{unit=COUNT, value=2}, 
        totalDataManifests=CounterResult{unit=COUNT, value=1}, 
        totalDeleteManifests=CounterResult{unit=COUNT, value=1}, 
        scannedDataManifests=CounterResult{unit=COUNT, value=1}, 
        skippedDataManifests=CounterResult{unit=COUNT, value=0}, 
        totalFileSizeInBytes=CounterResult{unit=BYTES, value=10}, 
        totalDeleteFileSizeInBytes=CounterResult{unit=BYTES, value=20}, 
        skippedDataFiles=CounterResult{unit=COUNT, value=0}, 
        skippedDeleteFiles=CounterResult{unit=COUNT, value=0}, 
        scannedDeleteManifests=CounterResult{unit=COUNT, value=1}, 
        skippedDeleteManifests=CounterResult{unit=COUNT, value=0}, 
        indexedDeleteFiles=CounterResult{unit=COUNT, value=2}, 
        equalityDeleteFiles=CounterResult{unit=COUNT, value=1}, 
        positionalDeleteFiles=CounterResult{unit=COUNT, value=1}}, 
    metadata={
        iceberg-version=Apache Iceberg 1.4.0-SNAPSHOT (commit 4868d2823004c8c256a50ea7c25cff94314cc135)}}
``` 
and CommitReport
```
INFO org.apache.iceberg.metrics.LoggingMetricsReporter - Received metrics report: 
CommitReport{
    tableName=scan-planning-with-eq-and-pos-delete-files, 
    snapshotId=1, 
    sequenceNumber=1, 
    operation=append, 
    commitMetrics=CommitMetricsResult{
        totalDuration=TimerResult{timeUnit=NANOSECONDS, totalDuration=PT0.098429626S, count=1}, 
        attempts=CounterResult{unit=COUNT, value=1}, 
        addedDataFiles=CounterResult{unit=COUNT, value=1}, 
        removedDataFiles=null, 
        totalDataFiles=CounterResult{unit=COUNT, value=1}, 
        addedDeleteFiles=null, 
        addedEqualityDeleteFiles=null, 
        addedPositionalDeleteFiles=null, 
        removedDeleteFiles=null, 
        removedEqualityDeleteFiles=null, 
        removedPositionalDeleteFiles=null, 
        totalDeleteFiles=CounterResult{unit=COUNT, value=0}, 
        addedRecords=CounterResult{unit=COUNT, value=1}, 
        removedRecords=null, 
        totalRecords=CounterResult{unit=COUNT, value=1}, 
        addedFilesSizeInBytes=CounterResult{unit=BYTES, value=10}, 
        removedFilesSizeInBytes=null, 
        totalFilesSizeInBytes=CounterResult{unit=BYTES, value=10}, 
        addedPositionalDeletes=null, 
        removedPositionalDeletes=null, 
        totalPositionalDeletes=CounterResult{unit=COUNT, value=0}, 
        addedEqualityDeletes=null, 
        removedEqualityDeletes=null, 
        totalEqualityDeletes=CounterResult{unit=COUNT, value=0}}, 
    metadata={
        iceberg-version=Apache Iceberg 1.4.0-SNAPSHOT (commit 4868d2823004c8c256a50ea7c25cff94314cc135)}}
```

This PR incorporates this metrics as separate Iceberg facets. 

Technically this is done by injecting OpenLineage Iceberg Metric Reporter, collecting limited amount of the reports in memory and attaching them to OpenLineage events based on the dataset version that is already present within OL event. If user already has `MetricsReporter` configured, an OpenLineage reported calls its methods as well to preserve existing scenario.


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/TableContentChangeDatasetBuilder.java
```diff
@@ -31,9 +35,12 @@
 @Slf4j
 public class TableContentChangeDatasetBuilder
     extends AbstractQueryPlanOutputDatasetBuilder<LogicalPlan> {
+  private final DatasetFactory<OutputDataset> factory;
 
-  public TableContentChangeDatasetBuilder(OpenLineageContext context) {
+  public TableContentChangeDatasetBuilder(
+      OpenLineageContext context, DatasetFactory<OutputDataset> factory) {
     super(context, false);
+    this.factory = Objects.requireNonNull(factory, "parameter: factory");
```
mobuchowski, line 43: `@NonNull`?

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/CatalogMetricsReporterHolder.java
```diff
@@ -0,0 +1,150 @@
...
+    return catalogMetricsReporter.get(catalogName);
+  }
+
+  @VisibleForTesting
+  void clear() {
+    catalogMetricsReporter.clear();
+  }
+
+  private static class SingletonHolder {
+    public static final CatalogMetricsReporterHolder instance = new CatalogMetricsReporterHolder();
```
mobuchowski, line 144: Rather than doing singleton here, put it into `OpenLineageContext`

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/CommitReportsFacetBuilder.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private static IcebergCommitMetrics getIcebergCommitMetrics(CommitMetricsResult metricsResult) {
+    IcebergCommitMetrics.IcebergCommitMetricsBuilder builder = IcebergCommitMetrics.builder();
+
+    if (metricsResult == null) {
+      return builder.build();
+    }
+
+    Optional.ofNullable(metricsResult.totalDuration())
```
mobuchowski, line 35: Pretty sure it's possible to avoid all of those...

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/IcebergMetricsReporterInjector.java
```diff
@@ -0,0 +1,121 @@
...
+    if (!catalog.isPresent()) {
+      return false;
+    }
+
+    // check catalog class starts with org.apache.iceberg
+    return catalog.get().getClass().getCanonicalName().startsWith("org.apache.iceberg");
+  }
+
+  private Optional<TableCatalog> getCatalog(LogicalPlan plan) {
+    try {
```
mobuchowski, line 70: Would love an explanation why we're using reflection here.

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/IcebergMetricsReporterInjector.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Override
+  public List<D> apply(LogicalPlan x) {
+    SparkCatalog catalog = (SparkCatalog) getCatalog(x).get();
+    // hack catalog to inject OpenLineageMetricsReporter
+    Catalog icebergCatalog = catalog.icebergCatalog();
+    if (icebergCatalog instanceof CachingCatalog) {
+      // get root catalog of a caching catalog
+      Field catalogField = FieldUtils.getField(icebergCatalog.getClass(), "catalog", true);
```
mobuchowski, line 103: Same here, this reflection behavior seems crucial to understand what's going on here, and is pretty opaque when all of those could fail, or why are we even using reflection instead of direct calls here.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/IcebergCommitMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+  Long totalRecords;
+  Long addedFilesSizeInBytes;
+  Long removedFilesSizeInBytes;
+  Long totalFilesSizeInBytes;
+  Long addedPositionalDeletes;
+  Long removedPositionalDeletes;
+  Long totalPositionalDeletes;
+  Long addedEqualityDeletes;
+  Long removedEqualityDeletes;
+  Long totalEqualityDeletes;
```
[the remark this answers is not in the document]
pawel-big-lebowski: Because we don't want to serialize empty values to `0`.

What we do, is basically expressing Iceberg Metrics classes like this -> https://github.com/apache/iceberg/blob/apache-iceberg-1.7.1/core/src/main/java/org/apache/iceberg/metrics/ScanMetricsResult.java with our facet. Iceberg metrics reports contain dozens of nullable `CounterResult`. In case of null, we don't want to set the value within our class.  Then Jackson will skip serialisation of those fields and serialize only the non-null metrics from icberg report. 

Screen below comes from debugging facets produced locally. `CommitMetric` contain only 11 values, although `IcebergCommitMetrics` has over 20 fields. 

Screenshot 2024-12-10 at 09 06 57

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/CommitReportsFacetBuilder.java
pawel-big-lebowski: Would love to hear one ;)

The problem is: there is Iceberg interface with ~20 methods, each of the methods returns `Counter` which can be null or may have long value. We need to translate it to our class as we don't want to build facets based on Iceberg classes directly.

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/IcebergMetricsReporterInjector.java
```diff
@@ -66,24 +67,38 @@ public boolean isDefinedAt(LogicalPlan plan) {
...
+      return getCatalogFromCaseClass(((BinaryCommand) plan).left());
+    }
+
+    return Optional.empty();
+  }
+
+  /**
+   * Runs catalog method on LogicalPlan class. This is done through reflection as the case class
+   * with catalog property differs across different Spark versions. For example,
+   * `ResolvedDBObjectName` is available only for Spark 3.3
```
mobuchowski, line 92: In the cases where we're using reflection, we need to have complete explanation for where we're going into some different conditions, since we can't just ctrl+click on those using IDE. For example isn't good enough I believe, the explanation needs to be exhaustive, at least for those cases when we're not falling back to some default behavior.

I know it wasn't always done that way in the past but it's extremely confusing later, and there's no better moment to document that behavior than now.

integration/spark/vendor/iceberg/src/main/java/io/openlineage/spark/agent/vendor/iceberg/metrics/CatalogMetricsReporterHolder.java
pawel-big-lebowski: Let me look into this again.

pawel-big-lebowski: What we need to do is: register OpenLineage metrics reporter within the Iceberg catalog while still having a handle to it. Singleton is the easiest handle to go. Whenever a dataset's version facet is added, we call custom facet builders which attach iceberg metrics facets based on the version. 

Including metrics reporter in `OpenLineageContext` is more complex, mainly bcz for each run (each run a new context is created), we need to get the catalog from the logical plan, get the metrics reporter attached to a catalog and link it with current context. I believe this is doable and better, but it's not a straight-forward code change.

mobuchowski (approved):
Amazing work Paweł. LGTM.

Issue #3260, opened 2024-11-19
[FEATURE] DBT Column Level Lineage support


mayurmadnani:
### Motivation

Currently Column Level Lineage is only limited to Spark. It would be great to introduce it for dbt integration as well

### Do you plan to make this contribution yourself?

- [X] I am interested in doing this work

PR #3261, opened 2024-11-19
[Integration][DBT] Support for CLL in DBT integration

mayurmadnani:
### Problem

DBT integration does not have CLL support

Closes: #3260 

### Solution

DBT gives table level lineage in manifest. Using the SQL parser and dbt compiled code, CLL is generated.

#### One-line summary:

Adds CLL support using openlineage SQL parser to derive column lineage

### Checklist

----

### DIFF
--- a/integration/common/tests/dbt/test_dbt_local.py
+++ b/integration/common/tests/dbt/test_dbt_local.py
@@ -268,3 +268,88 @@ def test_build_target_path(test_name, dbt_project, expected):
         job_namespace="ol-namespace",
     )
     assert processor.build_target_path(dbt_project) == expected
+
+
+@pytest.mark.parametrize(
+    "path",
+    [
+        "tests/dbt/compiled_code",
+    ],
+)
+def test_dbt_column_lineage(path, parent_run_metadata):
+    """
+    Test that column-level lineage is correctly extracted from dbt artifacts
+    """
+    processor = DbtLocalArtifactProcessor(
+        producer="https://github.com/OpenLineage/OpenLineage/tree/0.0.1/integration/dbt",
+        job_namespace="job-namespace",
+        project_dir=path,
+    )
+    processor.dbt_run_metadata = parent_run_metadata
+    dbt_events = processor.parse()
+    events = [
+        attr.asdict(event, value_serializer=serialize)
+        for event in dbt_events.starts + dbt_events.completes + dbt_events.fails
+    ]
+    with open(f"{path}/result.json") as f:
+        assert match(json.load(f), events)
+
+
+@pytest.mark.parametrize(
+    "path",
+    [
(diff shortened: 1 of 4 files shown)


mobuchowski:
Thanks for the contribution @mayurmadnani. It would be great to add tests that check validity of the solution here. The tests are located here: https://github.com/OpenLineage/OpenLineage/tree/0399867fc3cf29015c7de1f8923b49ec0b263617/integration/common/tests/dbt

mayurmadnani:
Thanks @mobuchowski 
Added tests. Also, here is a sample run in local mode

image

JDarDagran:
@mayurmadnani do you have knowledge about how compiled SQL is created in results file? Do database object always stay the same in SQL statement or do they have database / schema added if they're missing in the model definition?

On the other hand, maybe this is very rare case because usually one does refer to db objects using {{ ref }}?

mayurmadnani:
> @mayurmadnani do you have knowledge about how compiled SQL is created in results file? Do database object always stay the same in SQL statement or do they have database / schema added if they're missing in the model definition?
> 
> On the other hand, maybe this is very rare case because usually one does refer to db objects using {{ ref }}?

@JDarDagran when you use {{ ref }} dbt will compile and generate the compiled SQL with db and schema info. see this- 
image

JDarDagran:
Is this the same for {{ source }} and tests? Sorry for being too inquisitive, just trying to make sure we don't miss some common use case.
I think your contribution is great 🚀

mayurmadnani:
> Is this the same for {{ source }} and tests? Sorry for being too inquisitive, just trying to make sure we don't miss some common use case. I think your contribution is great 🚀

Yes, holds for source and tests as well
image
image

JDarDagran:
Awesome, thanks for checking

integration/common/tests/dbt/test_dbt_local.py
```diff
@@ -268,3 +268,88 @@ def test_build_target_path(test_name, dbt_project, expected):
...
+        project_dir=path,
+    )
+    processor.dbt_run_metadata = parent_run_metadata
+    dbt_events = processor.parse()
+    events = [
+        attr.asdict(event, value_serializer=serialize)
+        for event in dbt_events.starts + dbt_events.completes + dbt_events.fails
+    ]
+    with open(f"{path}/result.json") as f:
+        assert match(json.load(f), events)
```
mobuchowski, line 295: @mayurmadnani there is no column-level lineage in the result event: https://github.com/OpenLineage/OpenLineage/blob/main/integration/common/tests/dbt/compiled_code/result.json

The comparison won't check existence of CLL in the event your change is generating.

mobuchowski (changes requested):

Issue #3284, opened 2024-11-28
[BUG] GCS path is not correctly stripped of leading slash


kacpermuda:
### OpenLineage integration/client

Spark integration

### OpenLineage version

1.24.2

### Problem details

When reading from GCS e.g. 

from `gs://bucket/dir1/dir2/*.csv` we receive input dataset with name= `/dir1/dir2`,
from `gs://bucket/co*` with 1 file matching the pattern we receive input dataset with name= `/copy.csv`.

I think the leading slashes should not be present.

For s3, also in spark integration, the leading slashes are removed correctly:
with `s3://bucket/test/` we get name=`test`,
with `s3://bucket/test/pre` we get name=`test/pre`,

In Airflow integration we also get rid of leading slashes, so i think we should adjust spark integration's behaviour to avoid problems with dataset matching between integrations.

Probably the code responsible for that behaviour is [here](https://github.com/OpenLineage/OpenLineage/blob/66a0b8bb45a6a5c7802dc44152e4a37b7772d5e2/client/java/src/main/java/io/openlineage/client/utils/filesystem/FilesystemDatasetUtils.java#L16)

### How to reproduce

Just a simple read/write from one cloud storage to another, with the paths mentioned above:
```
df = spark.read.csv("gs://bucket/co*", header=True)
df.write.mode("overwrite").csv(f"s3a://bucket/test/pre/")
```

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

kacpermuda:
@dolfinus Maybe you'll have some idea on why it's happening?

dolfinus:
FilesystemDatasetUtils have tests for GCP, leading slash is stripped from the path:
https://github.com/OpenLineage/OpenLineage/blob/1.25.0/client/java/src/test/java/io/openlineage/client/utils/filesystem/FilesystemDatasetUtilsTestForGCS.java#L23-L33

But it is possible that code producing this OL event does not call `FilesystemDatasetUtils.fromLocation` in the first place.

pawel-big-lebowski:
@kacpermuda thanks for opening this. I've searched Spark integration's code for the places `new DatasetIdentifier(` which call `DatasetIdentifier`. Apart from tests, there are few places that identify datasets without `FilesSystemDatasetUtils` logic applied. `DatasetFactory` may be the one responsible for the issue you have:
[...]

PR #3285, opened 2024-11-29
Fix spark inconsitent dataset naming

pawel-big-lebowski:
https://github.com/OpenLineage/OpenLineage/issues/3284 -> not sure if this will help, but definitely the naming logic should go through PathUtils in those cases.


kacpermuda:
It seems to solve the problem, at least the one described in the issue. Thank you @pawel-big-lebowski for the quick fix !

kacpermuda (approved):

PR #3301, opened 2024-12-04
introduce transform transport

pawel-big-lebowski:
Implement new transport type to allow easy event modification before emitting. 

The `TransformTransport` is designed to enable event manipulation before emitting the event. Together with `CompositeTransport`, it can be used to 
send different events into multiple backends.

#### Configuration

- `type` - string, must be "transform". Required.
- `transformerClass` - class name of the event transformer. Class has to implement `io.openlineage.client.transports.transform.EventTransformer` interface and provide public no-arg constructor. Class needs to be available on the classpath. Required.
- `transformerProperties` - Extra properties that can be passed into `transformerClass` based on the configuration. Optional.
- `transport` - Transport configuration to emit modified events. Required.

#### Behavior

- The configured `transformerClass` will be used to alter events before the emission.
- Modified events will be passed into the configured `transport` for further processing.

#### `EventTransformer` interface

```java
public class CustomEventTransformer implements EventTransformer {
  @Override
  public void initialize(Map<String, String> properties) { ... }

  @Override
  public RunEvent transform(RunEvent event) { ... }

  @Override
  public DatasetEvent transform(DatasetEvent event) { .. }

  @Override
  public JobEvent transform(JobEvent event) { ... }
}
```

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
@@ -40,7 +40,10 @@ public final class CompositeConfig implements TransportConfig, MergeConfig<Compo
       // Handle List<Map<String, Object>> case
       this.transports =
           ((List<Map<String, Object>>) transports)
-              .stream().map(this::createTransportConfig).collect(Collectors.toList());
+              .stream()
+                  .map(this::createTransportConfig)
+                  .sorted(TransportConfig::compareTo)
+                  .collect(Collectors.toList());
     } else if (transports instanceof Map) {
       // Handle Map<String, Object> case
       Map<String, Object> transportMap = (Map<String, Object>) transports;
@@ -54,6 +57,7 @@ public final class CompositeConfig implements TransportConfig, MergeConfig<Compo
                     return nestedMap;
                   })
               .map(this::createTransportConfig)
+              .sorted(TransportConfig::compareTo)
               .collect(Collectors.toList());
     } else {
       throw new IllegalArgumentException("Invalid transports type");
--- a/client/java/src/main/java/io/openlineage/client/transports/EventTransformer.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/EventTransformer.java
@@ -31,7 +31,9 @@ public interface EventTransformer {
    * @param event
    * @return
    */
-  RunEvent transform(RunEvent event);
+  default RunEvent transform(RunEvent event) {
+    return event;
+  }
 
   /**
    * Transforms the DatasetEvent into a new DatasetEvent.
@@ -39,7 +41,9 @@ public interface EventTransformer {
    * @param event
    * @return
    */
-  DatasetEvent transform(DatasetEvent event);
+  default DatasetEvent transform(DatasetEvent event) {
+    return event;
+  }
 
   /**
    * Transforms the JobEvent into a new JobEvent.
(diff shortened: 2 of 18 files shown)


client/java/src/main/java/io/openlineage/client/transports/composite/CompositeConfig.java
```diff
@@ -3,16 +3,18 @@
 /* SPDX-License-Identifier: Apache-2.0
 */
 
-package io.openlineage.client.transports;
+package io.openlineage.client.transports.composite;
```
pawel-big-lebowski, line 6: @mobuchowski @JDarDagran  This PR moves transports into separate packages which, in my opinion, clears the burden within the codebase. On the other hand, this may be seen as breaking public API and potential java client users other than spark and flink integration will have to modify their `import` in case they manually instantiate transport classes. Let me know if you have other opinion on that.

Spark and Flink users, using Spark/Flink conf or yaml files, won't see the difference.

client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
```diff
@@ -54,6 +63,12 @@ public CompositeConfig(
                     return nestedMap;
                   })
               .map(this::createTransportConfig)
+              .sorted(
+                  (t1, t2) -> {
+                    return Optional.ofNullable(t1.getName())
+                        .orElse("")
+                        .compareTo(Optional.ofNullable(t2.getName()).orElse(""));
+                  })
```
ddebowczyk92, line 71: is this needed? a bit cumbersome to me, wouldn't be better to make the `CompositeConfig` implement the `Comparable` interface?

client/java/src/main/java/io/openlineage/client/transports/EventTransformer.java
```diff
@@ -0,0 +1,51 @@
...
+ * <p>Class implementing it needs to provide no-arg constructor
+ */
+public interface EventTransformer {
+
+  /**
+   * Initialize the transformer with properties.
+   *
+   * @param properties
+   */
+  void initialize(Map<String, String> properties);
```
ddebowczyk92, line 26: does not feel like a part of an interface to me, breaks contract behavior principle, such a method fits better to abstract class

ddebowczyk92 (changes requested):

client/java/src/test/java/io/openlineage/client/transports/TransformTransportTest.java
```diff
@@ -0,0 +1,236 @@
...
+      return null;
+    }
+
+    @Override
+    public JobEvent transform(JobEvent event) {
+      return null;
+    }
+  }
+
+  public static class SuccessfulEventTransformer implements EventTransformer {
```
mobuchowski, line 213: We could do with a test that actually changes something in the event.

client/java/src/main/java/io/openlineage/client/transports/EventTransformer.java
```diff
@@ -0,0 +1,51 @@
...
+   */
+  default void initialize(Map<String, String> properties) {}
+
+  /**
+   * Transforms the RunEvent into a new RunEvent.
+   *
+   * @param event
+   * @return
+   */
+  RunEvent transform(RunEvent event);
```
mobuchowski, line 34: Might be better with default impl for all of those returning passed event.

mobuchowski (approved):

PR #3305, opened 2024-12-06
Parallel event emitting for composite transport

pawel-big-lebowski:
Make `composite` transport emit the events in parallel for Java client. 
The behaviour should be default if `continueOnFailure` is `true`

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
@@ -28,13 +28,16 @@ public final class CompositeConfig implements TransportConfig, MergeConfig<Compo
 
   @Getter @Setter private List<TransportConfig> transports;
 
-  @Getter @Setter private boolean continueOnFailure;
+  @Getter @Setter private Boolean continueOnFailure;
+
+  @Getter @Setter private Boolean withThreadPool;
 
   @JsonCreator
   @SuppressWarnings("unchecked")
   public CompositeConfig(
       @JsonProperty("transports") Object transports,
-      @JsonProperty("continueOnFailure") boolean continueOnFailure) {
+      @JsonProperty("continueOnFailure") Boolean continueOnFailure,
+      @JsonProperty("withThreadPool") Boolean withThreadPool) {
 
     if (transports instanceof List) {
       // Handle List<Map<String, Object>> case
@@ -63,14 +66,16 @@ public final class CompositeConfig implements TransportConfig, MergeConfig<Compo
       throw new IllegalArgumentException("Invalid transports type");
     }
 
-    this.continueOnFailure = continueOnFailure;
+    this.continueOnFailure = (continueOnFailure == null) ? true : continueOnFailure;
+    this.withThreadPool = (withThreadPool == null) ? true : withThreadPool;
   }
 
   public static CompositeConfig createFromTransportConfigs(
-      List<TransportConfig> transports, boolean continueOnFailure) {
+      List<TransportConfig> transports, boolean continueOnFailure, boolean withThreadPool) {
     CompositeConfig compositeConfig = new CompositeConfig();
     compositeConfig.setTransports(transports);
     compositeConfig.setContinueOnFailure(continueOnFailure);
+    compositeConfig.setWithThreadPool(withThreadPool);
     return compositeConfig;
   }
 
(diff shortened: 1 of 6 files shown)


client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
```diff
@@ -36,41 +40,83 @@ public List<Transport> getTransports() {
...
+    genericEmit(jobEvent);
+  }
+
+  /**
+   * Emit events in parallel using a thread pool. Gets Object as param since different event types
+   * do not share a common base class nor interface.
+   *
+   * @param event
+   */
+  private void genericEmit(Object event) {
```
mobuchowski, line 62:
```suggestion
  private void doEmit(BaseEvent event) {
```

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
```diff
@@ -36,41 +40,83 @@ public List<Transport> getTransports() {
...
+      for (Transport transport : transports) {
+        emit(transport, event);
+      }
+    } else {
+      // Emit events in parallel
       try {
-        transport.emit(jobEvent);
-      } catch (Exception e) {
-        handleEmissionFailure(transport, e);
+        Executors.newCachedThreadPool()
```
mobuchowski, line 71: Do not create new thread pool on each emit. Create it on transport constructor, and maybe allow configuration of it.

client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
```diff
@@ -28,13 +28,13 @@ public final class CompositeConfig implements TransportConfig, MergeConfig<Compo
 
   @Getter @Setter private List<TransportConfig> transports;
 
-  @Getter @Setter private boolean continueOnFailure;
+  @Getter @Setter private Boolean continueOnFailure;
```
mobuchowski, line 31: why the need to have boxed values?

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
```diff
@@ -36,41 +40,83 @@ public List<Transport> getTransports() {
...
+                                  emit(t, event);
+                                  return null;
+                                })
+                    .collect(Collectors.toList()))
+            .forEach(
+                f -> {
+                  try {
+                    f.get();
+                  } catch (InterruptedException | ExecutionException e) {
+                    log.warn("Failed to emit event", e);
```
mobuchowski, line 87: Indicate for which transport the emission failed

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
```diff
@@ -36,41 +40,83 @@ public List<Transport> getTransports() {
...
 
-  private void handleEmissionFailure(Transport transport, Exception e) {
-    if (!config.isContinueOnFailure()) {
-      throw new RuntimeException(
-          "Transport " + transport.getClass().getSimpleName() + " failed to emit event", e);
+  /**
+   * @param transport
+   * @param event
+   */
+  private void emit(Transport transport, Object event) {
```
mobuchowski, line 101:
```suggestion
  private void emit(Transport transport, BaseEvent event) {
```

client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
arturowczarek: If someone doesn't pass the value in the configuration it defaults to false. We want to be able to tell if the value was empty (and we should replace it with `true`)

client/java/src/main/java/io/openlineage/client/transports/CompositeConfig.java
```diff
@@ -28,13 +28,16 @@ public final class CompositeConfig implements TransportConfig, MergeConfig<Compo
...
+
+  @Getter @Setter private Boolean threadPoolActive;
```
dolfinus, line 33: IMHO this naming is a bit confusing. What do you think about `useThreadPool` or `withThreadPool`?

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
pawel-big-lebowski: I've also included test to verify a pool is shutdown when needed.

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
```diff
@@ -36,46 +50,94 @@ public List<Transport> getTransports() {
...
+      } else if (event instanceof OpenLineage.DatasetEvent) {
+        transport.emit((OpenLineage.DatasetEvent) event);
+      } else if (event instanceof OpenLineage.JobEvent) {
+        transport.emit((OpenLineage.JobEvent) event);
+      } else {
+        throw new IllegalArgumentException("Unsupported event type: " + event.getClass().getName());
+      }
+    } catch (Exception e) {
+      log.warn("Transport {} failed to emit event", transport.getClass().getSimpleName(), e);
+      if (!config.getContinueOnFailure()) {
```
mobuchowski, line 129: Why in both cases we're throwing an exception, but different ones?

mobuchowski (approved):
One nit to explain but otherwise looks good to me

PR #3321, opened 2024-12-13
Add extension system for the GCPLineage transport

jphalip:
This would allow further customization of the GRPC connection settings


harels:
thanks for your contribution @jphalip !
a few notes:
1. would you mind adding some more context to the description on what kind of customization this enables, for reference.
2. can you also follow the DCO guidelines?
3. the build for the client is failing.

jphalip:
Thanks @harels 

1) This would allow customizing the gRPC connection, for example by setting specific request headers in some environments.
2) I've just resubmitted the commit with the signoff.
3) I'm not sure I could see the failures. Could you please share the error messages?

jphalip:
UPDATE: Actually, I could see that it was failing because of Spotless. I've just updated the commit to included the spotlessApply'ed code. Hopefully that fixes it.

pawel-big-lebowski:
All the customisations can be achieved with ServiceLoader and loading custom classes. On the other hand, I think we should try to limit the amount of extensions allowed and try to implement them as generic as possible. The interface:
```
public interface GCPLineageTransportExtension {
  void updateSettings(com.google.cloud.datacatalog.lineage.v1.LineageSettings.Builder builder);
}
```
seems tightly-coupled with gcp classes. 

Perhaps you can write few more lines about the problem, so that we could think if there is some more generic solution to this. 

If a problem is: in my runtime I would like to inject extra transport settings which cannot be passed through standard configuration way (like env properties, spark config or yml file), then the solution would be to allow providing custom extensions to overwrite `SparkOpenlineageConfig`, which contains transport config entries you need.

jphalip:
> in my runtime I would like to inject extra transport settings which cannot be passed through standard configuration way (like env properties, spark config or yml file)

This is indeed the problem we're trying to solve for.

I'm unclear about the solution you're proposing:

> the solution would be to allow providing custom extensions to overwrite SparkOpenlineageConfig, which contains transport config entries you need.

Could you clarify?

pawel-big-lebowski:
Sure, I can clarify on this. 

Within Spark integration, there is `ArgumentParser` class with a parse method:
https://github.com/OpenLineage/OpenLineage/blob/1403415145213da5a7f76fadd29e473a1966ef24/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java#L67

This methods tries reading configuration from the yaml file:
https://github.com/OpenLineage/OpenLineage/blob/1403415145213da5a7f76fadd29e473a1966ef24/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java#L71

and environment variables:
https://github.com/OpenLineage/OpenLineage/blob/1403415145213da5a7f76fadd29e473a1966ef24/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java#L71

and Spark config as well:
https://github.com/OpenLineage/OpenLineage/blob/1403415145213da5a7f76fadd29e473a1966ef24/integration/spark/app/src/main/java/io/openlineage/spark/agent/ArgumentParser.java#L77

Configuration objects are merged by `mergeWith` method. 

One could create `CustomConfigExtractor` interface with a method:
```
SparkOpenLineageConfig extract()
``` 
and try instantiating it with ServiceLoader. In case it was present, one can merge `SparkOpenLineageConfig` returned from `CustomConfigExtractor`  with the already created `SparkOpenLineageConfig` instance.

jphalip:
Thanks for the clarification. If I understand correctly, this would require adding attributes to the `SparkOpenLineageConfig` class (and other descendants of `OpenLineageConfig` for other integrations), is that correct? If so, then I would prefer not using that route as this could pollute those classes with lots of attributes that no one should really ever care about. Do I make sense?

pawel-big-lebowski:
It's not clear to me what config entries you are trying to inject. If these are only related to Gcp lineage transport, then it would be enough to add some fields into the `GcpLineageTransportConfig` class:
https://github.com/OpenLineage/OpenLineage/blob/main/client/java/transports-gcplineage/src/main/java/io/openlineage/client/transports/gcplineage/GcpLineageTransportConfig.java
In this case, you don't need to add any attributes to `SparkOpenLineageConfig`. 

Values for the added fields would be merged with other fields taken from user's configuration. 
Perhaps I could prepare some PoC of this code, so that you can get better feeling of what my idea is about. 

Additionally, in contrast to my previous message, I would replace ServiceLoader approach with extra config entry that specifies class to read extra config from. 

Pls let me know having extra properties within `GcpLineageTransportConfig` works for you.

pawel-big-lebowski:
Prepared initial PR to have a better understanding of what I was proposing:
https://github.com/OpenLineage/OpenLineage/pull/3373/files

@mobuchowski Does this make sense to you?
@jphalip Will this solve the issue you're having?

jphalip:
Thanks for working on that other PR. Could you explain how this would work in practice? How/who would set up the extra custom config? Would it have to be the user directly, or could it be the platform runner (e.g. Dataproc)?

Also let me try to clarify our use case. We would like to set up some special gRPC request headers for all requests going through the GCPLineageTransport when run from Dataproc. The issue I'm seeing with relying on the TransportConfig object is that the user could easily inadvertently mess up that config. Ideally we'd like all of this to be completely transparent to the user. If the user chooses to configure their environment to use the GCPLineageTransport, then our Dataproc environment would automatically and reliably set those headers. Does that make sense?

pawel-big-lebowski:
First, dataproc environment would need to provide an environment variable:
```
spark.openlineage.customConfigProviderClass=com.google.CustomConfigProvider
```
Additionally, dataproc would need to provide `com.google.CustomConfigProvider` class on classpath. gPRC request headers should be configurable within `GcpLineageTransportConfig`. `CustomConfigProvider` should return `SparkOpenlineageConfig` containing only `TransportConfig` which should contain only `gPRC` related field of `GcpLineageTransportConfig`. This should be possible as configuration objects are merged, so custom config provider would overwrite only the `gPRC` field while leaving other config entries intact. 

Is this transparent to the user? Yes. This injects extra properties via config that a user does not have to be aware of.

kacpermuda:
@jphalip @pawel-big-lebowski It his PR still relevant or some work-around has been implemented already?

harels:
closing due to inactivity, please feel free to reopen otherwise.

PR #3342, opened 2024-12-18
Add extra test to verify custom run and job facets are created for application events

pawel-big-lebowski:
* Write extra test that verifies if custom facets are built and attached onto the event
 * Modify existing tests which checked every element of the collection without verifying if collection was non-empty. In fact collection was empty, as argument captor was not called.

### DIFF
--- a/integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/SparkApplicationExecutionContextTest.java
+++ b/integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/SparkApplicationExecutionContextTest.java
@@ -15,17 +15,24 @@ import static org.mockito.Mockito.when;
 
 import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import io.openlineage.client.OpenLineage;
+import io.openlineage.client.OpenLineage.DefaultJobFacet;
+import io.openlineage.client.OpenLineage.DefaultRunFacet;
+import io.openlineage.client.OpenLineage.JobFacet;
 import io.openlineage.client.OpenLineage.RunEvent;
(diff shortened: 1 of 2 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
```diff
@@ -421,7 +426,10 @@ private OpenLineage.JobFacets buildJobFacets(
                         .timer(
                             "openlineage.spark.facets.job.execution.time",
                             "facet.builder",
-                            fn.getClass().getCanonicalName())
+                            Optional.ofNullable(fn)
+                                .map(Object::getClass)
+                                .map(Class::getCanonicalName)
+                                .orElse(""))
```
pawel-big-lebowski, line 432: Added as this was throwing NullPointerException for inline custom builders

integration/spark/app/src/test/java/io/openlineage/spark/agent/lifecycle/SparkApplicationExecutionContextTest.java
```diff
@@ -111,14 +120,16 @@ void testRunAndJobAreSet(SparkSession spark) {
       context.start(mock(SparkListenerApplicationStart.class));
       context.end(mock(SparkListenerApplicationEnd.class));
     }
+    verify(eventEmitter, times(2)).emit(lineageEvent.capture());
```
pawel-big-lebowski, line 123: I think the test was not working properly without this line added. No lineage events were captured and assertions in the for loop were not called at all.

mobuchowski (approved):

PR #3360, opened 2024-12-24
[Maintain][dbt] Add unit tests for structured logs parsing

MassyB:
### Problem

Follow-up to this PR https://github.com/OpenLineage/OpenLineage/pull/3314.
Test coverage can be better.

### Solution

This PR add unit tests to verify:

1. the relationships between START/COMPLET/FAIL OL events, to check that they have the same `runId`
2. the relationships between parent/child OL events, to check that the `parent` facet has the appropriate `runId`
3. the usage of the `consume_struct_logs_function`

#### One-line summary:

Add unit tests for dbt structured logs parsing

### Checklist

- [x] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [x] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [x] Your changes are accompanied by tests (_if relevant_)
- [ ] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] Your comment includes a one-liner for the changelog about the specific purpose of the change (_not required for changes to tests, docs, or CI config_)
- [ ] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_if relevant_)
- [ ] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----
SPDX-License-Identifier: Apache-2.0\
Copyright 2018-2024 contributors to the OpenLineage project

PR #3362, opened 2024-12-26
[Feature][dbt] Consume dbt structured logs for `test` and `build` commands

MassyB:
### Problem

Follow-up to this PR https://github.com/OpenLineage/OpenLineage/pull/3314.
Here I add the `test` and `build` commands.
The test command is interesting, it adds the `dataQualityAssertions`. 

I'm  using (and adding more) unit tests defined in this PR https://github.com/OpenLineage/OpenLineage/pull/3360

### Solution

Compared to how it's done in the other dbt processors (`Local` and `Cloud`), we have a single assertion in the facet.
This follows the execution model of dbt, each test is executed independently from the other tests. In addition, progress is reported on that test alone without waiting for other tests (those linked to a single node) to finish.

You can have a look at the unit tests ([example](https://github.com/MassyB/OpenLineage/blob/3af8c6a7a9b50a2d17bbf441500c2f75725b4a81/integration/common/tests/dbt/structured_logs/postgres/test/results/test_ol_events.json#L1)) to see which OL events are generated and their shape when users execute a `dbt test` command.

#### One-line summary:
Consume dbt structured logs for `test` and `build` commands.

### Checklist

----


integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -278,6 +299,25 @@ def parse_node_finished_event(self, event) -> RunEvent:
             outputs=outputs,
         )
 
+    def _get_attached_dataset(self, test_node_id: str) -> InputDataset:
+        """
+        gets the input the data test is attached to
+        """
+        all_nodes = {**self.compiled_manifest["nodes"], **self.compiled_manifest["sources"]}
+        test_node = all_nodes[test_node_id]
+        attached_node_id = test_node["attached_node"]
```
MassyB, line 308: Here the test is linked to its attached node whereas in the current implementation it has the arbitrary value of the last node in the parent map

https://github.com/MassyB/OpenLineage/blob/b7837beeb0154cf084cb1913c8d46688ae2eea24/integration/common/openlineage/common/provider/dbt/processor.py#L394-L397

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -265,6 +272,20 @@ def parse_node_finished_event(self, event) -> RunEvent:
             for model_input in self._get_model_inputs(node_unique_id)
         ]
         outputs = [self.node_to_output_dataset(node=self._get_model_node(node_unique_id), has_facets=True)]
+        if resource_type == "test":
+            success = node_status == "pass"
+            assertion = self._get_assertion(node_unique_id, success)
+            assertion_facet = data_quality_assertions_dataset.DataQualityAssertionsDatasetFacet(
+                assertions=[assertion]
```
MassyB, line 279: A notable difference here. There is a single asseration per test since there is only one assertion that is tested in every dbt test. Have a look at the unit [tests](https://github.com/MassyB/OpenLineage/blob/b7837beeb0154cf084cb1913c8d46688ae2eea24/integration/common/tests/dbt/structured_logs/postgres/events/results/successful_test_NodeFinished_OL.yaml#L226)

all the assertions that pertain to the same node (model) have the same input dataset.

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -424,6 +464,9 @@ def _get_job_name(self, event) -> str:
         elif node_unique_id.startswith("snapshot"):
```
MassyB, line 464: I repeated what was done in the `processor.py`. IMO this should be as simple as the `node_id` it's unique across all dbt nodes of the project.

Same as for Airflow task for example https://openlineage.io/docs/spec/naming/#job-naming

LMK what you think !

integration/common/openlineage/common/test.py
```diff
@@ -80,7 +80,7 @@ def setup_jinja() -> Environment:
 env = setup_jinja()
 
 
-def match(expected, result) -> bool:
+def match(expected, result, ordered_list=False) -> bool:
```
MassyB, line 83: I have use cases where I want to test list equality (order and every element in it)

mobuchowski: I don't remember why the "by name" ordering was added (it was added in 2021 :) ), but it looks like nothing breaks now - maybe this should be the only mode, and if someone needs non-ordered match they can readd this later?

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -253,7 +260,7 @@ def parse_node_finished_event(self, event) -> RunEvent:
 
         event_type = RunState.COMPLETE
 
-        if node_status != "success":
+        if node_status not in ("success", "pass"):
```
mobuchowski, line 263: Not necessarily a thing to do now, but might be worth to have those statuses as an enum

integration/common/openlineage/common/provider/dbt/structured_logs.py
mobuchowski: This works right now because the tests are always attached to only one model.

`attached_node` approach is better, but it's working only as of manifest v9 - dbt 1.5, meanwhile current approach works as far as dbt 0.20.0 AFAIK.

Manifest v8 does not have that property https://schemas.getdbt.com/dbt/manifest/v8/index.html#tab-pane_nodes_additionalProperties_oneOf_i6 

I think doing this via `attached_node` is fine, since this is a new functionality, and even dbt 1.6 is out of support https://docs.getdbt.com/docs/dbt-versions/core

mobuchowski (approved):

integration/common/openlineage/common/test.py
MassyB: I can go ahead and make it the only mode.

integration/common/openlineage/common/provider/dbt/structured_logs.py
MassyB: I'll go ahead and put the `node_id` as the OL job name

integration/common/openlineage/common/test.py
MassyB: changed here https://github.com/OpenLineage/OpenLineage/pull/3362/commits/3695f168b2fe16336ff4f9695b225d7790e80036

integration/common/openlineage/common/provider/dbt/structured_logs.py
MassyB: changed here https://github.com/OpenLineage/OpenLineage/pull/3362/commits/209ee89c0eac2b167f7903d98acd9fc9ecca767c

integration/common/openlineage/common/test.py
MassyB: it's actually making some tests [fail](https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/13910/workflows/8149afa4-678d-4561-aca7-aa4730d29448/jobs/346141). I don't want to overload this PR. I'll keep the "by name" behavior.

mobuchowski (approved):

PR #3375, opened 2025-01-04
Add `oleander.dev` as ol consumer/producer

wslulciuc:
This PR adds [`oleander.dev`](https://oleander.dev) as an OpenLineage consumer/producer.


merobi-hub (approved):
@wslulciuc awesome to see this. Have you checked how the logo looks on the cards? As is (without padding), it might be inconsistent with the other ones. And are you open to possibly adding more detail to the description strings?

merobi-hub:
Here's how it looks edited:

Screenshot from 2025-01-03 23-02-06

merobi-hub:
@wslulciuc may I suggest adding the name to the card? I took the liberty of creating a logo with the name added in case you agree with this approach. I used the font from header logo on the oleander site. It's included here as a png file:
Screenshot from 2025-01-04 14-02-19

wslulciuc:
@merobi-hub thanks for the suggestions and helping out with the PR! I've updated the description with more context.

PR #3392, opened 2025-01-10
Dataset partitioning proposal

pawel-big-lebowski:
Google docs discussion document into formal proposal.

### DIFF
--- a/proposals/3392/dataset_partitioning.md
+++ b/proposals/3392/dataset_partitioning.md
@@ -1,6 +1,6 @@
 ---
-Authors: Paweł Leszczyński, Maxim Martynov
-Created: January tenth, 2025 
+Authors: Paweł Leszczyński, Maxim Martynov, Maciej Obuchowski
+Created: February 18, 2025 
 ---
 
 # OpenLineage dataset partitioning proposal
@@ -10,12 +10,14 @@ Lineage metadata should contain information needed to answer the following quest
 2. What partitions have been read by a particular run?
 3. What partitions have been updated by a particular run?
 
-Partition filtering is just a sub problem of describing a subset of the input dataset that has been used.
-We propose `SubsetDatasetFacet` to answer (2) and (3) while keeping in mind, that it may be used in other contexts as well.
+Partition filtering is just a sub-problem of describing a subset of the input dataset that has been used.
+We propose subset dataset facet to answer (2) and (3) while keeping in mind, that it may be used in other contexts, like for example
+query filtering
 
 ## PartitionDatasetFacet
 
-Create `PartitionDatasetFacet`
+We introduce: `PartitionDatasetFacet`:
+
 ```json
 {
   "PartitionDatasetFacet": {
@@ -42,9 +44,9 @@ Create `PartitionDatasetFacet`
               }
             }
           },
-         "partitionedBySqlClause": {
+         "description": {
             "type": "string",
-            "description": "Optional SQL clause definition like: '(days(ts), category)' for SQL clause `PARTITIONED BY (days(ts), category)` used."
+            "description": "Partitioning description, can be SQL definition clause like `PARTITIONED BY (days(ts), category)`."
          }
         },
         "required": ["dimensions"]
@@ -75,7 +77,7 @@ Examples:
         "transform": "identity"
       }
     ], 
-     "sqlClause": "('business_date', 'country')" 
+     "description": "PARTITIONED BY ('business_date', 'country')" 
   }
 }
 ```
@@ -90,7 +92,8 @@ Examples:
         ],
         "transform": "date(event_time)"
       }
-    ]
+    ],
+    "description": "PARTITIONED BY (date(event_time))"
   }
 }
 ```
@@ -112,107 +115,53 @@ Examples:
         ],
         "transform": "bar(col_c)"
       }
-    ]
+    ], 
+    "description": "PARTITIONED BY foo(col_a, col_b), bar(col_c)"
   }
 }
 ```
 
-## SubsetDatasetFacet
+## Describing subset of a dataset
 
-It can be argued if a `SubsetDatasetFacet` should be a part of partitioning proposal. 
-The counterargument to this, is that it makes more sense to discuss partitioning 
-with a facet that is capable of describing partitions being read or written to.
-Otherwise, it is easy to miss the practical context of discussing partitioning.
+It can be argued if a facet to describe dataset's subset should be a part of partitioning proposal. 
+However, it makes more sense to discuss partitioning facets
+together with a facet that is capable of describing partitions being read or written to.
+Otherwise, it is easy to miss the practical context of partitioning metadata.
 
-`SubsetDatasetFacet` may describe either `input` or `output` dataset. 
-Possible scenarios to cover include: 
+Subset facet may describe either `input` or `output` dataset. 
+Common scenarios to cover include: 
  * SQL query filtering: `WHERE a = '...' AND b = '...'`,
- * Job reading a path defined with regex `/some/path/a=*/b=1`,
- * Job writing to `s3://bucket/dataset/business_date=20241101`,
- * Job reading to a list of partitions,
- * Job writing to a list of partitions,
+ * Job reading/writing to specified path `s3://bucket/dataset/business_date=20241101`,
+ * Job reading a list of partitions,
+ * Job writing to a specified partition,
  * Job reading a dataset by a predicate on partitioned fields.
 
-### Base subset dataset facet
-
-From the examples above, we can distinguish logical and physical definitions
-of a subset. A proposed facet should be capable of storing both definitions at
-the same time as sometimes reading a path `/some/b=1` may be equivalent
-to logical filtering `b=1`. The physical notion describes the actual location while
-the logical describes a condition applied. 
+We start with logical subset definitions:
 
 ```json
 {
-   "BaseSubsetDatasetFacet": {
-      "allOf": [
-         {
-            "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/DatasetFacet"
-         },
-         {
-            "type": "object",
-            "properties": {
-               "logical": {
-                  "$ref": "#/$defs/LogicalSubsetDefinition"
-               },
-               "physical": {
-                  "$ref": "#/$defs/PhysicalSubsetDefinition"
-               }
-            }
-         }
-      ],
+   "BaseSubsetCondition": {
       "type": "object",
       "properties": {
-         "subset": {
-            "$ref": "#/$defs/BaseSubsetDatasetFacet"
-         }
+         "type": "string"
       }
-   }
-}
-```
-
-A subset dataset facet can be used either as `InputDatasetFacet` and `OutputDatasetFacet`,
-however it cannot extend both objects at the same time.
-To solve this, we introduce:
-```json
-{
-   "InputSubsetDatasetFacet": {
-      "allOf": [
-         {
-            "$ref": "#/$defs/InputDatasetFacet"
-         },
-         {
-            "$ref": "#/$defs/BaseSubsetDatasetFacet"
-         }
-      ]
    },
-   "OutputSubsetDatasetFacet": {
-      "allOf": [
-         {
-            "$ref": "#/$defs/OutputDatasetFacet"
-         }
-      ]
-   }
-}
-```
-
-## Logical subset definition
-
-```json
-{
-   "LogicalSubsetDefinition": { },
    "BinarySubsetCondition": {
       "allOf": [
          {
-            "$ref": "LogicalSubsetDefinition"
+            "$ref": "BaseSubsetCondition"
          },
          {
             "type": "object",
             "properties": {
                "left": {
-                  "type": "LogicalSubsetDefinition"
+                  "type": "BaseSubsetCondition"
                },
                "right": {
-                  "type": "LogicalSubsetDefinition"
+                  "type": "BaseSubsetCondition"
+               },
+               "type": {
+                  "const": "binary"
                },
                "operator": {
                   "enum": [
@@ -224,19 +173,22 @@ To solve this, we introduce:
          }
       ]
    },
-   "LiteralSubsetCondition": {
+   "CompareSubsetCondition": {
       "allOf": [
          {
-            "$ref": "LogicalSubsetDefinition"
+            "$ref": "BaseSubsetCondition"
          },
          {
             "type": "object",
             "properties": {
-               "field": {
-                  "type": "String"
+               "type": {
+                  "const": "compare"
                },
-               "value": {
-                  "type": "String"
+               "left": {
+                  "type": "CompareExpression"
+               },
+               "right": {
+                  "type": "CompareExpression"
                },
                "comparison": {
                   "enum": [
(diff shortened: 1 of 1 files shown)


proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,301 @@
...
+      "field": "col1",
+      "value": "7",
+      "comparison": "EQUAL"
+    },
+    "right": {
+      "field": "col2",
+      "value": "9",
+      "comparison": "EQUAL"
+    },
+    "operator": "AND"
```
dolfinus, line 231: If there are multiple `AND`s, how this should be represented?

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,301 @@
...
+                "transform": {
+                  "type": "string"
+                }
+              }
+            }
+          },
+         "sqlClause": {
+            "type": "string",
+            "description": "Optional SQL clause definition like: 'PARTITIONED BY (days(ts), category)'."
+         }
```
dolfinus, line 51: What is the purpose of storing only `PARTITIONED BY` part of table's DDL? I mean, if there is already a full DDL in another facet, could if be reused?

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,301 @@
...
+```json
+{
+  "partitions": {
+    "updated": [
+      {
+        "business_date": "2024-10-15",
+        "country": "PL"
+      },
+      {
+        "business_date": "2024-10-15",
```
dolfinus, line 282: This describes only Hive-style partitioning, but other datasources may have another partitioning schemas. For example, Postgres allows `LIST`, `RANGE` and `HASH` partitioning. So instead of including value of partitioning columns (which can be arbitrary, and different for each row in the dataset), users may want to know just the partition name (in Postgres partition is just another table bound to main one) or a list of partition column constraints (e.g. `business_at BETWEEN '2024-01-01' AND '2025-01-01' for `RANGE` partitioning).

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,301 @@
...
+
+ * (1) Simple partitioning by two columns: `business_date` and `country`
+
+```json
+{
+  "partitions": {
+    "dimensions": [
+      {
+        "fields": [
+          "busniess_date"
```
dolfinus, line 70:
```suggestion
          "business_date"
```

proposals/3392/dataset_partitioning.md
pawel-big-lebowski: `AND` is binary condition but `left` and `right` can be composed of other binary conditions as well.

proposals/3392/dataset_partitioning.md
pawel-big-lebowski: How about renaming this into `partitionedBySqlClause` and omitting `PARTITIONED BY`? 
I think facets should be self-contained and some redundancy is Ok, even if there is another facet with a full DDL.

dolfinus: So for PostgreSQL's `PARTITION BY RANGE (field)` it will be `RANGE(field)`?

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+An example of this would be a dataset partitioned by `business_date` and `country`. A job run affecting many countries for a single date would affect multiple partitions and this would be described with a facet:
+
+```json
+{
+  "partitions": {
+    "updated": [
+      {
+        "business_date": "2024-10-15",
```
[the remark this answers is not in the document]
pawel-big-lebowski: I've included `operation`, `location` and `identifier` fields.

pawel-big-lebowski:
@dolfinus Thanks for the input. I've rearranged output dataset partitioning section to support some of the cases you mentioned. For others, I added `BasePartitionOutputDatasetFacet` which I think allows to extend easily. 

I've added you to a list of authors of this proposal. I think your input is substantial. I hope you don't mind.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,331 @@
...
+   }
+}
+```
+
+An example of this would be a dataset partitioned by `business_date` and `country`. A job run affecting many countries for a single date would affect multiple partitions and this would be described with a facet:
+
+```json
+{
+   "$id": "https://openlineage.io/spec/facets/1-0-0/SelectedPartitionsOutputDatasetFacet.json",
+   "operation": "UPDATE",
```
dolfinus, line 299: Theoretically, several operations could be performed in the same time. For example, writing dataframe with `df.write.mode("overwrite")` and `SparkSession.builder.config("spark.sql.sources.partitionOverwriteMode", "dynamic")` may:
* drop existing partitions, already presented in a filesystem and matched partition column values in the dataframe
* create new partitions, if partition column values were present in the dataframe only, but not in the filesystem

But if I understand it correctly, Spark currently doesn't report which partitions were dropped/created in this case. So the only way to capture exact partitions info is to parse commands manually, like AddPartitionCommand, AlterPartitionCommand and so on, which usually perform only one operation with specific list of partitions.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,331 @@
...
+                        "properties": {
+                           "location": {
+                              "type": "string",
+                              "description": "Optionally provided location of the partition specified"
+                           },
+                           "identifier": {
+                              "type": "string",
+                              "description": "Optionally provided identifier of the partition specified"
+                           }
+                        },
```
dolfinus, line 276: This theoretically may interfere with partition column values:
```json
{
  "$id": "https://openlineage.io/spec/facets/1-0-0/SelectedPartitionsOutputDatasetFacet.json",
  "operation": "UPDATE",
  "partitions": [
    {
      "location: "hdfs:///warehouse/schema.db/table/location=Europe/business_dt=2025-01-01",
      "location: "Europe",
      "business_dt: "2025-01-01",
    },
  ]
}
```

What do you think about schema like this:
```json
{
  "$id": "https://openlineage.io/spec/facets/1-0-0/SelectedPartitionsOutputDatasetFacet.json",
  "operation": "UPDATE",
  "partitions": [
    {
      "location: "hdfs:///warehouse/schema.db/table/location=Europe/business_dt=2025-01-01",
      "values: {
        "location": "Europe",
        "business_dt: "2025-01-01",
      },
    },
  ]
}
```
So these fields will not mix.

proposals/3392/dataset_partitioning.md
dolfinus: For me, this looks like a part of another proposal leaked to partitioning spec. Is that really needed to describe a list of partitions read by Spark or other integration?

proposals/3392/dataset_partitioning.md
pawel-big-lebowski: I am not sure if one could support dynamic partitioning automatically. 
But it would definitely make sense to change `operation` property location and have it per partitions item.

proposals/3392/dataset_partitioning.md
dolfinus: Ok

proposals/3392/dataset_partitioning.md
dolfinus: Could we use the same facet for input & output partitions? And move this facet with filters to another proposal.

dolfinus:
I like how SelectedPartitionsOutputDatasetFacet looks like, but have [doubts](https://github.com/OpenLineage/OpenLineage/pull/3392#discussion_r1913153387) about FilterInputDatasetFacet.

Also I don't have developer permissions here, so I cannot slap on Approve button :)

proposals/3392/dataset_partitioning.md
mobuchowski: I agree with @dolfinus - the proposal makes sense even without filtering, and it's a special case that we might want to think about separately, with more clear use case.

pawel-big-lebowski: I missed that, sry.  I will remove this part onto another PR.

pawel-big-lebowski: https://github.com/OpenLineage/OpenLineage/pull/3470

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,236 @@
...
+}
+```
+
+### Output partitions' dataset facet
+
+```json
+{
+   "$schema": "https://json-schema.org/draft/2020-12/schema",
+   "$defs": {
+      "BasePartitionOutputDatasetFacet": {
```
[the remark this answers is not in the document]
pawel-big-lebowski: It would make more sense to me now to include subset dataset facet again within this proposal. 
Output partitions' dataset facet should be a subset dataset facet as well.
A single partition facet or selected partition facet should extend subset facet. 
Does it make sense?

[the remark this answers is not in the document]
dolfinus: > include subset dataset facet again within this proposal

Sorry, don't get what facet do you mean. Could you give an example?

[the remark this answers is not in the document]
pawel-big-lebowski: https://github.com/OpenLineage/OpenLineage/pull/3470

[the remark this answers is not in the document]
dolfinus: I mean, why? `FileScanRDD` has attribute `filePartitions: FilePartition` with `files: Array[PartitionedFile]`, `HadoopFsRelation` has attribute `location: FileIndex` with nested `inputFiles: Array[String]`, which can be used to extract partition paths/names from.
https://jaceklaskowski.gitbooks.io/mastering-spark-sql/content/spark-sql-SparkPlan-FileSourceScanExec.html

Yes, `FileSourceScanExec` contains `partitionFilters`, ``dataFilters` which can be used to build `SubsetDatasetFacet`, but it could be hard to interpret by both users and OL events consumer. For example, `spark.read.csv("/some/path/a=*/b=1")` will have filters like `b=1`, which does not point to exact partition. But result of file scanning will have exact paths to each file matching this filter.

pawel-big-lebowski:
@mobuchowski @dolfinus Thank you for your support on this. I agree it makes sense to have same data structure to describe the subset of inputs and outputs. I included subset dataset facets within this proposal, because this is necessary to describe partitions read or written. If you think it's too much for one go, then I think we should start with dataset's subset definition first. 

I tried to put as much examples of subset conditions as possible.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,474 @@
...
+SELECT * FROM table WHERE col1 = 7 and col2 = 9
+``` 
+would be:
+```json
+{
+   "namespace": "dataset-namespace",
+   "name": "dataset-name",
+   "inputFacets": {
+     "subset": {
+       "$id": "https://openlineage.io/spec/facets/1-0-0/BinarySubsetCondition.json",
```
mobuchowski, line 371: I don't really like having OpenApi refs as semantically relevant part of the spec. 

I don't know what the better option would be if we want to make sure it's extensible, but maybe `type` field would be a better option?

Like, this example would be

```json
{
   "namespace": "dataset-namespace",
   "name": "dataset-name",
   "inputFacets": {
     "subset": {
       "type": "binary",
       "operator": "AND"
       "left": {
          "type": "condition", 
          "field": "col1",
          "value": "7",
          "comparison": "EQUAL"
       },
       "right": {
          "type": "condition",
          "field": "col2",
          "value": "9",
          "comparison": "EQUAL"
       },
     }
   }
}
``` 

Do we have to have `literal condition` operator, or would it be resolved by nesting?
Like, we'd just have a comparison operator, and left and right sides would be represented by nested references/literals/whatever else.

This would also solve the issue with 

```
SELECT * FROM table WHERE 7 >= col1 and col2 = 9
```

even if you normalize the order, you can very easily mess up comparison operator. 

I think the whole idea is good, but rather tricky and possibly complicated to implement: would have to support a large subset of SQL for it to work reliably.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,474 @@
...
+ * Job reading to a list of partitions,
+ * Job writing to a specified partition,
+ * Job reading a dataset by a predicate on partitioned fields.
+
+From the examples above, we can distinguish logical and physical definitions
+of a subset. A proposed facet should be capable of storing both definitions at
+the same time as sometimes as reading a path `/some/b=1` may be equivalent
+to logical filtering `b=1`. This can happen for Hive model where partitioning correlates
+with directory structure. The physical notion describes the actual location while
+the logical describes a condition applied. To make it even more complex, for Spark jobs it is
```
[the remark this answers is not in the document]
dolfinus: Isn't each kafka topic a separate dataset?

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,474 @@
...
+  .filter(col("c") == 2)
+```
+
+would result in:
+
+```json
+{
+   "namespace": "dataset-namespace",
+   "name": "dataset-name",
+   "outputFacets": {
```
dolfinus, line 434:
```suggestion
   "inputFacets": {
```

dolfinus:
I see that current approach tries to collect as much information as possible - raw SQL filters, paths, paths with globs, set of partition columns and so on.

But IMHO these should be different facets, and not the part of the same facet. For example, Spark has both logical and physical plan representations. Using code like:
```python
df = spark.read.format("orc").load("s3://bucket/some/path/a=*/b=1")
df.count()
```
will produce logical plan with filter "s3://bucket/some/path/a=*/b=1" against input path, and physical plan containing the exact partitions list `["s3://bucket/some/path/a=1/b=1", "s3://bucket/some/path/a=2/b=1"]` (or `namespace="s3://bucket", name="/some/path", partitions=[{"a": 1, "b": 1"}, {"a": 2, "b": 1}]`) which are actually read by Spark.

User _may_ be interested in both of these representations - to see a raw filter value helps to understand what was intention of ETL process, and seeing a list of resolved paths helps to understand which data satisfied the filter. IMHO the latter may be more useful in terms of data lineage, showing that specific ETL process assessed specific dataset + partitions list, and those data was written somewhere else.

Also there may be cases like:
```python
df = spark.read.format("orc") \
  .load("s3://bucket/some/path/a=*/b=1") \
  .where("a=1")
df.count()
```
So there will be 2 logical filters, and another set of physical partitions.

pawel-big-lebowski:
@dolfinus I am not sure if OL should distinguish between logical and physical plan representations. Especially that a subset definition can be a mix of those two, like: 
```
spark.load("s3://bucket/some/path/a=*/b=3").where("c=1")
```
For this case, looking at the logical condition (`c=1`) only, can be misleading

Instead, we could provide a best-effort hierarchy for describing a subset. In general, there are dozens of complex filtering definitions, available in `WHERE` clause, that OL won't be able to understand and it's OK to admit that. For those cases, we can just put `WHERE` condition string into the subset definition, as this is the best we can do. However, there are cases where OL can do much better.

Proposed hierarchy: 
1. If a subset can be identified by a single partition definition or a partition list, use this as subset definition first. A single partition should be a minimal granularity we care. If a single partition is additionally filtered by some other condition, we don't care about it. 
2. If a subset can be identified by a list of physical locations, like directories or files (with no wildcards), use this as subset definitions. We don't care about additional logical filtering in this case as well. 
3. If a subset definition is a combination of physical location with wildcard and logical operation, we include both definitions in subset definition.   
4. If a logical definition is a combination of straight-forward literal conditions like: ((a>1 and b<3) or (c=4), we support it through a combination of binary and literal conditios.
5. In case of more complex condition syntax, OL adds string representation like the content of where clause.

dolfinus:
> 3. If a subset definition is a combination of physical location with wildcard and logical operation, we include both definitions in subset definition.

Can OL call some internal Spark method to convert this physical location + a bunch of logical filters to get resolved paths of partitions? To get the info which partitions are read by current Spark job/execution after all filters already applied. If this information is available, there is no need to track logical filters in the first place and add facets representing it.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,474 @@
...
+      {
+        "fields": [
+          "col_c"
+        ],
+        "transform": "bar(col_c)"
+      }
+    ]
+  }
+}
+```
```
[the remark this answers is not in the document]
pawel-big-lebowski: No, it's just a different partitioning example. Added `description` field with SQL clause.

proposals/3392/dataset_partitioning.md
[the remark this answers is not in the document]
pawel-big-lebowski: Current version of the proposal (18Feb) is missing `operation` information. I don't think this is suits `OutputSubsetDatasetFacet` and perhaps deserves separate facet in future.

pawel-big-lebowski:
> > 3. If a subset definition is a combination of physical location with wildcard and logical operation, we include both definitions in subset definition.
> 
> Can OL call some internal Spark method to convert this physical location + a bunch of logical filters to get resolved paths of partitions? To get the info which partitions are read by current Spark job/execution after all filters already applied. Or maybe extract this information from command/exec objects, like I've mentioned here [#3392 (comment)](https://github.com/OpenLineage/OpenLineage/pull/3392#discussion_r1946104820)
> 
> If this information is available, there is no need to track logical filters in the first place and add facets representing it.

I removed (3.) from the hierarchy. I removed path wildcards from the proposal.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,555 @@
...
+A job affecting many countries for a single date would affect multiple 
+partitions and this would be described with a facet:
+
+```json
+{
+   "namespace": "dataset-namespace",
+   "name": "dataset-name",
+   "outputFacets": {
+      "subset": {
+         "$id": "https://openlineage.io/spec/facets/1-0-0/PartitionListSubsetCondition.json",
```
dolfinus, line 466: Probably outdated reference

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,555 @@
...
+## Reference implementation with Apache Spark
+
+ * Support extracting dataset partitioning info from:
+   * `DataSourceV2Relation objects`, see partitioning field ([see code](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/java/org/apache/spark/sql/connector/catalog/Table.java#L71)),
+   * [Iceberg partitioning](https://iceberg.apache.org/docs/1.6.1/partitioning/#problems-with-hive-partitioning).
+ * Verify how dataset filtering is reflected in `LogicalPlan`. Produce subset condition input facets based on filtering proposal prepared.
+ * Support configuration based partitioning:
+   * Provide configuration to manually specify how datasets are partitioned within the organization.
+   * For example, if a configuration contains `business_date` partitioning definition, then a Spark job reading data from `s3://bucket/dataset/business_date=20241101` would replace dataset identifier with `s3://bucket/dataset` and `FilterInputDatasetFacet` determining `business_date` specified.
+   * Similarly, a job writing a location `s3://bucket/dataset/business_date=20241101` would identify a dataset `s3://bucket/dataset` and enrich it with `PartitionOutputDatasetFacet`.
```
dolfinus, line 543: Can we infer partitioning info from the path? For example, split location by `/`, all items without `=` are a part of dataset, and items with `=` are dataset subsets

pawel-big-lebowski: I think we can infer partitioning from the path. One can allow detecting partition dimensions from substrings `/a=b/`, but it may be also useful to limit this to datasets matching a specified regular expression. One may also disable auto partitioning detection for specific `a` values, or enable only for selected ones.  

It can become complex to describe all the partitioning strategies within the organisation, but looks doable.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,555 @@
...
+A job affecting many countries for a single date would affect multiple 
+partitions and this would be described with a facet:
+
+```json
+{
+   "namespace": "dataset-namespace",
+   "name": "dataset-name",
+   "outputFacets": {
+      "subset": {
+         "$id": "https://openlineage.io/spec/facets/1-0-0/PartitionSubsetCondition.json",
```
dolfinus, line 466:
```suggestion
         "type": "partition",
```

proposals/3392/dataset_partitioning.md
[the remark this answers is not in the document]
dolfinus: Well, for the most cases each Spark command performs one operations for all partitions in the list (bulk drop, create/alter only one partition at the time), so for now it's okay to not overcomplicate things. Also command name is already a part of job name, and maybe even don't require explicit facet

proposals/3392/dataset_partitioning.md
dolfinus: I'm not very optimistic about `describe all the partitioning strategies within the organisation`, but this is about Spark integration implementation detail, and not about facet proposal, so we can move this discussion elsewhere

proposals/3392/dataset_partitioning.md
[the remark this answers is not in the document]
dolfinus: Resolved by https://github.com/OpenLineage/OpenLineage/pull/3392#issuecomment-2659356747, items 1 and 2. How this information should be extracted from Spark is actually another story, and not the part of facet proposal

dolfinus:
LGTM, thanks @pawel-big-lebowski!

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,555 @@
...
+               "right": {
+                  "type": "CompareExpression"
+               },
+               "comparison": {
+                  "enum": [
+                     "EQUAL",
+                     "GREATER_THAN",
+                     "GREATER_EQUAL_THAN",
+                     "LESS_THAN",
+                     "LESS_EQUAL_THAN"
```
mobuchowski, line 199: Enums in JSON schema are not a good idea - they are very hard to change later.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,555 @@
...
+               "right": {
+                  "type": "BaseSubsetCondition"
+               },
+               "type": {
+                  "const": "binary"
+               },
+               "operator": {
+                  "enum": [
+                     "OR",
+                     "AND"
```
mobuchowski, line 169: Same for enums here.

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,550 @@
...
+            "/some/path/a=2/b=1",
+            "/some/path/a=3/b=1",
+            "/some/path/a=4/b=1"
+         ]
+      }
+   }
+}
+```
+
+If the number of locations exceeds a configurable threshold, the facet would be:
```
asolimando, line 509: It would be probably good to express that the list is not exhaustive, especially in large organizations there is a high chance that the person who sets the threshold won't be the (only) one consuming the lineage metadata, having partial data without an explicit notice might give the impression that the list is indeed complete

proposals/3392/dataset_partitioning.md
```diff
@@ -0,0 +1,550 @@
...
+```json
+{
+   "subset": {
+      "type": "description",
+      "description": "EXISTS (SELECT * FROM EmployeeReviews er  WHERE er.EmployeeID = Employees.EmployeeID AND er.Rating > 4)"
+   }
+}
+```
+
+### Subset dataset facet
```
asolimando, line 359: Q: since the notion of subset is more generic than that of partition (be it logical or physical), do you see the present proposals covering also bucketing?

asolimando (commented):
The proposal looks very interesting, we are working with massive partitioned data with thousands of files, these new facets would allow to have a better grasp of our data lineage. I have left a few comments but being new to OpenLineage I hope they make sense.

proposals/3392/dataset_partitioning.md
pawel-big-lebowski: Including bucketing within the proposal makes sense only if we know this is achievable for some OL producers, like for example Spark. I am not sure if a list of updated/read buckets can be extracted with spark connector. And I would like to avoid verifying this when working on the proposal. 

However, I am OK to add a comment that hierarchy model can be extended with `bucket` subset type in future.

proposals/3392/dataset_partitioning.md
pawel-big-lebowski: Reaching the threshold should not trim the list. It will skip this subset definition type instead.

asolimando: I have misread the example, my bad. What if we added a counter of the identified locations in case they exceed the threshold? Unless you have in mind a lazy enumeration of such locations, I guess we would have this information to decide on the threshold. I think it would be interesting to know how many locations we are reading, especially when they are too many to be listed.

PR #3403, opened 2025-01-14
Allowing events from SerializedFromObject for non-delta plans

ssanthanam185:
### Problem

Events from sql collect are being filtered even its not run in a Databricks env/delta write command. 
Closes: #ISSUE-NUMBER

### Solution

Filter out SerializedFromObjectPlans only for Data Bricks. This ensure we get complete lineage coverage for all other usecases.

### Checklist

- [ x] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [x ] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [x ] Your changes are accompanied by tests (_if relevant_)

----

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DatabricksEventFilter.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DatabricksEventFilter.java
@@ -38,7 +38,7 @@ public class DatabricksEventFilter implements EventFilter {
 
   @Override
   public boolean isDisabled(SparkListenerEvent event) {
-    return isSerializeFromObject() || isWriteIntoDeltaCommand() || isDisabledDatabricksPlan(event);
+    return isWriteIntoDeltaCommand() || isDisabledDatabricksPlan(event);
   }
 
   public boolean isDisabledDatabricksPlan(SparkListenerEvent ignoredEvent) {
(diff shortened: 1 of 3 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/EventFilterUtils.java
```diff
@@ -28,7 +30,22 @@ public static boolean isDisabled(OpenLineageContext context, SparkListenerEvent
             new SparkNodesFilter(context),
             new CreateViewFilter(context),
             new AdaptivePlanEventFilter(context))
-        .anyMatch(filter -> filter.isDisabled(event.getClass().cast(event)));
+        .anyMatch(
+            filter -> {
+              boolean isDisabled = filter.isDisabled(event.getClass().cast(event));
+              if (isDisabled) {
```
pawel-big-lebowski, line 36: Could you please enclose the whole block within `if (log.isDebugEnabled()`? 
This makes sure no method calls are performed when information is not needed.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DatabricksEventFilter.java
```diff
@@ -38,7 +38,7 @@ public DatabricksEventFilter(OpenLineageContext context) {
 
   @Override
   public boolean isDisabled(SparkListenerEvent event) {
-    return isSerializeFromObject() || isWriteIntoDeltaCommand() || isDisabledDatabricksPlan(event);
+    return isWriteIntoDeltaCommand() || isDisabledDatabricksPlan(event);
```
pawel-big-lebowski, line 41: The main problem of this class is that it's not doing what the name suggests. 
It should filter events for Databricks runtime only. 

I think `isDisabled` method should start with:
```
if (!DatabricksUtils.isRunOnDatabricksPlatform(context)  || !context.getQueryExecution().isPresent()) {
      return false;
}
```
This should solve your issue as you get an event filtered outside databricks runtime.

pawel-big-lebowski (commented):
Class contains a bug and thank you for spotting this. 
I left a comment with an alternative approach to solve this.
Please let me know what do you think .

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DatabricksEventFilter.java
ssanthanam185: The existing check is essentially doing if (!DatabricksUtils.isRunOnDatabricksPlatform(context)
        || !context.getQueryExecution().isPresent()) & contains("sql.transaction.tahoe.commands.WriteIntoDeltaCommand")) 
        
        
 Are you suggesting  we dont need to care if its WriteIntoDeltaCommand ?

pawel-big-lebowski: Sorry, my previous comment was inaccurate. 
I am suggesting that this class should have no effect outside databricks runtime. 

So `isDisabled` method should start with `if (!DatabricksUtils.isRunOnDatabricksPlatform(context) { return false }`

pawel-big-lebowski (approved):

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DatabricksEventFilter.java
pawel-big-lebowski: Bcz we're close to release time, I found it's easier to include proposed change within the commit: https://github.com/OpenLineage/OpenLineage/pull/3403/commits/b31404ca2a6d55386b1bc5a211170811ed2d1428

Thanks for contributing

Issue #3398, opened 2025-01-13
[BUG][SPARK] When using a Kafka source and foreachBatch sink, the integration produces 2 disconnected events


d-m-h:
### OpenLineage integration/client

Spark integration

### OpenLineage version

```
io.openlineage:openlineage-spark_2.12: >=1.21.x
io.openlineage:openlineage-spark_2.13: >=1.21.x
```

### Problem details

As the title suggests, when one runs a Spark job that reads from Apache Kafka and then uses the `foreachBatch` sink function, the integration produces 2 disconnected events. The test `io.openlineage.spark.agent.SparkStreamingTest#testKafkaSourceToBatchSink` already captures the setup, however it makes some incorrect assertions.

What happens is that the connector currently produces lineage that looks like this:

```json
[
  {
    "run": {
      "runId": "0194606c-a34f-7c00-89e6-ed68b075ce73"
    },
    "job": {
      "namespace": "default",
      "name": "test_kafka_source_to_batch_sink.project"
    },
    "inputs": [
      {
        "namespace": "kafka://prod-cluster:56163",
        "name": "source-topic-809"
      }
    ],
    "outputs": []
  },
  {
    "run": {
      "runId": "0194606c-a3a8-7136-99de-ceb8e5c98ab6"
    },
    "job": {
      "namespace": "default",
      "name": "test_kafka_source_to_batch_sink.execute_insert_into_hadoop_fs_relation_command.tmp_batch_sink"
    },
    "inputs": [],
    "outputs": [
      {
        "namespace": "file",
        "name": "/tmp/batch_sink"
      }
    ]
  }
]
```

This breaks lineage in our application, has the run identifiers and job identifiers are different.

### What you think should happen instead

The input and output dataset should be present in the same event.

### How to reproduce

Run the `io.openlineage.spark.agent.SparkStreamingTest#testKafkaSourceToBatchSink` test and examine the output.

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

pawel-big-lebowski:
@d-m-h I was looking into this and prepared draft PR, which passes an improved version of `testKafkaSourceToBatchSink` test. Unfortunately, at the moment it's breaking other tests.

PR #3405, opened 2025-01-15
fix Spark structured streaming query context

pawel-big-lebowski:
Problem: https://github.com/OpenLineage/OpenLineage/issues/3398 - when a stream processing query is started, Spark can still run batch SQL jobs. Openlineage connector treats those jobs as separate which is breaking a lineage.

This PR modifies OpenLineage listener in a way that, when streaming query is started, all the SQL jobs are getting the same runId and jobName as a streaming query. For the inner SQLs `start` and `complete` events are not emitted (events are of a `running` type). When a streaming job terminates, `complete` event is published. 

The alternative approach to this are:
 * Do nothing and let backend correlate the events through application id.
 * Add another logical parent run hierarchy level for the streaming jobs. So a streaming job would be a child of an application job, while all the SQLs run during the streaming job would be its children. On one hand this makes sense, but on the other hand I think this adds unnecessary complexity on the backend side. Example: if a job reads events from Kafka, windows them, and modifies a table with a window aggregation, then I think that reading a stream and writing to a table is the same job. Breaking this into two jobs is what Spark implementation does, but does not reflect what the jobs is about.

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
@@ -53,7 +53,9 @@ class SparkSQLExecutionContext implements ExecutionContext {
   private final OpenLineageContext olContext;
   private final EventEmitter eventEmitter;
   private final OpenLineageRunEventBuilder runEventBuilder;
+  private final boolean isInStreamingQueryMode;
 
+  private boolean shouldEmitStartEvent = true;
   private boolean emittedOnSqlExecutionStart = false;
   private boolean emittedOnSqlExecutionEnd = false;
   private boolean emittedOnJobStart = false;
@@ -67,11 +69,13 @@ class SparkSQLExecutionContext implements ExecutionContext {
       long executionId,
       EventEmitter eventEmitter,
       OpenLineageContext olContext,
-      OpenLineageRunEventBuilder runEventBuilder) {
+      OpenLineageRunEventBuilder runEventBuilder,
+      boolean isInStreamingQueryMode) {
     this.executionId = executionId;
     this.eventEmitter = eventEmitter;
     this.olContext = olContext;
     this.runEventBuilder = runEventBuilder;
+    this.isInStreamingQueryMode = isInStreamingQueryMode;
   }
 
   @Override
@@ -90,7 +94,7 @@ class SparkSQLExecutionContext implements ExecutionContext {
 
     olContext.setActiveJobId(activeJobId);
     // only one START event is expected, in case it was already sent with jobStart, we send running
-    EventType eventType = emittedOnJobStart ? RUNNING : START;
+    EventType eventType = !shouldEmitStartEvent || emittedOnJobStart ? RUNNING : START;
     emittedOnSqlExecutionStart = true;
 
     RunEvent event =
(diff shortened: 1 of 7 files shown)


integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -135,6 +138,9 @@ public void end(SparkListenerSQLExecutionEnd endEvent) {
     if (emittedOnJobStart && !emittedOnJobEnd) {
       // expecting jobEnd event later on
       eventType = RUNNING;
+    } else if (isInStreamingQueryMode) {
+      // streaming query mode should never emit COMPLETE event for SQL
+      eventType = RUNNING;
```
d-m-h, line 143: This isn't entirely true. When using the `foreachBatch` method, the source is a streaming source but the sink is a batch sink. It is a strange situation.

d-m-h:
Something else that I discovered by debugging the application, the sequence of events goes something like this:

1. SparkListenerSQLExecutionStart (containing the `Project` node)
2. SparkListenerSQLExecutionStart (containing the `InsertIntoHadoopsFsRelationCommand` node)
3. SparkListenerJobStart, which corresponds with the `InsertIntoHadoopsFsRelationCommand` node

I wonder if there is a way to link these things together.

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
pawel-big-lebowski: Well, this is what the issue is about. Spark starts streaming job and within this job it is spawning extra SQL queries. We can either treat those as separate jobs, and accept the lineage gets broken. One can support this on the backend by correlating lineage through applicationId.

The other approach is to say that if a streaming query is started, then all SQL queries run at the time it is active should be the part of the same job. And this is actually what this PR does. After getting `QueryStartedEvent` event, all the sql jobs belong the same run. To make it work this way, while keeping aligned with the standard, we need to make sure start and complete events are sent only one.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkStreamingTest.java
```diff
@@ -296,10 +308,46 @@ void testKafkaSourceToBatchSink() throws TimeoutException, StreamingQueryExcepti
...
+                    e ->
+                        e.getRun().getFacets() == null
+                            || e.getRun().getFacets().getParent()
+                                != null) // ignore application events
+                .map(x -> Pair.of(x.getRun().getRunId(), x.getJob().getName()))
+                .distinct()
+                .collect(Collectors.toList()))
+        .hasSize(1);
+
+    List<String> eventTypes =
```
pawel-big-lebowski, line 324: One thing that this PR is not solving is that, events generated for the in-the-middle SQL queries should not contain job and run facets, at least the processing type facet. To be fixed in case a community has a consensus on which approach to solve.

pawel-big-lebowski:
Not sure if this is the right way to go. 
@d-m-h Do you have draft PR of your approach in case someone would like to take it over?

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkStreamingTest.java
wobu: @pawel-big-lebowski i want to pick up this PR or at least get a running version for our needs.

i know that you have already closed this PR, but i wanted to ask if there has any consensus been reached to this or where the discussion is made.

Issue #3407, opened 2025-01-15
[BUG] ClassNotFoundException when using openlineage-spark integration and Spark connector


ddebowczyk92:
### OpenLineage integration/client

Spark integration

### OpenLineage version

openlineage-spark >= 1.18.0

### Deployment details

Dataproc Spark cluster with ` 2.2-ubuntu22` image

### Problem details

When using the `openlineage-spark` integration alongside a Spark connector that implements the `spark-extension-interfaces` (`spark-bigquery`), a `ClassNotFoundException` or `NoClassDefFoundError` can occur due to class loader conflicts.

This happens when the `openlineage-spark` integration and the Spark connector are loaded using different class loaders, such as:

- The `openlineage-spark` integration is loaded using the --jars parameter, while the Spark connector is placed in the `/usr/lib/spark/jars` directory.
- The Spark connector is loaded using the --jars parameter, while the `openlineage-spark` integration is placed in the `/usr/lib/spark/jars` directory.

In these scenarios, the `OpenLineageExtensionProvider` interface is only accessible to the class loader that loaded the `openlineage-spark` integration. If the Spark integration attempts to access this interface implementation from a different class loader, a `ClassNotFoundException` is thrown.

### How to reproduce

1. Add the `openlineage-spark` integration JAR via the --jars parameter.
2. Place a Spark connector (implementing spark-extension-interfaces) in the /usr/lib/spark/jars directory.
3. Run a Spark application that initializes both components.
4. Observe the `ClassNotFoundException` in the logs, indicating the `OpenLineageExtensionProvider` class is not found.

### Do you plan to make this contribution yourself?

- [X] I am interested in doing this work

PR #3408, opened 2025-01-15
[SPARK] Fix ClassLoader handling for OpenLineageExtensionProvider

ddebowczyk92:
### Problem

Closes: #3407

### Solution

The solution works around class loader conflicts by transferring the `OpenLineageExtensionProvider` class between class loaders, albeit using an illegal reflective access. Here's how it works:

1. Serialization: The bytecode of the OpenLineageExtensionProvider class is retrieved from the source class loader and serialized into a ByteBuffer.
2. Reflective Access: Using reflection, the defineClass method of the target class loader is invoked to load the serialized bytecode into its context.
3. Usage: Once defined, the class becomes accessible in the target class loader, resolving the visibility issue.

### Checklist

----

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/SparkOpenLineageExtensionVisitorWrapper.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/SparkOpenLineageExtensionVisitorWrapper.java
@@ -13,20 +13,27 @@ import io.openlineage.client.OpenLineage.InputDataset;
 import io.openlineage.client.OpenLineageClientUtils;
 import io.openlineage.client.utils.DatasetIdentifier;
 import io.openlineage.client.utils.DatasetIdentifier.Symlink;
+import io.openlineage.spark.agent.util.ExtensionClassloader;
 import io.openlineage.spark.api.SparkOpenLineageConfig;
 import io.openlineage.spark.extension.OpenLineageExtensionProvider;
+import java.io.IOException;
+import java.io.InputStream;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.reflect.MethodUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
-import org.jetbrains.annotations.NotNull;
 
 /**
  * A helper class that uses reflection to call all methods of SparkOpenLineageExtensionVisitor,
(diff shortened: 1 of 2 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/SparkOpenLineageExtensionVisitorWrapper.java
```diff
@@ -190,10 +199,28 @@ private Optional<ImmutablePair<Object, Method>> getMethod(
...
-      throws ClassNotFoundException, InstantiationException, IllegalAccessException {
+      throws ClassNotFoundException, IOException, InstantiationException, IllegalAccessException {
     List<Object> objects = new ArrayList<>();
+    // The following sequence of operations must be preserved as is.
+    // We cannot use ResourceFinder or ServiceLoader to determine if there are any
+    // OpenLineageExtensionProvider implementations because doing so involves classloading
+    // machinery.
+    // As a result, there is no reliable way to bypass the entire mechanism, even if
+    // no OpenLineageExtensionProvider implementations are present.
+    ClassLoader currentThreadClassloader = Thread.currentThread().getContextClassLoader();
```
pawel-big-lebowski, line 210: why is it needed to instantiate `currentThreadClassloader` here and pass to another method? 
It is not used in this context at all, isn't it?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/SparkOpenLineageExtensionVisitorWrapper.java
```diff
@@ -190,10 +199,28 @@ private Optional<ImmutablePair<Object, Method>> getMethod(
   }
 
   private static List<Object> init(String testExtensionProvider)
```
pawel-big-lebowski, line 201: Isn't it that `init` method is created anytime OpenLineage context is created and each time `ByteBuffer` with class bytes is read?

pawel-big-lebowski (commented):
My understanding is that this code contains machinery to support corner-case scenario for loading classes with different classloaders. I think it would be great to write a code in a way that clearly distinguishes flow for a happy path from a corner case (code comments, separate methods, etc.)

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/SparkOpenLineageExtensionVisitorWrapper.java
ddebowczyk92: done

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/SparkOpenLineageExtensionVisitorWrapper.java
ddebowczyk92: fixed

ddebowczyk92:
> My understanding is that this code contains machinery to support corner-case scenario for loading classes with different classloaders. I think it would be great to write a code in a way that clearly distinguishes flow for a happy path from a corner case (code comments, separate methods, etc.)

@pawel-big-lebowski it’s not entirely true; loading OpenLineageExtensionProvider to other classloaders isn’t a corner-case scenario. We load it to every classloader that hasn’t loaded it yet.

I added some comments and addressed your suggestions

PR #3442, opened 2025-01-27
[SPARK] configurable node filtering

pawel-big-lebowski:
Provides configurable way to modify default list of Spark nodes which are filtered.


integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkOpenLineageConfig.java
```diff
@@ -150,6 +156,14 @@ public static class VendorsConfig {
     private final Map<String, String> additionalProperties = new HashMap<>();
   }
 
+  @Getter
+  @Setter
+  @ToString
+  public static class FilterConfig {
+    private final List<String> allowedSparkNodes = new ArrayList<>();
+    private final List<String> deniedSparkNodes = new ArrayList<>();
```
mobuchowski, line 164: Maybe this should be more of a "lower level" config, where we can define which listeners get which nodes, rather than fully allowing/blocking all nodes of this type for all of the interfaces? 

When it's internal interface, we could not think about it, but if we expose it, we're stuck with it for a bit.

pawel-big-lebowski: Not following this Maciej, there is only one listener `OpenLineageSparkListener` and filtering is applied only at the root level of a logical plan. 

Let me give an example use case for this. A spark query has an `Aggregate` node plan as root which can be some group by and `collect` performed. The result of the query is not written to any other dataset, it just stays within Spark session. In the past we turned off generating OL events for this case: if a job is missing output dataset, its lineage may have little value. At some point later, we introduced application events and now this may have sense again as for some users it may be valuable to know if a certain Spark app reads from datasets. 

Making the thing configurable may be useful bcz different behaviours make more sense in the context of notebooks or regular Spark jobs.

mobuchowski: I was thinking more in a line of enabling/disabling the particular configuration on the particular environment. However, we can do this in two ways even with current design. Either implement the EventFilter yourself, or set up different node filter configuration on different environments (clusters). So I'm okay with this, maybe my idea was overengineering.

mobuchowski (approved):

surjikal:
From the docs, it's implied that there are nodes that are filtered out by default. Where can I find that list?

pawel-big-lebowski:
@surjikal You can find it in the code: https://github.com/OpenLineage/OpenLineage/tree/main/integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters

mostly here:
https://github.com/OpenLineage/OpenLineage/blob/7f43016a8f01dea38c9cd4b96685bf1d7790d933/integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/SparkNodesFilter.java#L29-L40

PR #3444, opened 2025-01-28
Enable SSL context for Java client

pawel-big-lebowski:
Enables TLS configuration on the client side.


mobuchowski (approved):
Few nits, looking good otherwise.

PR #3449, opened 2025-01-31
fix multi version search

### LINKED ISSUE #3448: website - search should work only for a single, currently selected version
When looking for something that repeats across many versions - such as given integration documentation - you get a list of same doc, but in different versions. Ideally this would not happen.

Image

pawel-big-lebowski:
Should fix #3448

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -131,7 +131,13 @@ const config = {
         path: "./blogs",
       },
     ],
-    require.resolve("docusaurus-lunr-search"),
+    [
+      "docusaurus-lunr-search",
+      {
+        disableVersioning: false,
+        excludeRoutes: ["**/next/**", "docs/+([0-9]).+([0-9]).+([0-9])/**"],
+      },
+    ],
   ],
 
   themeConfig:


website/docusaurus.config.js
```diff
@@ -131,7 +131,13 @@ const config = {
         path: "./blogs",
       },
     ],
-    require.resolve("docusaurus-lunr-search"),
+    [
+      "docusaurus-lunr-search",
+      {
+        disableVersioning: false,
+        excludeRoutes: ["**/next/**", "**/docs/1.*/**"],
```
JDarDagran, line 138:
```suggestion
        excludeRoutes: ["**/next/**", "docs/+([0-9]).+([0-9]).+([0-9])/**"],
```

I would rather go with broader versioning pattern.

JDarDagran (approved):
I'm not sure if searching on latest released version is the solution. Not saying it's good or wrong, I'd do the same probably. One can still view latest unreleased and other versions without searching after all.
Another benefit of this approach is that there's no need to customize search plugin.

mobuchowski:
@JDarDagran I'd do the same too - this decision does not prohibit us from doing something better in the future, and immediately improves experience.

mobuchowski (approved):

PR #3454, opened 2025-02-03
update installation section of the docs

pawel-big-lebowski:
It makes no sense for the latest docs to refer to Spark artefact without Scala version in the name that used to be present prior to 1.8. I think we can remove it, while leaving a tiny warning at the end of section .


pawel-big-lebowski:
[installation.pdf](https://github.com/user-attachments/files/18639729/installation.pdf) -> pdf of the rendered website

mobuchowski (approved):

JDarDagran (approved):

PR #3455, opened 2025-02-03
fix docs markdown

pawel-big-lebowski:
Screenshot 2025-02-03 at 09 55 43

Screenshot 2025-02-03 at 10 29 30

This will not fix historical versions, only the future one. Is it OK?


JDarDagran:
> This will not fix historical versions, only the future one. Is it OK?

We can merge this one and post PR fixing this in historical versions in the `openlineage-site` repo.

JDarDagran (approved):

PR #3456, opened 2025-02-03
expose metadata from circuit breaker as run facets

pawel-big-lebowski:


pawel-big-lebowski:
I'll leave it for now and try different approach with giving a circuit breaker an access to `MeterRegistry` instead.

PR #3464, opened 2025-02-05
Enable filtering through transform transport

pawel-big-lebowski:
Currently a `transform` transport throws an exception If a transformer class returns null. It makes more sense to log warning in this case while not throwing an exception. This will allow filtering the events within transformer classes.


JDarDagran:
Could you provide some example in the docs how filtering could be implemented?

mobuchowski:
@JDarDagran looking at what the code does, by `EventTransformer` returning null 🙂

JDarDagran (approved):

Issue #3414, opened 2025-01-17
[PROPOSAL] User-supplied Tags for the Client


leogodin217:
**Purpose:**
With the [implementation of tags](https://github.com/OpenLineage/OpenLineage/pull/3399) in the OLin spec, integrations can now add tags to runs, jobs and datasets. One problem with this implementation is it generally requires integrations to pull tags from code. Yet, some tags are dependent on the environment code is run in. User-supplied tags will allow the client to inject new tags or override tags provided by the integrations for jobs and runs.

Examples:
* set environment=production/staging/dev for jobs and runs
* set adhoc=true for runs that are not scheduled
* Easily override pipeline=billing for jobs and runs 

Note: This proposal specifically does not include datasets. If there is a need for user-supplied dataset tags, please comment with use cases and examples. 

**Proposed implementation**

OLin already supports environment variables for configuration, it makes sense to use the same functionality for user-supplied tags. We can use a similar schema as the config variables. 

OPENLINEAGE_TAGS_JOB__key=value
OPENLINEAGE_TAGS_RUN__key=value

For example, if we want to set environment=production, we would use `export OPENLINEAGE_TAGS_JOB__environment=production`

Rules for adding tags to job and run facets:

* If the tags facet does not exist, create it and add all user-supplied tags
* If the tags facet exists and none of the user-supplied tag keys match integration-supplied keys, add all user-supplied tags to the facet
* If the tags facet exists and one or more of the user-supplied tag keys matches an integration supplied key, override the integration-supplied key and add the rest. 
* Source for each tag will either by PYTHON_CLIENT or JAVA_CLIENT

*Python Client*

Updating tags facets can follow the pattern used in `add_environment_facets()`

```python
    def emit(self, event: Event) -> None:
        ...
        event = self.add_environment_facets(event)
        event = self.update_job_tag_facet(event) # Inject job tags
        event = self.update_run_tag_facet(event,) # inject run tags
        ...
        self.transport.emit(event)
```

```python
    def _get_tag_env_vars(self, tag_type) -> Event:

        # get os.environ.items only starting with OPENLINEAGE_TAG_JOB prefix
        tag = sorted(
            filter(lambda k: k[0].startswith(f'OPENLINEAGE_TAG_{tag_type}'), os.environ.items())
        )

        return list(tag_vars)
```

```python
    def update_job_tag_vars(self, event) -> Event:
        tag_vars = self._get_tag_env_vars("job")
[...]

PR #3399, opened 2025-01-13
Feature/tags facets v2

leogodin217:
### Problem

Arbitrary metadata, often referred to as tagging, concerns adding additional datapoints that are not part of the OL spec to events. Several use cases exist for this type of data.

- Environment information for jobs
- Setting PII/security information for columns
- Grain for datasets
- Technology-specific attributes i.e. (dbt Cloud project, Airflow tags, etc.)
- Downstream notification information like Slack channels, email addresses, etc.
- Configuration information for downstream automation (A message to re-run a pipeline, create Jira ticket, kick off reverse ETL, etc.)
- Enriching events with domain-specific tags
- etc.

No matter how many fields OpenLineage adds to the spec, there will always be a need for additional metadata both supplied by the client and the user. In order to allow OpenLineage to support more use cases, we should consider a generalized and automatable approach to capturing this type of metadata.

See these discussions for additional context:

https://github.com/OpenLineage/OpenLineage/issues/2779
https://github.com/OpenLineage/OpenLineage/issues/2748
https://github.com/OpenLineage/OpenLineage/pull/1630

Closes: #3169

### Solution

For this implementation I created separate facets for runs, jobs and datasets. Each facet has a facet fields object that holds the properties for the tag. We do this instead of inheriting a single TagsFacetField object because the code generator requires external refs to already be published. 

TagsDatasetFacetFields:
- key - Name of the tag
- value - Value of the tag
- field - Dataset field the tag applies to if it applies to a field
- source - Where the tag is coming from i.e. (DBT, SPARK, etc...)

TagsJobFacetFields:
- key - Name of the tag
- value - Value of the tag
- source - Where the tag is coming from i.e. (DBT, SPARK, etc...)

TagsRunFacetFields:
- key - Name of the tag
- value - Value of the tag
- source - Where the tag is coming from i.e. (DBT, SPARK, etc...)

Issue: [Tags for arbitrary metadata](https://github.com/OpenLineage/OpenLineage/issues/3169) for context.

- [ ] Your change modifies the [core](https://github.com/OpenLineage/OpenLineage/blob/main/spec/OpenLineage.json) OpenLineage model
- [ ] Your change modifies one or more OpenLineage [facets](https://github.com/OpenLineage/OpenLineage/tree/main/spec/facets)

### Checklist

- [x] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
[...]

PR #3471, opened 2025-02-06
feat(tagsfacet): User-supplied Tags in the Python Client

leogodin217:
### Problem

With the https://github.com/OpenLineage/OpenLineage/pull/3399 in the OLin spec, integrations can now add tags to runs, jobs and datasets. One problem with this implementation is it generally requires integrations to pull tags from code. Yet, some tags are dependent on the environment code is run in. User-supplied tags will allow the client to inject new tags or override tags provided by the integrations for jobs and runs.

There are three scenarios where user-supplied tags and integration-supplied tags can be integrated. 

* No tags facet exists - Create tags facet and add all user-supplied tags
* Tags facet exists and no tags have the same key as the user supplied tags (Add all user-supplied tags to the existing tags)
* Tags facet exists and one or more tags has the same key as a user-supplied tag (Add all user-supplied tags and replace any integration-supplied tags with the same key)

Closes: #3414

### Solution

In this solution, we collect user-supplied tags from environment variables in the Client constructor. Tag facets for runs and jobs are updated in the `client.emit()` function. Some of the tags might override an integration-supplied tag. For this , we replace any tags with the same key as a user-supplied tag.  

### Checklist

----


client/python/openlineage/client/transport/no_output.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class NoOutputTransport(Transport):
+    """Used only for writing tests. Simply stores the event for assertions"""
+
+    kind = "no_output"
+    config_class = NoOutputConfig
+
+    def __init__(self, config: NoOutputConfig) -> None:  # noqa: ARG002
+        self.event = None
```
leogodin217, line 25: I created this transport to make testing easier. Another pattern would be to mock the transport in the test and validate how client.emit() is called. I prefer the new-transport approach, because it makes tests easier to read. Now, we can verify the exact event sent.

client/python/tests/test_client.py
```diff
@@ -846,3 +848,177 @@ def test_config_loader(self, env_vars, expected_config):
...
+        assert transport.event.job.facets["tags"].tags == tags
+
+
+def test_update_job_tag_facet_creates_updates_existing_facet():
+    tag_environment_variables = {
+        "OPENLINEAGE__TAG__JOB__ENVIRONMENT": "PRODUCTION",
+        "OPENLINEAGE__TAG__JOB__pipeline": "SALES",
+    }
+
+    run_event = RunEvent(
```
leogodin217, line 927: Do we have a preference on creating full events for each test vs using fixtures? Shorter code from fixtures, easier troubleshooting from creating full events.

client/python/openlineage/client/client.py
```diff
@@ -115,6 +119,9 @@ def __init__(  # noqa: PLR0913
             if _filter:
                 self._filters.append(_filter)
 
+        tag_env_vars = self._collect_tag_environment_variables()
+        self.tags = self._create_tags(tag_env_vars)
```
leogodin217, line 123: Should this be injected into the config object or stay separate?

client/python/openlineage/client/transport/no_output.py
mobuchowski: There already is a `NoopTransport`, do we need second one? 🙂

client/python/openlineage/client/client.py
mobuchowski: I think conceptually those aren't really part of config - we're not configuring anything, but collecting them in the runtime and passing them to the event. I think it's fine where it is

client/python/tests/test_client.py
mobuchowski: In general, mix of both - few e2e tests with most tests using fixtures or just checking if some particular field in resulting JSON meets expectations. 

Having too many e2e tests failing is not helpful when trying to find a bug when accidentally breaking a one thing compared to more "unit" tests, but no tests that capture full events can more easily miss some stuff.

client/python/openlineage/client/transport/no_output.py
leogodin217: NoopTransport doesn't save the event. I'm using this in place of assert_called_with. I could add an event property to NoopTransport to do the same thing or just use assert_called_with.

client/python/openlineage/client/client.py
leogodin217: I ended up putting it in the config on advice from @JDarDagran. Do you see that as a problem?

client/python/openlineage/client/transport/no_output.py
mobuchowski: If we need it just for tests, maybe just keep it in the test file?

client/python/openlineage/client/client.py
mobuchowski: My mistake - I did not get that it's only for user-supplied tags. It's fine now.

client/python/openlineage/client/client.py
```diff
@@ -72,6 +80,8 @@ def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
 
 class OpenLineageClient:
     DYNAMIC_ENV_VARS_PREFIX = "OPENLINEAGE__"
+    DYNAMIC_ENV_VARS_TAG_JOB_PREFIX = "OPENLINEAGE__TAG__JOB__"
+    DYNAMIC_ENV_VARS_TAG_RUN_PREFIX = "OPENLINEAGE__TAG__RUN__"
```
kacpermuda, line 84: I’m not sure if we need those two lines. What’s their purpose?

JDarDagran: That's probably leftover when refactoring to OL config.

leogodin217: Sure is. I'll remove it.

client/python/openlineage/client/transport/no_output.py
leogodin217: Done. I added it to conftest. I think we can use this to clean up a lot of deprecation warnings.

client/python/openlineage/client/client.py
kacpermuda: @leogodin217 I think this is still pending ? 😄

leogodin217: D'oh. Fixed.

website/docs/client/python.md
```diff
@@ -951,3 +950,137 @@ for event in events:
 The resulting lineage events received by Marquez would look like this.
 
 ![the Marquez graph](./mqz_graph_example.png)
+
+
+##### User-supplied Tags with Environment Variables 
+
+Integrations can add [tag facets](https://github.com/OpenLineage/OpenLineage/blob/main/proposals/3169/tags_facet.md) to runs, jobs and datasets. To allow more control over tags, users can add to and override integration-supplied tags through environment variables supplied to the client. The following rules apply to user-supplied tags. 
+
+* Tag environment variables follow this format. OPENLINEAGE__TAGS__JOB/RUN__KEY="VALUE"
```
JDarDagran, line 959:
```suggestion
* Tag environment variables follow this format. OPENLINEAGE__TAG__JOB/RUN__KEY="VALUE"
```

This follows proposed naming. However, I'm not convinced the section in config should be called `tag` instead of `tags`.

leogodin217: I changed it twice while coding. By config standards it should probably be tags. But the environment variable is only passing a single tag so I made it singular. I'm happy to put it bag to TAGS. 

This brings up another concern. To use the simplest solution, I created two properties: tags_job and tags_run. 
```python
@attr.s
class OpenLineageConfig:
    transport: dict[str, Any] | None = attr.ib(factory=dict)
    facets: FacetsConfig = attr.ib(factory=FacetsConfig)
    filters: list[FilterConfig] = attr.ib(factory=list)
    tags_job: list[TagsJobFacetFields] = attr.ib(factory=list)
    tags_run: list[TagsRunFacetFields] = attr.ib(factory=list)

    @classmethod
    def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
        ...
        if tags_job := params.get("tag", {}).get("job"):
            config.tags_job = [TagsJobFacetFields(key, value, "USER") for key, value in tags_job.items()]
        if tags_run := params.get("tag", {}).get("run"):
            config.tags_run = [TagsRunFacetFields(key, value, "USER") for key, value in tags_run.items()]
        return config
```

Would you prefer that the config matches the envar format?  

```python
@attr.s
class OpenLineageConfig:
   ...
    tags: {
        "job": [...]
        "run": [...]
    }
```

Or maybe something like this? 

```python
@attr.s
class OpenLineageConfig:
   ...
    tags: SomeNewTagsClass(
        @property
        def job():
            ...

        @property
        def run():
            ...
    )
```

leogodin217: The more I think about it, the more I like option 3. It makes sense to keep the config in the same format as the confit file.

JDarDagran: I'd go with option 3 - nested `tags` field that may hold either job or run tags info.

JDarDagran: my bad, I meant both

client/python/openlineage/client/client.py
```diff
@@ -404,3 +415,41 @@ def _collect_environment_variables(self) -> dict[str, str]:
...
+        """
+        Handles updating tags in an existing tag facet
+        """
+
+        # Get tags from the facet that will not be updated (Do not have the same key as a user tag)
+        user_tag_keys = [tag.key for tag in user_tags]
+        keep_tags = list(filter(lambda x: x.key.lower() not in user_tag_keys, tags_facet.tags))
+
+        # Update tag key for any user tags that should override a facet tag.
+        # This preserves case of the facet tag key while updating value and source.
```
JDarDagran, line 447: How would that work with existing tags that are not lowercase?

client/python/openlineage/client/client.py
```diff
@@ -63,6 +67,13 @@ def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
             config.facets = FacetsConfig(**params["facets"])
         if "filters" in params:
             config.filters = [FilterConfig(**filter_config) for filter_config in params["filters"]]
+        if "tags" in params:
+            job_tags = params["tags"].get("job", {})
+            run_tags = params["tags"].get("run", {})
+            config.tags = TagsConfig(
```
JDarDagran, line 73: maybe there could be validation whether value of the tag is not empty?

website/docs/client/python.md
```diff
@@ -951,3 +950,137 @@ for event in events:
...
+* Source for a user-supplied tag is always set to "USER"
+* If an integration-supplied tag has the same key as a user tag (case-insensitive), the tag value and source will be overridden.
+
+
+###### Examples
+
+Using this environment variable, an event with no tags facets will create a tag facet and add the following tag. 
+
+```sh
+OPENLINEAGE__TAG__JOB__ENVIRONMENT="PRODUCTION"
```
JDarDagran, line 971:
```suggestion
OPENLINEAGE__TAGS__JOB__ENVIRONMENT="PRODUCTION"
```

website/docs/client/python.md
JDarDagran, line 959: It's explained somewhere else so I guess it should be suggested to use it in a way described here.
FYI you can also set:
`OPENLINEAGE__TAGS='{"job":{"key":"value"},"run":{"key":"value"}}'` as well. The benefit it is you can do it in single env var.

mobuchowski (approved):

PR #3477, opened 2025-02-07
spark: bump tested version

mobuchowski:
Spark 3.5.4 is out, we should use it for building and testing purposes.


pawel-big-lebowski:
Nice improvement @mobuchowski 
Could you do the same for `3.4.3` -> `3.4.4`?

mobuchowski:
@pawel-big-lebowski unfortunately, no Bitnami image for 3.4.4 which we use as a base for our own images.

pawel-big-lebowski (approved):

PR #3481, opened 2025-02-11
Set java toolchain & fix spark build deps script

pawel-big-lebowski:
* Turning of tests for java client and spark extensions. For example java client has a docker based s3 test which is not working on latest Mac and of course requires running docker environment.
 * Setting toolchain java 17 for client-java and spark extensions makes sure, the projects are build with a proper java version regardless of default `java` set.


mobuchowski (approved):

mobuchowski:
@pawel-big-lebowski are you sure this will keep the compatibility with Java 8?

We have 
```
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
```

for client for a reason.

pawel-big-lebowski:
@mobuchowski Sure Maciej.

PR #3487, opened 2025-02-13
Async breaker timeout setting is not a real timeout.

pawel-big-lebowski:
Recent version of OpenLineage introduced `TaskQueueCircuitBreaker` with a config entry `timeoutSeconds`. 
Problem: this is not a timeout setting.

Although a circuit breaker is async, it allows a configurable blocking time after submitting a callable. In some scenarios, this can be helpful to improve the ordering of the events. For example, if the Spark listeners is called twice roughly at the same time, then running OL code in parallel could cause race condition, improper even ordering and sending termination event prior to other events. This cannot be fully solved for `async` listener, however we can mitigate the risk by allowing user to provide a configurable blocking time at the beginning of the async call.

The feature was already implemented but the config entry `timeoutSeconds` was misleading. Within this PR we modify the config entry name, fix the default value and modify docs,


JDarDagran (approved):
As this is not a breaking change I think we can rename it that way.
💯  current name is misleading.

mobuchowski (approved):

PR #3489, opened 2025-02-14
Single package for Flink 1 & 2

pawel-big-lebowski:
**Idea**: It is reasonable to have both Flink listeners in a single package. This is because they share common configuration and some code. This is also a better user experience because of a single `openlineage-jar`. It's a configuration which chooses listener to pick up. Although a mechanism of extracting lineage changes between Flink 1 and 2, from a user perspective, this is just another version supported, like in `openlineage-spark`.

Most of dependencies are compiled-only and the largest included one is a shared `openlineage-java`. So the approach should not lead to extensive package size. 

**Solution**
Gradle build contains `flink.version` param which describes version of Flink that a compiled code will be tested against. This should not affect the code within a jar. This PR renames `app` module to `flink1` and introduces `flink2` module. 
Both modules are always compiled but compilation of their tests happens only when necessary. In other words, flink1 tests are not compiled when `flink-version` is `2`.


.circleci/continue_config.yml
```diff
@@ -15,6 +15,10 @@ parameters:
   nightly-run:
     type: string
     default: "inactive"
+  integration-type:
+    type: enum
+    default: "*"
+    enum: [ "*", "airflow", "dagster", "dbt", "publish", "python" ]
```
mobuchowski, line 21: Why is "publish" an integration type?

integration/flink/flink2/build.gradle
```diff
@@ -0,0 +1,183 @@
...
+    id 'com.github.johnrengelman.shadow'
+}
+
+group = "io.openlineage"
+
+archivesBaseName='openlineage-flink2-app'
+
+ext {
+    compileflinkVersion = project.getProperty('flink.flink2.version')
+    testedflinkVersion = project.getProperty('flink.version')
```
mobuchowski, line 25: Nice solution.

mobuchowski (approved):
One question, besides it looks good.

.circleci/continue_config.yml
pawel-big-lebowski: enum content is just copied from another config.

PR #3494, opened 2025-02-18
Unify configs and context for flink versions

pawel-big-lebowski:
Creates `OpenLineageContext` object that is shared for all Flink versions' integration. 
Makes Flink 2 integration use shared config and emitter classes.


mobuchowski (approved):
Amazing job on removing unnecessary code @pawel-big-lebowski.

PR #3495, opened 2025-02-18
deprecate naming spec

harels:
### Problem

we deprecated this naming spec in favor of docs, but it's not very clear.

### Solution
removing the content and linking out

----


merobi-hub (approved):

JDarDagran:
Just curious: having docs as source of truth for naming means it can be versioned as well?

JDarDagran:
> @JDarDagran I think naming scheme is one thing we don't want to be versioned.

We should move it from docs into unversioned section of the webpage, shouldn't we?

PR #3518, opened 2025-02-27
Add Iceberg's RestCatalog integration test & fix issues

pawel-big-lebowski:
This PR adds docker based integration test to verify behaviour of Scan & Commit metrics report facets when using `RestCatalog`. It uses `tabulario/iceberg-rest` docker image. 

When running the test, an unsupported scenario was encountered. When creating a new table within RestCatalog, it's impossible to create DatasetIdentifier on `START` events, as table location is empty (will be filled after rest request) and warehouse location may be empty as well. In this case `new Path(null)` was run which thrown an exception. 

As a workaround `MissingDatasetIdentifierCatalogException` is introduced. It is thrown when there is no proper way to identify a dataset. It's caught in `PlanUtils3` and logged with debug level.


integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkRestIcebergIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+                  new StructField("a", LongType$.MODULE$, false, Metadata.empty()),
+                  new StructField("b", LongType$.MODULE$, false, Metadata.empty())
+                }))
+        .repartition(1);
+  }
+
+  @Test
+  @SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
+  void testMetricsFacets() {
+    if (JAVA_VERSION.startsWith("1.8") && System.getProperty(SPARK_VERSION).startsWith("3.5")) {
```
mobuchowski, line 136: nit: use EnabledIfSystemProperty or EnabledIfEnvironmentVariable for those checks instead of doing it inside test

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkRestIcebergIntegrationTest.java
```diff
@@ -0,0 +1,206 @@
...
+      // This test will not work as Iceberg classes used are Java 11
+      assertThat(true).isTrue();
+      return;
+    }
+
+    clearTables("temp", "scan_source", "scan_target");
```
mobuchowski, line 142: BeforeEach/BeforeAll?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
```diff
@@ -299,6 +306,12 @@ private String getCatalogType(Map<String, String> catalogConf) {
           "Default the catalog type to 'glue' because the catalog impl is {}",
           catalogConf.get(CATALOG_IMPL));
       return "glue";
+    } else if (catalogConf.containsKey(CATALOG_IMPL)
+        && catalogConf.get(CATALOG_IMPL).endsWith("RESTCatalog")) {
```
mobuchowski, line 310: Any reason we're not doing exact string matching but endsWith? 
We can potentially capture something we do not want here, for example `DeltaRESTCatalog` when we want `IcebergRESTCatalog`.

mobuchowski (approved):
Approved but with some nits for tests

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
pawel-big-lebowski: I took same approach as with glue three lines above. The scenario you describe is impossible bcz is it's `IcebergHandler` and its methods are called only for `org.apache.iceberg.spark.SparkCatalog`. `CATALOG_IMPL` is canonical name, so perhaps the intention was to handle same implementation in different packages. Leaving this as is for now.

Issue #3529, opened 2025-03-04
[BUG] Duplicate BigQuery Inputs


luke-hoffman1:
### OpenLineage integration/client

Spark integration

### OpenLineage version

`io.openlineage:openlineage-spark_2.12:1.29.0`, `io.openlineage:transports-gcs:1.29.0`

### Technology and package versions

Dataproc - `2.1.82-debian11`
SPARK_BQ_CONNECTOR_URL - `gs://spark-lib/bigquery/spark-3.3-bigquery-0.42.0.jar`

### Environment configuration

"spark.jars.packages": "io.openlineage:openlineage-spark_2.12:1.29.0,io.openlineage:transports-gcs:1.29.0",
"spark.extraListeners": "io.openlineage.spark.agent.OpenLineageSparkListener",
"spark.openlineage.transport.type": "gcs",
"spark.openlineage.transport.bucketName": gcs_bucket,
"spark.openlineage.transport.fileNamePrefix": f"dataproc/openlineage/{test_name}/events/",
"spark.openlineage.transport.projectId": project_id,
"spark.openlineage.columnLineage.datasetLineageEnabled": "true",
"spark.openlineage.facets.spark.logicalPlan.disabled": "false",
"spark.openlineage.facets.debug.disabled": "false"

### Deployment details

Dataproc

### Problem details

I am seeing duplicate `project.dataset.table` input datasets in the `inputs` facet. e.g., 

```
"inputs": [
        {
            "namespace": "bigquery",
            "name": "google-project.bq-dataset.table1"
        },
        {
            "namespace": "bigquery",
            "name": "google-project.bq-dataset.table1"
        },
        {
            "namespace": "bigquery",
            "name": "google-project.bq-dataset.table2"
        },
        {
            "namespace": "bigquery",
            "name": "google-project.bq-dataset.table2"
        }
],
```

### What you think should happen instead

There shouldn't be any duplicates.

### How to reproduce

Here is the code snippet you can run (you will need to fill in the three variables at the top):

```
from pyspark.sql import SparkSession
from pyspark.sql.functions import col

# provide your Google information
google_project_id = ''
google_bq_dataset = ''
gcs_bucket = ''

# Create a SparkSession
spark =  SparkSession.builder.appName("SparkOpenLineageReproIssue") \
        .config("viewsEnabled", "true") \
        .config("temporaryGcsBucket", gcs_bucket) \
        .config("spark.sql.broadcastTimeout", "1200") \
[...]

PR #3548, opened 2025-03-11
Spark - input duplicates for subquery alias

pawel-big-lebowski:
Closes: https://github.com/OpenLineage/OpenLineage/issues/3529

Subquery alias seems to be duplicating inputs. Turning off `SubqueryAliasInputDatasetBuilder` seems to be solving the issue. However, it was introduced to support some marge into scenario for delta.

Solution: Subquery alias input builder should not delegate input visitors.


mobuchowski (approved):
Awesome bug fix @pawel-big-lebowski.

PR #3580, opened 2025-03-24
[dbt][structured logs][fix] Read dbt log file incrementally

MassyB:
### Problem

For some reason I don't [know](https://github.com/dbt-labs/dbt-common/blob/e9efd8343ce4dd9e694814716a115cab27e8bcf7/dbt_common/events/event_manager.py#L76) dbt process is writing incomplete JSONs to the dbt log file. 
Consequently, the integration is reading incomplete lines (incomplete JSONs) and it skips some events.
image
The resulting error is a `KeyError`. This happens very rarely which suggests that some race condition is causing it.

I added some piggybacks to this PR too.

### Solution

Instead of relying on python's `readlines` I added a class dedicated to reading the log file chunk by chunk. That class only returns complete lines.

#### One-line summary:
Consume only complete lines from dbt log file

### Checklist

----

### DIFF
--- a/integration/common/openlineage/common/provider/dbt/structured_logs.py
+++ b/integration/common/openlineage/common/provider/dbt/structured_logs.py
@@ -149,7 +149,7 @@ class DbtStructuredLogsProcessor(DbtLocalArtifactProcessor):
             dbt_event = json.loads(line)
         except ValueError:
             # log that can't be consumed
-            self.logger.info(f"The following log is not valid JSON:\n{line}")
+            self.logger.error(f"The following log is not valid JSON:\n{line}")
             return None
 
         dbt_event_name = dbt_event["info"]["name"]
@@ -512,7 +512,6 @@ class DbtStructuredLogsProcessor(DbtLocalArtifactProcessor):
         self._open_dbt_log_file()
         incremental_reader = IncrementalFileReader(self._dbt_log_file)
         process = subprocess.Popen(dbt_command_line, stdout=sys.stdout, stderr=sys.stderr, text=True)
-        exception = None
         parse_manifest = True
         try:
             while process.poll() is None:
@@ -526,14 +525,13 @@ class DbtStructuredLogsProcessor(DbtLocalArtifactProcessor):
             if self._dbt_log_file is not None:
                 yield from incremental_reader.read_lines()
 
-        except Exception as e:
-            exception = e
+        except Exception:
+            self.logger.exception("An exception occurred in OL code. dbt is still running.")
+            while process.poll() is None:
+                pass  # wait for the process to finish
         finally:
             if self._dbt_log_file is not None:
                 self._dbt_log_file.close()
-            if exception:
-                process.kill()
-                raise exception
 
     def _open_dbt_log_file(self):
         """
--- a/integration/common/openlineage/common/provider/dbt/structured_logs.py
+++ b/integration/common/openlineage/common/provider/dbt/structured_logs.py
@@ -36,10 +36,10 @@ from openlineage.common.provider.dbt.utils import (
     get_parent_run_metadata,
 )
 from openlineage.common.utils import (
+    IncrementalFileReader,
     add_command_line_args,
     add_or_replace_command_line_option,
     get_from_nullable_chain,
-    get_next_lines,
     has_lines,
 )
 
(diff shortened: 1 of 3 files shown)


integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -509,25 +510,30 @@ def _run_dbt_command(self) -> Generator[str, None, None]:
...
         self._open_dbt_log_file()
+        incremental_reader = IncrementalFileReader(self._dbt_log_file)
         process = subprocess.Popen(dbt_command_line, stdout=sys.stdout, stderr=sys.stderr, text=True)
-
+        exception = None
+        parse_manifest = True
         try:
             while process.poll() is None:
-                if has_lines(self._dbt_log_file) > 0:
+                if parse_manifest and has_lines(self._dbt_log_file) > 0:
```
MassyB, line 519: piggyback: we try to parse the manifest only once

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -149,6 +149,7 @@ def _parse_structured_log_event(self, line: str) -> Optional[RunEvent]:
             dbt_event = json.loads(line)
         except ValueError:
             # log that can't be consumed
+            self.logger.info(f"The following log is not valid JSON:\n{line}")
```
MassyB, line 152: piggyback: added a log when the line is not a valid JSON

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -509,25 +510,30 @@ def _run_dbt_command(self) -> Generator[str, None, None]:
...
         except Exception as e:
-            process.kill()
-            raise e
+            exception = e
         finally:
             if self._dbt_log_file is not None:
                 self._dbt_log_file.close()
+            if exception:
+                process.kill()
+                raise exception
```
MassyB, line 536: piggyback: if there is an exception the dbt log files is closed before it's raised

MassyB (commented):
pointed on some things.

integration/common/openlineage/common/provider/dbt/structured_logs.py
mobuchowski: Should we kill the underlying process anyway? I don't think so - we should log and probably stop processing data, but the underlying dbt process should survive us failing to read log lines IMO.

integration/common/openlineage/common/utils.py
```diff
@@ -161,9 +161,35 @@ def has_lines(text_file: TextIO):
...
+class IncrementalFileReader:
+    """
+    dbt process writes to the log file incrementally. Sometimes lines are written incomplete.
+    This class is responsible of reading complete lines only and returns them via its methods.
+    """
+
+    def __init__(self, text_file: TextIO):
+        self.text_file = text_file
+        self.incomplete_line = ""
+        self.chunk_size = 1024
```
mobuchowski, line 173: Default block size in most linux filesystems is 4096 - probably no reason to use smaller value

integration/common/openlineage/common/utils.py
```diff
@@ -161,9 +161,35 @@ def has_lines(text_file: TextIO):
...
+        incomplete line chars are saved in the incomplete_line member
+        """
+
+        chunk = self.incomplete_line + self.text_file.read(self.chunk_size)
+        while chunk and chunk != self.incomplete_line:
+            line = []
+            next_line_start = 0
+            for i in range(len(chunk)):
+                char = chunk[i]
+                if char != os.linesep:
```
chendatadog, line 187: Can this be a string index() instead of a for loop?

integration/common/openlineage/common/utils.py
MassyB: I'm reading the file in text mode (utf-8 encoded, 1 to 4 bytes for each char). I made the assumption of 4 bytes per char. 

I can bump it to 4096. It will read more than one FS blocks.

integration/common/openlineage/common/utils.py
MassyB: I need the index below to keep track of when the next_line starts 
``` python
next_line_start = i + 1
```

integration/common/openlineage/common/utils.py
mobuchowski: Okay, 1024 is good then.

MassyB: I bumped it. It's not going to do any harm IMO.

integration/common/openlineage/common/provider/dbt/structured_logs.py
MassyB: updated it. I'm not killing the process

integration/common/openlineage/common/utils.py
mobuchowski: I think it can be somehow optimized - string methods like `.index()` or `.find()` drop down to C to be faster. However we can do this in follow-up PR, and merge the fix faster.

mobuchowski (approved):

PR #3594, opened 2025-04-01
Fix metrics sent through DebugFacet

pawel-big-lebowski:
DebugFacet contains only zero valued metrics if no micrometer-registry is declared.

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
@@ -11,6 +11,7 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
 import io.openlineage.client.Environment;
 import io.openlineage.client.OpenLineageConfig;
 import io.openlineage.client.circuitBreaker.CircuitBreaker;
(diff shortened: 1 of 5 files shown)


mobuchowski (commented):
Shouldn't we register it only if debug facet is enabled? The point of not doing it by default was to not have any overhead if user did not use the metrics in any way.

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -381,6 +383,19 @@ private void initializeContextFactoryIfNotInitialized(SparkConf sparkConf, Strin
   private static void initializeMetrics(OpenLineageConfig<?> openLineageConfig) {
     meterRegistry =
         MicrometerProvider.addMeterRegistryFromConfig(openLineageConfig.getMetricsConfig());
+
+    // register SimpleMeterRegistry if no other registries are present and debug facet is enabled
+    if (meterRegistry instanceof CompositeMeterRegistry
+        && ((CompositeMeterRegistry) meterRegistry).getRegistries().isEmpty()
```
mobuchowski, line 389: Top level registry coming from `addMeterRegistryFromConfig` will always be CompositeMeterRegistry, no need to check that.

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -381,6 +383,19 @@ private void initializeContextFactoryIfNotInitialized(SparkConf sparkConf, Strin
...
+        && openLineageConfig.getFacetsConfig() != null
+        && openLineageConfig.getFacetsConfig().getDisabledFacets() != null
+        && openLineageConfig.getFacetsConfig().getDisabledFacets().containsKey("debug")
+        && !openLineageConfig.getFacetsConfig().getDisabledFacets().get("debug")) {
+      meterRegistry =
+          new CompositeMeterRegistry(
+              Clock.SYSTEM, Collections.singletonList(new SimpleMeterRegistry()));
```
mobuchowski, line 396: Please don't overwrite `meterRegistry` - it won't be visible for anyone else who uses `MicrometerProvider.getMeterRegistry`. You can simply add that `SimpleMeterRegistry` using `MicrometerProvider.addMeterRegistry`.

mobuchowski (changes requested):

mobuchowski (approved):

PR #3570, opened 2025-03-18
add dbt glue adapter support

mehdimld:
### Problem

This PR adds support for the dbt-glue adapter to the OpenLineage dbt integration. The dbt-glue adapter is a community adapter that allows dbt to work with AWS Glue, and currently, it's not supported in OpenLineage's dbt integration.

### Solution

This change extends the OpenLineage dbt integration to support the dbt-glue adapter, allowing users who use dbt with AWS Glue to track their data lineage through OpenLineage. The implementation follows the same pattern as other supported dbt adapters (like dbt-spark, dbt-athena, etc.).

This change does not modify the OpenLineage core model or facets, as it simply adds support for a new adapter within the existing dbt integration framework.

#### Integration Scope:
- Adds support for dbt-glue adapter in the OpenLineage dbt integration
- Tested with AWS Glue and dbt-glue adapter
- Compatible with existing dbt integration patterns

#### One-line summary:
Add dbt-glue adapter support to OpenLineage dbt integration

### Checklist

- [x] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [x] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [ ] Your changes are accompanied by tests (_if relevant_)
- [x] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [x] You've updated any relevant documentation (_if relevant_)
- [x] Your comment includes a one-liner for the changelog about the specific purpose of the change
- [ ] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_not relevant - no schema changes_)
- [x] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----
SPDX-License-Identifier: Apache-2.0\
Copyright 2018-2025 contributors to the OpenLineage project

PR #3625, opened 2025-04-09
followup: add dbt glue adapter support

mobuchowski:
This is a followup for https://github.com/OpenLineage/OpenLineage/pull/3570/commits with whitespace fixed.


harels (approved):
thanks!

Issue #3547, opened 2025-03-11
[BUG][SPARK] self-join results in duplicate inputs


pawel-big-lebowski:
### OpenLineage integration/client

Spark integration

### OpenLineage version

unrelated

### Technology and package versions

self-join results in duplicate inputs

### Environment configuration

can be reproduced within integration tests 

### Problem details

Self-join results in duplicate inputs

### What you think should happen instead

a single input should be detected

### How to reproduce

This can be reproduced within `SparkIcebergIntegrationTest` with: 
```
@Test
  void testInputDuplicatesFromSelfJoin() {
    clearTables("tbl_duplicates1", "tbl_duplicates2", "temp");
    createTempDataset(2).createOrReplaceTempView("temp");

    spark.sql("CREATE TABLE tbl_duplicates1 USING iceberg AS SELECT * FROM temp");
    spark.sql("CREATE TABLE tbl_duplicates2 USING iceberg SELECT t1.a as a1, t2.a as a2 FROM tbl_duplicates1 t1 JOIN tbl_duplicates1 t2 ON t1.a = t2.b");

    MockServerUtils
        .getEventsEmitted(mockServer)
        .stream()
        .map(RunEvent::getInputs)
        .map(inputs -> inputs.stream().map(InputDataset::getName).collect(Collectors.toList()))
        .forEach(
            // make sure inputs name do not repeat
            inputs -> assertThat(inputs).doesNotHaveDuplicates()
        );
  }
```

resulting in:
```
Found duplicate(s):
  ["/tmp/iceberg/default/tbl_duplicates1"]
in:
  ["/tmp/iceberg/default/tbl_duplicates1", "/tmp/iceberg/default/tbl_duplicates1"]
```

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

PR #3630, opened 2025-04-14
[Spark] Remove duplicates when building Input Datasets

luke-hoffman1:
### Problem

Spark's logical and optimized plans can contain multiple distinct nodes (e.g., RelationV2) that reference the same underlying logical data source. This commonly occurs in scenarios like self-joins or when joining the same table multiple times using different aliases, as demonstrated in the following tests `testInputDuplicatesFromSelfJoin` `testInputDuplicatesFromJoiningSameTableTwice` respectively.

The current logic in `OpenLineageRunEventBuilder.buildInputDatasets` traverses the LogicalPlan and applies visitors to each node. While the visitors correctly identify each relevant data source node, the results are aggregated using `Stream.collect(Collectors.toList())` without considering the logical identity (namespace + name) of the generated InputDataset objects. This leads to duplicate entries appearing in the `RunEvent.inputs` list when the same data source is represented by multiple nodes in the plan.

Closes: #3547

### Solution

This change modifies the `OpenLineageRunEventBuilder.buildInputDatasets` method to correctly handle potential duplicate datasets identified during logical plan traversal.

Instead of collecting the stream of potential InputDataset objects directly into a List, the code now uses `java.util.stream.Collectors.toMap`. This collector uses a key derived from the dataset's namespace and name (via a small, compatible helper class DatasetKey) to ensure uniqueness. If multiple InputDataset objects are generated for the same key (i.e., the same logical dataset identified multiple times in the plan), the merge function (ds1, ds2) -> ds1 keeps only the first instance encountered.

The unique InputDataset objects (the values from the resulting map) are then converted back into a List. This deduplicated list is used for the subsequent facet merging logic and for populating the final RunEvent.inputs.

This approach correctly reflects the distinct logical datasets involved in the operation, even when they are referenced multiple times in the Spark plan.

#### One-line summary:

Fix duplicate input datasets when Spark plan contains multiple relations for the same source.

### Checklist

----


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
```diff
@@ -234,6 +237,29 @@ RunFacetsBuilder constructRunFacetsBuilder(
     return facetBuilder;
   }
 
+  private static class DatasetKey {
```
pawel-big-lebowski, line 240: Could you use `io.openlineage.client.utils.DatasetIdentifier` class instead?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
```diff
@@ -266,7 +292,20 @@ private List<InputDataset> buildInputDatasets(List<Object> nodes) {
...
+            .collect(
+                Collectors.toMap(
+                    // Key Mapper: Create a DatasetKey from namespace and name
+                    dataset -> new DatasetKey(dataset.getNamespace(), dataset.getName()),
+
+                    // Value Mapper: The dataset itself
+                    Function.identity(),
+
+                    // Merge Function: If duplicates found, keep the first one encountered
+                    (ds1, ds2) -> ds1));
```
pawel-big-lebowski, line 304: Would it make sense to use `io.openlineage.client.OpenLineageClientUtils#mergeFacets` instead of picking the first one?

pawel-big-lebowski (commented):
Nice improvement, thank you for working on this. 
Please take a look at the comments and let me know if this makes sense  to you.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
ddebowczyk92: wouldn't be easier to use Set instead of Map? You get merge for free

mobuchowski:
@luke-hoffman1 PMD rules are failing:
```
/home/circleci/openlineage/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkIcebergIntegrationTest.java:739: JUnitTestsShouldIncludeAssert:  JUnit tests should include assert() or fail()
/home/circleci/openlineage/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkIcebergIntegrationTest.java:756: JUnitTestsShouldIncludeAssert:  JUnit tests should include assert() or fail()
/home/circleci/openlineage/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkIcebergIntegrationTest.java:794: SystemPrintln:  System.out.println is used
```

mobuchowski:
@luke-hoffman1 sounds like the fix works - could we fix those PMD configs to merge the PR and release it maybe next week?

luke-hoffman1:
@mobuchowski, before we merge this, I am also working with @ddebowczyk92 and @pawel-big-lebowski on an issue where duplicates are getting multiplied. This solution only de-dups them after the fact, but we are investigating a potentially better solution closer to the root of the issue.

Is that okay if we keep this open until we find out if there is a better alternative? Will hopefully have an answer sometime next week.

mobuchowski:
Sure. Let's keep this open then as a draft PR - I was under impression this was ready to be merged outside of the PMD issues.

luke-hoffman1:
At first I was too 🙂 but after further investigation I was observing a deeper issue..

PR #3636, opened 2025-04-14
Flink 2 documentation doc

pawel-big-lebowski:
Refactors existing Flink documentation and adds Flink 2.x section.


website/docs/integrations/flink/flink1.md
```diff
@@ -0,0 +1,82 @@
+---
+sidebar_position: 4
+title: Flink 1.x
+---
+
```
mobuchowski, line 5: Can we add warning here that the doc here is for Flink 1? I know it will be in the sidebar but would not hurt us to have this.

mobuchowski (approved):

JDarDagran (approved):

PR #3639, opened 2025-04-14
spark,iceberg: add handling of JDBC catalog

mobuchowski:
Add handling of Iceberg's JDBC Catalog.


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
```diff
@@ -163,6 +163,9 @@ private Optional<DatasetIdentifier> getSymlinkIdentifier(
     } else if ("glue".equals(catalogType)) {
       log.debug("Getting symlink for glue");
       value = getGlueIdentifier(tableName, session);
+    } else if ("jdbc".equals(catalogType)) {
```
pawel-big-lebowski, line 166: I hope someone will find some time extract this into several `IcebergWhateverHandler` classes.

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifier.java
```diff
@@ -41,6 +42,12 @@ public DatasetIdentifier withSymlink(Symlink symlink) {
     return this;
   }
 
+  public DatasetIdentifier merge(DatasetIdentifier ident) {
+    symlinks.add(new Symlink(ident.getName(), ident.getNamespace(), SymlinkType.UNKNOWN));
+    symlinks.addAll(ident.getSymlinks());
```
pawel-big-lebowski, line 47: What if there are common symlinks?

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandlerTest.java
```diff
@@ -168,6 +169,52 @@ void testGetDatasetIdentifierForRest() {
         .hasFieldOrPropertyWithValue("type", DatasetIdentifier.SymlinkType.TABLE);
   }
 
+  @ParameterizedTest
+  @CsvSource({
+    "jdbc:mysql://test.1234567890.us-west-2.rds.amazonaws.com:3306/database,mysql://test.1234567890.us-west-2.rds.amazonaws.com:3306",
+    "jdbc:postgresql://localhost:5432/mydb?ssl=true&applicationName=myApp,postgres://localhost:5432",
+    "jdbc:oracle:thin:@//myhost:1521/myservice?TNS_ADMIN=/path/to/tns,oracle://myhost:1521",
+    "jdbc:sqlserver://localhost:1433;databaseName=mydb;user=me;password=secret,sqlserver://localhost:1433",
```
pawel-big-lebowski, line 177: Wouldnt a single test scenario be enough? Looks like this is testing `JdbcDatasetUtils` here and duplicating tests should be avoided.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
```diff
@@ -135,8 +136,7 @@ public DatasetIdentifier getDatasetIdentifier(
         maybeTableLocation.orElseGet(
             () -> reconstructDefaultLocation(new Path(warehouseLocation), identifier));
     DatasetIdentifier di = PathUtils.fromPath(tableLocation);
-    maybeSymlink.ifPresent(
-        symlink -> di.withSymlink(symlink.getName(), symlink.getNamespace(), SymlinkType.TABLE));
+    maybeSymlink.ifPresent(di::merge);
```
pawel-big-lebowski, line 139: `di` will contain dataset name and namespace extracted from table location and use it as a symlink. 
Why do you want to assign symlink type as `unknown`. Wouldn't  `TABLE_PATH` be a better choice? It looks like the code knows where the symlink comes from.

mobuchowski: In this concrete case, yes - although I would not name it `TABLE_PATH` - it's either a table or a path :) 

However, I plan to add more types that we miss - for example, we don't capture catalog name, and we probably should - under type `CATALOG`.

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandlerTest.java
mobuchowski: I disagree - here I care purely about namespace part and stripping of unnecessary params. `JdbcDatasetUtils` uses those to determine database name that's used as part of dataset name.

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifier.java
mobuchowski: Good catch - however, how would you determine that? IMO that's still on the caller of this method.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
mobuchowski: Yup, that code isn't the prettiest...

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifier.java
pawel-big-lebowski: What's on the caller of this method? I thought just about converting `List` into `Set`

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
pawel-big-lebowski: so what's your opinion: stay with `unknown` or add other types?

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifier.java
mobuchowski: Then we need good equals and hashcode 🙂 
I think those are the questions we need to answer, list vs set is just a technicality

mobuchowski: I left `TABLE` for now, however, will revisit that later.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandler.java
mobuchowski: I left TABLE for now, however, will revisit that later.

client/java/src/main/java/io/openlineage/client/utils/DatasetIdentifier.java
dolfinus: I don't see there is this method used, and why `di1.merge(di2)` expects that d2 is always a TABLE symlink.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkJdbcIcebergIntegrationTest.java
```diff
@@ -0,0 +1,167 @@
...
+    assertThat(
+            runEvents.stream()
+                .flatMap(e -> e.getInputs().stream())
+                .filter(
+                    e ->
+                        hasSymlinkWith(
+                            e,
+                            symlink ->
+                                symlink.getName().endsWith("scan_source")
+                                    && symlink.getNamespace().startsWith("sqlite"))))
```
dolfinus, line 123: Could we compare symlink to exact name & namespace?

mobuchowski: Updating this part right now actually.

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandlerTest.java
```diff
@@ -168,6 +169,52 @@ void testGetDatasetIdentifierForRest() {
...
+            new HashMap<>());
+
+    assertThat(datasetIdentifier)
+        .hasFieldOrPropertyWithValue("namespace", "s3://lakehouse")
+        .hasFieldOrPropertyWithValue("name", "warehouse/database/table");
+
+    assertThat(datasetIdentifier.getSymlinks())
+        .singleElement()
+        .hasFieldOrPropertyWithValue("namespace", namespace)
+        .hasFieldOrPropertyWithValue("name", "database.table")
```
dolfinus, line 214: IMHO this could be confusing and not consistent with other implementations.
The symlink namespace points to Postgres/MySQL/Oracle instance, like user read/write to JDBC source (which is not the case). At the same time, dataset name corresponds to Iceberg metastore info (Iceberg database+name), but not the Postgres database/schema/name, so user actually is not able to access this table by connecting to specified namespace.

For Hive metastore it doesn't matter if underlying metastore is Thrift or JDBC, it always have `hive://`` protocol, to distinguish between Postgres namespace+name and Hive metastore namespace+name:
https://github.com/OpenLineage/OpenLineage/blob/bf9a7b729062eaa6b9b1968c238cb60deb3b73d0/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/PathUtilsTest.java#L145-L155

mobuchowski: If we really want to hold the property of being able to connect to the dataset via specified identifier, I don't think there is other option than having specific namespaces for each Iceberg catalog type.

dolfinus: Maybe this should be changed to:
```
namespace: s3://bucket
name: warehouse/database.db/table
symlinks:
  - namespace: iceberg://host:5432
    name: database.table
```
This is also a bit confusing, as Iceberg is a file format, and not a some kind of server. Also other file formats, like Parquet, Hudi or Delta don't produce `delta://` symlinks, so why Iceberg should?

I can move this discussion to a dedicated issue.

mobuchowski: I want to do something similar to 
```
symlinks:
  - namespace: iceberg://host:5432
    name: database.table
```
as a followup, but also introducing `CATALOG` symlink type.

I think the solution in this case would be to be more explicit about the fact that we present a "logical" dataset name that makes sense in the context of the job. If you use particular catalog, the logical dataset name makes sense in this context. However, it does not hold the property of being "globally unique dataset name" - and it can't, but we should handle it neverthless.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkJdbcIcebergIntegrationTest.java
mobuchowski: Done.

dolfinus: No commits were pushed :smile:

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkJdbcIcebergIntegrationTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+    assertThat(
+            runEvents.stream()
+                .flatMap(e -> e.getOutputs().stream())
+                .filter(
+                    e ->
+                        hasSymlinkWith(
+                            e,
+                            symlink ->
+                                symlink.getName().endsWith("default.scan_target")
```
dolfinus, line 134:
```suggestion
                                symlink.getName().equals("default.scan_target")
```

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandlerTest.java
dolfinus: Could you please provide an example with `CATALOG` symlink type?

dolfinus: Opened a discussion #3642

mobuchowski: @dolfinus relating to the discussion above - in this particular case, what would you think of 

```
namespace: s3://bucket
name: warehouse/database.db/table
symlinks:
  - namespace: jdbc://host:5432
    name: database.table
    type: TABLE
  - namespace: my_spark_catalog://host:5432 
    name: database.table
    type: CATALOG
```

here we lose information on what database the particular JDBC catalog points to unfortunately.

dolfinus: I don't see how the second symlink could be used by end users. And also how `jdbc://` is applied to the comment https://github.com/OpenLineage/OpenLineage/pull/3639#discussion_r2046477198

mobuchowski: >I don't see how the second symlink could be used by end users.

I think many users would expect that if they query data from `my_spark_catalog.database.table` they would see this in their lineage tool, instead of `s3://bucket/warehouse/database.db/table` which is just implementation noise for them. Even if it's canonical location for the table.

mobuchowski:
This is paused until we have consensus on issue #3642 and https://github.com/OpenLineage/OpenLineage/pull/3659

Issue #3596, opened 2025-04-01
[CLIENTS] add HTTP retries


merobi-hub:
### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

mobuchowski:
It's done.

PR #3646, opened 2025-04-17
refactor python HTTP transport

mobuchowski:
- set headers in one place
- set session params in one place
- add default retry config

### DIFF
--- a/client/python/openlineage/client/transport/http.py
+++ b/client/python/openlineage/client/transport/http.py
@@ -11,10 +11,11 @@ from typing import TYPE_CHECKING, Any
 from urllib.parse import urljoin
 
 import attr
+import urllib3.util
 
 if TYPE_CHECKING:
     from openlineage.client.client import Event, OpenLineageClientOptions
-    from requests.adapters import HTTPAdapter, Response
+    from requests import Response
 
 import http.client as http_client
 
@@ -22,6 +23,7 @@ from openlineage.client.serde import Serde
 from openlineage.client.transport.transport import Config, Transport
 from openlineage.client.utils import get_only_specified_fields, try_import_from_string
 from requests import Session
+from requests.adapters import HTTPAdapter
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


client/python/openlineage/client/transport/http.py
```diff
@@ -205,8 +192,31 @@ def _auth_headers(self, token_provider: TokenProvider) -> dict:  # type: ignore[
...
+            session.mount(self.url, self.config.adapter)
+        else:
+            session.mount(self.url, self._prepare_adapter())
+
+    def _prepare_adapter(self) -> HTTPAdapter:
+        retry = urllib3.util.Retry(
+            total=3,
+            read=3,
+            connect=3,
+            backoff_factor=0.3,
```
kacpermuda, line 206: I think we should allow the user to overwrite this with config values. Can we add that to HTTP config? At least total retries and maybe backoff_factor?

mobuchowski: User can do that today by providing their own HTTP adapter, do you think it's not enough? Or is config-level override better for you?

kacpermuda: I was thinking about config-level configuration so that it's easier to provide with env variables or yaml file. Adding few retries with a very low sleep time in between should rarely be an issue, but it might be nice from a user perspective to easily adjust it, without the need to write my own adapter (as we do f.e. with timeout, where it's very easy to provide).

kacpermuda (approved):

kacpermuda:
related #3596

client/python/tests/test_client.py
```diff
@@ -866,6 +866,30 @@ class TestOpenLineageConfigLoader:
                     }
                 },
             ),
+            (
+                {
+                    "OPENLINEAGE__TRANSPORT__TYPE": "http",
+                    "OPENLINEAGE__TRANSPORT__URL": "http://localhost:5050",
+                    "OPENLINEAGE__TRANSPORT__RETRY__TOTAL": "5",
+                    "OPENLINEAGE__TRANSPORT__RETRY__BACKOFF_FACTOR": "0.2",
```
dolfinus, line 874: Should new parameters be documented?

mobuchowski: Updated the docs.

PR #3647, opened 2025-04-17
explain where dbt error code comes from

mobuchowski:
Explain to the user that error code returned by `dbt-ol` comes from dbt execution, not from OL.

### DIFF
--- a/integration/dbt/openlineage/dbt/__init__.py
+++ b/integration/dbt/openlineage/dbt/__init__.py
@@ -190,6 +190,7 @@ def consume_structured_logs(
         return_code = 1
 
     logger.info("Emitted %d OpenLineage events", emitted_events)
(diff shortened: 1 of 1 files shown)


kacpermuda (approved):

integration/dbt/openlineage/dbt/__init__.py
```diff
@@ -308,6 +309,7 @@ def consume_local_artifacts(
                 exc_info=True,
             )
     logger.info("Emitted %d OpenLineage events", emitted_events)
+    logger.info("Underlying dbt execution returned %d", return_code)
```
pawel-big-lebowski, line 312: any reason for not having a single log message?

mobuchowski: Those messages are about different things. I don't think having them separate can cause any issues, and is more readable.

PR #3650, opened 2025-04-18
Support schema facet for Avro

pawel-big-lebowski:
* Work on schema extraction for Avro messages sent through Kafka. 
 * upgrade flink-kafka-connector to an officially published version of (released on 23rd April)


integration/flink/flink2/src/test/java/io/openlineage/flink/Flink2ContainerTest.java
```diff
@@ -182,6 +189,34 @@ void testKafkaTopicPatternJob() {
...
+    // check schema facet from Avro app for input topic
+    Optional<InputDataset> input = events.stream().flatMap(e -> e.getInputs().stream()).findFirst();
+    assertThat(input.get().getFacets().getSchema()).isNotNull();
+    List<SchemaDatasetFacetFields> fields = input.get().getFacets().getSchema().getFields();
+
+    assertThat(fields).hasSize(2);
+    assertThat(fields.get(0)).extracting("name", "type").contains("id", "string");
+    assertThat(fields.get(1)).extracting("name", "type").contains("version", "long");
+
+    // output will work after -> https://github.com/apache/flink/pull/26507
```
mobuchowski, line 207: Will we support only the version after `https://github.com/apache/flink/pull/26507` gets released or will this be optional?

mobuchowski (approved):
Approved with one question

integration/flink/flink2/src/test/java/io/openlineage/flink/Flink2ContainerTest.java
pawel-big-lebowski: OpenLineage code translates Flink's `TypeInformation` into schema facet. 26507 implements interface to return TypeInformation. 

So: 
 * We don't need to upgrade OL connector to support this in future. It's just users who will rely on the updated flink Avro package. The feature is ready on our side.
 * At the same time, it makes sense to uncomment the added tests at some point in future to increase the coverage.

Issue #3642, opened 2025-04-16
[Discussion] Consistent approach with metastore symlinks


dolfinus:
**Dive-in:**

Some datasets are accessed as-is, e.g. paths on filesystem:
```yaml
namespace: s3://bucket
name: warehouse/somedb.db/sometable
```

But some engines have a concept of catalog or metastore which stores metadata about this dataset - DDL, schema, custom properties and so on. For example, Hive metastore:
```yaml
namespace: s3://bucket
name: warehouse/somedb.db/sometable
symlinks:
 - namespace: hive://somehost:10000
   name: somedb.sometable
```

But the current way of representing these catalogs/metastore items is not consistent within OpenLineage integrations.

For Hive metastore, metadata can be stored in external database providing JDBC API - Postgres, MySQL, Derby and so on. The namespace is always looks like `hive://host:port`, even if JDBC URL was `postgres://host:port`.

For Iceberg, there are Hive metastore `hive://host:port`, HDFS `hdfs://path/to/metadata`, REST API metastore `http://host:port`, JDBC metastore (ongoing discussion on #3639), [Nessie catalog](https://projectnessie.org/iceberg/spark/#configuration) `http://host:port`

For Flink2 SQL, [catalog](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/catalogs/#catalog-types) can be stored in-memory, Hive metastore or in JDBC (Postgres, MySQL only). But for some reason, symlinks always have format like:
```yaml
namespace: kafka://server:port
name: mytopic
symlinks:
 - namespace: kafka://server:port
   name: default_catalog.default_database.mytable
```
Although Kafka is not a catalog.
https://github.com/OpenLineage/OpenLineage/blob/1.31.0/integration/flink/flink2/src/main/java/io/openlineage/flink/visitor/identifier/KafkaTableLineageDatasetIdentifierVisitor.java

There are also generic metastore/catalog implementations, like:
* [Apache Gravitino](https://gravitino.apache.org/docs/0.6.1-incubating/iceberg-rest-service/) - for Iceberg same REST API catalog with address like `http://host:port/iceberg`, but also capable of tracking objects in Postgres, MySQL, Doris, Kafka and HDFS.
* [UnityCatalog](https://docs.unitycatalog.io/integrations/unity-catalog-spark/) - mainly used for Spark + Delta, but also can be used by Trino, DuckDB and so on.

**Questions:**

[...]

PR #3657, opened 2025-04-22
Fix kafka table symlink

pawel-big-lebowski:
Symlink dataset identifier for Flink2 contains unnecessary catalog_name as identifier prefix.


dolfinus:
My concern in #3642 was not only about catalog name as dataset name prefix, but also about namespace format `kafka://...` which is misleading because Flink metastore is not Kafka.

dolfinus:
> there is no external catalog nor Flink metastore.

Yes, this particular example uses Flink [GenericInMemoryCatalog|https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/catalogs/#genericinmemorycatalog]. So I thought about namespace `flink://...`, just like OL jobs.

integration/flink/flink2/src/test/java/io/openlineage/flink/visitor/identifier/catalog/GenericInMemoryCatalogSymlinkProviderTest.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Test
+  void testGetSymlinkRemovesOnlyCatalogFromDatasetName() {
+    when(lineageDataset.name()).thenReturn("catalogName.catalogName.tableName");
+    when(catalogContext.getCatalogName()).thenReturn("catalogName");
+    assertThat(provider.getSymlink(catalogContext, lineageDataset))
+        .isPresent()
+        .get()
+        .hasFieldOrPropertyWithValue("namespace", "flink://")
```
dolfinus, line 50: Doesn't flink have an address?

pawel-big-lebowski: Perhaps it's available, but still this something like the address of Spark driver. There is not much value out of it.

dolfinus: In this case Flink instances running on different hosts will produce the same symlink, although they can be independent, and lineage graph could become misleading

pawel-big-lebowski: I still think it's the same as with local Spark pointing to `file`. Globally this can be misleading, but it is convenient for testing.

pawel-big-lebowski: The other solution would be to remove the symlink in this scenario for `GenericInMemoryCatalog` while leaving the `CatalogSymlinkProvider` for future extensions.

dolfinus: Could we discuss this in #3642?

harels:
closing due to inactivity, please feel free to reopen otherwise.

PR #3659, opened 2025-04-22
spec: add CatalogDatasetFacet

mobuchowski:
Introduces a new CatalogDatasetFacet to the OpenLineage spec. 

More relevant discussion: [issue #3642](https://github.com/OpenLineage/OpenLineage/issues/3642) and [pull request #3639](https://github.com/OpenLineage/OpenLineage/pull/3639).
In short, we don't handle Iceberg tables accessed via JDBC-based or custom catalogs in Spark. Additionally, we don't capture catalog name - and ​this is how users most frequently interact with those tables.

Previously, the approach in PR #3639 attempted to handle JDBC catalogs by creating symlinks with non-compliant dataset information, which was not ideal. The new `CatalogDatasetFacet` standardizes way to include catalog-specific metadata in the lineage events without relying on symlinks.

Drawback of this approach is that consumers who wish to use this information as a dataset identifier need to look at one additional facet: not only top-level dataset name and namespace and symlink facet, but also this one.​

### DIFF
--- a/client/python/openlineage/client/generated/catalog_dataset.py
+++ b/client/python/openlineage/client/generated/catalog_dataset.py
@@ -1,0 +1,32 @@
+# Copyright 2018-2025 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import attr
+from openlineage.client.generated.base import DatasetFacet
+
+
+@attr.define
+class CatalogDatasetFacet(DatasetFacet):
+    framework: str
+    """The storage framework for which the catalog is configured"""
+
+    type: str
+    """Type of the catalog."""
+
+    name: str
+    """Name of the catalog, as configured in the source system."""
+
+    metadataUri: str | None = attr.field(default=None)  # noqa: N815
+    """URI or connection string to the catalog, if applicable."""
+
+    warehouseUri: str | None = attr.field(default=None)  # noqa: N815
+    """URI or connection string to the physical location of the data that the catalog describes."""
+
+    source: str | None = attr.field(default=None)
+    """Source system where the catalog is configured."""
+
+    @staticmethod
+    def _get_schema() -> str:
+        return "https://openlineage.io/spec/facets/1-0-0/CatalogDatasetFacet.json#/$defs/CatalogDatasetFacet"
--- a/spec/facets/CatalogDatasetFacet.json
+++ b/spec/facets/CatalogDatasetFacet.json
@@ -1,0 +1,56 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/CatalogDatasetFacet.json",
+  "$defs": {
+    "CatalogDatasetFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/DatasetFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "framework": {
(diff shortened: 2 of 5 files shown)


client/python/openlineage/client/generated/catalog_dataset.py
```diff
@@ -0,0 +1,29 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import attr
+from openlineage.client.generated.base import DatasetFacet
+
+
+@attr.define
+class CatalogDatasetFacet(DatasetFacet):
```
dolfinus, line 11: Java classes are generated while building .jar, am I right?

spec/facets/CatalogDatasetFacet.json
```diff
@@ -0,0 +1,51 @@
...
+              "type": "string",
+              "example": "spark, flink, hive"
+            },
+            "type": {
+              "description": "Type of the catalog.",
+              "type": "string",
+              "example": "jdbc, glue, polaris"
+            },
+            "uri": {
+              "description": "URI or connection string to the catalog, if applicable.",
```
dolfinus, line 29: This description means that uri can be not applicable to some catalogs, but this field is required in JSON schema

spec/facets/CatalogDatasetFacet.json
```diff
@@ -0,0 +1,51 @@
...
+        },
+        {
+          "type": "object",
+          "properties": {
+            "technology": {
+              "description": "The storage framework for which the catalog is configured",
+              "type": "string",
+              "example": "iceberg, delta, hive"
+            },
+            "source": {
```
dolfinus, line 18: Well, OpenLineage events already have processing engine facet, what's the difference?

client/python/openlineage/client/generated/catalog_dataset.py
mobuchowski: Yes.

We could move this to runtime and add pre-commit checks, to make those changes more visible in PRs.

mobuchowski:
@dolfinus @pawel-big-lebowski I've updated the proposed spec:
- renamed `technology` to framework, as in table type framework or catalog framework. IMO it's slightly better, although I feel there are no short, descriptive, non-overloaded names for that
- only required fields are framework, type, and name: in case where name or type does not make sense, for example because some system only allows to provide you a singular catalog, or where there is only single type of catalog for given framework, it's okay to provide some default value. 
- distinguished between `uri` and `warehouseUri` - the first provides URI of a catalog - for example JDBC connection string - and the second provides the physical location of the data in this catalog. In some cases this does not make sense, like for `delta`, and those fields are optional.
- make the `source` field optional. Yes, it duplicates `processing_engine` facet for Spark, but I think there are cases (like `DatasetEvent` cases) where this would be useful.

client/python/openlineage/client/generated/catalog_dataset.py
dolfinus: This is just a question :)

website/static/spec/facets/1-0-0/CatalogDatasetFacet.json
```diff
@@ -0,0 +1,56 @@
...
+              "description": "Type of the catalog.",
+              "type": "string",
+              "example": "jdbc, glue, polaris"
+            },
+            "name": {
+              "description": "Name of the catalog, as configured in the source system.",
+              "type": "string",
+              "example": "my_iceberg_catalog"
+            },
+            "uri": {
```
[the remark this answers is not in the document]
mobuchowski: Ok, done.

dolfinus:
LGTM, thanks @mobuchowski

JDarDagran (approved):

PR #3674, opened 2025-05-01
feature: Support for Unity Catalog OSS catalog

### LINKED ISSUE #4654: [FEATURE] Unity Catalog OSS support
### Motivation

The open source Unity Catalog is something completely different than Unity Catalog in Databricks (although Databricks pretends they are the same). When you use it in your projects the events are incomplete and are missing tables symlinks (because you are using a different catalog class).

### Description

We should provide a handler that supports https://github.com/unitycatalog/unitycatalog/blob/main/connectors/spark/src/main/scala/io/unitycatalog/spark/UCSingleCatalog.scala
The change should be properly tested. It should provide integration test that is testing with real Unity Catalog using local file system and some external storage (for example abfs).

### Related issues

_No response_

arturowczarek:
### Problem

Open source Unity Catalog (https://github.com/unitycatalog/unitycatalog) uses its own Spark catalog `io.unitycatalog.spark.UCSingleCatalog`. There is not handler processing this catalog, so the output dataset facets are not being produced when we use OSS Unity Catalog.

### Solution

We should have a handler to support OSS Unity Catalog Spark catalog.

Underneath `UCSingleCatalog` may (but not always does) use `org.apache.spark.sql.delta.catalog.DeltaCatalog` but `io.openlineage.spark3.agent.lifecycle.plan.catalog.DeltaHandler` uses some of the `DeltaCatalog` methods which are not available in `UCSingleCatalog`, so we can't reuse it. We could use reflection to access the delegate `DeltaCatalog` but it may not always be present.

#### One-line summary:

Support for Unity Catalog open source

### Checklist

- [x] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained (ok, it is not small, but is self-contained)

Closes: #4654

----


arturowczarek:
This is a work in progres. I've copied most of the code from `DeltaHandler` and checked that it produces results in my case. But I need to determine which parts (except the ones I've already deleted) are truly necessary.

client/java/src/main/java/io/openlineage/client/utils/filesystem/ObjectStorageDatasetExtractor.java
```diff
@@ -12,9 +12,21 @@
...
+
+  /**
+   * @param scheme the scheme prefix this extractor matches (e.g. {@code abfs}, which also matches
+   *     {@code abfss} since matching is by prefix)
+   * @param canonicalScheme the scheme emitted in the resulting namespace; matched variants are
+   *     normalized to it (e.g. {@code abfs} and {@code abfss} both normalize to {@code abfss})
+   */
+  public ObjectStorageDatasetExtractor(String scheme, String canonicalScheme) {
     this.scheme = scheme;
+    this.canonicalScheme = canonicalScheme;
```
arturowczarek, line 29: As in description. I introduce this change because I test the change with abfss for the table location and noticed that if someone uses `abfs://` (without extra `s`), it is not normalized to `abfss`.

integration/spark/app/src/test/java/io/openlineage/spark/agent/DynamicParameter.java
arturowczarek, line 1: I've moved this class out of app into a new `test-commons` module where it is reachable by test suites from different modules.

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkTestUtils.java
```diff
@@ -82,10 +81,10 @@ public static HttpServer createHttpServer(HttpHandler handler) throws IOExceptio
 
   static SparkSession createSparkSession(Integer httpServerPort, String appName) {
     Path tmpDir = Paths.get(System.getProperty("java.io.tmpdir"));
-    UUID testUuid = UUID.randomUUID();
+    String testId = TestIds.randomHex();
```
arturowczarek, line 84: I noticed this pattern appears in many places. In the places where I was using it it created noise. I extracted it and asked AI to find other places where it could be useful. It found this one so I thought "why not"

integration/spark/app/build.gradle
```diff
@@ -112,6 +112,7 @@ dependencies {
     compileOnly("org.apache.spark:spark-sql-kafka-0-10_${scala}:${spark}")
     compileOnly("io.micrometer:micrometer-core:${micrometerVersion}")
 
+    testImplementation(testFixtures(project(path: ":test-commons")))
```
arturowczarek, line 115: We extracted `DynamicParameter` out of current (`app`) module into this one

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/catalog/CatalogHandler.java
```diff
@@ -3,7 +3,7 @@
 /* SPDX-License-Identifier: Apache-2.0
 */
 
-package io.openlineage.spark3.agent.lifecycle.plan.catalog;
+package io.openlineage.spark.agent.lifecycle.plan.catalog;
```
arturowczarek, line 6: Unity Catalog handler is a handler that is not working in Spark 3 anymore (it used to in the early days of development, but now it it is Delta 4 based, etc). So I had to move the Catalog Handler into a module (`shared`) where it could be reached by other spark modules.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/DatasetBuilderFactory.java
```diff
@@ -35,4 +36,13 @@ default Collection<ColumnLevelLineageVisitor> getColumnLevelLineageVisitors(
       OpenLineageContext context) {
     return Collections.emptyList();
   }
+
+  /**
+   * Catalog handlers contributed by this Spark version. Lets version-specific modules (e.g. Spark 4
+   * for open source Unity Catalog) add their {@link CatalogHandler}s to the shared catalog
+   * resolution flow without lower modules depending on them.
+   */
+  default List<CatalogHandler> getCatalogHandlers(OpenLineageContext context) {
```
arturowczarek, line 64: This is quite important change. In the first version I used SPI to load different catalog handlers automatically, but Tomasz suggested this approach is better (I agree). We pass this list to OpenLineageContext from where it is available in `CatalogUtils3` (the name could be changed BTW)

integration/spark/shared/src/main/java/io/openlineage/spark/api/OpenLineageContext.java
```diff
@@ -53,6 +54,11 @@ public class OpenLineageContext {
   // filled up only for SparkListenerApplication{Start,End} events
   @Setter @Getter UUID applicationUuid;
 
+  // The Spark-version-specific DatasetBuilderFactory, used to source version-specific catalog
+  // handlers (see CatalogUtils3). Null when the context is not built via the agent's
+  // ContextFactory.
+  @Getter DatasetBuilderFactory datasetBuilderFactory;
```
arturowczarek: I put the factories for the catalogs that could be potentially used around in this field

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/iceberg/IcebergHandler.java
```diff
@@ -65,10 +65,10 @@ public IcebergHandler(OpenLineageContext context) {
   @Override
   public boolean hasClasses() {
     try {
-      IcebergHandler.class.getClassLoader().loadClass("org.apache.iceberg.catalog.Catalog");
+      IcebergHandler.class.getClassLoader().loadClass("org.apache.iceberg.spark.SparkCatalog");
```
arturowczarek, line 68: `org.apache.iceberg.catalog.Catalog` is a generic, core catalog that is not specific to Spark. It was carried along with Unity Catalog. It was activating this handler but it was not necessary. I swapped this class for Spark specific `org.apache.iceberg.spark.SparkCatalog`

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/AbstractDatabricksHandler.java
```diff
@@ -6,6 +6,7 @@
 package io.openlineage.spark3.agent.lifecycle.plan.catalog;
 
 import io.openlineage.client.utils.DatasetIdentifier;
+import io.openlineage.spark.agent.lifecycle.plan.catalog.CatalogHandler;
```
arturowczarek, line 9: All the catalogs have to update the location of `CatalogHandler` because it was moved to be available in different Spark versions modules

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
```diff
@@ -32,9 +35,22 @@ private static List<CatalogHandler> getHandlers(OpenLineageContext context) {
             new DatabricksUnityV2Handler(context),
             new JdbcHandler(context),
             new V2SessionCatalogHandler(context));
+    // Handlers contributed by other modules (e.g. Unity Catalog OSS, which requires Spark 4) are
+    // discovered via ServiceLoader so that spark3 does not have to depend on them.
+    handlers.addAll(getProvidedHandlers(context));
```
arturowczarek: This is where we load catalog handlers from other modules

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
```diff
@@ -32,9 +35,22 @@ private static List<CatalogHandler> getHandlers(OpenLineageContext context) {
...
     return handlers.stream().filter(CatalogHandler::hasClasses).collect(Collectors.toList());
   }
 
+  private static List<CatalogHandler> getProvidedHandlers(OpenLineageContext context) {
+    // Catalog handlers contributed by the running Spark version's DatasetBuilderFactory (provided
+    // explicitly by the app module). Null when the context was not built via the agent's
+    // ContextFactory (e.g. some unit tests).
+    DatasetBuilderFactory builderFactory = context.getDatasetBuilderFactory();
```
arturowczarek: The default implementation of additional builder factories is an empty list. Then it is passed to `OpenLineageContext`. This is an extra piece of defence in case someone forgets to do it. It is better not to provide lineage than to fail an application I think.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
```diff
@@ -129,44 +127,7 @@ public Optional<OpenLineage.StorageDatasetFacet> getStorageDatasetFacet(
...
-      }
-    } else if (MethodUtils.getAccessibleMethod(deltaTable.getClass(), "initialSnapshot") != null) {
-      try {
-        return Optional.of((Snapshot) MethodUtils.invokeMethod(deltaTable, "initialSnapshot"));
-      } catch (InvocationTargetException | NoSuchMethodException | IllegalAccessException e) {
-        log.error("Could not invoke method", e);
-      }
-    }
-    return Optional.empty();
+    return DeltaVersionUtils.getDatasetVersion(deltaCatalog.loadTable(identifier));
```
arturowczarek, line 130: This thing was exactly the same in `UnityCatalogHandler` so I extracted it instead of copying.

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/column/InputFieldsCollectorTest.java
```diff
@@ -238,7 +238,7 @@ void collectWhenGrandChildNodeIsHadoopRelation() {
     verify(builder, times(1))
         .addInput(
             exprId,
-            new DatasetIdentifier("/path", "abfss://tmp@storage.dfs.core.windows.net"),
+            new DatasetIdentifier("path", "abfss://tmp@storage.dfs.core.windows.net"),
```
arturowczarek, line 241: Before I registered in `ObjectStorageDatasetExtractor` an extractor for `abfss`, no extractor used to claim`abfs`/`abfss` and it was a generic extractor that was handling them. The generic extractor leaves the `/` in front of the dataset name. It wasn't consistent with other location identifiers like s3 where the initial slash is dropped.
**THIS IS NOT BACKWARD COMPATIBLE**
Tell me what you think. I think we aim at consistency.

client/java/src/test/java/io/openlineage/client/utils/filesystem/FilesystemDatasetUtilsTestForWASBS.java
```diff
@@ -37,6 +37,27 @@ void testFromLocation() {
         .hasFieldOrPropertyWithValue("name", "warehouse/location");
   }
 
+  @Test
+  @SneakyThrows
+  void testFromLocationWithDifferentProtocol() {
```
arturowczarek, line 42: **THIS CHANGE IS NOT BACKWARD COMPATIBLE**
I think it makes sense, though.

client/java/src/test/java/io/openlineage/client/utils/filesystem/FilesystemDatasetUtilsTestForABFSS.java
```diff
@@ -0,0 +1,113 @@
...
+            FilesystemDatasetUtils.fromLocation(
+                new URI("abfss://container@account.dfs.core.windows.net")))
+        .hasFieldOrPropertyWithValue("namespace", "abfss://container@account.dfs.core.windows.net")
+        .hasFieldOrPropertyWithValue("name", "/");
+
+    assertThat(
+            FilesystemDatasetUtils.fromLocation(
+                new URI("abfss://container@account.dfs.core.windows.net/warehouse")))
+        .hasFieldOrPropertyWithValue("namespace", "abfss://container@account.dfs.core.windows.net")
+        .hasFieldOrPropertyWithValue("name", "warehouse");
```
arturowczarek, line 29: **THIS IS NOT BACKWARD COMPATIBLE**
Before this change, `abfss` and `abfs` used to be handled by generic dataset indentifier extractor that used to leave the preceding `/`. In this case the name would be `/warehouse` but it is not consistent with other schemes like s3.

client/java/src/test/java/io/openlineage/client/utils/filesystem/FilesystemDatasetUtilsTestForABFSS.java
```diff
@@ -0,0 +1,113 @@
...
+   * addresses the same data, so it is normalized to the canonical {@code abfss} scheme defined by
+   * the OpenLineage naming spec
+   */
+  @Test
+  @SneakyThrows
+  void testFromLocationWithDifferentProtocol() {
+    assertThat(
+            FilesystemDatasetUtils.fromLocation(
+                new URI("abfs://container@account.dfs.core.windows.net/warehouse")))
+        .hasFieldOrPropertyWithValue("namespace", "abfss://container@account.dfs.core.windows.net")
```
arturowczarek, line 55: **THIS IS NOT BACKWARD COMPATIBLE**
Here we normalize `abfs` to `abfss` scheme. The identifier we use in OpenLineage is just an agreed identifier. It doesn't say you can only access data with TLS or without. It could be anything else `azuredatalake://` as long as everybody accepts it. `abfs://` and `abfss://` with the same rest point to the same files anyway.

integration/spark/spark40/src/main/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogHandler.java
```diff
@@ -0,0 +1,181 @@
...
+    } catch (NoClassDefFoundError | Exception e) {
+      // If the class does not exist, or its loading fails for some reason, we handle that failure
+      // by returning false
+    }
+    return false;
+  }
+
+  @Override
+  public boolean isClass(TableCatalog tableCatalog) {
+    return UC_SINGLE_CATALOG_CLASS_NAME.equals(tableCatalog.getClass().getCanonicalName());
```
arturowczarek, line 59: I rely on name, because I want to couple `isClass` and `hasClasses` with the same class name. I don't use the actual class in this file anyway so there is no point importing it.

integration/spark/spark40/src/main/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogHandler.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /**
+   * Namespace scheme for the Unity Catalog naming convention {@code unitycatalog://{host}[:port]}.
+   */
+  private static final String UNITY_CATALOG_NAMESPACE_SCHEME = "unitycatalog";
```
arturowczarek, line 65: This is the namespace scheme I suggest. I've updated it in documentation

integration/spark/spark40/src/main/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogHandler.java
```diff
@@ -0,0 +1,181 @@
...
+      // Spark instantiates UCSingleCatalog from the same conf, so a functioning Unity Catalog
+      // without this key should not be possible.
+      log.error(
+          "No Unity Catalog server URI configured for catalog {}; returning the location-based "
+              + "dataset identifier without a Unity Catalog symlink",
+          tableCatalog.name());
+      return locationIdentifier;
+    }
+
+    return locationIdentifier.withSymlink(
```
arturowczarek, line 121: We produce physical location + table symlink

integration/spark/spark40/src/main/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogHandler.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+
+  @SneakyThrows
+  @Override
+  public Optional<String> getDatasetVersion(
+      TableCatalog tableCatalog, Identifier identifier, Map<String, String> properties) {
+    Table table = tableCatalog.loadTable(identifier);
+    if (table instanceof DeltaTableV2) {
+      return DeltaVersionUtils.getDatasetVersion(table);
```
arturowczarek, line 163: This is where I use the code determining dataset version that I extracted from Delta catalog handler

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogAbfssIntegrationTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Opt-in integration test that verifies {@link UnityCatalogHandler} produces an {@code abfss://}
+ * primary identifier for an Azure Data Lake Storage Gen2 backed table, with the Unity Catalog
+ * three-part name attached as a TABLE symlink.
+ *
+ * <p>Tagged {@code azure}: runs only with {@code -Pazure} against a real storage account.
+ */
+class UnityCatalogAbfssIntegrationTest extends AbfssUnityCatalogTestBase {
```
arturowczarek, line 24: Integration test that is checking the dataset identifier is produced when we have a table that is stored in Unity Catalog and we use `abfss` as the external table location. You have to activate Gradle profile `azure` to run it. You can use Taskfile task `abfss-test` to run it. It is going to complain with messages what environmental variables are missing. You can provide them in `.env` file (use `.env.template` for inspiration).

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogAbfssStreamingIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Opt-in integration test that verifies {@link UnityCatalogHandler} resolves an {@code abfss://}
+ * identifier (with a Unity Catalog TABLE symlink) for a Delta table populated by a <b>streaming</b>
+ * query.
+ *
+ * <p>Tagged {@code azure}: runs only with {@code -Pazure} against a real storage account.
+ */
+class UnityCatalogAbfssStreamingIntegrationTest extends AbfssUnityCatalogTestBase {
```
arturowczarek: In my case, it was important the handler works with streaming.

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogAbfssStreamingIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+  protected void provisionTable() throws Exception {
+    // Stream a bounded rate source into a Delta table at the abfss location, with an abfss
+    // checkpoint. AvailableNow processes the currently-available data and then stops.
+    Dataset<Row> stream =
+        spark.readStream().format("rate").option("rowsPerSecond", "1").load().select("value");
+    StreamingQuery query =
+        stream
+            .writeStream()
+            .format("delta")
+            .option("checkpointLocation", checkpointsManager.randomLocation())
```
arturowczarek: In streaming, it is important to keep the state of the stream in checkpoints. These are files stored somewhere. I've introduced checkpointsManager that can give us a safe, random location for storing them and is registered to automatically clean the files when we don't need them anymore (even when tests fail (JVM hook))

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogAbfssStreamingIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+            .trigger(Trigger.AvailableNow())
+            .start(tableLocation);
+    boolean terminated = query.awaitTermination(120_000);
```
arturowczarek: I used random timeout here. It shouldn't take that much of course.

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogAbfssStreamingIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+    // checkpoint. AvailableNow processes the currently-available data and then stops.
+    Dataset<Row> stream =
+        spark.readStream().format("rate").option("rowsPerSecond", "1").load().select("value");
+    StreamingQuery query =
+        stream
+            .writeStream()
+            .format("delta")
+            .option("checkpointLocation", checkpointsManager.randomLocation())
+            .trigger(Trigger.AvailableNow())
+            .start(tableLocation);
```
arturowczarek: The UnityCatalogTable tests typically use just single table. I pre-create a catalog, schema and prepare location where it will be stored. It is just for convenience. If someone wants to create additional table, there are also means to do it.

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/utils/AbfsStorageBackend.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * A {@link StorageBackend} rooted at {@code <containerUri>/<runId>} on Azure ADLS Gen2 (abfss). The
+ * run id keeps cleanup scoped to this run, never the container root. close() deletes the run root
+ * via the Hadoop {@link FileSystem} built from the given configuration, which must carry the {@code
+ * fs.azure.*} credentials (a snapshot of the SparkSession's hadoop configuration). The
+ * configuration is captured eagerly so close() also works from the shutdown hook, after the shared
+ * SparkContext may already be stopped.
+ */
+@Slf4j
+public final class AbfsStorageBackend implements StorageBackend {
```
arturowczarek, line 23: This is a backend that we use in ABFS tests. It is making sure the resources are removed from abfs after they are not needed anymore.

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/utils/AbfssUnityCatalogTestBase.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.junit.jupiter.api.Tag;
+
+/**
+ * Base for Unity Catalog table tests that run against Azure ADLS Gen2 (abfss). Provisions tables
+ * and checkpoints under the configured writable container ({@link
+ * AzureDynamicParameter#ContainerUri}) and registers the account credentials on the session. Tagged
+ * {@code azure} (inherited by subclasses), so these run only with {@code -Pazure}.
+ */
+@Tag("azure")
```
arturowczarek, line 21: The tests are switched of when we don't pass this profile. Someone must pass the environmental variables for it to work so it can't run automatically.

.pre-commit-config.yaml
```diff
@@ -93,7 +93,7 @@ repos:
         types: [text]
         args:
           - --ignore-words=spelling_wordlist.txt
-          - --skip=*.css,*.js,*.svg,*.xml,*/**/test/*,*/**/tests/*,*/yarn.lock,**/openapi/*.html,*/package-lock.json,*/spec/*
+          - --skip=*.css,*.js,*.svg,*.xml,*/**/test/*,*/**/tests/*,*/**/testScala*/*,*/yarn.lock,**/openapi/*.html,*/package-lock.json,*/spec/*
```
arturowczarek, line 96: The newly-added classes in spark40 testing Unity Catalog are scala 2.13 specific so they can't be put in `test` source set. They are put in `testScala213` source set and this new source sec was not covered by this exception. It should be because this check wrongly detects JUnit `@AfterAll` annotation as not following the conventino.

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/utils/ExternalTablesManager.java
```diff
@@ -0,0 +1,31 @@
...
+ * Computes external-table storage locations ({@code <catalog>/<schema>/<table>}) over a backend.
+ */
+@AllArgsConstructor
+public final class ExternalTablesManager {
+
+  StorageBackend backend;
+
+  /** {@code <root>/<catalog>/<schema>/<table>}; does not create anything. */
+  public String getLocation(String catalog, String schema, String table) {
+    validate(catalog, "catalog");
```
arturowczarek, line 20: I migrated the `ExternalTablesManager` from some other project and learned that it saved me a lot when I have these validators because nulls and empty values tend to sneak in during development.

integration/spark/spark40/src/testScala213/java/io/openlineage/spark40/agent/utils/ReservedPort.java
```diff
@@ -0,0 +1,55 @@
...
+import lombok.Getter;
+
+/**
+ * Represents a reserved TCP port that can be released and then used by a service.
+ *
+ * <p>Binding a {@link ServerSocket} first guarantees the port is free; releasing it just before the
+ * service binds minimises the race window that causes "address already in use" errors in tests.
+ */
+@AllArgsConstructor
+public class ReservedPort implements AutoCloseable {
```
arturowczarek, line 22: As in the description. Unity Catalog requires some specific port we bind to. This class is a trick that allows me to reserve a random port and almost atomically release it and use in Unity Catalog.

integration/spark/app/src/test/java/io/openlineage/spark/agent/EmrDynamicParameter.java
```diff
@@ -102,11 +99,6 @@ public enum EmrDynamicParameter implements DynamicParameter {
   EmrDynamicParameter(String parameterName, String defaultValue) {
     this.parameterName = parameterName;
     this.defaultValue = defaultValue;
-    this.prefix = "openlineage.test";
-  }
-
-  @Override
-  public Logger getLog() {
-    return log;
+    this.prefix = "openlineage.tests.emr";
```
arturowczarek, line 102: I've changed the prefix for properties to that it is consistent with databricks and abfss unity catalog

.github/workflows/integration-tests-spark-azure.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  workflow_dispatch:
+
+defaults:
+  run:
+    working-directory: integration/spark
+
+jobs:
+  run-integration-tests-azure:
```
arturowczarek, line 30: Example run: https://github.com/arturowczarek/OpenLineage/actions/runs/27822534760/job/82338851926

arturowczarek:
@mobuchowski @pawel-big-lebowski @tnazarew If you could kindly have a look :)

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
tnazarew: could we move all the hardcoded handlers to `DatasetBuilderFactory` so we have something like 
`return context.getDatasetBuilderFactory().getCatalogHandlers(context)`.stream().filter(...).collect(...)`?

for each version of the factory we have lists of the visitors and handlers in `get.+putBuilders` so why not do that for handlers?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
```diff
@@ -32,9 +35,22 @@ private static List<CatalogHandler> getHandlers(OpenLineageContext context) {
...
+    return builderFactory == null
```
tnazarew: is there a scenario (beside unit tests) where this happens?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
arturowczarek: Are you thinking about moving the handlers to module `spark3` and then everything that is Spark 3-based should use the same handlers? I was thinking about it but it is not that simple. In the OpenLineage context we have only one `DatasetBuilderFactory` (for specific Spark version). We would have to copy the handlers list for `DatasetBuilderFactory` for each Spark version or introduce a new abstract base that they would extend. It is a little bit of the code and I'm not sure we should make this change even bigger.

integration/spark/spark40/src/main/java/io/openlineage/spark40/agent/lifecycle/plan/catalog/UnityCatalogHandler.java
```diff
@@ -0,0 +1,181 @@
...
+          "Unity Catalog table {} has unsupported type {}; expected a Delta or V1 table",
+          identifier,
+          table.getClass().getName());
+      throw new UnsupportedCatalogException(
+          "Unsupported Unity Catalog table type: " + table.getClass().getName());
+    }
+
+    // Physical storage location (e.g. abfss://..., s3://..., file:/...). FilesystemDatasetUtils
+    // resolves the proper object-storage namespace (Azure ADLS Gen2 abfss/abfs included).
+    URI locationUri =
```
tnazarew: we have private method in PathUtils `getLocationUri` which does basically that, maybe make it public and use it here?

arturowczarek: Done

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
arturowczarek: I don't thinks so. I've replaced it with an `EMPTY` DatasetBuilderFactory singleton as default for unit tests.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/CatalogUtils3.java
```diff
@@ -25,13 +27,17 @@ public class CatalogUtils3 {
...
   private static List<CatalogHandler> getHandlers(OpenLineageContext context) {
     List<CatalogHandler> handlers =
-        Arrays.asList(
-            new IcebergHandler(context),
-            new DeltaHandler(context),
-            new DatabricksDeltaHandler(context),
-            new DatabricksUnityV2Handler(context),
-            new JdbcHandler(context),
-            new V2SessionCatalogHandler(context));
+        new ArrayList<>(
```
arturowczarek: We have to wrap the list with ArrayList, because we mutate it with addAll. Without it the list created by `Arrays.asList` complains it is not modifiable

pawel-big-lebowski:
Hey, great work on this! The UC OSS handler itself is clean and well-tested. One thought before this lands — it looks like the PR is doing quite a few things at once beyond the core feature:
  - CatalogHandler getting moved from spark3 to shared (needed as a prerequisite, but a lot of import churn on its own)
  - abfs/wasb → abfss/wasbs normalization in ObjectStorageDatasetExtractor (useful independently)
  - DeltaVersionUtils extracted from DeltaHandler (nice cleanup, unrelated to UC OSS)
  - DynamicParameter refactored and moved to test-commons, with the EMR property prefix renamed from openlineage.test → openlineage.tests.emr (this one's a breaking change for external users — probably
   worth its own changelog entry at least)
  - IcebergHandler.hasClasses() quietly changed to probe SparkCatalog instead of Catalog — this changes behavior in envs where iceberg-core is on the classpath but iceberg-spark isn't, and it kind of
  snuck in here
  - A pre-existing bug fix in InputFieldsCollectorTest ("/path" → "path")
 
If this understanding is correct, perhaps we could separate this into multiple PRs.

mobuchowski:
Awesome to see this effort being resurrected. Will try to find some time to review soon

arturowczarek:
@pawel-big-lebowski I've extracted the following pull requests. They are all independent but are required for current pull request to succeed.
- https://github.com/OpenLineage/OpenLineage/pull/4661
- https://github.com/OpenLineage/OpenLineage/pull/4662
- https://github.com/OpenLineage/OpenLineage/pull/4663
- https://github.com/OpenLineage/OpenLineage/pull/4664
- https://github.com/OpenLineage/OpenLineage/pull/4665

mobuchowski:
>Will try to find some time to review soon

Scratch that, @pawel-big-lebowski and @tnazarew are already on it, I won't push to cause mess :)

.github/workflows/integration-tests-spark-azure.yml
```diff
@@ -0,0 +1,55 @@
...
+          OPENLINEAGE_TESTS_AZURE_CLIENT_ID: ${{ secrets.AZURE_TESTS_CLIENT_ID }}
+          OPENLINEAGE_TESTS_AZURE_CLIENT_SECRET: ${{ secrets.AZURE_TESTS_CLIENT_SECRET }}
+          OPENLINEAGE_TESTS_AZURE_TENANT_ID: ${{ secrets.AZURE_TESTS_TENANT_ID }}
+        run: task abfss-test -- -Pjava.compile.home=${JAVA_HOME}
+      - name: Upload test report
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: test-report
+          path: integration/spark/spark40/build/reports/tests
```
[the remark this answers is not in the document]
arturowczarek: Done

tnazarew (approved):

PR #3685, opened 2025-05-07
Add Amazon DataZone transport

Shinabel:
### Problem

DataZone and SageMaker Unified Studio(SMUS) currently accepts OpenLineage events via the [PostLineageEvent API](https://docs.aws.amazon.com/datazone/latest/APIReference/API_PostLineageEvent.html). However, the existing OpenLineage framework does not have a transport that can directly send OpenLineage events to Amazon DataZone / SageMaker Unified Studio.

### Solution
Adding DataZone transport where OpenLineage events can be sent directly to DataZone / SageMaker Unified Studio.

The Java transport has been tested in the AWS Glue job / EMR / local pyspark. This was done by adding the following spark config. The lineage events were successfully emitted to Amazon DataZone / SMUS.
```
--conf spark.extraListeners=io.openlineage.spark.agent.OpenLineageSparkListener
--conf spark.openlineage.transport.type=amazon_datazone_api
--conf spark.openlineage.transport.domainId=dzd_domain_id
```

The Python transport has been tested in the Amazon Managed Workflows for Apache Airflow (MWAA) environment. This was done by loading the updated openlineage-python.whl package as a [custom dependency](https://docs.aws.amazon.com/mwaa/latest/userguide/configuring-dag-import-plugins.html). The lineage events were successfully emitted to Amazon DataZone / SMUS.
```
transport:
  type: amazon_datazone_api
  domainId: dzd-domain-id
``` 

#### One-line summary:

Add Amazon DataZone transport

### Checklist

----

### DIFF
--- a/client/java/transports-datazone/build.gradle
+++ b/client/java/transports-datazone/build.gradle
@@ -1,0 +1,38 @@
+/*
+/* Copyright 2018-2025 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+plugins {
+    id 'eclipse'
+    id 'jacoco'
+    id 'java'
+    id 'java-library'
+    id 'maven-publish'
+    id 'signing'
+    id 'com.adarshr.test-logger' version '3.2.0'
+    id 'com.diffplug.spotless' version '7.0.2'
+    id 'com.gradleup.shadow' version '8.3.6'
+    id 'pmd'
+    id 'io.freefair.lombok' version '8.13.1'
+}
+
+ext {
+    projectDescription = 'DataZone OpenLineage transport library'
+    awsSdkVersion = '2.31.15'
+}
+
+dependencies {
+    compileOnly('com.google.code.findbugs:jsr305:3.0.2')
+    implementation("software.amazon.awssdk:auth:${awsSdkVersion}")
+    implementation("software.amazon.awssdk:datazone:${awsSdkVersion}")
+    implementation("software.amazon.awssdk:apache-client:${awsSdkVersion}")
+}
+
+shadowJar {
+    relocate 'software.amazon', 'io.openlineage.client.shaded.software.amazon'
+    relocate 'org.apache', 'io.openlineage.client.shaded.org.apache'
+    exclude 'org.slf4j'
+}
+
+apply from: '../transports.build.gradle'
--- a/client/python/openlineage/client/transport/__init__.py
+++ b/client/python/openlineage/client/transport/__init__.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
 
+from openlineage.client.transport.amazon_datazone import AmazonDataZoneConfig, AmazonDataZoneTransport
 from openlineage.client.transport.composite import CompositeTransport
 from openlineage.client.transport.console import ConsoleTransport
 from openlineage.client.transport.factory import DefaultTransportFactory
@@ -25,6 +26,7 @@ _factory.register_transport(ConsoleTransport.kind, ConsoleTransport)
 _factory.register_transport(NoopTransport.kind, NoopTransport)
 _factory.register_transport(FileTransport.kind, FileTransport)
 _factory.register_transport(TransformTransport.kind, TransformTransport)
+_factory.register_transport(AmazonDataZoneTransport.kind, AmazonDataZoneTransport)
 
 
 def get_default_factory() -> DefaultTransportFactory:
@@ -44,6 +46,8 @@ __all__ = [
     "Config",
     "Transport",
     "TransportFactory",
+    "AmazonDataZoneConfig",
+    "AmazonDataZoneTransport",
     "CompositeTransport",
     "ConsoleTransport",
     "FileTransport",
--- a/client/python/openlineage/client/transport/amazon_datazone.py
+++ b/client/python/openlineage/client/transport/amazon_datazone.py
@@ -1,0 +1,77 @@
+# Copyright 2018-2025 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+from openlineage.client import event_v2
+from openlineage.client.run import RunEvent
+from openlineage.client.serde import Serde
+from openlineage.client.transport.transport import Config, Transport
+
+if TYPE_CHECKING:
+    from openlineage.client.client import Event
+
+log = logging.getLogger(__name__)
+
+
+@dataclass
+class AmazonDataZoneConfig(Config):
+    domain_id: str
+    endpoint_override: str | None = None
+
+    @classmethod
+    def from_dict(cls, params: dict[str, Any]) -> AmazonDataZoneConfig:
+        if "domainId" not in params:
+            msg = "`domainId` key not passed to AmazonDataZoneConfig"
+            raise RuntimeError(msg)
+        return cls(domain_id=params["domainId"], endpoint_override=params.get("endpointOverride"))
+
+
+class AmazonDataZoneTransport(Transport):
+    kind = "amazon_datazone_api"
+    config_class = AmazonDataZoneConfig
+
+    def __init__(self, config: AmazonDataZoneConfig) -> None:
+        self.config = config
+        self.datazone = None
+
+        self._setup_datazone(self.config.endpoint_override)
+        log.debug(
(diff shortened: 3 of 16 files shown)


client/python/openlineage/client/transport/__init__.py
```diff
@@ -43,6 +45,8 @@ def register_transport(clazz: type[Transport]) -> type[Transport]:
     "MSKIAMTransport",
     "MSKIAMConfig",
     "ConsoleTransport",
+    "AmazonDataZoneTransport"
```
Shinabel, line 48: Not sure how `,` got dropped. Will update

client/java/transports-datazone/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+    id 'io.freefair.lombok' version '8.13.1'
+}
+
+ext {
+    projectDescription = 'DataZone OpenLineage transport library'
+}
+
+dependencies {
+    compileOnly('com.google.code.findbugs:jsr305:3.0.2')
+    implementation('software.amazon.awssdk:auth:2.31.15')
```
pawel-big-lebowski, line 26: looks like `awssdk` version can be moved to a variable.

client/java/transports-datazone/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+    implementation('software.amazon.awssdk:datazone:2.31.15')
+    implementation('software.amazon.awssdk:apache-client:2.31.15')
+}
+
+shadowJar {
+    relocate 'software.amazon', 'io.openlineage.client.shaded.software.amazon'
+    relocate 'org.apache', 'io.openlineage.client.shaded.org.apache'
+    relocate 'org.slf4j', 'io.openlineage.client.shaded.org.slf4j'
```
pawel-big-lebowski, line 34: Is it necessary to relocate? Can we exclude it from trainsitive dependency?

pawel-big-lebowski (commented):
Java code looks good to me, thanks for bringing this 💯 

Two minor comments added. 
Update to the docs can be done within the same PR, which makes sense in this case. 

I didn't verify python code.

client/python/openlineage/client/transport/amazon_datazone.py
```diff
@@ -0,0 +1,70 @@
...
+                response["id"], response["domainId"])
+        except Exception as error: # noqa: BLE001
+            msg = f"Failed to send lineage event to DataZone Domain {self.config.domain_id}: {event}"
+            raise RuntimeError(msg) from error
+
+    def _setup_datazone(self, endpoint_url: str | None = None) -> None:
+        try:
+            import boto3
+
+            self.datazone = boto3.client(
```
kacpermuda, line 60: Maybe we can initialize `self.datazone` or even make it private `self._datazone` as None in init? Not sure why mypy is not catching that it's defined outside of init.

client/python/openlineage/client/transport/amazon_datazone.py
```diff
@@ -0,0 +1,70 @@
...
+                "datazone",
+                endpoint_url=endpoint_url
+            )
+        except ModuleNotFoundError:
```
kacpermuda, line 64: I like this check ! Should we also catch ImportError? Not sure if it's possible to be thrown here.

client/python/openlineage/client/transport/amazon_datazone.py
```diff
@@ -0,0 +1,70 @@
...
+    config_class = AmazonDataZoneConfig
+
+    def __init__(self, config: AmazonDataZoneConfig) -> None:
+        self.config = config
+        self._setup_datazone(self.config.endpoint_override)
+
+    def emit(self, event: Event) -> None:
+        if not isinstance(event, (RunEvent, event_v2.RunEvent)):
+            # DataZone only supports RunEvent
+            log.debug("DataZone only supports RunEvent")
```
kacpermuda, line 44: I think we can even make it a warning, since we are not emitting the event. That's pretty important thing to log.

client/python/openlineage/client/transport/amazon_datazone.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class AmazonDataZoneTransport(Transport):
+    kind = "amazon_datazone_api"
+    config_class = AmazonDataZoneConfig
+
+    def __init__(self, config: AmazonDataZoneConfig) -> None:
+        self.config = config
+        self._setup_datazone(self.config.endpoint_override)
+
```
kacpermuda, line 40:
```suggestion
log.debug(
            "Constructing OpenLineage transport that will send events "
            "to Amazon Datazone domain `%s`.",
            config.domain_id,
        )
```
Can we add something like this? It's present in all other transports, and can help with debugging.

client/python/openlineage/client/transport/amazon_datazone.py
```diff
@@ -0,0 +1,70 @@
...
+    @classmethod
+    def from_dict(cls, params: dict[str, Any]) -> AmazonDataZoneConfig:
+        if "domainId" not in params:
+            msg = "`domainId` key not passed to AmazonDataZoneConfig"
+            raise RuntimeError(msg)
+        return cls(domain_id=params["domainId"], endpoint_override=params.get("endpointOverride", None))
+
+
+class AmazonDataZoneTransport(Transport):
+    kind = "amazon_datazone_api"
```
kacpermuda, line 34: I am not familiar with the DataZone, but is the `_api` necessary here? Is there some other way to use it? Do we expect f.e. `amazon_datazone_notapi` transport?

client/python/tests/test_amazon_datazone.py
```diff
@@ -0,0 +1,104 @@
...
+@pytest.fixture()
+def job_event() -> JobEvent:
+    return JobEvent(
+        eventTime="2024-08-20T11:08:01.123456",
+        job=Job(namespace="test-namespace", name="test-job"),
+        producer="prod",
+        schemaURL="schema",
+    )
+
+def test_datazone_config_raises_when_missing_domain_id() -> None:
```
kacpermuda, line 59: Could you also add tests for missing boto3 library and error while emitting the event? Just to test if the error is not being silenced.

kacpermuda (commented):
Just some minor comments to python code, in general it looks good !

client/java/transports-datazone/build.gradle
Shinabel: Different version of AWS Products (i.e. Glue spark, EMR spark) use different aws sdk version underneath which may not contain `PostLineageEvent` we are looking for. It's safer to shade `software.amazon` imo - I will double check if we need to shade `org.apache` and `org.slf4j` as we are using it minimally

client/python/openlineage/client/transport/amazon_datazone.py
Shinabel: An ImportError could happen if boto3 is found but can't be properly loaded; permission issues or a corrupted installation. While this scenario is relatively rare for a stable package like boto3, we can add it

client/python/openlineage/client/transport/amazon_datazone.py
Shinabel: Good question! We are expecting to create another transports for DataZone in future (i.e. dispatching directly to some streaming endpoint DZ has as opposed to going through PostLineageEvent api) hence added `_api` as a distinction

client/java/transports-datazone/build.gradle
pawel-big-lebowski: Shading 'software.amazon' is fine. My question relates mostly to `org.slf4j` which perhaps can be excluded like we do in other places:
https://github.com/OpenLineage/OpenLineage/blob/2cfa49654a301893da11195299a61b23ad3326b2/integration/spark/build.gradle#L177

Shinabel: This comes with the assumption that the consumer will have this dependency (org.slf4j) already in their application - However, this could be problematic for applications using Log4j 2 directly (without SLF4J) or Apache Commons Logging, as they might experience conflicting initialization behavior.

For Flink and Spark specifically (which supports OL java integration), since they already include org.slf4j in their dependencies, we could potentially exclude it from our shadow JAR. Have you seen any dependency conflicts related to org.slf4j from OL across different version of Spark or Flink (if so, shading may be a safer option)?

kacpermuda (commented):
Python part looks good, but it's missing the documentation. @Shinabel could you add a similar doc as you did for java client? Thanks !

pawel-big-lebowski (approved):

kacpermuda (approved):

kacpermuda:
Thanks @Shinabel for the contribution, looking forward to the next ones ! 🔥

PR #3689, opened 2025-05-09
Update debug facet information

pawel-big-lebowski:


website/docs/integrations/spark/configuration/spark_conf.md
```diff
@@ -27,7 +27,6 @@ The following parameters can be specified:
...
 | spark.openlineage.run.tags                                | List of run-level tags. Tags are passed in a string, with key:value information separated by colon `:`, and tags being separated by semicolon `;`                                                                                                                                                                                                                      | "key:value;label;another:tag"                  |
-| spark.openlineage.debugFacet                              | Determines whether debug facet shall be generated and included within the event. Set `enabled` to turn it on. By default, facet is disabled.                                                                                                                                                                                                                           | enabled                                       |
```
kacpermuda, line 30: Should we remove it from the list of all possible configurations? I understand that it's not a production use case, but shouldn't it stay here just to say "this is also a permitted option"?

mobuchowski (approved):

website/docs/integrations/spark/configuration/spark_conf.md
kacpermuda: Ah, I see, this was deprecated and is now configured separately. Sorry.

mobuchowski:
Few other places to fix: 
https://github.com/OpenLineage/OpenLineage/blob/e04800f928bc56472f6b35ca7c471f533d4498bd/integration/spark/cli/spark-conf.yml#L8
https://github.com/OpenLineage/OpenLineage/blob/a31cc8dc9702d0798e21f834c4f32d7ecc2819f8/integration/spark/cli/spark-conf-docker.yml#L11

website/docs/integrations/spark/configuration/spark_conf.md
mobuchowski: AFAIK Paweł's idea is to use `spark.openlineage.facets.debug.disabled=false`

pawel-big-lebowski: The problem is that this option is not working. It used to work like this for some time, and now it's working with `spark.openlineage.facets.debug.disabled=false`.

pawel-big-lebowski:
@mobuchowski Can I prepare a separate PR to remove the setting from Spark integration as this ain't working.

mobuchowski:
Sure.

pawel-big-lebowski:
https://github.com/OpenLineage/OpenLineage/pull/3690 follow-up

PR #3695, opened 2025-05-12
spark: Implement CatalogDatasetFacet for CatalogHandler

mobuchowski:
This PR implements adding CatalogDatasetFacet for Spark `CatalogHandler` and few other places that are using `TableCatalog`:  `DropTableVisitor` and `CreateReplaceDatasetBuilder`.

### DIFF
--- a/integration/spark/buildDependencies.sh
+++ b/integration/spark/buildDependencies.sh
@@ -3,6 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 set -x -e
-(cd ../../client/java && ./gradlew clean publishToMavenLocal -x test)
+(cd ../../client/java && ./gradlew generateCode publishToMavenLocal -x test --rerun-tasks)
 (cd ../spark-extension-interfaces && ./gradlew clean publishToMavenLocal -x test)
 (cd ../sql/iface-java && ./script/compile.sh && ./script/build.sh)
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
@@ -32,7 +32,16 @@ public class AwsUtils {
                   getGlueCatalogId(sparkConf, hadoopConf)
                       .map(glueCatalogId -> "arn:aws:glue:" + region + ":" + glueCatalogId));
     } else {
-      return Optional.empty();
+      try {
+        return awsRegion()
+            .flatMap(
+                region ->
+                    getGlueCatalogId(sparkConf, hadoopConf)
+                        .map(glueCatalogId -> "arn:aws:glue:" + region + ":" + glueCatalogId));
+      } catch (Exception e) {
+        log.error("Failed to retrieve Glue", e);
+        return Optional.empty();
+      }
     }
   }
 
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/JdbcHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/JdbcHandler.java
@@ -5,6 +5,7 @@
 
 package io.openlineage.spark3.agent.lifecycle.plan.catalog;
 
+import io.openlineage.client.OpenLineage;
 import io.openlineage.client.dataset.namespace.resolver.DatasetNamespaceCombinedResolver;
 import io.openlineage.client.utils.DatasetIdentifier;
 import io.openlineage.client.utils.jdbc.JdbcDatasetUtils;
@@ -12,9 +13,10 @@ import io.openlineage.spark.api.OpenLineageContext;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.reflect.FieldUtils;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.connector.catalog.Identifier;
@@ -22,10 +24,13 @@ import org.apache.spark.sql.connector.catalog.TableCatalog;
 import org.apache.spark.sql.execution.datasources.jdbc.JDBCOptions;
 import org.apache.spark.sql.execution.datasources.v2.jdbc.JDBCTableCatalog;
 
+@Slf4j
 public class JdbcHandler implements CatalogHandler {
+  private final OpenLineageContext context;
   private final DatasetNamespaceCombinedResolver namespaceResolver;
 
   public JdbcHandler(OpenLineageContext context) {
+    this.context = context;
     namespaceResolver = new DatasetNamespaceCombinedResolver(context.getOpenLineageConfig());
   }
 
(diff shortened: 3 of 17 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/JdbcHandler.java
```diff
@@ -39,23 +44,57 @@ public boolean isClass(TableCatalog tableCatalog) {
...
             .collect(Collectors.toList());
 
     return namespaceResolver.resolve(
         JdbcDatasetUtils.getDatasetIdentifier(
-            jdbcOptions.url(), parts, jdbcOptions.asConnectionProperties()));
+            options.url(), parts, options.asConnectionProperties()));
+  }
+
+  @Override
+  public Optional<OpenLineage.CatalogDatasetFacet> getCatalogDatasetFacet(
```
dolfinus, line 71: Does it make sense to add this facet for JDBCRelation? There is no catalog registered in Spark session

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/JdbcHandler.java
```diff
@@ -39,23 +44,57 @@ public boolean isClass(TableCatalog tableCatalog) {
...
+      TableCatalog tableCatalog, Map<String, String> properties) {
+    Optional<JDBCOptions> options = getJdbcOptions(tableCatalog);
+
+    OpenLineage.CatalogDatasetFacetBuilder builder =
+        context
+            .getOpenLineage()
+            .newCatalogDatasetFacetBuilder()
+            .name(tableCatalog.name())
+            .framework("jdbc")
+            .type("jdbc") // TODO(mobuchowski) handle different databases here? Based on dialect
```
dolfinus, line 81: IMHO, there is no need in this, metastore URI already has all required information

mobuchowski: Removed.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
```diff
@@ -26,13 +26,24 @@ public class AwsUtils {
   @SneakyThrows
   public static Optional<String> getGlueArn(SparkConf sparkConf, Configuration hadoopConf) {
     if (isHiveUsingGlue(sparkConf, hadoopConf)) {
+      log.info("Hive using Glue");
```
pawel-big-lebowski, line 29: Isn't this a debug?

integration/spark/buildDependencies.sh
```diff
@@ -3,6 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 set -x -e
-(cd ../../client/java && ./gradlew clean publishToMavenLocal -x test)
+(cd ../../client/java && ./gradlew clean generateCode publishToMavenLocal -x test --rerun-tasks)
```
pawel-big-lebowski, line 6: why do you want to use this together with `clean` ?

mobuchowski: You risk a situation when you add a facet, but when using local maven repository you don't see it. The previous line worked if there was regular code change, but not when `OpenLineage.java` was changed.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
mobuchowski: Removed.

pawel-big-lebowski (approved):

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
```diff
@@ -32,7 +32,16 @@ public static Optional<String> getGlueArn(SparkConf sparkConf, Configuration had
...
+      try {
+        return awsRegion()
+            .flatMap(
+                region ->
+                    getGlueCatalogId(sparkConf, hadoopConf)
+                        .map(glueCatalogId -> "arn:aws:glue:" + region + ":" + glueCatalogId));
+      } catch (Exception e) {
+        log.error("Failed to retrieve Glue", e);
+        return Optional.empty();
+      }
```
Jorricks, line 44: Is it on purpose that we connect amazon s3 even if we don't have s3 enabled?
I'm trying to setup OpenLineage with S3 Transport (over the Datahub acryl-spark-lineage package which we used earlier), but my log is getting filled with this error specifically.
I tried setting the `hive.metastore.client.factory.class` but because of this change, even when I set it to different values, it's still trying to connect to Amazon s3, which we are only using for the transport of the events. Our Hive is a local Hive metastore with HDFS.

Note: our production environment is heavily firewalled environment so we will get a bunch of timeouts and weird firewall drops. Locally I'm getting a bunch of 403's with invalid token as it detects the AWS credentials I am trying to use for the s3 transport.

pawel-big-lebowski: How do you know this function connects anything? It looks it just browses Spark config entries. 
How come your flow enters this code? You mentioned you're using `acryl-spark-lineage` (https://central.sonatype.com/artifact/io.acryl/acryl-spark-lineage/versions) which is 6 months old while the PR you comments is few weeks old.

Jorricks: Because I attached a debugger locally 😁

Furthermore, what I tried to mention but maybe wasn't clear enough, the datahub acryl-spark-lineage package isn't working for us (some problems with shading), hence I'm trying this package. 

To be more explicit: I got hit with the new compatibility issue that was introduced in AWSSDK 2.30. Here they introduced request & response checksums, which is incompatible with Ceph S3 (which we use). 

To debug, I attached a debugger and then got super confused when it tried to connect to Amazon S3 even though I set an endpoint overwrite using the open lineage settings. Then I found out there were two different parts connecting to s3:
- this flow trying to get bucket information
- the transport S3 package that actually tries to upload info

pawel-big-lebowski: Are you able to find out which line or method of this class resulted in the requests that caused problem on your side? 

From what I see in the code, there is:
```
} else {
      log.debug("Fetching current account ID to use as the catalog ID.");
      return Optional.of(AwsAccountIdFetcher.getAccountId());
    }
```
which does a call to get `accountId` and perhaps OL integration could do better in persisting account Id after the first call. 

Please mind that setting `spark.glue.accountId` could be a workaround to avoid this call.

Jorricks: From what I remember, it's indeed the fetching of this ID. Then because I'm setting up authentication in a generic way, this call also uses those credentials and hence the 403.

If I set the value you mentioned, would it avoid doing any api calls afterwards as well?

pawel-big-lebowski: The class `software.amazon.awssdk.services.sts.StsClient` is used only within `AwsAccountIdFetcher`. Method `io.openlineage.spark.agent.util.AwsAccountIdFetcher#getAccountId` is called only when no `hive.metastore.glue.catalogid` nor `spark.glue.accountId` provided. So, this should help. 

Alternative approach would be to extend `io.openlineage.spark.agent.util.AwsAccountIdFetcher` to configure authentication.

Jorricks: I can confirm the `hive.metastore.glue.catalogid` indeed worked. Thanks for pointing that out!

Non-the-less, I don't think this should be the default as new users will get confused by the error logs (with a non-obvious fix). In my case I had two issues, given they were both showing up, it was more difficult for me to figure out what was happening exactly. Explicitely needing to enable Glue lookups is probably an even better solution then creating an option to disable it. 

The issues were;
- The above mentioned glue issue
- I was also facing issues with custom S3 which is mentioned here; https://github.com/aws/aws-cli/issues/9214#issue-2791559399. Setting the environment variables `AWS_REQUEST_CHECKSUM_CALCULATION=when_required` and `AWS_RESPONSE_CHECKSUM_VALIDATION=when_required` fixed those.

Jorricks: To whomever finds this and needs to submit the catalogid to a Python spark job, you should prefix it;
https://stackoverflow.com/a/58634843

PR #3696, opened 2025-05-13
DBT: Generate new runId for each start event

dolfinus:
### Problem

In Python, default arguments like:
```python
def func(arg = "something"):
  ...
```

are effectively static values. Using some function result as a default value, like:
```python
def func(arg = generate_default_value()):
  ...
```
Will call `generate_default_value()` only once, while .py file is parsed and evaluated, not at each call of `func()`.

This pattern should be replaced with a guard like:
```python
def func(arg = None):
  arg = arg or generate_default_value()
  ...
```

### Solution

#### One-line summary:

DBT: Generate new runId for each start event

### Checklist

----

### DIFF
--- a/integration/dagster/tests/conftest.py
+++ b/integration/dagster/tests/conftest.py
@@ -164,8 +164,8 @@ class DagsterRunLatestProvider(DagsterRunProvider):
 def make_test_event_log_record(
     event_type: Optional[DagsterEventType] = None,
     pipeline_name: str = "a_job",
-    pipeline_run_id: str = str(generate_new_uuid()),
-    timestamp: float = time.time(),
+    pipeline_run_id: Optional[str] = None,
+    timestamp: Optional[float] = None,
     step_key: Optional[str] = None,
     storage_id: int = 1,
 ):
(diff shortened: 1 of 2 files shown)


kacpermuda:
I'm assuming this is a bug and not a desired behavior for keeping the run_id the same @mobuchowski ?

kacpermuda:
Maybe we should put it as some constant variable to make it clearer that it's supposed to be generated only once?

dolfinus:
> it's supposed to be generated only once?

Don't get it. It should be generated once per dbt project run.

kacpermuda:
I'm not sure tho, but this is the current behavior right? I'm just assuming it's not faulty, don't have much experience with this dbt integration.

dolfinus:
> I'm not sure tho, but this is the current behavior right?

Currently dbt-ol integration is a script which runs dbt as a child process, generates OL events for it, and then stops. So runId is always new for each script run.

But from the source code perspective (e.g. calling this integration programmatically) runId should be generated by each `dbt_run_event(...)` call, and not during import.

mobuchowski (approved):
yes, this is a valid fix. No bugs were surfaced to the users because we never actually called the thing twice - other callers than the first one passed the run id explicitely.

Issue #3711, opened 2025-05-20
[FEATURE][SPARK] Smart debug facet


pawel-big-lebowski:
### Motivation

Currently OpenLineage gives very little guarantees on the completeness of lineage graph detected. Although exact guarantees may be hard to achieve, there are few improvements that could help orgs in measuring the coverage and constantly increasing it. One of the improvements proposed is to automatically turn on debug facet.

Debug facet may be extremely useful in case of a series of OpenLineage events that did not capture any input or output datasets. It does not make much sense to turn debug facet on by default for all the jobs as this is not the intended way for debugging to work. However, in case of any problem symptoms detected, it's great to have debug facet information without re-running the job.

The proposed feature would be optional and configured with:
 * `openlineage.debug.smart=enabled`
 * `openlineage.debug.smart.mode=any-missing|output-missing`
 
`OpenLineage` context would store additional boolean flags determining if inputs/outputs were emitted. In case of a terminal event, and if required datasets are missing, debug facet will be turned on and included in the event. 

Additionally, debug facet would contain additional field, determining the reason why it was triggered. 

The benefits of the approach are:
 * Orgs can measure lineage completeness based on the ratio of terminal events with debug facet triggered bcz of empty datasets and the amount of all terminal events. Note: this does not consider lineage events that were not emitted bcz of other issues.
 * Investigating such events can be helpful to increase the coverage. First, one can add extra filters for commands that should be filtered. Secondly, debug facet gives an information about classes used in logical tree. 

### Do you plan to make this contribution yourself?

- [x] I am interested in doing this work

pawel-big-lebowski:
Hey @mobuchowski, thanks for sharing your concerns and this feature should be opt-in. You're right that in most use cases logs should do the job. However, imagine a scenario of an external OL consumer like a data catalog. In this case a consumer does not have the access to producers logs out-of-the-box. So, the more observability information is sent within OL event, the better. 

I don't agree with memory concerns in this case. Storing two boolean values (was input emitted/was output emitted) within OpenLineageContext should not be a problem. BTW we already store information if a `START` event was emitted.

pawel-big-lebowski:
Yeah, debug works only when the event gets emitted. 
For the lost events, I have another issue in mind - smart timeouts ;) which I plan to create and work next week. 
[...]

PR #3715, opened 2025-05-23
[SPARK] Smart debug facet

pawel-big-lebowski:
[https://github.com/OpenLineage/OpenLineage/issues/3711] Add the ability to automatically include debug facet, when no inputs and outputs datasets detected. This can be useful to assure lineage completeness. Please read the docs within the PR to get more details on the feature. 

Changes implemented: 
 * extend `OpenLineageContext` with `OpenLineageRunStatus` to store information whether any inputs/outputs have been collected through all the events for the given run. 
 * turn on `DebugFacet` on `COMPLETE` event when the configurable criteria are met: output dataset missing or any of the input/output is missing.
 * add `logs` field to debug facet determining if inputs or outputs are empty.
 * make sure `DebugFacet` json does not exceed 100KB, or other configurable value. 
 * In case the value is exceeded, replece debug facet with an empty facet and a single log message attached telling a debug facet size exceeded the limit. 
 * Add configuration entry to turn off metrics sent via debug facet. Metrics can be huge, bcz of metrics tags passed in JSON. 
 * Remove legacy `spark.openlineage.debugFacet` setting to turn on/off the debug facet.

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/DebugRunFacet.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/DebugRunFacet.java
@@ -6,9 +6,11 @@
 package io.openlineage.spark.agent.facets;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import io.micrometer.core.instrument.Tag;
 import io.openlineage.client.OpenLineage;
 import io.openlineage.spark.agent.Versions;
+import io.openlineage.spark.agent.facets.serializer.DebugRunFacetSerializer;
 import io.openlineage.spark.api.SparkOpenLineageConfig;
 import java.util.List;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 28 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/api/DebugConfig.java
```diff
@@ -0,0 +1,67 @@
...
+@Setter
+@Getter
+@Slf4j
+public class DebugConfig {
+
+  public static String ANY_MISSING = "any-missing";
+  public static String OUTPUT_MISSING = "output-missing";
+
+  @JsonProperty("smart")
+  String smartDebugEnabled;
```
mobuchowski: why not boolean?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/serializer/DebugRunFacetSerializer.java
```diff
@@ -0,0 +1,97 @@
...
+    int payloadSize =
+        OpenLineageClientUtils.toJson(new DebugRunFacetWithStandardSerializer(facet))
+                .getBytes(StandardCharsets.UTF_8)
+                .length
+            / 1024;
+
+    if (payloadSize > facet.getPayloadSizeLimitInKilobytes()) {
+      log.warn(
+          "DebugRunFacetSerializer skipping serialization of DebugRunFacet due to payload size: {} kilobytes",
+          payloadSize);
```
mobuchowski, line 50: If you serialize then check, you might already blow up the memory consumption by holding the JSON

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/DebugRunFacet.java
```diff
@@ -22,25 +24,32 @@
...
  */
 @Getter
+@JsonSerialize(using = DebugRunFacetSerializer.class)
 public class DebugRunFacet extends OpenLineage.DefaultRunFacet {
   private final ClasspathDebugFacet classpath;
   private final SystemDebugFacet system;
   private final SparkConfigDebugFacet config;
   private final LogicalPlanDebugFacet logicalPlan;
   private final MetricsDebugFacet metrics;
+  private final List<String> logs;
```
mobuchowski, line 34: I don't understand how you fill this field - should this not be very much dependent of logger config and Spark distribution?

pawel-big-lebowski: This will not contain slf4j logs. It can just contain some extra information about debug facet like: 
 - debug facet serialization size exceeds certain size
 - no inputs or no output dataset have been detected.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/serializer/DebugRunFacetSerializer.java
pawel-big-lebowski: Debug facet does not contain Spark's LogicalPlan classes and serialises only OpenLineage classes explicitly defined. Logical plan's nodes are contained, but by using:
```
String.format("%s@%s", node.nodeName(), node.hashCode())
```

So, this is not aiming to protect against gigabyte size logical plan nodes serialization, but as a limiter to make sure event size does not exceed backend boundaries.

integration/spark/shared/src/main/java/io/openlineage/spark/api/DebugConfig.java
pawel-big-lebowski: sure, let's make it boolean then.

PR #3722, opened 2025-05-28
Filter temp inner jobs for bigquery indirect mode

pawel-big-lebowski:
Draft PR to solve this: https://github.com/OpenLineage/OpenLineage/discussions/3702

I think this is the best way to filter temp jobs events for indirect big query write method. 
It works, the code is clean, but it's kind of hacky. To be fair, similar hacks are already done for `DeltaEventFilter`


integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/BigQueryIntermediateJobFilter.java
```diff
@@ -0,0 +1,61 @@
...
+    if (!logicalPlan.isPresent()) {
+      return false;
+    }
+
+    if (!(logicalPlan.get() instanceof InsertIntoHadoopFsRelationCommand)) {
+      return false;
+    }
+
+    InsertIntoHadoopFsRelationCommand command =
+        (InsertIntoHadoopFsRelationCommand) logicalPlan.get();
```
ddebowczyk92: using isPresent()->get() is considered as an anti-pattern, is it possible to use map().orElseGet()?

ddebowczyk92: the whole logic operating on InsertIntoHadoopFsRelationCommand could be in separate method

ddebowczyk92 (changes requested):

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/BigQueryIntermediateJobFilter.java
pawel-big-lebowski: applied changes proposed

ddebowczyk92 (approved):

PR #3725, opened 2025-05-28
[DBT] Add processing_engine facet

dolfinus:
### Problem

Currently dbt integration uses custom facet `dbt_version` instead of standard facet `processing_engine` reported by all other integrations.

#### One-line summary:

DBT: Add `processing_engine` facet

### Checklist

----

### DIFF
--- a/integration/common/openlineage/common/provider/dbt/processor.py
+++ b/integration/common/openlineage/common/provider/dbt/processor.py
@@ -21,11 +21,13 @@ from openlineage.client.facet_v2 import (
     documentation_dataset,
     job_type_job,
     output_statistics_output_dataset,
+    processing_engine_run,
     schema_dataset,
     sql_job,
 )
 from openlineage.client.uuid import generate_new_uuid
 from openlineage.common.provider.dbt.facets import DbtVersionRunFacet, ParentRunMetadata
+from openlineage.common.provider.dbt.utils import __version__ as openlineage_version
 from openlineage.common.provider.snowflake import fix_account_name
 from openlineage.common.utils import get_from_multiple_chains, get_from_nullable_chain
 from openlineage_sql import parse as parse_sql
@@ -643,7 +645,10 @@ class DbtArtifactProcessor:
             )
 
     def get_run(self, run_id: str) -> Run:
-        run_facets = {"dbt_version": self.dbt_version_facet()}
+        run_facets = {
+            "dbt_version": self.dbt_version_facet(),
+            "processing_engine": self.processing_engine_facet(),
+        }
         if self._dbt_run_metadata:
             run_facets["parent"] = self._dbt_run_metadata.to_openlineage()
         return Run(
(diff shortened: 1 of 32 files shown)


integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -643,17 +645,28 @@ def extract_namespace(self, profile: Dict) -> str:
...
         if self._dbt_run_metadata:
             run_facets["parent"] = self._dbt_run_metadata.to_openlineage()
         return Run(
             runId=run_id,
             facets=run_facets,
         )
 
+    # TODO: remove
     def dbt_version_facet(self):
         return DbtVersionRunFacet(version=self.run_metadata["dbt_version"])
```
kacpermuda, line 661: Maybe let's emit a debug/info log here that this facet it's deprecated, and users should use the ProcessingEngineRunFacet (both here and in the other file)? It will also be more clear for developers, what this `#TODO Remove` means. DeprecationWarning is to much IMO because users will not be able to do anything with it, but some logging should be fine for now.

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -515,8 +538,16 @@ def _setup_dbt_run_metadata(self, start_event):
             root_parent_job_namespace=root_parent_job_namespace,
         )
 
+    # TODO: remove
     def dbt_version_facet(self) -> DbtVersionRunFacet:
-        return {"dbt_version": DbtVersionRunFacet(version=self.dbt_version)}
+        return DbtVersionRunFacet(version=self.dbt_version)
```
kacpermuda, line 543: This is probably just used internally, but either the typing or the function behavior was wrong until this change. Just letting you know about this change @MassyB , in case you rely on it somewhere else.

kacpermuda (approved):
Leaving nit comments. Also not sure if we should hardcode versions in tests, as well as use `any` instead of some regex for version validation, but given that this is not introduced in this PR we can improve it in separate PR.

integration/common/openlineage/common/provider/dbt/structured_logs.py
kacpermuda: I think it's fine, it's consistent with the method in other class that was there for a while and with the typing. We can probably consider it a fix and not a breaking change. cc @mobuchowski

integration/common/openlineage/common/provider/dbt/processor.py
dolfinus: I could add information about version there this facet will be removed, but what version should it be? 1.33 + 3 versions? 1.40?

kacpermuda: I don't think we have a clear policy on removal (@mobuchowski @harels did we settle on something ? I remember discussing this policy some time ago). We may have to keep it until major release, so maybe I'd be more general here and say that it will be removed `in the future`? Not very informative, but I'm not sure how else to handle it not to break somebody's workflow 😄

integration/common/openlineage/common/provider/dbt/structured_logs.py
dolfinus: I've reverted this change and updated method annotation

Issue #2774, opened 2024-06-14
[PROPOSAL] standard dataset facet for DBMS objects


davidjgoss:
**Purpose:**
Many datasets in an OpenLineage landscape are likely to be objects (tables, views) within a DBMS (in which I'm including Postgres, MySQL, Oracle etc, plus cloud warehouses like Snowflake etc).

Such objects have properties which may be of interest to an OpenLineage-ingesting system, but would not be applicable to other kinds of dataset (blob storage, APIs, etc). Specifically:

- Hierarchy - OpenLineage models a dataset name as a single string which is reasonable. DBMS objects tend to have a name derived from a hierarchy e.g. `database` -> `schema` -> `table` become `database.schema.table`. A consuming system may want to model those parts separately to e.g. make the hierarchy navigable in a tree fashion by a user, or assist searching. They could parse the single name string, but this would be fussy and error-prone - just splitting on the period is not enough for some object names, where parts may have to have escape characters added to ensure they are unambiguous. Also, although `database.schema.table` is the most common, several vendors have a different number/naming of items e.g. MySQL, Databricks.
- Object type - Both tables and views can be represented as datasets; it may benefit a consuming system's user to know which type a dataset is.

**Proposed implementation**
I'd propose a new `dbmsObject` facet, applicable to both input and output datasets, and static datasets.

Example dataset:
```json
{
  "namespace": "snowflake://my-account.eu-central-1",
  "name": "MY_DATABASE.MY_SCHEMA.\"InterestinglyNamedTable\"",
  "facets": {
    "dbmsObject": {
      "_producer": "...",
      "_schemaURL": "...",
      "objectType": "TABLE",
      "hierarchy": [
        {
          "name": "MY_DATABASE"
        },
        {
          "name": "MY_SCHEMA"
        },
        {
          "name": "InterestinglyNamedTable"
        }
      ]
    }
  }
}
```

Schema:
```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "...",
  "type": "object",
  "properties": {
    "type": {
      "type": "string",
      "description": "The type of object in the DBMS. Valid values may include `TABLE` and `VIEW`."
    },
    "hierarchy": {
      "type": "array",
      "description": "The names of items in the hierarchy within the namespace (e.g. database, schema, table), from highest to lowest level",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string",
            "description": "The name of the item in the hierarchy"
[...]

PR #3747, opened 2025-06-03
[SPEC] Add HierarchyDatasetFacet

dolfinus:
### Problem

Closes: #2774
Related: https://github.com/OpenLineage/OpenLineage/issues/3642#issuecomment-2875290259

#### One-line summary:

[SPEC] Add HierarchyDatasetFacet

### Checklist

----


dolfinus:
`objectType` is required because we can know it without knowing the hierarchy itself - e.g. all datasets within MongoDB are collections, all datasets within Kafka are topics and so on.
`hierarchy` can be empty as some databases have no such concept at all (e.g. Kafka topics are flat)

website/docs/spec/facets/dataset-facets/database_hierarchy.md
```diff
@@ -0,0 +1,54 @@
...
+sidebar_position: 12
+---
+
+# Database Hierarchy Dataset Facet
+
+The facet contains information about the hierarchy of objects within database.
+
+Fields description:
+- `objectType`: Type of the object within the hierarchy (e.g. `TABLE`, `VIEW`, `TOPIC`).
+- `hierarchy`: Database hierarchy levels (e.g. `DATABASE` -> `SCHEMA` -> `TABLE`), from highest to lowest level. Array of objects:
```
mobuchowski: I think it's worth mentioning the order of the array is significant here

website/docs/spec/facets/dataset-facets/database_hierarchy.md
```diff
@@ -0,0 +1,54 @@
...
+  - `type`: Hierarchy level type (e.g. `DATABASE`, `SCHEMA`, `TABLE`).
+  - `name`: Hierarchy level object name (e.g. `my_db`, `my_schema`, `my_table`).
+
+`objectType` is required field. `hierarchy` is not, because some databases have no hierarchy concept (e.g. Kafka topics).
```
mobuchowski: Why include the facet then? For `objectType` alone we can have https://openlineage.io/spec/facets/1-0-0/DatasetTypeDatasetFacet.json#/$defs/DatasetTypeDatasetFacet

dolfinus: I didn't know that there is a DatasetTypeDatasetFacet, as it is not present in OpenLineage docs :)

In this case, `objectType` can be removed, and `hierarchy` will be required field

dolfinus: Done

website/docs/spec/facets/dataset-facets/database_hierarchy.md
dolfinus: Added a note

website/docs/spec/facets/dataset-facets/database_hierarchy.md
dolfinus: Regarding missing `DatasetTypeDatasetFacet` documentation, created #3752

pawel-big-lebowski:
I tried to follow the discussion but perhaps missed something. What is the reason for having `hierarchy` as an array instead of having clearly defined properties: CATALOG, DATABASE, SCHEMA, TABLE? A producer could simply fill the fields present. The agility on the producers' side will bring confusion on the consumer being uncertain about the fields can be filled. 

I think I would be in favour of naming properties explicitly or having a better understanding of the currently proposed version.

dolfinus:
Different DBMS/catalogs/metastores can have quite different hierarchy:
* Postgres has SCHEMA, TABLE. It can also include DATABASE, but users cannot select data from another database unless they're created a fdw/dblink, so this is questionable.
* Oracle has SCHEMA (=USER), TABLE. It can also include DATABASE, but users cannot select data from another database unless they're created a dblink, so this is questionable.
* Teradata has only TABLE.
* Spark, Flink and Trino have CATALOG, SCHEMA, TABLE.
* Iceberg have NAMESPACE (which is actually [an array](https://iceberg.apache.org/javadoc/1.2.1/org/apache/iceberg/catalog/Namespace.html#of-java.lang.String...-)), TABLE

Are 4 described properties enough for all DBMS types?

pawel-big-lebowski:
@dolfinus thx for bringing this, Iceberg's array of namespaces is actually something that justifies the array approach. 

Given a Flink example:
```
CREATE TABLE my_catalog.mydb.mysql_table (
  ...
) WITH (
   'connector' = 'jdbc',
   'url' = 'jdbc:mysql://localhost:3306/target_db',
   'table-name' = 'target_table'
);
```
we can't have multiple DatabaseHierarchy facets for the same dataset. 
Would it make sense to add extra field to a facet to know which dataset hierarchy it describes?

dolfinus:
Yes, I've highlighted this in https://github.com/OpenLineage/OpenLineage/issues/3642#issuecomment-2875569598

pawel-big-lebowski:
@dolfinus ok, I like the idea of `source: catalog` field within the facet. Otherwise, the consumer can't know if DatabaseHierarchy facet and Catalog facet correspond to the same hierarchy, right?

dolfinus:
Right, but this still remains unsolved:

> we can't have multiple DatabaseHierarchy facets for the same dataset.

mobuchowski:
Do we want to address that or should we have a separate conversation? I think there is a pattern of having information that is relevant to only one naming scheme/symlink for dataset and we should probably solve it once rather than have potential disjoint solutions for different facets

dolfinus:
As discussed above, there can be a discrepancy between catalog information and actual dataset hierarchy in metastore - Flink's `CREATE TABLE my_catalog.mydb.mysql_table` doesn't mean that there is a table in MySQL with this specific database and table name.

Maybe we should detach this PR from #3642 and leave this facet just for real dataset hierarchy in the database. For #3642 we can add field describing object inside catalog to CatalogDatasetFacet.

heron--:
@dolfinus @mobuchowski Do you know if this change will be merged in the near future? We have a use case where we'd like to represent hierarchy for a producer's proprietary data store. This facet seems like the best fit for that.

We can implement it before this is officially a part of the spec, but it'd be better long term if it's officially supported.

mobuchowski:
I think we got hang up on the "catalog" vs "real database" hierarchy. Thanks @heron-- for getting us to move forward.

And I think we should go with 
- DatabaseHierarchy facet -> describe "real" database
- add `hierarchy` field to CatalogDatasetFacet - describe catalog's view on the hierarchy
basically agreeing with @dolfinus.
This would solve both issues without losing information, and without "maps of hierarchies" or "source" fields, but the information is contained in a relevant facets

I think we should merge this as is - @dolfinus do you want to rebase/update the PR?
And we should add CatalogDatasetFacet support in the subsequent PR.

dolfinus:
Let me rebase this PR and exclude CATALOG from examples

website/docs/spec/facets/dataset-facets/database_hierarchy.md
```diff
@@ -0,0 +1,47 @@
...
+  - `name`: Hierarchy level object name (e.g. `my_db`, `my_schema`, `my_table`).
+
+Example:
+
+```json
+{
+    ...
+    "inputs": {
+        "facets": {
+          "database_hierarchy": {
```
mobuchowski: JSON Schema defines it as `databaseHierarchy` rather than `database_hierarchy`

website/docs/spec/facets/dataset-facets/database_hierarchy.md
```diff
@@ -0,0 +1,47 @@
...
+            "_producer": "https://some.producer.com/version/1.0",
+            "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/DatabaseHierarchyDatasetFacet.json",
```
mobuchowski: https://openlineage.io/spec/facets/1-0-0/DatabaseHierarchyDatasetFacet.json#/$defs/DatabaseHierarchyDatasetFacet

website/docs/spec/facets/dataset-facets/hierarchy.md
```diff
@@ -0,0 +1,47 @@
...
+  Array of objects, the order is important:
+  - `type`: Hierarchy level type (e.g. `DATABASE`, `SCHEMA`, `TABLE`).
+  - `name`: Hierarchy level object name (e.g. `my_db`, `my_schema`, `my_table`).
+
+Example:
+
+```json
+{
+    ...
+    "inputs": {
```
mobuchowski:
```
  "inputs": [
    {
      "namespace": "...",
      "name": "...",
      "facets": { ... }
    }
  ]
```
rather than 

```
  "inputs": {
    "facets": { ... }
  }
```

dolfinus:
Thanks, fixed

mobuchowski:
@jakub-moravec 

>What about the case when the [documented best practices](https://openlineage.io/docs/spec/naming) say that db name should be part of the namespace? (Currently, only Azure Cosmos DB it seems)

I don't get where we get that? It seems to me that Azure Cosmos DB is the outlier for the opposite reason.

>Does this facet reflect/document the format of the Dataset Name? Akka, if there are three levels documented in this facet (db/schema/table), does that imply that the name of the dataset should have three segments (db.schema.table)?

I believe the point is also the opposite. The solution presented here is flexible enough to not prescribe the depth of the hierarchy, and would work with arbitrary depth - from two, like in MySQL to unbounded number as in Iceberg namespaces.  

>Why is the scope of the facet limited to databases/datasets only? The ability to describe the semantics of different levels of hierarchy for jobs would be very useful as well

One of our original mistakes was that we used the name/namespace convention for both the jobs and datasets, suggesting that they are the same - while they are quite different, with different requirements. And more often they are user provided in jobs case.

I think it would be nice to have something like this for jobs too. However, while the database hierarchy is a "static" property of a dataset at this moment, something like job hierarchy might not. Imagine job that spawns other jobs, but something fails and the subsequent jobs aren't spawned. That would not be represented in a hierarchy like that well. So we probably should discuss that separately.

dolfinus:
@jakub-moravec you're looking at an example of the facet. Nothing restricts databases from having 2 levels (schema.table), 1 level (MongoDB collection, Kafka topic) or more than 3 levels (nested Iceberg namespaces). Level type is not enum, it's an arbitrary string.

jakub-moravec:
Thanks for the comments @mobuchowski @dolfinus 

- Levels of depth - I'm not saying we should be restricting levels of depth or the number of name segments. But what I would expect from this facet (and how we could efficiently use it) is - if I see that the Hierarchy facet has 3 levels, I expect that the dataset name has 3 segments, I would parse it accordingly, and treat the individual levels with respective semantical meaning. Is this how the hierarchy facet is supposed to _interact_ with the dataset name? 

- Datasets vs Jobs - on the consumer side, we treat them the same. We try to create the hierarchy for them, and assign semantical meaning to the name segments, you might remember my presentation on that. Currently, we have that solution in our product, but this facet allows pushing the logic to the OL payloads / OL producers (which is a good direction) - but for us to fully leverage that, we'd need that to work across the technologies - databases, non-db datasets (filesystems, etc), jobs. So would it make sense to take a bit more time and try making the solution more generic now, rather than adding it to the spec for DBs only and then doing a similar thing for other cases with a different facet?

dolfinus:
> I expect that the dataset name has 3 segments, I would parse it accordingly

The main purpose of having such facet is to _avoid_ parsing dataset name into segments, and instead rely on already present information about hierarchy (level type+name). For example, database, schema or table name can contain dots, and instead of nice format `mydb.myschema.mytable` you'll get `"my.database"."my.schema"."my.table"` (depending on database). Having dedicated segments inside a facet:
```json
[
  {"type": "DATABASE", "name": "my.database"},
  {"type": "SCHEMA", "name": "my.schema"},
  {"type": "TABLE", "name": "my.table"}
]
```
allows to avoid parsing altogether.

Regarding other dataset types, e.g. file paths, API methods and so on, those usually have a `/some/path` form, and `/` delimiter cannot be a part of any directory/file name/url (filesystem or HTTP specification prohibits this). So there is no such issue in the first place, and thus this facet is not very helpful.

Regarding Job hierarchy - this is not a part of current proposal. It should be discussed in a dedicated issue. But there is already a `parent` run facet, and `jobDependencies` facet as well, which allow to express complex interactions between jobs.
Also, unlike datasets in some DB, which knows everything about dataset hierarchy, jobs within hierarchy can be run by different engines, and there will be no one source of truth for their hierarchy. I don't think there are any similarities here.

mobuchowski:
DatabaseHierarchy facet alone is also useful for static lineage - dataset events.

I also think that while deduplicating and making things more generic feels good, we'll end up at most at two facets - for jobs and datasets. I think it's fine to work on it separately.

>Also, unlike datasets in some DB, which knows everything about dataset hierarchy, jobs within hierarchy can be run by different engines, and there will be no one source of truth for their hierarchy. I don't think there are any similarities here.

I would expect different integrations to make their hierarchy more official and standarized if we adopt the JobHierarchy facet. At Datadog we do take advantage of the implicit standard job hierarchy - like we know that there is Airflow DAG -> Airflow Task event connection. We can't assume the same for more generic jobs and have to deal with that.

The different engines taking part of single execution is a valid point. We need to make sure this is "dynamic" enough. I don't have good answer to that yet.

jakub-moravec:
Thanks for [this](https://github.com/OpenLineage/OpenLineage/pull/3747#issuecomment-3944135484) comment @dolfinus. I misrepresented some aspects of the proposed facet. But I still think the Hierarchy facet should be correlated with the OL name segments
- Why would I have a segment in the name that isn't documented in the Hierarchy facet? 
- Why would I have an entity in the hierarchy facet that isn't part of the dataset name? 

That just creates space for having datasets that have partially interpretable hierarchies, and require consumers to have a more edge-case aware handling. 

On the re-usability of the facet - makes complete sense to discuss the Jobs in a separate ticket. But to keep the door open for re-usability, not only for Jobs, but also for other dataset types, I'd suggest renaming the facet to something more generic like `Hierarchy facet`. Just to give you some examples of technologies we are using OpenLineage for and where this would be useful: 

Filesystems: 
- bucket or equivalent (not always present)
- folder 
- folder (can be nested)
- file

Confluence 
- Space
- Document

ElasticSearch
- Index

Slack
- Workspace
- Channel
- Document

... there are many more where the hierarchy is important, but use of `DatabaseHierarchy facet` would seem a bit off.

dolfinus:
> Why would I have a segment in the name that isn't documented in the Hierarchy facet?
> Why would I have an entity in the hierarchy facet that isn't part of the dataset name?

Both dataset name and hierarchy should contain the same number of segments. I don't see any case when these may be different, other that bugs in implementation/naming convention.

> That just creates space for having datasets that have partially interpretable hierarchies, and require consumers to have a more edge-case aware handling.

I've mentioned before case with escaping dots in database/schema/table name. But also there are examples then dataset name contains dot (e.g. Kafka topic), and this doesn't mean that there is some kind of hierarchy in the database (e.g. there are no databases/schemas/namespaces in Kafka at all). This is _already_ an edge case which is hard to interpret on consumer side - OL event may lack proper context about dataset name to deduce this.

It's always better to _explicitly_ provide a list of hierarchy levels in specific database than to rely on name parsing. New facet provides more information than just name alone.

> I'd suggest renaming the facet to something more generic like Hierarchy facet

Ok, let's rename facet

heron--:
Thanks for all the continued discussion! I'm a big fan of the current approach - I can see broader adoption of this facet making it much simpler for OL event consumers to process datasets.

When do you think this will get merged and released? I'd rather use this facet once it's officially a part of the OL spec.

mobuchowski:
BTW @jakub-moravec - due to JSON Schema deficiencies, we'd have to have separate `HierarchyJobFacet` or `HierarchyRunFacet` if we want to attach this to those two. The same happened with `Tag*Facet`. 

@heron-- from my point of view this looks good - I think we have consensus here. I'd be happy to merge and release this in the next release. Would be good if any other committer looked at this too.

jakub-moravec:
@mobuchowski yeah, that is understood. From my PoV, it's a good start to have this usable universally for any dataset.

website/docs/spec/facets/dataset-facets/hierarchy.md
```diff
@@ -0,0 +1,49 @@
...
+  - `type`: Hierarchy level type (e.g. `DATABASE`, `SCHEMA`, `TABLE`, etc).
+  - `name`: Hierarchy level object name (e.g. `my_db`, `my_schema`, `my_table`).
+
+Example:
+
+```json
+{
+    ...
+    "inputs": {
+        "namespace": "posgres://some.host:5432",
```
mobuchowski:
```suggestion
        "namespace": "postgres://some.host:5432",
```

spec/tests/HierarchyDatasetFacet/1.json
```diff
@@ -0,0 +1,20 @@
+{
```
mobuchowski, line 1: we should rename the file

website/docs/spec/facets/dataset-facets/hierarchy.md
```diff
@@ -0,0 +1,49 @@
...
+  Array of objects, the order is important:
+  - `type`: Hierarchy level type (e.g. `DATABASE`, `SCHEMA`, `TABLE`, etc).
+  - `name`: Hierarchy level object name (e.g. `my_db`, `my_schema`, `my_table`).
+
+Example:
+
+```json
+{
+    ...
+    "inputs": {
```
mobuchowski: inputs should be an array

website/docs/spec/facets/dataset-facets/hierarchy.md
dolfinus: Fixed

spec/tests/HierarchyDatasetFacet/1.json
dolfinus: Done

website/docs/spec/facets/dataset-facets/hierarchy.md
dolfinus: Fixed

mobuchowski:
Thanks for contribution @dolfinus. I think this wins the award for longest open PR that was subsequently merged.

Issue #1701, opened 2023-03-07
spark: support for 'select * from ...' for Spark JDBC queries


tnazarew:
Current column level lineage doesn't support `select * from ...` syntax. That is because only information we have about fields come from parsed query, if there is no info about fields in query, we can't really tell anything about the lineage. 

So the real issue is getting additional information about fields from external source, which is complicated because you would often have to request multiple tables and make those requests as separate call(s) to database. Furthermore each database can have different syntax for it so it can no longer be really generic. Fortunately all the details for jdbc connection are present while collecting lineage. So it shouldn't be that much of a problem to connect to source.

dolfinus:
Solved in #3763

Issue #1758, opened 2023-04-05
Spark JDBC star column lineage support


perttus:
Please add column lineage support also for Spark JDBC sources as discussed also in https://github.com/OpenLineage/OpenLineage/issues/1584

dolfinus:
Solved in #3763

PR #3763, opened 2025-06-11
Column level lineage for jdbc queries load

pawel-big-lebowski:
Problem: Sql parser does not support column level lineage for wildcards like `select *` queries. Although the problem in general is hard to solve, often times there is a workaround to this. 

For example, a JDBC table can be read in Spark by using ` .option("dbtable", "authors")` option, but this doesn't allow adding even a simple filtering which can be achieved with: ` .option("query", "select author_id, author_name from authors where author_name like 'Marquez'")`. Usecases like this can be solved with the following logic: if a sql parsed contains only a single input table and does not contain a column lineage, and query performed in lower case starts with `select * from`, then we can use LogicalRelation to obtain fields and names as a starting point for column level lineage.


dolfinus:
Closes #1701 and #1758

Thanks @pawel-big-lebowski !

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/column/JdbcColumnLineageInputCollectorTest.java
```diff
@@ -73,32 +87,65 @@ void setup() {
...
+      when(InputFieldsCollector.extractDatasetIdentifier(
+              any(ColumnLevelLineageContext.class), any(JDBCRelation.class)))
+          .thenReturn(Arrays.asList(datasetIdentifier1, datasetIdentifier2));
+      visitor.collectInputs(context, logicalRelation);
+      verify(builder, never())
+          .addInput(any(ExprId.class), any(DatasetIdentifier.class), any(String.class));
+    }
+  }
+
+  @Test
```
tnazarew, line 128: we could do sth about these tests... so much is mocked that even though I know what the test is about, it's not easy to see what exactly does it check. So maybe a line of comment that we check that method is executed so we know that some inputs are being added?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/JdbcColumnLineageVisitor.java
```diff
@@ -0,0 +1,66 @@
...
+    this.context = context;
+  }
+
+  @Override
+  public void collectInputs(ColumnLevelLineageContext context, LogicalPlan node) {
+    if (!isJDBCNode(node)) {
+      return;
+    }
+    JdbcColumnLineageVisitorDelegate delegate =
+        new JdbcColumnLineageVisitorDelegate(
```
tnazarew, line 32: ~what~ why do we use delegate here?

pawel-big-lebowski: We want to have the code split into several methods. 
In order to limit amount of those method's arguments, it's good to introduce a new class to share the common parameters like `sqlMeta`, `jdbcOptions`,  `datasetIdentifiers`, column lineage context, etc. However, we can't achieve this within the `JdbcColumnLineageVisitor` -> it should not contain any state apart from `OpenLineageContext`.

integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/column/JdbcColumnLineageInputCollectorTest.java
pawel-big-lebowski: will do

tnazarew (approved):

PR #3770, opened 2025-06-13
python: async HTTP client

mobuchowski:
This HTTP client addresses performance bottlenecks when emitting thousands of OpenLineage events (particularly from dbt integrations) by introducing an optional async HTTP transport. The current synchronous transport blocks on each event, creating significant delays when processing large numbers of events sequentially.

Key improvements:
  - Non-blocking event emission with configurable concurrency (default: 100 parallel requests)
  - Event ordering guarantees (START events sent before COMPLETE/FAIL/ABORT) if send before the 
  - Bounded queue management to prevent memory exhaustion
  - Maintains backward compatibility with existing synchronous transport

One drawback is that integration needs to support new client method (wait_for_completion), otherwise the HTTP client could lose data if there are events in transit - this matters particularly for Airflow.


website/docs/client/python.md
```diff
@@ -232,7 +232,11 @@ Allows sending events to HTTP endpoint, using [requests](https://requests.readth
   - `apiKey` - string setting the Authentication HTTP header as the Bearer. Required if `type` is `api_key`.
 - `compression` - string, name of algorithm used by HTTP client to compress request body. Optional, default value `null`, allowed values: `gzip`. Added in v1.13.0.
 - `custom_headers` - dictionary of additional headers to be sent with each request. Optional, default: `{}`.
-- `retry` - dictionary of additional configuration options passed to [`urllib3.util.Retry`](https://urllib3.readthedocs.io/en/1.26.20/reference/urllib3.util.html) object. Added in v1.33.0. Defaults are below; those are non-exhaustive options, but the ones that are set by default. Look at  [`urllib3.util.Retry`](https://urllib3.readthedocs.io/en/1.26.20/reference/urllib3.util.html) options for full reference.
+- `async_config` - dictionary specifying asynchronous transport configuration. Optional, by default async mode is disabled.
```
dolfinus, line 235: Maybe there should be `http` and `http_async` transports, each with specific configuration?
Changing behaviour of existing client (the most used one, probably) doesn't seem right.

integration/dbt/openlineage/dbt/__init__.py
```diff
@@ -107,14 +108,44 @@ def dbt_run_event_failed(
...
+        }
+    )
+    logger = logging.getLogger("openlineage.dbt")
+    custom_logging_level = os.getenv("OPENLINEAGE_CLIENT_LOGGING", None)
+    if custom_logging_level:
+        logger.setLevel(custom_logging_level)
+    return logger
+
+
+logger = set_up_logger()
```
dolfinus: I'd recommend moving this method call into `def main():`

integration/common/setup.py
```diff
@@ -22,7 +22,7 @@
 requirements = [
     "attrs>=19.3.0",
     f"openlineage-python=={__version__}",
-    f"openlineage_sql=={__version__}",
+    "openlineage_sql>=1.33.0",
```
dolfinus, line 25: Probably a local testing artifact

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         self.dbt_command_line: List[str] = dbt_command_line
         self.profiles_dir: str = get_dbt_profiles_dir(command=self.dbt_command_line)
         self.dbt_log_file_path: str = get_dbt_log_path(command=self.dbt_command_line)
-        self.dbt_log_dirname: str = self.dbt_log_file_path.split("/")[-2]
+        self.dbt_log_dirname: str = "/".join(self.dbt_log_file_path.split("/")[:-1])
```
dolfinus, line 67: Why not using `os.path.split()`?

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -618,33 +618,61 @@ def _run_dbt_command(self) -> Generator[str, None, None]:
         dbt_command_line = add_or_replace_command_line_option(
             dbt_command_line, option="--write-json", replace_option="--no-write-json"
         )
+
         self._open_dbt_log_file()
+        self.logger.debug(f"Opened log file `{self.dbt_log_file_path}`")
```
dolfinus: Please use `{}` or `%` formatting instead of f-strings for log messages, as the they may be used by some integrations as unique log message key (e.g. Sentry)

website/docs/client/python.md
dolfinus: Also, I have an idea how this approach with queue-based transport can be generalized: #3771

mobuchowski: I thought of this, but moved to single transport because `httpx` can be used in both modes, removing need for duplicate HTTP request libraries. And `sync` is default.
Another benefit that convinced me is that for people using just `OPENLINEAGE_URL` and there's a lot of them, I can add single flag that enables that behavior: `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP__ASYNC_CONFIG__ENABLED`

like in the test
```
@patch.dict(
    os.environ,
    {
        "OPENLINEAGE_URL": "http://example.com",
        "OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP__ASYNC_CONFIG__ENABLED": "true",
    },
)
def test_async_transport_with_overwritten_enabled_async_flag() -> None:
    transport: HttpTransport = OpenLineageClient().transport
    assert transport.kind == HttpTransport.kind
    print(transport.config)
    assert transport.config.async_config.enabled is True
```

However I'm open to doing it as a separate transport, I don't think it's a blocker to have to use it like this.

dolfinus: My concern is:
* `.emit(event)` method is no longer blocking, and users have to call `.wait_for_completion()` to ensure that all events were sent. This is change in behavior, not expected by users, and also not implemented in any other transport.
* HttpAdapter and Session are not just deprecated, but completely ignored by new transport implementation, but someone may rely on them (e.g. custom timeouts or proxy support).

dolfinus: `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP__ASYNC_CONFIG__ENABLED=true` can be replaced with `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP__TYPE=http_async`

mobuchowski: Yeah, new transport can read the `OPENLINEAGE_URL` if needed too. 

>.emit(event) method is no longer blocking, and users have to call .wait_for_completion() to ensure that all events were sent. This is change in behavior, which other transports do not implement.

It is still blocking if async is disabled. Other transports don't support that, but if we don't introduce that interface, we'd never be able to do something like this. 

BTW, batch transport would have to rely on pretty much the same behavior - with the difference that it might send a batch in the meantime if we bump into full batch threshold. But at the end of execution we might not, and we'll have to have a way to flush/wait_for_completion/shutdown rest of the events.

dolfinus: > It is still blocking if async is disabled.

I see, but from the developer's perspective this can be inconvenient. Transports are designed to match specific interface, so they can be switched in any time, and almost nothing will change from developer's perspective. And user can configure only some details (auth, compression, custom headers) which influence where and how _data_ will be send, but this is still one `.emit()` call in code.

But now configuration can influence behavior - if `async_enabled=True` developer have to call `.emit()` and `.wait_for_completion()`, if `async_enabled=False` developed should know that `.emit()` is blocking operation. This can make things more complicated, especially in case of custom integrations.

dolfinus: IMHO, this behaviour should be chosen by a developer, not by user. For example, by passing `.emit(event, blocking=False)` in source code.
The ultimate solution will be another transport interface I've described in the issue above, but designing and implementing this will definitely take some time.

mobuchowski: Agree with you that it complicates things. This is not done without purpose though - current sync interface just does not allow us to have the desired throughput. 
So I think we need to have that kind of behavior anyway, and forcing integrations to shutdown/flush explicitely if they care about data completeness is not that large tradeoff. Alternatives - like, for example, separate `emit_async` call returning something like Future, or otherwise managing this by integration itself seem to me like much more work for potentially not better results.

mobuchowski: >.emit(event, blocking=False)

That's interesting idea, however suffers from the same problem - most of the transports just implement the blocking interface.

>IMHO, this behaviour should be chosen by a developer, not by user. For example, by passing .emit(event, blocking=False) in source code.

Not sure why - we already expose a lot of connection-level options like retries to be user-configurable. Developers can choose to build transport manually and create client with that transport, or defer to config/env variables - which makes sense in the most cases, because the public integrations will get used with different backends.

integration/common/setup.py
mobuchowski: Yes, will separate the dbt level changes from this PR

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,544 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+    def emit(self, event: Event) -> None:
+        event_str = Serde.to_json(event)
+
+        run_id = None
+        event_type = None
+
+        if isinstance(event, (RunEvent, RunEventV2)):
+            event_id = event.run.runId + "-" + str(event.eventType.value)
+            run_id = event.run.runId
+            event_type = str(event.eventType.value) if hasattr(event, "eventType") else None
```
dolfinus: event.eventType is already used above to get event_id

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,544 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+                            len(pending_events),
+                        )
+                        break
+
+                    if resp.status_code not in self.config.retry.get(
+                        "status_forcelist", [500, 502, 503, 504]
+                    ):
+                        handle_failure(response=resp)
+                        break
+                    elif attempt == self.config.retry.get("total", 5):
```
dolfinus: Let's move default values to class level constants

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,544 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+    def mark_failed(self, event_id: str):
+        with self.lock:
+            log.debug("Marking event %s as failed", event_id)
+            self.stats["pending"] -= 1
+            self.stats["failed"] += 1
+            if self.events.get(event_id) == EventStatus.PENDING:
+                del self.events[event_id]
+
+    def add_pending_completion_event(
+        self, run_id: str, event_id: str, body: bytes | str, headers: dict[str, str]
```
dolfinus: Maybe this set of arguments should be moved to a dedicated class Event or Request?

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,544 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+            "AsyncEmitter for %s started with configured queue size: %s (actual: %s), max concurrent: %s",
+            self.config.url,
+            self.configured_queue_size,
+            self.configured_queue_size * 2,
+            self.max_concurrent,
+        )
+
+    def _setup_loop(self) -> None:
+        self.loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self.loop)
```
dolfinus, line 343: This makes event loop a global object, which may interfere with other async libraries or frameworks. Also it will be impossible to use CombinedTransport with multiple async HttpTransport, as each one will create it's own loop, replacing current one.
Could we call methods like `loop.create_task(...)` directly? If not, could we create event loop only if there was none, and never close it?

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,544 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+            timeout=self.config.timeout, verify=self.config.verify, follow_redirects=True, transport=transport
+        ) as client:
+            resp = client.post(
+                url=urljoin(self.url, self.endpoint),
+                content=body,
+                headers=headers,
+            )
 
-    return Session()
+        http_client.HTTPConnection.debuglevel = prev_debuglevel
```
dolfinus: This is probably should be set within try..finally block, otherwise raising an exception will left current debuglevel instead of resetting it

website/docs/client/python.md
dolfinus: > That's interesting idea, however suffers from the same problem - most of the transports just implement the blocking interface.

This is just a temporary solution for Python HttpTransport, to allow using it async way, if integration requires so.

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,544 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+                    break
+                try:
+                    resp = await client.post(
+                        url=urljoin(self.config.url, self.config.endpoint),
+                        content=body,
+                        headers=headers,
+                    )
+
+                    if 200 <= resp.status_code < 300:
+                        pending_events = self.event_tracker.mark_success(event_id, run_id, event_type)
```
dolfinus: Using threading.Lock in combination with asyncio may lead to event loop stuck, because another coroutine within same thread already locked it, and Lock object is not reentrant. This part have to be written **very** carefully.

client/python/openlineage/client/transport/http.py
mobuchowski: It's run in a worker thread, that only handles that loop. It is created by `AsyncEmitter`:

```
        self.worker_thread = threading.Thread(target=self._setup_loop, daemon=True)
```

Nothing else should run in that thread, or it should not reuse event loop from some other library. 

We can manage event loop manually but I don't know what would be the advantage of that in this case

dolfinus: The main question here is "does asyncio.set_event_loop() modifies thread-local object or a global one?". I didn't find an answer in [python docs](https://docs.python.org/3/library/asyncio-eventloop.html)

mobuchowski: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.set_event_loop
```
Set loop as the current event loop for the current OS thread.
```

mobuchowski: https://github.com/python/cpython/blob/main/Lib/asyncio/events.py#L706-L727

dolfinus: Great, thanks! If you don't mind, please add a comment that this is a thread-local loop

client/python/openlineage/client/transport/http.py
mobuchowski: Agreed with the sentiment. The Python coroutines can suspend only on await/yield, and `EventTracker` is pretty much self-contained class wrapping some tracking resources, that does not do any function calls, and only interacts with dicts it owns - IMO it's okay

dolfinus: As discussed above, asyncio event loops are thread-local, so even if event loop stuck because of waiting threading.Lock, only worker thread will be locked, not all coroutines in the Python process. So yeah, it's okay, comment can be resolved.

client/python/openlineage/client/transport/composite.py
```diff
@@ -83,9 +83,15 @@ def emit(self, event: Event) -> None:
...
                 if self.config.continue_on_failure:
                     log.warning("Transport %s failed to emit event with error: %s", transport, e)
                 else:
                     msg = f"Transport {transport} failed to emit event"
                     raise RuntimeError(msg) from e
+
+    def wait_for_completion(self, timeout: float = 10.0) -> bool:
+        # This can wait longer than timeout if multiple transports are slow, but acceptable
+        for transport in self.transports:
+            transport.wait_for_completion(timeout)
```
kacpermuda, line 96:
```suggestion
        # This can wait longer than timeout if multiple transports are slow, but acceptable
        for transport in self.transports:
            transport.wait_for_completion(timeout \ len(self.transports))
```
Shouldn't we try to limit the timeout for each transport? In environments that limit OL's time for execution this can be crucial. Not a blocker, but something to consider, since we're introducing the feature now. I'd expect that when passing timeout to some method, the whole method will be done within that time, regardless of number of transports.

client/python/openlineage/client/transport/http.py
```diff
@@ -70,10 +83,546 @@ def create_token_provider(auth: dict[str, str]) -> TokenProvider:
...
+    def emit(self, event: Event) -> httpx.Response:
+        # If anyone overrides debuglevel manually, we can potentially leak secrets to logs.
+        # Override this setting to make sure it does not happen.
+        prev_debuglevel = http_client.HTTPConnection.debuglevel
+        try:
+            http_client.HTTPConnection.debuglevel = 0
+
+            event_str = Serde.to_json(event)
+            body, headers = self._prepare_request(event_str)
+            transport = httpx.HTTPTransport(retries=self.config.retry.get("total", 5))
```
kacpermuda, line 609: Are we not using all other settings for retry like `status_forcelist` ? I see them in async, but not in sync. This changes the current behavior, and even though it was recently introduced, I think we should either not change it or be verbose about it.

client/python/openlineage/client/transport/http.py
```diff
@@ -113,22 +657,31 @@ def from_dict(cls, params: dict[str, Any]) -> HttpConfig:
...
+                category=DeprecationWarning,
+                stacklevel=3,
+            )
         return cls(
             url=url,
             timeout=options.timeout,
             verify=options.verify,
             auth=ApiKeyTokenProvider({"api_key": options.api_key}) if options.api_key else TokenProvider({}),
-            session=session,
-            adapter=options.adapter,
```
kacpermuda, line 131: I feel like, if we're doing a breaking change here, maybe it's time to remove options entirely from http and client, since It's been deprecated for a long time now?

client/python/openlineage/client/client.py
```diff
@@ -293,42 +307,52 @@ def _find_yaml_config_path() -> str | None:
...
+            config_dict = deep_merge_dicts(config_dict, env_config["transport"])
+
+        return HttpTransport(HttpConfig.from_dict(config_dict))
 
     @staticmethod
     def _http_transport_from_url(
         url: str,
         options: OpenLineageClientOptions | None,
-        session: Session | None,
+        session: Session | None,  # Deprecated: no longer used with httpx
```
kacpermuda, line 341: I have similar feelings as @dolfinus here from a compatibility perspective. If we have to introduce breaking changes (ignoring options, session etc.) to implement this new feature, it does not seem right. We've been keeping a lot of legacy stuff around for the sole purpose of not breaking people's workflows so let's not break it now. I think requiring two http requests libraries is a small price for this. Either that, or let's remove the options and other legacy code entirely, since it's been deprecated for long time now.

client/python/openlineage/client/transport/composite.py
mobuchowski: If you really want to do this in this way, the way would be to keep track of how much time is left after first, second... executions and set timeout to the rest value. This way you will emit as much as you can with given timeout, using given order of transports by the user.

PR #3793, opened 2025-06-18
dbt: fix log path, more precise file reading

mobuchowski:
This PR fixes issue where dbt-ol read incorrect structured log file, reads log file more precisely, and adds better debug logging.

### DIFF
--- a/integration/common/openlineage/common/utils.py
+++ b/integration/common/openlineage/common/utils.py
@@ -1,7 +1,10 @@
 # Copyright 2018-2025 contributors to the OpenLineage project
 # SPDX-License-Identifier: Apache-2.0
+import logging
 import os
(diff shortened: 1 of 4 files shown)


integration/common/openlineage/common/utils.py
```diff
@@ -172,14 +175,18 @@ def __init__(self, text_file: TextIO):
         self.incomplete_line = ""
         self.chunk_size = 4096
 
-    def read_lines(self):
+    def read_lines(self, max_read_size: int) -> Generator[str, Any, None]:
```
kacpermuda, line 178: Is that a breaking change or it's not a public interface function?

mobuchowski: It's not a public interface. We should assume nothing is that is not otherwise mentioned.

PR #3796, opened 2025-06-19
[DBT] Make invocation_id facet optional

dolfinus:
### Problem

According to [dbt manifest schema](https://github.com/dbt-labs/dbt-core/blob/main/schemas/dbt/manifest/v12.json) and [dbt-core sources](https://github.com/dbt-labs/dbt-core/blob/v1.10.1/core/dbt/context/base.py#L618-L623), `invocation_id` can be null. If this field is missing in dbt log or manifest for some reason, this will lead to exception:
https://github.com/apache/airflow/pull/51916

#### One-line summary:

[DBT] Make `invocation_id` field optional

### Checklist

----

### DIFF
--- a/integration/common/openlineage/common/provider/dbt/processor.py
+++ b/integration/common/openlineage/common/provider/dbt/processor.py
@@ -6,7 +6,7 @@ import datetime
 import logging
 from abc import abstractmethod
 from enum import Enum
-from typing import Dict, List, Optional, Sequence, Tuple
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 import attr
 from openlineage.client.event_v2 import Dataset, InputDataset, Job, OutputDataset, Run, RunEvent, RunState
@@ -147,7 +147,7 @@ class DbtArtifactProcessor:
         self.job_namespace = job_namespace
         self.dataset_namespace = ""
         self.skip_errors = skip_errors
-        self.run_metadata = None
+        self.run_metadata: dict[str, Any] = {}
         self.command = None
         self.models = models or []
         self.selector = selector
@@ -681,9 +681,9 @@ class DbtArtifactProcessor:
 
     def get_run(self, run_id: str) -> Run:
         run_facets = {
-            "dbt_version": self.dbt_version_facet(),
-            "dbt_run": self.dbt_run_run_facet(),
-            "processing_engine": self.processing_engine_facet(),
+            **self.dbt_version_facet(),
+            **self.dbt_run_run_facet(),
+            **self.processing_engine_facet(),
         }
         if self._dbt_run_metadata:
             run_facets["parent"] = self._dbt_run_metadata.to_openlineage()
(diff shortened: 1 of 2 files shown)


integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -700,8 +702,11 @@ def dbt_version_facet(self) -> DbtVersionRunFacet:
         )
         return DbtVersionRunFacet(version=self.run_metadata["dbt_version"])  # type: ignore[index]
 
-    def dbt_run_run_facet(self) -> DbtRunRunFacet:
-        return DbtRunRunFacet(invocation_id=self.run_metadata["invocation_id"])  # type: ignore[index]
+    def dbt_run_run_facet(self) -> Optional[DbtRunRunFacet]:
+        invocation_id = self.run_metadata.get("invocation_id")  # type: ignore[attr-defined]
```
kacpermuda: I know it's unlikely to happen, but if called before `self.run_metadata` is set within parse, it will also throw error. Maybe we should initialize the run_metadata in init as dict instead? Or check if it's None before retrieval? Not sure if type ignore is a good approach here. I see the same in `dbt_version_facet` so it's probably been there for a while with no issues, but if we're making it optional, we might as well prevent some other failures.

kacpermuda (approved):
Apart from the nit above, it looks good, thanks !

integration/common/openlineage/common/provider/dbt/processor.py
dolfinus: I can convert `*_facet` methods to return dict, which are then merged together

kacpermuda: I meant that `self.run_metadata` is `None` before `parse` is called, so this `self.run_metadata.get("invocation_id")` will cause error as `None.get(...)` won't work. So I think changing method output to dict is not necessary, maybe additional check if self.run_metadata is in fact, a dict ?

dolfinus: > so this self.run_metadata.get("invocation_id") will cause error as None.get(...)

I got your point, but multiple facets are using `run_metadata` content, and making all of them return `None` with later checking `if optional_facet is not None: facets[key] = optional_facet` is IMHO harder to read than `facets.update(*optional_facet)`

kacpermuda: Sure, whatever solution is fine with me as long as it solves the problem. I was hoping we can avoid changing the return types of the functions as it may potentially break something in people's workflows. WDYT?

kacpermuda: I'm just not sure how "public" these methods are, so I wanted to be cautious.

kacpermuda: We can just change the self.run_metadata = {} in init, that solves the problem, right?

dolfinus: > We can just change the self.run_metadata = {} in init, that solves the problem, right?

If dict is empty, accessing value by key will still raise an exception, just `KeyError` instead of `AttributeError`

kacpermuda: Yes I meant together with switch to .get(), then it's fine.

dolfinus: > I'm just not sure how "public" these methods are, so I wanted to be cautious.

Well, any change in return type of `def dbt_version_facet` could be breaking for someone - it always was `DbtVersionRunFacet` but become `DbtVersionRunFacet | None` or `dict[str, DbtVersionRunFacet]`.
For `def dbt_run_run_facet` and `def processing_engine_facet` it's easier as both methods appeared in OL 1.34 few days ago

kacpermuda: Makes sense, implement it as you see fit.

dolfinus: Set `self.run_metadata = {}` by default, and changed facet methods to return dict. In either way, DbtArtifactProcessor returned facet objects, and DbtStructuredLogsProcessor which inherits from DbtArtifactProcessor sometimes returned dicts, so this already was a bit surprising from developers perspective

PR #3808, opened 2025-06-23
run Java SQL tests when building package

mobuchowski:


kacpermuda (approved):

integration/sql/iface-java/script/build.sh
```diff
@@ -70,3 +70,6 @@ else
     printf "Expected %s\n Got %s" "$EXPECTED" "$OUTPUT"
     exit 1
 fi
+
+printf "\n------ Running gradle tests ------\n"
+"$ROOT"/gradlew test
```
fetta: shouldn't this be ran before publishToMavenLocal? or during the same gradle built?
```
# Install to maven
"$ROOT"/gradlew test -x javadoc publishToMavenLocal
```

mobuchowski: It should not really matter, as it's run during the same CI job  - will fail it in any case https://github.com/OpenLineage/OpenLineage/blob/68c96f9ae2ce1bc5ac21ba054bd29c8c1b02cec3/.circleci/continue_config.yml#L237

fetta: If the gradle publish completes, you're publishing an artifact that hasn't been tested. Yes, the CI would fail, but you can publish an incorrect maven package to a local directory. This can be easily avoided by reversing the order of operations.

mobuchowski: Makes sense, reversed the order

PR #3811, opened 2025-06-24
Spark: fix & upgrade databricks test

pawel-big-lebowski:


integration/spark/spark35/src/main/java/io/openlineage/spark35/agent/lifecycle/plan/CreateReplaceOutputDatasetBuilder.java
```diff
@@ -63,6 +63,7 @@ public boolean isDefinedAt(SparkListenerEvent event) {
 
   @Override
   protected List<OpenLineage.OutputDataset> apply(SparkListenerEvent event, LogicalPlan plan) {
+    log.info(String.format("Applying CreateReplaceOutputDatasetBuilder to %s", plan));
```
mobuchowski: `{}` instead of String.format for lazy formatting, also do we need info level here?

mobuchowski (approved):
One nit, besides looks good

integration/spark/spark35/src/main/java/io/openlineage/spark35/agent/lifecycle/plan/CreateReplaceOutputDatasetBuilder.java
pawel-big-lebowski: Will remove this line. It was added throughout the testing  with autocomplete. 
Thanks for spotting this!

Issue #3771, opened 2025-06-13
[FEATURE] Non-blocking Transport interface


dolfinus:
### Motivation

Currently all transports can emit just one event per `.emit(...)` method call. As almost all transports perform some kind of IO (sending data to a network, writing to a file) sending a batch of events requires making a request for first event, waiting for response, sending another one, and so on.

But some transports, like Kafka, have a natural way to send add a batch of events to queue, and yhem flush them in one operation. Probably Transport interface could be updated to support that.

### Description

Implement `BatchTransport` interface with methods like:

```java
interface BatchTransport {
  void emitBatch(List<RunEvent> batch);  // handle a batch of events & wait for response
  void emitBatch(List<DatasetEvent> batch);
  void emitBatch(List<JobEvent> batch);
}
```

or:
```java
interface BatchTransport {
  void addToBatch(RunEvent event);  // add event to batch/queue
  void addToBatch(DatasetEvent event);
  void addToBatch(JobEvent event);
  void sendBatch();  // block until batch/queue is handled
}
```

If some transport does support sending a batch of events, it can implement this interface in addition to Transport.

For queue-based transport, like Kafka, this is as natural as calling `producer.send(record)` + `producer.flush()`.

For HTTP it can be either:
* parallel sending requests to the same endpoint + using queue to keep user-defined order.
* enhancing HTTP API to send not a JSON object, but a list of JSON objects, in the same order as user passed.

### Related issues

#3770

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

dolfinus:
There can be overlaps between batch/queue based transport and `Task Queue based Async CircuitBreaker` feature of Java client.

mobuchowski:
That was my second idea on to how to solve this problem I had with dbt.

The issue is with compatibility - while async HTTP transport (given that integration sets it well) works with any HTTP backend, and if we updated the integrations (like airflow) to properly close the transport it can be a drop it replacement, the batch endpoint is something new and we don't have a way to know whether it's supported by the configured backend. 
Sending an event and falling back to sending one-by-one sync events if we receive 400-level response is not something that would be a good user experience.  

It can be an option or different http transport type, however. 

However with Kafka it makes a lot of sense and we should support that.

dolfinus:
> Sending an event and falling back to sending one-by-one sync events if we receive 400-level response is not something that would be a good user experience.

[...]

PR #3812, opened 2025-06-25
python: async HTTP transport v2

mobuchowski:
This is a followup to https://github.com/OpenLineage/OpenLineage/pull/3770 - opened as a new PR because I've made significant changes - make it as a separate `async_http` transport.

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -184,10 +184,36 @@ class OpenLineageClient:
         log.debug("OpenLineage event successfully emitted.")
 
     def shutdown(self, timeout: float = -1) -> bool:
-        return self.transport.shutdown(timeout)
+        """
+        Shutdown the transport, waiting for all events to complete or until the timeout is reached.
+
+        Params:
+            timeout: Timeout in seconds. `-1` means to block until last event is processed.
+
+        Returns:
+            bool: True if all events were processed before transport was closed,
+                False if some events were not processed.
+        """
+        result = self.wait_for_completion(timeout)
+        self.close()
+        return result
 
     def close(self) -> None:
-        self.transport.close()
+        """
+        Closes down the transport immediately, without waiting for the transport to process remaining events.
+        """
+        pass
+
+    def wait_for_completion(self, timeout: float = -1) -> bool:
+        """
+        Block until all events are processed or timeout is reached.
+        If the transport is fully synchronous, this method should be a no-op and return True.
+        Params:
+          timeout: Timeout in seconds. `-1` means to block until last event is processed.
+        Returns:
+            bool: True if all events were processed, False if some events were not processed.
+        """
+        return True
 
     @property
     def config(self) -> OpenLineageConfig:
@@ -305,50 +331,8 @@ class OpenLineageClient:
         return None
 
     def _http_transport_from_env_variables(self) -> HttpTransport:
-        """Create HTTP transport from legacy environment variables with modern config merging.
-
-        This method handles the transition from legacy OPENLINEAGE_* environment variables
-        to the newer OPENLINEAGE__TRANSPORT__* pattern. Legacy variables are used as the
-        base configuration and can be overridden or extended by the newer transport-specific
-        environment variables.
-
-        The legacy environment variables (OPENLINEAGE_URL, OPENLINEAGE_API_KEY, etc.) are
-        also aliased as the default HTTP transport configuration. This allows users to
-        seamlessly upgrade from legacy configs to modern transport configurations by simply
-        adding transport-specific environment variables.
-
-        Environment Variables Priority (higher priority overrides lower):
-        1. OPENLINEAGE__TRANSPORT__* (modern transport config)
-        2. OPENLINEAGE_* (legacy config used as base)
-
-        Example Use Cases:
-
-        1. Legacy configuration:
-           OPENLINEAGE_URL=http://openlineage-server:5000
-           OPENLINEAGE_API_KEY=secret-key
-           # Results in: standard HTTP transport
-
-        2. Upgrading to async transport:
-           OPENLINEAGE_URL=http://openlineage-server:5000  # legacy base config
-           OPENLINEAGE_API_KEY=secret-key                  # legacy auth
-           OPENLINEAGE__TRANSPORT__TYPE=async_http          # upgrade to async
-           OPENLINEAGE__TRANSPORT__MAX_CONCURRENT_REQUESTS=50  # async-specific config
-           # Results in: async HTTP transport with legacy URL/auth + async features
-
-        3. Override legacy endpoint:
-           OPENLINEAGE_URL=http://old-server:5000           # legacy URL
-           OPENLINEAGE__TRANSPORT__URL=http://new-server:8080  # override with new URL
-           # Results in: transport uses new-server:8080
-
-        Returns:
-            HttpTransport: Configured HTTP transport (could be standard or async based on
-                          merged configuration)
-
-        Note:
-            This method assumes OPENLINEAGE_URL is set (checked by caller). The resulting
-            transport type depends on the merged configuration - if type=async_http is
-            specified in OPENLINEAGE__TRANSPORT__*, an AsyncHttpTransport will be created
-            instead of the standard HttpTransport.
+        """
+        Create HTTP transport from legacy environment variables
         """
         # Start with basic config from legacy environment variables
         config_dict = {
@@ -359,12 +343,6 @@ class OpenLineageClient:
             },
             "endpoint": os.environ.get("OPENLINEAGE_ENDPOINT", "api/v1/lineage"),
         }
-
-        # Merge with any OPENLINEAGE__TRANSPORT__* environment variables
-        env_config = self._load_config_from_env_variables()
-        if env_config and "transport" in env_config:
-            config_dict = deep_merge_dicts(config_dict, env_config["transport"])
-
         return HttpTransport(HttpConfig.from_dict(config_dict))
 
     @staticmethod
@@ -396,40 +374,38 @@ class OpenLineageClient:
                 )
                 return
 
-            os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__TYPE"] = "http"
-
             api_key = os.environ.get("OPENLINEAGE_API_KEY")
             endpoint = os.environ.get("OPENLINEAGE_ENDPOINT")
 
-            if transport_type := os.environ.get("OPENLINEAGE__TRANSPORT__TYPE"):
+            os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__TYPE"] = "http"
+            os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__URL"] = url
+            if api_key:
+                os.environ[
+                    f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__AUTH"
+                ] = json.dumps(
+                    {
+                        "type": "api_key",
+                        "apiKey": api_key,
+                    }
+                )
+            if endpoint:
+                os.environ[
+                    f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__ENDPOINT"
+                ] = endpoint
+
+            if os.environ.get("OPENLINEAGE__TRANSPORT__TYPE") == "async_http":
                 # Special case - for seamless switch to async transport
-                if transport_type == "async_http":
-                    os.environ["OPENLINEAGE__TRANSPORT__TYPE"] = "async_http"
+                if not os.getenv("OPENLINEAGE__TRANSPORT__URL"):
                     os.environ["OPENLINEAGE__TRANSPORT__URL"] = url
-                    if api_key:
-                        os.environ["OPENLINEAGE__TRANSPORT__AUTH"] = json.dumps(
-                            {
-                                "type": "api_key",
-                                "apiKey": api_key,
-                            }
-                        )
-                    if endpoint:
-                        os.environ["OPENLINEAGE__TRANSPORT__ENDPOINT"] = endpoint
-                else:
-                    os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__URL"] = url
-                    if api_key:
-                        os.environ[
-                            f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__AUTH"
-                        ] = json.dumps(
-                            {
-                                "type": "api_key",
-                                "apiKey": api_key,
-                            }
-                        )
-                    if endpoint:
-                        os.environ[
-                            f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__ENDPOINT"
-                        ] = endpoint
+                if api_key and not os.getenv("OPENLINEAGE__TRANSPORT__AUTH"):
+                    os.environ["OPENLINEAGE__TRANSPORT__AUTH"] = json.dumps(
+                        {
+                            "type": "api_key",
+                            "apiKey": api_key,
+                        }
+                    )
+                if endpoint and not os.getenv("OPENLINEAGE__TRANSPORT__ENDPOINT"):
+                    os.environ["OPENLINEAGE__TRANSPORT__ENDPOINT"] = endpoint
 
     @classmethod
     def _load_config_from_env_variables(cls) -> dict[str, Any] | None:
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -28,7 +28,7 @@ from openlineage.client.transport import (
     TransportFactory,
     get_default_factory,
 )
-from openlineage.client.transport.http import HttpConfig, HttpTransport, create_token_provider
+from openlineage.client.transport.http import HttpConfig, HttpTransport
 from openlineage.client.transport.noop import NoopConfig, NoopTransport
 from openlineage.client.utils import deep_merge_dicts
 
@@ -183,6 +183,12 @@ class OpenLineageClient:
         self.transport.emit(event)
         log.debug("OpenLineage event successfully emitted.")
 
+    def shutdown(self, timeout: float = -1) -> bool:
+        return self.transport.shutdown(timeout)
+
+    def close(self) -> None:
+        self.transport.close()
+
     @property
     def config(self) -> OpenLineageConfig:
         """
@@ -238,7 +244,7 @@ class OpenLineageClient:
 
         # 2. Check if transport is provided explicitly
         if kwargs.get("transport"):
-            return cast(Transport, kwargs["transport"])
+            return cast("Transport", kwargs["transport"])
 
         # 3. Check if transport configuration is provided in YAML config file
         if self.config.transport and self.config.transport.get("type"):
@@ -292,28 +298,74 @@ class OpenLineageClient:
                     return path
                 if path and verbose:
                     log.debug("OpenLineage config file is missing or not readable: `%s`.", path)
-            except Exception:  # noqa: BLE001
+            except Exception:
                 # We can get different errors depending on system
                 if verbose:
                     log.exception("Couldn't check if OpenLineage config file is readable: `%s`", path)
         return None
 
-    @staticmethod
-    def _http_transport_from_env_variables() -> HttpTransport:
-        config = HttpConfig(
-            url=os.environ["OPENLINEAGE_URL"],
-            auth=create_token_provider(
-                {
-                    "type": "api_key",
-                    "apiKey": os.environ.get("OPENLINEAGE_API_KEY", ""),
-                },
-            ),
-        )
-        endpoint = os.environ.get("OPENLINEAGE_ENDPOINT", None)
-        if endpoint is not None:
-            config.endpoint = endpoint
+    def _http_transport_from_env_variables(self) -> HttpTransport:
+        """Create HTTP transport from legacy environment variables with modern config merging.
 
-        return HttpTransport(config)
+        This method handles the transition from legacy OPENLINEAGE_* environment variables
+        to the newer OPENLINEAGE__TRANSPORT__* pattern. Legacy variables are used as the
+        base configuration and can be overridden or extended by the newer transport-specific
+        environment variables.
+
+        The legacy environment variables (OPENLINEAGE_URL, OPENLINEAGE_API_KEY, etc.) are
+        also aliased as the default HTTP transport configuration. This allows users to
+        seamlessly upgrade from legacy configs to modern transport configurations by simply
+        adding transport-specific environment variables.
+
+        Environment Variables Priority (higher priority overrides lower):
+        1. OPENLINEAGE__TRANSPORT__* (modern transport config)
+        2. OPENLINEAGE_* (legacy config used as base)
+
+        Example Use Cases:
+
+        1. Legacy configuration:
+           OPENLINEAGE_URL=http://openlineage-server:5000
+           OPENLINEAGE_API_KEY=secret-key
+           # Results in: standard HTTP transport
+
+        2. Upgrading to async transport:
+           OPENLINEAGE_URL=http://openlineage-server:5000  # legacy base config
+           OPENLINEAGE_API_KEY=secret-key                  # legacy auth
+           OPENLINEAGE__TRANSPORT__TYPE=async_http          # upgrade to async
+           OPENLINEAGE__TRANSPORT__MAX_CONCURRENT_REQUESTS=50  # async-specific config
+           # Results in: async HTTP transport with legacy URL/auth + async features
+
+        3. Override legacy endpoint:
+           OPENLINEAGE_URL=http://old-server:5000           # legacy URL
+           OPENLINEAGE__TRANSPORT__URL=http://new-server:8080  # override with new URL
+           # Results in: transport uses new-server:8080
+
+        Returns:
+            HttpTransport: Configured HTTP transport (could be standard or async based on
+                          merged configuration)
+
+        Note:
+            This method assumes OPENLINEAGE_URL is set (checked by caller). The resulting
+            transport type depends on the merged configuration - if type=async_http is
+            specified in OPENLINEAGE__TRANSPORT__*, an AsyncHttpTransport will be created
+            instead of the standard HttpTransport.
+        """
+        # Start with basic config from legacy environment variables
+        config_dict = {
+            "url": os.environ["OPENLINEAGE_URL"],
+            "auth": {
+                "type": "api_key",
+                "apiKey": os.environ.get("OPENLINEAGE_API_KEY", ""),
+            },
+            "endpoint": os.environ.get("OPENLINEAGE_ENDPOINT", "api/v1/lineage"),
+        }
+
+        # Merge with any OPENLINEAGE__TRANSPORT__* environment variables
+        env_config = self._load_config_from_env_variables()
+        if env_config and "transport" in env_config:
+            config_dict = deep_merge_dicts(config_dict, env_config["transport"])
+
+        return HttpTransport(HttpConfig.from_dict(config_dict))
 
     @staticmethod
     def _http_transport_from_url(
@@ -343,21 +395,41 @@ class OpenLineageClient:
                     default_transport_name,
                 )
                 return
+
             os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__TYPE"] = "http"
-            os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__URL"] = url
-            if api_key := os.environ.get("OPENLINEAGE_API_KEY"):
-                os.environ[
-                    f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__AUTH"
-                ] = json.dumps(
-                    {
-                        "type": "api_key",
-                        "apiKey": api_key,
-                    }
-                )
-            if endpoint := os.environ.get("OPENLINEAGE_ENDPOINT"):
-                os.environ[
-                    f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__ENDPOINT"
-                ] = endpoint
+
+            api_key = os.environ.get("OPENLINEAGE_API_KEY")
+            endpoint = os.environ.get("OPENLINEAGE_ENDPOINT")
+
+            if transport_type := os.environ.get("OPENLINEAGE__TRANSPORT__TYPE"):
+                # Special case - for seamless switch to async transport
+                if transport_type == "async_http":
+                    os.environ["OPENLINEAGE__TRANSPORT__TYPE"] = "async_http"
+                    os.environ["OPENLINEAGE__TRANSPORT__URL"] = url
+                    if api_key:
+                        os.environ["OPENLINEAGE__TRANSPORT__AUTH"] = json.dumps(
+                            {
+                                "type": "api_key",
+                                "apiKey": api_key,
+                            }
+                        )
+                    if endpoint:
+                        os.environ["OPENLINEAGE__TRANSPORT__ENDPOINT"] = endpoint
+                else:
+                    os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__URL"] = url
+                    if api_key:
+                        os.environ[
+                            f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__AUTH"
+                        ] = json.dumps(
+                            {
+                                "type": "api_key",
+                                "apiKey": api_key,
+                            }
+                        )
+                    if endpoint:
+                        os.environ[
+                            f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__ENDPOINT"
+                        ] = endpoint
 
     @classmethod
     def _load_config_from_env_variables(cls) -> dict[str, Any] | None:
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -183,37 +183,25 @@ class OpenLineageClient:
         self.transport.emit(event)
         log.debug("OpenLineage event successfully emitted.")
 
-    def shutdown(self, timeout: float = -1) -> bool:
+    def close(self, timeout: float = -1.0) -> bool:
         """
-        Shutdown the transport, waiting for all events to complete or until the timeout is reached.
-
+        Closes down the transport until all events are processed or timeout is reached.
         Params:
-            timeout: Timeout in seconds. `-1` means to block until last event is processed.
+          timeout: Timeout in seconds. `-1` means to block until last event is processed, 0 means no timeout.
 
-        Returns:
-            bool: True if all events were processed before transport was closed,
-                False if some events were not processed.
         """
-        result = self.wait_for_completion(timeout)
-        self.close()
-        return result
+        return self.transport.close(timeout)
 
-    def close(self) -> None:
-        """
-        Closes down the transport immediately, without waiting for the transport to process remaining events.
-        """
-        pass
-
-    def wait_for_completion(self, timeout: float = -1) -> bool:
+    def wait_for_completion(self, timeout: float = -1.0) -> bool:
         """
         Block until all events are processed or timeout is reached.
         If the transport is fully synchronous, this method should be a no-op and return True.
         Params:
-          timeout: Timeout in seconds. `-1` means to block until last event is processed.
+          timeout: Timeout in seconds. `-1` means to block until last event is processed, 0 means no timeout.
         Returns:
             bool: True if all events were processed, False if some events were not processed.
         """
-        return True
+        return self.transport.wait_for_completion(timeout)
 
     @property
     def config(self) -> OpenLineageConfig:
--- a/client/python/openlineage/client/transport/async_http.py
+++ b/client/python/openlineage/client/transport/async_http.py
@@ -1,0 +1,551 @@
+# Copyright 2018-2025 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import asyncio
+import gzip
+import hashlib
+import logging
+import queue
+import threading
+import time
+from typing import TYPE_CHECKING, Any
+from urllib.parse import urljoin
+
+import attr
+import httpx
+from openlineage.client.event_v2 import RunEvent as RunEventV2
+from openlineage.client.run import RunEvent
+from openlineage.client.serde import Serde
+from openlineage.client.transport.http import HttpCompression, TokenProvider, create_token_provider
+from openlineage.client.transport.transport import Config, Transport
+from openlineage.client.utils import get_only_specified_fields
+
+if TYPE_CHECKING:
+    from openlineage.client.client import Event
+
+
+"""
+Async HTTP Transport for OpenLineage Events
+
+This module provides asynchronous HTTP transport mechanisms for sending
+OpenLineage events to remote endpoints using httpx. The AsyncHttpTransport implements a sophisticated
+event ordering and delivery system with race-condition-free task management.
+
+## Transport Architecture
+
+The AsyncHttpTransport provides:
+- **High-throughput asynchronous transport**: Non-blocking event emission
+- **Event ordering guarantees**: Ensures START events are sent before terminal events
+- **Concurrent processing**: Configurable parallel HTTP requests
+- **Advanced error handling**: Retry logic with exponential backoff
+
+## AsyncHttpTransport Architecture
+
+### 1. Event Ordering Guarantee
+- **Conditional START-before-completion ordering**: COMPLETE, FAIL, and ABORT events are held
+  only if their corresponding START event is currently in progress (pending submission)
+- **No START event**: If no START event was ever submitted, completion events are processed immediately
+- **Pending event queue**: Completion events are stored in memory only while their START event is pending
+- **Automatic release**: When a START event succeeds, all pending completion events for that run
+  are immediately queued for processing
+
+### 2. Concurrent Processing
+- **Bounded parallelism**: Configurable maximum concurrent HTTP requests (default: 100)
+- **Semaphore-controlled**: Uses asyncio.Semaphore to limit concurrent operations
+- **Task-based architecture**: Each event is processed as an independent asyncio task
+- **Race-condition-free**: Immediate task removal prevents double-processing of completed tasks
+- **Reserved queue capacity**: 2x configured queue size to accommodate released completion events
+
+### 3. Event Tracking and Statistics
+- **Simplified tracking**: Uses direct counters for pending, success, and failed events
+- **Real-time statistics**: Thread-safe access to event processing status
+- **Memory efficient**: Tracks only essential state without complex run metadata
+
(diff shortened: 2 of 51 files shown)


client/python/openlineage/client/transport/composite.py
```diff
@@ -83,10 +83,16 @@ def emit(self, event: Event) -> None:
...
                     log.debug("OpenLineage emission failure details:", exc_info=True)
                 else:
                     msg = f"Transport {transport} failed to emit event"
                     raise RuntimeError(msg) from e
+
+    def wait_for_completion(self, timeout: float = 10.0) -> bool:
+        # This can wait longer than timeout if multiple transports are slow, but acceptable
+        for transport in self.transports:
+            transport.wait_for_completion(timeout)
+        return True
```
kacpermuda: I think we should return a bool here that will be False if any of transports return False. We can also add logging in the loop if transport returns False. Right now, event if all transports return False - which means some events were lost, we return True.

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,628 @@
...
+import warnings
+from enum import Enum
+from typing import TYPE_CHECKING, Any
+from urllib.parse import urljoin
+
+import attr
+import httpx
+
+if TYPE_CHECKING:
+    from openlineage.client.client import Event, OpenLineageClientOptions
```
kacpermuda: Let's move type checking to the end of imports, it can break sorting.

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,628 @@
...
+
+        return self._all_processed()
+
+    def get_stats(self) -> dict[str, int]:
+        return self.event_stats
+
+    def shutdown(self, wait: bool = True, timeout: float = 30.0) -> bool:
+        if wait:
+            result = self.wait_for_completion(timeout)
+            self.should_exit.set()
```
dolfinus: should_exit should be moved to the beginning of the method, to notify threads about pending shutdown

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,628 @@
...
+        request = Request(event_id=event_id, run_id=run_id, event_type=event_type, body=body, headers=headers)
+        self._add_event(request)
+
+        # Wait for space in the regular capacity portion of the queue
+        while self.event_queue.qsize() >= self.configured_queue_size:
+            time.sleep(0.01)
+
+        # Add to queue - include run_id and event_type for START event tracking
+        self.event_queue.put(request)
+        log.debug(f"Queued event with event_id {event_id} for async emission to {self.config.url}")
```
dolfinus: Please use % formatting in log messages

mobuchowski: Fixed this and overwritten the changes with the rebase 😞

website/docs/client/python.md
```diff
@@ -218,7 +218,7 @@ environment variables:
 
 ### HTTP
 
-Allows sending events to HTTP endpoint, using [requests](https://requests.readthedocs.io/).
+Allows sending events to HTTP endpoint, using [httpx](https://www.python-httpx.org/). Supports both synchronous and asynchronous transport modes for different performance requirements.
```
dolfinus: Documentation should be updated to match HttpTransport vs. AsyncHttpTransport difference

client/python/openlineage/client/transport/async_http.py
mobuchowski: This being after `wait_for_completion` lets the transport process events from the queue - not only those actively being emitted. `should_exit` blocks the worker thread from getting next event from the queue.

I think with supplied timeout this behavior is desired, we can imagine some integration emitting event, then immediately calling shutdown with 10 second timeout - I would imagine this should try to emit the event.

client/python/openlineage/client/transport/transport.py
```diff
@@ -44,6 +44,16 @@ class Transport:
     def emit(self, event: Event) -> Any:
         raise NotImplementedError
 
+    def wait_for_completion(self, timeout: float = 10.0) -> bool:
```
dolfinus: Should `shutdown` be added here as well?

client/python/openlineage/client/transport/async_http.py
dolfinus: Ok, good point

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,628 @@
...
+
+            # Brief sleep before checking again
+            time.sleep(0.01)
+
+        return self._all_processed()
+
+    def get_stats(self) -> dict[str, int]:
+        return self.event_stats
+
+    def shutdown(self, wait: bool = True, timeout: float = 30.0) -> bool:
```
dolfinus: Should `wait=False` case be moved to a dedicated method, e.g. `.close()`?

```python
def shutdown(self, timeout: float = 30.0):
  self.wait_for_completion(timeout)
  self.close()

def close(self):
  self.should_exit.set()
  if self.worker_thread.is_alive():
    self.worker_thread.join()
```

mobuchowski: We can do that.

website/docs/client/python.md
mobuchowski: Done

client/python/openlineage/client/transport/async_http.py
mobuchowski: Done.

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -243,7 +243,7 @@ def on_starting(component):
 
 @hookimpl
 def before_stopping(component):
-    executor.shutdown(wait=False)
+    executor.close()
```
dolfinus: Probably a result of bulk text replacement, there is no such method in ThreadPoolExecutor

website/docs/client/python.md
```diff
@@ -216,9 +216,9 @@ environment variables:
 
 ## Built-in Transport Types
 
-### HTTP
+### HTTP Transport
 
-Allows sending events to HTTP endpoint, using [requests](https://requests.readthedocs.io/).
+The HTTP transport provides synchronous, blocking event emission using [httpx](https://www.python-httpx.org/). This is the default transport implementation suitable for most use cases where immediate event delivery and error handling are preferred.
```
dolfinus: Well, sync http transport still uses `requests`

website/docs/client/python.md
```diff
@@ -291,6 +291,120 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
 
 </Tabs>
 
+### Async HTTP Transport
+
+The Async HTTP transport provides high-performance, non-blocking event emission with advanced queuing and ordering guarantees. Use this transport when you need high throughput or want to avoid blocking your application on lineage event delivery.
+
+#### Configuration
+
+- `type` - string, must be `"openlineage.client.transport.async_http.AsyncHttpTransport"` or use direct instantiation. Required.
```
dolfinus: Shouldn't we have an alias `async_http`?

website/docs/client/python.md
```diff
@@ -291,6 +291,120 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+1. Events are queued for processing (START events immediately, completion events wait if their START is pending)
+2. Worker thread processes events using configurable parallelism
+3. Successful START events trigger release of pending completion events
+4. Event statistics are tracked and available via `get_stats()`
+
+#### Additional Methods
+
+- `wait_for_completion(timeout)` - Wait for all events to be processed
+- `get_stats()` - Get processing statistics (`{"pending": 0, "success": 10, "failed": 0}`)
+- `shutdown(wait, timeout)` - Graceful shutdown with optional wait
```
dolfinus:
```suggestion
- `shutdown(timeout: float)` - Graceful shutdown, wait untill pending events are send
- `close()` - Shutdown immediately and skip pending events
```

client/python/openlineage/client/transport/composite.py
```diff
@@ -83,10 +83,16 @@ def emit(self, event: Event) -> None:
...
-            except Exception as e:  # noqa: BLE001
+            except Exception as e:
                 if self.config.continue_on_failure:
                     log.warning("Transport %s failed to emit event with error: %s", transport, e)
                     log.debug("OpenLineage emission failure details:", exc_info=True)
                 else:
                     msg = f"Transport {transport} failed to emit event"
                     raise RuntimeError(msg) from e
+
+    def wait_for_completion(self, timeout: float = 10.0) -> bool:
```
dolfinus: `def close(self):` should be implemented too

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,630 @@
...
+                )
+            else:
+                log.error(
+                    "Failed to emit event (event_id) %s to %s",
+                    request.event_id,
+                    self.config.url,
+                )
+
+        async with semaphore:  # Limit concurrent requests
+            for attempt in range(self.config.retry.get("total", 5) + 1):
```
dolfinus: Let's either add a class-level constant with defaults for `total` and `status_forcelist`, or use `self.config.retry["total"]` syntax, as defaults are already set in config constructor

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,630 @@
...
+
+        # Wait for space in the regular capacity portion of the queue
+        while self.event_queue.qsize() >= self.configured_queue_size:
+            time.sleep(0.01)
+
+        # Add to queue - include run_id and event_type for START event tracking
+        self.event_queue.put(request)
+        log.debug("Queued event with event_id %s for async emission to %s", event_id, self.config.url)
+
+    def wait_for_completion(self, timeout: float = 30.0) -> bool:
```
dolfinus: Default queue size is 1 million events. Default timeout for completion is just 30 seconds. This is 30K RPS for just one-threaded worker, IMHO this is not very realistic, even for asyncio.

Maybe `timeout=-1` or `timeout=math.inf` should be an indicator for infinite timeout, and set as a default value? So users will be sure that all events were actually send.

website/docs/client/python.md
mobuchowski: Fixed the doc

website/docs/client/python.md
mobuchowski: Removed that part, what we use should be just an implementation detail.

integration/airflow/openlineage/airflow/listener.py
mobuchowski: Fixed.

website/docs/client/python.md
mobuchowski: Added that part.

client/python/openlineage/client/transport/async_http.py
mobuchowski: Went with `self.config.retry["total"]`

client/python/openlineage/client/transport/composite.py
dolfinus: Ok, can be resolved

client/python/openlineage/client/transport/transport.py
dolfinus: Ok, can be resolved

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,559 @@
...
+                            "pending completion events.",
+                            request.event_id,
+                            self.config.url,
+                            len(pending_events),
+                        )
+                        break
+
+                    if resp.status_code not in self.config.retry.get(
+                        "status_forcelist", [500, 502, 503, 504]
+                    ):
```
dolfinus:
```suggestion
                    if resp.status_code not in self.config.retry["status_forcelist"]:
```

client/python/openlineage/client/transport/async_http.py
dolfinus: @mobuchowski What do you think about making `-1` default value for all transport implementations?

client/python/openlineage/client/client.py
```diff
@@ -343,7 +353,17 @@ def _alias_env_vars(self) -> None:
...
             os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__TYPE"] = "http"
+
+            # Special case - for seamless switch to async transport
+            if transport_type := os.environ.get("OPENLINEAGE__TRANSPORT__TYPE"):
+                if transport_type == "async_http":
+                    os.environ["OPENLINEAGE__TRANSPORT__TYPE"] = "composite"
+                    os.environ[
+                        f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__TYPE"
+                    ] = "async_http"
+
```
kacpermuda: I don't understand this special case. What is the reasoning here?

client/python/openlineage/client/client.py
```diff
@@ -292,28 +298,32 @@ def _find_yaml_config_path() -> str | None:
...
-            url=os.environ["OPENLINEAGE_URL"],
-            auth=create_token_provider(
-                {
-                    "type": "api_key",
-                    "apiKey": os.environ.get("OPENLINEAGE_API_KEY", ""),
-                },
-            ),
-        )
+    def _http_transport_from_env_variables(self) -> HttpTransport:
+        # Start with basic config from legacy environment variables
```
kacpermuda, line 325: Can we add some docstring here, that the legacy variables are also aliased as default_http transport and this is why we are also trying to merge with other env var configs? Also maybe provide some example use case here, that if somebody is using legacy config with openlineage_url, we want to be able to wswitch to async_http with one more env var.

client/python/openlineage/client/client.py
```diff
@@ -292,28 +298,32 @@ def _find_yaml_config_path() -> str | None:
...
+        config_dict = {
+            "url": os.environ["OPENLINEAGE_URL"],
+            "auth": {
+                "type": "api_key",
+                "apiKey": os.environ.get("OPENLINEAGE_API_KEY", ""),
+            },
+        }
+
         endpoint = os.environ.get("OPENLINEAGE_ENDPOINT", None)
```
kacpermuda: Another case, since we're aliasing this env vars and there is very little change that somebody is setting both `OPENLINEAGE_URL` and `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP__URL`, maybe we should use this new env vars `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP...` here and the legacy ones would only stay to be aliased? Maybe not the scope of the PR, but since we're already merging configs here, I think we should make it more clear on why the `OPENLINEAGE_URL` config and `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP__ENDPOINT` may end up being used together.

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,559 @@
...
+from openlineage.client.event_v2 import RunEvent as RunEventV2
+from openlineage.client.run import RunEvent
+from openlineage.client.serde import Serde
+from openlineage.client.transport.transport import Config, Transport
+from openlineage.client.utils import get_only_specified_fields
+
+log = logging.getLogger(__name__)
+
+
+"""
```
kacpermuda, line 28: Can we set this nice explanation as module docstring and not string in the middle of the module?

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,559 @@
...
+import threading
+import time
+from typing import TYPE_CHECKING, Any
+from urllib.parse import urljoin
+
+import attr
+import httpx
+from openlineage.client.transport.http import HttpCompression, TokenProvider, create_token_provider
+
+if TYPE_CHECKING:
```
kacpermuda: Please move type checking to the end of imports, it's not automatically moved by ruff.

client/python/openlineage/client/transport/async_http.py
kacpermuda: +1 to that ^

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,559 @@
...
+
+        if isinstance(event, (RunEvent, RunEventV2)):
+            event_type = str(event.eventType.value)  # type: ignore[union-attr]
+            event_id = event.run.runId + "-" + event_type
+            run_id = event.run.runId
+
+            # We don't want to emit terminal events until we have a successful START event
+            # Only queue as pending if START has been scheduled but not yet successful
+            # If START was never scheduled, process COMPLETE normally (fall through)
+            if event_type in ["COMPLETE", "FAIL", "ABORT"]:
```
kacpermuda:
```suggestion
            if event_type != "START":
```
What about `RUNNING` events or `OTHER`, also allowed by spec? I think we should generally process START first and then all other

client/python/openlineage/client/transport/async_http.py
kacpermuda: +1 ^

client/python/tests/integration/test_server/test_server.py
kacpermuda, line 1: Awesome work on this server, will be very helpful for testing. ❤️

client/python/tests/test_client.py
```diff
@@ -604,6 +640,45 @@ def test_composite_transport_with_aliased_url_and_overriden_alias() -> None:
...
+    from openlineage.client.transport.async_http import AsyncHttpTransport
+
+    client = OpenLineageClient()
+    transport: AsyncHttpTransport = client.transport
+    try:
+        assert transport.kind == "async_http"
+    finally:
+        # Fast shutdown without waiting
+        transport.should_exit.set()
+        transport.worker_thread.join(timeout=0.1)
```
kacpermuda: Shouldn't `.close()` do that? If not, I think we should modify the close method and call it here.

client/python/tests/test_client.py
```diff
@@ -604,6 +640,45 @@ def test_composite_transport_with_aliased_url_and_overriden_alias() -> None:
...
+def test_async_transport_with_overwritten_enabled_async_flag() -> None:
+    from openlineage.client.transport.async_http import AsyncHttpTransport
+
+    client = OpenLineageClient()
+    transport: AsyncHttpTransport = client.transport
+    assert transport.kind == "composite"
+    assert transport.transports[0].kind == "async_http"
+    assert transport.transports[0].url == "http://example.com"
+
+
```
kacpermuda: Already asked about it above, I don't think this is how it's supposed to work. If you want async, you should probably set `OPENLINEAGE__TRANSPORT__TRANSPORTS__DEFAULT_HTTP_TYPE` to `async_http` and then check for that in `_http_transport_from_env_variables`.

client/python/tests/test_facet.py
```diff
@@ -58,9 +58,17 @@ def event() -> dict[str, Any]:
...
-    client = OpenLineageClient(url="http://example.com", session=session)
+@patch("openlineage.client.transport.http.Session")
+def test_symlink_dataset_facet(mock_client_class, event: dict[str, Any]) -> None:
+    # Mock the context manager and post method
+    mock_client = MagicMock()
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_client.post.return_value = mock_response
+    mock_client_class.return_value.__enter__.return_value = mock_client
+    mock_client_class.return_value.__exit__.return_value = None
```
kacpermuda: Should we move this logic to some fixture? It looks like it's copy/pasted in multiple places.

client/python/tests/test_http_integration.py
```diff
@@ -0,0 +1,470 @@
...
+    create_multiple_runs,
+    create_run_lifecycle_events,
+)
+
+
+@pytest.mark.integration
+class TestBasicEventFlow:
+    """Test basic event emission and reception."""
+
+    def test_sync_transport_single_event(self, sync_transport, server_helper):
```
kacpermuda, line 25: It would be great if we could add just two more tests for simple flow for sync http transport 🔥

kacpermuda (commented):
That's a really impressive PR 🚀  Added some comments, mostly about cosmetics and documentation, but also some questions about env vars setup special case.

client/python/openlineage/client/client.py
mobuchowski: You can specify `OPENLINEAGE_URL` with `OPENLINEAGE__TRANSPORT__TYPE=async_http` - which significantly improves ease of moving to async http transport for people who use shorthand http config env var notation.

client/python/openlineage/client/transport/async_http.py
mobuchowski: Done

client/python/openlineage/client/transport/composite.py
mobuchowski: You're right, fixed.

client/python/openlineage/client/transport/async_http.py
mobuchowski: This conflicts with out license declaration. IMO this is not useful for anyone other than someone working on it, and we have separate documentation in website.

client/python/openlineage/client/client.py
mobuchowski: Added endpoint there. I don't think it should be a special case.

client/python/tests/test_client.py
mobuchowski: >I don't think this is how it's supposed to work

I disagree, this is how I want it to work. Do you have a reason to make it harder to switch to the async_http transport?

kacpermuda: But is there a reason why we end up with composite transport with single inner transport instead of just this inner transport, without being wrapped in composite? This is the behavior that's not clear to me, user never chose to use composite transport in this scenario, they want async_http.

kacpermuda: \+ by default, composite transport do not raise errors on failure, the [doc](https://openlineage.io/docs/client/python#composite) is wrong, `continue_on_failure=True` was default from the start

client/python/openlineage/client/transport/composite.py
dolfinus: Probably commit wasn't pushed yet

client/python/openlineage/client/client.py
mobuchowski: Added detailed docstring

client/python/tests/test_facet.py
mobuchowski: Done

client/python/tests/test_http_integration.py
mobuchowski: Added two tests, but let's keep more advanced sync test use cases out of this PR, as it's giant enough

client/python/tests/test_client.py
mobuchowski: Changed this to do regular async_http transport. Code is more complex and behavior is the same (as async transport also does not raise errors on failure), but if you prefer that code organization I'm ok with it

client/python/tests/test_client.py
mobuchowski: Those tests are heavily mocked and using close would wait till processing ends. It's easier to just join the thread here.

client/python/openlineage/client/transport/async_http.py
mobuchowski: Done

client/python/openlineage/client/transport/async_http.py
mobuchowski: Done

client/python/openlineage/client/transport/async_http.py
mobuchowski: Done.

client/python/openlineage/client/transport/async_http.py
kacpermuda: Changed to -1, resolving.

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,551 @@
...
+        # Add to queue - include run_id and event_type for START event tracking
+        self.event_queue.put(request)
+        log.debug("Queued event with event_id %s for async emission to %s", event_id, self.config.url)
+
+    def wait_for_completion(self, timeout: float = -1.0) -> bool:
+        # Block until all events are processed or timeout is reached.
+        # Special -1 value indicates that wait_for_completion will wait till all events are processed.
+        log.debug("Waiting for events to be sent in wait_for_completion.")
+        start_time = time.time()
+        while time.time() - start_time < timeout or timeout == -1:
```
dolfinus:
```suggestion
        while timeout < 0 or time.time() - start_time < timeout:
```

client/python/openlineage/client/transport/transport.py
```diff
@@ -44,6 +44,30 @@ class Transport:
...
+
+    def wait_for_completion(self, timeout: float = -1) -> bool:
+        """
+        Block until all events are processed or timeout is reached.
+        If the transport is fully synchronous, this method should be a no-op and return True.
+
+        Returns:
+            bool: True if all events were processed, False if some events were not processed
+        """
+        return True
```
dolfinus:
```suggestion
    def shutdown(self, timeout: float = -1) -> bool:
        """
        Shutdown the transport, waiting for all events to complete or untill the timeout is reached.

        Params:
          timeout: Timeout in seconds. `-1` means to block until last event is processed.

        Returns:
            bool: True if all events were processed before transport was closed, False if some events were not processed.
        """
        result = self.wait_for_completion(timeout)
        self.close()
        return result

    def close(self) -> None:
        """
        Closes down the transport immediately, without waiting for the transport to process remaining events.
        """
        pass

    def wait_for_completion(self, timeout: float = -1) -> bool:
        """
        Block until all events are processed or timeout is reached.
        If the transport is fully synchronous, this method should be a no-op and return True.

        Params:
          timeout: Timeout in seconds. `-1` means to block until last event is processed.

        Returns:
            bool: True if all events were processed, False if some events were not processed.
        """
        return True
```

client/python/openlineage/client/client.py
```diff
@@ -183,6 +183,12 @@ def emit(self, event: Event) -> None:
         self.transport.emit(event)
         log.debug("OpenLineage event successfully emitted.")
 
+    def shutdown(self, timeout: float = -1) -> bool:
+        return self.transport.shutdown(timeout)
+
+    def close(self) -> None:
+        self.transport.close()
```
dolfinus:
```suggestion
    def shutdown(self, timeout: float = -1) -> bool:
        """
        Shutdown the underlying transport, waiting for all events to complete or utill the timeout is reached.

        Params:
          timeout: Timeout in seconds. `-1` means to block until last event is processed.

        Returns:
            bool: True if all events were processed before transport was closed, False if some events were not processed.
        """
        return self.transport.shutdown(timeout)

    def close(self) -> None:
        """
        Closes down the underlying transport immediately, without waiting for the transport to process remaining events.
        """
        self.transport.close()
```

client/python/tests/test_client.py
```diff
@@ -604,6 +619,48 @@ def test_composite_transport_with_aliased_url_and_overriden_alias() -> None:
...
 
 
+@patch.dict(
+    os.environ,
+    {
+        "OPENLINEAGE_URL": "http://example.com",
+        "OPENLINEAGE__TRANSPORT": '{"type": "async_http", "url": "https://data-obs-intake.datadoghq.com", '
+        '"auth": {"type": "apiKey", "apiKey": "YOUR_API_KEY"}, '
+        '"async_config": {"enabled": true}}',
+    },
```
kacpermuda: What is this `async_config` doing? I don't see this option anywhere. Is it a leftover from then there was one config for both sync and async http transport?

client/python/openlineage/client/client.py
```diff
@@ -343,21 +395,41 @@ def _alias_env_vars(self) -> None:
                     default_transport_name,
                 )
                 return
+
```
kacpermuda, line 364: I think the whole function should look like below. We should always alias the legacy variables as `DEFAULT_HTTP` transport, and we should only add `OPENLINEAGE__TRANSPORT__URL`, api_key and endpoint if they are not already set. This ensures that we're not overwriting user modern config with legacy config.

With current approach, the below setup sends data to `example.com`:
```
os.environ["OPENLINEAGE_URL"] = "http://example.com"
os.environ["OPENLINEAGE__TRANSPORT__TYPE"] = "async_http"
os.environ["OPENLINEAGE__TRANSPORT__URL"] = "http://this.should.have.priority.com"
```

Full function code:
```
    def _alias_env_vars(self) -> None:
        default_transport_name = self.DEFAULT_URL_TRANSPORT_NAME.upper()
        if url := os.environ.get("OPENLINEAGE_URL"):
            if any(
                k.startswith(f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}")
                for k in os.environ
            ):
                log.warning(
                    "%s already found in environment variables, skipping aliasing OPENLINEAGE_URL",
                    default_transport_name,
                )
                return

            api_key = os.environ.get("OPENLINEAGE_API_KEY")
            endpoint = os.environ.get("OPENLINEAGE_ENDPOINT")

            os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__TYPE"] = "http"
            os.environ[f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__URL"] = url
            if api_key:
                os.environ[
                    f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__AUTH"
                ] = json.dumps(
                    {
                        "type": "api_key",
                        "apiKey": api_key,
                    }
                )
            if endpoint:
                os.environ[
                    f"OPENLINEAGE__TRANSPORT__TRANSPORTS__{default_transport_name}__ENDPOINT"
                ] = endpoint

            if os.environ.get("OPENLINEAGE__TRANSPORT__TYPE") == "async_http":
                # Special case - for seamless switch to async transport
                if not os.getenv("OPENLINEAGE__TRANSPORT__URL"):
                    os.environ["OPENLINEAGE__TRANSPORT__URL"] = url
                if api_key and not os.getenv("OPENLINEAGE__TRANSPORT__AUTH"):
                    os.environ["OPENLINEAGE__TRANSPORT__AUTH"] = json.dumps(
                        {
                            "type": "api_key",
                            "apiKey": api_key,
                        }
                    )
                if endpoint and not os.getenv("OPENLINEAGE__TRANSPORT__ENDPOINT"):
                    os.environ["OPENLINEAGE__TRANSPORT__ENDPOINT"] = endpoint
```

client/python/openlineage/client/client.py
```diff
@@ -292,28 +298,74 @@ def _find_yaml_config_path() -> str | None:
...
+                "apiKey": os.environ.get("OPENLINEAGE_API_KEY", ""),
+            },
+            "endpoint": os.environ.get("OPENLINEAGE_ENDPOINT", "api/v1/lineage"),
+        }
+
+        # Merge with any OPENLINEAGE__TRANSPORT__* environment variables
+        env_config = self._load_config_from_env_variables()
+        if env_config and "transport" in env_config:
+            config_dict = deep_merge_dicts(config_dict, env_config["transport"])
+
```
kacpermuda: I think we don't need that anymore. This function no longer returns async http transport, so there is no need to "enrich" http config.

If we apply the changes I proposed to the `_alias_env_vars` below, there will be an easy way to switch to http async transport with these legacy env vars.

client/python/openlineage/client/client.py
```diff
@@ -292,28 +298,74 @@ def _find_yaml_config_path() -> str | None:
...
+        Returns:
+            HttpTransport: Configured HTTP transport (could be standard or async based on
+                          merged configuration)
+
+        Note:
+            This method assumes OPENLINEAGE_URL is set (checked by caller). The resulting
+            transport type depends on the merged configuration - if type=async_http is
+            specified in OPENLINEAGE__TRANSPORT__*, an AsyncHttpTransport will be created
+            instead of the standard HttpTransport.
```
kacpermuda: this docstring is outdated, the code does not return asyn http anymore.

client/python/openlineage/client/client.py
kacpermuda: Also, these new `OPENLINEAGE__TRANSPORT__URL` env variables that we add will be checked in `_resolve_transport` before we even get to `OPENLINEAGE_URL`, so there is no need to modify the `_http_transport_from_env_variables` in any way.

client/python/tests/test_client.py
```diff
@@ -604,6 +619,48 @@ def test_composite_transport_with_aliased_url_and_overriden_alias() -> None:
...
+
+@patch.dict(
+    os.environ,
+    {
+        "OPENLINEAGE_URL": "http://example.com",
+        "OPENLINEAGE_ENDPOINT": "api/v2/lineage",
+        "OPENLINEAGE_API_KEY": "YOUR_API_KEY",
+        "OPENLINEAGE__TRANSPORT__TYPE": "async_http",
+    },
+)
```
kacpermuda, line 648: I mentioned it in the other comment, but I think we should add tests that makes sure this setup is treated correctly:

```
os.environ["OPENLINEAGE_URL"] = "http://example.com"
os.environ["OPENLINEAGE__TRANSPORT__TYPE"] = "async_http"
os.environ["OPENLINEAGE__TRANSPORT__URL"] = "http://this.should.have.priority.com"
```

kacpermuda (commented):
Left few more notes. I totally understand the need to provide users with `easy` way to switch to `async_http` and configure it, but I think we also need to make sure it's only hapenning when expected and there is no chance to overwrite modern setup in any way. With that in mind, I've added some comments to how the env variables are aliased and how legacy http transport is created.

client/python/openlineage/client/client.py
kacpermuda: Let's move it to the conversations below, they are more up to date.

client/python/openlineage/client/client.py
kacpermuda: The if/else you proposed is also fine, the key part is checking before overwriting:
```
if not os.getenv("OPENLINEAGE__TRANSPORT__URL"):
                    os.environ["OPENLINEAGE__TRANSPORT__URL"] = url
```

client/python/openlineage/client/client.py
kacpermuda: We can probably skip any changes to this func

client/python/openlineage/client/transport/async_http.py
```diff
@@ -0,0 +1,551 @@
...
+            time.sleep(0.01)
+
+        # Add to queue - include run_id and event_type for START event tracking
+        self.event_queue.put(request)
+        log.debug("Queued event with event_id %s for async emission to %s", event_id, self.config.url)
+
+    def wait_for_completion(self, timeout: float = -1.0) -> bool:
+        # Block until all events are processed or timeout is reached.
+        # Special -1 value indicates that wait_for_completion will wait till all events are processed.
+        log.debug("Waiting for events to be sent in wait_for_completion.")
```
dolfinus:
```suggestion
        log.debug("Waiting for events completion for %.3f seconds", timeout)
```

client/python/tests/test_client.py
mobuchowski: Yeah an artifact

JDarDagran (approved):
Great work @mobuchowski 🚀

website/docs/client/python.md
```diff
@@ -291,6 +291,121 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+
+- `type` - string, must be `"async_http"` or use direct instantiation. Required.
+- `url` - string, base url for HTTP requests. Required.
+- `endpoint` - string specifying the endpoint to which events are sent, appended to `url`. Optional, default: `api/v1/lineage`.
+- `timeout` - float specifying timeout (in seconds) value used while connecting to server. Optional, default: `5`.
+- `verify` - boolean specifying whether the client should verify TLS certificates from the backend. Optional, default: `true`.
+- `auth` - dictionary specifying authentication options. Optional, by default no authorization is used. If set, requires the `type` property.
+  - `type` - string specifying the "api_key" or the fully qualified class name of your TokenProvider. Required if `auth` is provided.
+  - `apiKey` - string setting the Authentication HTTP header as the Bearer. Required if `type` is `api_key`.
+- `compression` - string, name of algorithm used by HTTP client to compress request body. Optional, default value `null`, allowed values: `gzip`. Added in v1.13.0.
```
dolfinus:
```suggestion
- `compression` - string, name of algorithm used by HTTP client to compress request body. Optional, default value `null`, allowed values: `gzip`.
```

website/docs/client/python.md
```diff
@@ -291,6 +291,121 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+- `custom_headers` - dictionary of additional headers to be sent with each request. Optional, default: `{}`.
+- `max_queue_size` - integer specifying maximum events in processing queue. Optional, default: `1000000`.
```
dolfinus:
```suggestion
- `max_queue_size` - integer specifying maximum events in processing queue. Optional, default: `10000`.
```
Please also check documentation below, it uses old queue size

website/docs/client/python.md
```diff
@@ -291,6 +291,121 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+
+- **Event Ordering Guarantees**: START events are sent before their corresponding COMPLETE, FAIL, or ABORT events
+- **High Throughput**: Non-blocking event emission with configurable concurrent processing
+- **Queue Management**: Bounded queue prevents memory exhaustion with configurable size
+- **Advanced Error Handling**: Retry logic with exponential backoff for network and server errors
+- **Event Tracking**: Real-time statistics on pending, successful, and failed events
+
+#### Event Flow
+
+1. Events are queued for processing (START events immediately, completion events wait if their START is pending)
```
dolfinus:
```suggestion
1. Events are queued for processing (START events immediately, other events wait until corresponding START event is send)
```

website/docs/client/python.md
```diff
@@ -291,6 +291,121 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+
+# Wait for all events to complete
+client.transport.wait_for_completion(timeout=30)
+
+# Get processing statistics  
+stats = client.transport.get_stats()
+print(f"Pending: {stats['pending']}, Success: {stats['success']}, Failed: {stats['failed']}")
+
+# Graceful shutdown
+client.transport.shutdown(timeout=30)
```
dolfinus:
```suggestion
# Wait for all events to complete
client.wait_for_completion()

# Get processing statistics
stats = client.transport.get_stats()
print(f"Pending: {stats['pending']}, Success: {stats['success']}, Failed: {stats['failed']}")

# Graceful shutdown
client.shutdown()
```

dolfinus (approved):
Great work @mobuchowski

website/docs/client/python.md
mobuchowski: Done

website/docs/client/python.md
mobuchowski: Done

website/docs/client/python.md
mobuchowski: Done

website/docs/client/python.md
mobuchowski: Done

client/python/openlineage/client/transport/composite.py
```diff
@@ -83,10 +83,18 @@ def emit(self, event: Event) -> None:
...
                     log.debug("OpenLineage emission failure details:", exc_info=True)
                 else:
                     msg = f"Transport {transport} failed to emit event"
                     raise RuntimeError(msg) from e
+
+    def wait_for_completion(self, timeout: float = -1.0) -> bool:
+        # This can wait longer than timeout if multiple transports are slow, but acceptable
+        return all(transport.wait_for_completion(timeout) for transport in self.transports)
+
+    def close(self) -> None:
```
dolfinus:
```suggestion
    def shutdown(self, timeout: float = -1) -> bool:
        # This can wait longer than timeout if multiple transports are slow, but acceptable
        return all(transport.shutdown(timeout) for transport in self.transports)
        
    def wait_for_completion(self, timeout: float = -1) -> bool:
        return all(transport.wait_for_completion(timeout) for transport in self.transports)

    def close(self) -> None:
```
If some transport has custom `.shutdown()` method, it should be called.

Also TransformTransport should call `.shutdown()`, `.wait_for_completion()` and `.close()` of underlying transport as well.

dolfinus: Also, both shutdown and close should be more reliable - if some of underlying transport fails to close, then `.shutdown()` will not be called for the rest of them:
```python
def shutdown(self, timeout: float = -1):
  result = True
  last_exception: Exception | None = None
  for transport in self.transports:
    try:
      result = transport.shutdown(timeout) and result
    except Exception as e:
      logger.exception("Error while shutdown transport %s", transport)
      last_exception = e

  if last_exception:
    raise exception
  return result
```

dolfinus: This could be changed in another PR, thought

dolfinus:
After a thought, could we merge `.shutdown()` and `.close()` back into one method?
```python
class Transport:
  def wait_for_completion(self, timeout: float = -1) -> bool: ...

  def close(self, timeout: float = -1) -> None:
    events_not_send = self.wait_for_completion(timeout)
    self._close()
    if events_not_send:
        raise RuntimeException("Transport %s was closed forcely after timeout %.3f", self, timeout)

  def _close(self): ...  # actual closing


class OpenLineageClient:
  def close(self, timeout: float = -1) -> None:
    self.transport.close(timeout)
```

With this implementation users may use code like this for graceful shutdown:
```python
from contextlib import closing

with closing(transport):
  ...  # .close() will be executed automatically, with waiting for completion

with closing(client):
  ...
```

and code like this to immediate shutdown:
```python
transport.close(timeout=0)
client.close(timeout=0:
```

And also implementation will be similar with Java client there `void close()` is a part of Transport abstract class for some time already.

UPD: maybe `timeout` arg can be removed as well: https://github.com/OpenLineage/OpenLineage/pull/3839#issuecomment-3027021071

dolfinus:
IMHO this PR should be split to 3 ones:
* linter config update & fix applied
* changes to Transport interface for #3771
* AsyncHttpTransport

Most of my questions above are related to (2)

client/python/openlineage/client/transport/composite.py
mobuchowski: Fixed this for `close`.

mobuchowski:
@dolfinus removed separate shutdown. With negative timeout logic it makes sense to have a single method. 

I'm not okay with splitting the PR at this moment - with second PR, few rounds of reviews and receiving approvals - including yours. If it's too large to review, it should be indicated at the start. I'm not going to take offence as I'm working on this repo a lot, but imagine totally new external contributor getting that request after few weeks of trying to get the PR merged - most likely they'd drop trying to.

I've added a note in the website docs that

```
Async transport API is experimental, and can change over the next few releases.
```

If we have a good reason to modify this API, we can still do.

client/python/openlineage/client/transport/composite.py
```diff
@@ -83,10 +83,28 @@ def emit(self, event: Event) -> None:
...
                     log.debug("OpenLineage emission failure details:", exc_info=True)
                 else:
                     msg = f"Transport {transport} failed to emit event"
                     raise RuntimeError(msg) from e
+
+    def wait_for_completion(self, timeout: float = -1.0) -> bool:
+        # This can wait longer than timeout if multiple transports are slow, but acceptable
+        return all(transport.wait_for_completion(timeout) for transport in self.transports)
+
+    def shutdown(self, timeout: float = -1.0):
```
dolfinus:
```suggestion
    def close(self, timeout: float = -1.0):
```

mobuchowski: Fixed and added tests for this.

Issue #3727, opened 2025-05-29
[PROPOSAL]Formalize dataset naming and develop tooling for creation of names


ddebowczyk92:
**Purpose:**
Although dataset naming is documented in the [OpenLineage specification](https://openlineage.io/docs/spec/naming/#dataset-naming), there is currently no tooling or process to enforce or encourage its usage. As a result, producers may emit events with dataset names that do not comply with naming conventions. This can cause inconsistencies or issues on the consumer side, leading to frustration and potentially undermining OpenLineage’s reputation as the leading lineage solution.

This proposal aims to address that gap with a practical and extensible solution.

**Proposed Implementation:**
The proposed solution focuses on automatic code generation using the already defined dataset naming patterns. These naming formats are listed in the [dataset naming table](https://openlineage.io/docs/spec/naming/#dataset-naming), and we can generate code directly from them — no new definitions are introduced.

For each naming definition, we parse the template using a custom [ANTLR4 grammar](https://gist.github.com/ddebowczyk92/c7a58eb1733419030f6ee931cc3300b1#file-namingtemplate-g4), and then [generate](https://gist.github.com/ddebowczyk92/c7a58eb1733419030f6ee931cc3300b1#file-openlineagenaminggenerator-java) methods that construct valid dataset names.

The [example implementation](https://gist.github.com/ddebowczyk92/c7a58eb1733419030f6ee931cc3300b1#file-openlineagedatasetnaming-java) produces a single class with multiple static methods. For production use, this approach can be evolved into a more modular and extensible system. A key advantage is that it integrates well with the existing OpenLineage client code generation process.

**Alternative:**
An alternative would be to manually implement naming classes and methods for each dataset type. However, this would require manual effort every time a new dataset type is added and may not scale as easily. Automated code generation provides a more maintainable and future-proof approach.

kacpermuda:
Note: it would be good to remember and keep the naming convention source file in a single place and format, where it could be used for automated generation by both python and java client, to make sure they are consistent :) F.e. in python, facets are now generated from spec directory, so maybe we can also keep this naming convention somewhere in "shared" space

mobuchowski:
Can we just get the static class with generated name/namespace? I think using grammar for that task is overkill and introduces way more complexity than it solves

ddebowczyk92:
[...]

PR #3816, opened 2025-06-27
Formalize dataset naming for Python client

ddebowczyk92:
### Problem

Addresses https://github.com/OpenLineage/OpenLineage/issues/3727 but in Python

### Solution

### Checklist

----


client/python/openlineage/client/dataset.py
```diff
@@ -0,0 +1,499 @@
...
+    _: object, attribute: attr.Attribute[PubSubResourceType], value: PubSubResourceType
+) -> PubSubResourceType:
+    """Validator to check if an enum value is not None."""
+    if value is None:
+        msg = f"{attribute.name} cannot be None"
+        raise ValueError(msg)
+    return value
+
+
+class DatasetNaming(ABC):
```
kacpermuda: Do we need the ABC as base here and mixing it with `attr.define` later?  Or we can just inherit from class that will raise NotImplementedError?

Other option: just define the Protocol and not inherit, like:
```
from typing import Protocol
import attr

class DatasetNaming(Protocol):
    def get_namespace(self) -> str:
        ...

    def get_name(self) -> str:
        ...

@attr.define
class Athena:
    region_name: str = attr.field(validator=_check_not_empty)
    catalog: str = attr.field(validator=_check_not_empty)
    database: str = attr.field(validator=_check_not_empty)
    table: str = attr.field(validator=_check_not_empty)

    def get_namespace(self) -> str:
        return f"awsathena://athena.{self.region_name}.amazonaws.com"

    def get_name(self) -> str:
        return f"{self.catalog}.{self.database}.{self.table}"
```

Not saying it's wrong approach, just wondering. Personally I'd not expect attr class to carry some abstract obligations, but the code works and is valid 😄
cc @mobuchowski @JDarDagran WDYT?

client/python/openlineage/client/dataset.py
```diff
@@ -0,0 +1,499 @@
...
+    """Naming implementation for Snowflake."""
+
+    organization_name: str = attr.field(validator=_check_not_empty)
+    account_name: str = attr.field(validator=_check_not_empty)
+    database: str = attr.field(validator=_check_not_empty)
+    schema: str = attr.field(validator=_check_not_empty)
+    table: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return f"snowflake://{self.organization_name}-{self.account_name}"
```
kacpermuda: I know that it's just implementation of our naming convention, but there are cases where the same object can be identified with multiple identifiers. How do we want to approach it here with class naming, if we know that something like this exists? F.e. for [Snowflake](https://docs.snowflake.com/en/user-guide/admin-account-identifier#format-1-preferred-account-name-in-your-organization), the identifier can be `org_name-acc_name` but can also be `acc_locator.acc_region`. Should we create separate classes of Snowflake and SnowflakeLocator/SnowflakeLegacy etc. ? I think for Redshift it's the same case, multiple ways of connecting to the same cluster with either cluster_id, worskpace_id and something else.

client/python/tests/test_dataset.py
```diff
@@ -0,0 +1,276 @@
...
+    WASBS,
+    Athena,
+    AWSGlue,
+    AzureCosmosDB,
+    AzureDataExplorer,
+    AzureSynapse,
+    BigQuery,
+    Cassandra,
+    CrateDB,
+    Glue,
```
kacpermuda: This is breaking Python tests, this class is not there.

client/python/openlineage/client/dataset.py
```diff
@@ -0,0 +1,499 @@
...
+        return f"{self.database}.{self.table}"
+
+
+@attr.define
+class Redshift(DatasetNaming):
+    """Naming implementation for Redshift."""
+
+    cluster_identifier: str = attr.field(validator=_check_not_empty)
+    region: str = attr.field(validator=_check_not_empty)
+    port: str = attr.field(validator=_check_not_empty)
```
kacpermuda: Should we add default ports for known dbs?

client/python/openlineage/client/dataset.py
```diff
@@ -0,0 +1,499 @@
...
+
+@attr.define
+class GCS(DatasetNaming):
+    """Naming implementation for GCS."""
+
+    bucket_name: str = attr.field(validator=_check_not_empty)
+    object_key: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return f"gs://{self.bucket_name}"
```
kacpermuda: Maybe for separate PR, maybe not, but how about we extend this implementation with some additional validation? I think this implementation is helpful, but most of the time we have some discrepancies between how namespace and name are provided, it's now wrong scheme, it's something like different approach for leading/trailing slashes or wildcard (*) removal. F.e. in GCS or S3: should the object_key start with a slash or not? Should namespace end with slash or not? Should we remove wildcard and trim the path ? These are the problems I thought this would solve - maybe it's wrong assumption on my side, but do you plan to implement this as well?

kacpermuda (commented):
Leaving some thoughts, probably for open discussion. Also, CI needs fixing, the tests are failing.

client/python/openlineage/client/dataset.py
JDarDagran: I like the idea of using `Protocol`. This will avoid the overhead and rigidity of ABCs indeed.

client/python/openlineage/client/dataset.py
JDarDagran: Are there _better_ and _worse_ (that's probably wrong statement but don't know how to express that intuitively) identifiers? Or default ones and legacy/deprecated?
If not, I think we just should have separate classes that handle different scenarios.

client/python/openlineage/client/dataset.py
JDarDagran: We don't have them in https://openlineage.io/docs/spec/naming
I think that the purpose of formalizing dataset naming into Python/Java code should stay the same. We shouldn't aim in guessing what's the default port, that should be integration's task.

client/python/openlineage/client/dataset.py
JDarDagran: Adding a validation or normalization layer for paths is generally a good idea for consistency, but there are some of my concerns:

1. Should this be enforced or optional? Since current dataset naming conventions don’t specify things like slashes or wildcards, some users may have workflows that rely on their own conventions. Forcing normalization could discourage adoption of these classes.
2. Are we confident about the “correct” approach? For example, should wildcards always be removed from GCS paths, or are there cases where they’re needed? There isn’t always a clear standard probably.

An additional concern is that strict normalization could break existing integrations or introduce confusion if not clearly documented.

client/python/openlineage/client/dataset.py
mobuchowski: +1 for protocol

client/python/openlineage/client/dataset.py
mobuchowski: In general, this is unsolvable without DB connection to Snowflake. 

>Are there better and worse (that's probably wrong statement but don't know how to express that intuitively) identifiers? Or default ones and legacy/deprecated?

Not really, because Snowflake's account id naming is a big mess. There are new and old account ids, but the old accounts with acc_locator.acc_region don't have default org so you can use org_name-acc_name...

kacpermuda: Why would we need to connect to Snowflake here? Not sure I follow. We know that there are two different identifiers for the same thing, so we can prepare our naming convention to use both. Maybe Snowflake is not a good example here, as they have `preferred` and `old` identifiers, so we can just go with preferred for now, but there can be other similar cases. I think just separate classes are ok here.

client/python/openlineage/client/dataset.py
mobuchowski: Normalizing slashes yes, wildcards no. 
Wildcards aren't a feature of S3 paths itself, but shells or other systems using S3/GCS etc. 
Integrations should take care of normalizing wildcards, since they have the context of what the wildcard actually does.

client/python/openlineage/client/dataset.py
JDarDagran: > In general, this is unsolvable without DB connection to Snowflake.

This is solvable by having more than one convention per service type :)

client/python/openlineage/client/dataset.py
JDarDagran: Should it be on general level then? Or per class?

client/python/openlineage/client/dataset.py
```diff
@@ -0,0 +1,520 @@
...
+@attr.define
+class Kafka(DatasetNaming):
+    """Naming implementation for Kafka."""
+
+    bootstrap_server_host: str = attr.field(validator=_check_not_empty)
+    port: str = attr.field(validator=_check_not_empty)
+    topic: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return f"kafka://{self.bootstrap_server_host}:{self.port}"
```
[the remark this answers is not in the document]
kacpermuda: @ddebowczyk92 WDYT? Ih shorter `host` clear enough for Kafka?

client/python/openlineage/client/dataset.py
kacpermuda: Ok, we can provide separate classes in case there are different set of `args` needed for creation of identifier, like in the Snowflake case. Resolving this.

kacpermuda:
Can we also move this module to a separate directory like `openlineage.client.naming.dataset` ? I think it's good to future proof it, in case we do something similar for the job etc.

client/python/openlineage/client/dataset.py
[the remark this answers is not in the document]
ddebowczyk92: I'd stick with that as it is closer to the generally accepted convention regarding naming in Kafka

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,497 @@
...
+
+    Namespace is always "bigquery".
+    """
+
+    project_id: str = attr.field(validator=_check_not_empty)
+    dataset_name: str = attr.field(validator=_check_not_empty)
+    table_name: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return "bigquery"
```
kacpermuda, line 152: Is this a good place and moment to impose `bigquery://` ? I remember we wanted to make it uniform for all, and bq and file are the only exceptions I think. @JDarDagran @mobuchowski

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,497 @@
...
+    database: str = attr.field(validator=_check_not_empty)
+    schema: str = attr.field(validator=_check_not_empty)
+    table: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return f"db2://{self.host}:{self.port}"
+
+    def get_name(self) -> str:
+        return f"{self.database}.{self.schema}.{self.table}"
+
```
kacpermuda, line 222: I'm going side-to-side with naming convention [page](https://openlineage.io/docs/spec/naming) and Hive is missing. Is it intentional?

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,497 @@
...
+
+@attr.define
+class LocalFileSystem(DatasetNaming):
+    """Naming implementation for Local File System."""
+
+    path: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return "file"
```
kacpermuda, line 438: The same, should we go for: `file://` ?

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,497 @@
...
+    - topic:{project_id}:{topic_id}
+    - subscription:{project_id}:{subscription_id}
+    """
+
+    resource_type: PubSubResourceType = attr.field(validator=_check_enum_not_none)
+    project_id: str = attr.field(validator=_check_not_empty)
+    resource_id: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return "pubsub"
```
kacpermuda, line 509: `pubsub://` ?

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,497 @@
...
+
+
+def _check_enum_not_none(
+    _: object, attribute: attr.Attribute[PubSubResourceType], value: PubSubResourceType
+) -> PubSubResourceType:
+    """Validator to check if an enum value is not None."""
+    if value is None:
+        msg = f"{attribute.name} cannot be None"
+        raise ValueError(msg)
+    return value
```
kacpermuda: Can we return enum validation?

```
return PubSubResourceType(value)
```

If somebody passes some value unsupported by enum, it will also raise ValueError

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,497 @@
...
+    resource_type: PubSubResourceType = attr.field(validator=_check_enum_not_none)
+    project_id: str = attr.field(validator=_check_not_empty)
+    resource_id: str = attr.field(validator=_check_not_empty)
+
+    def get_namespace(self) -> str:
+        return "pubsub"
+
+    def get_name(self) -> str:
+        prefix = "topic" if self.resource_type == PubSubResourceType.TOPIC else "subscription"
+        return f"{prefix}:{self.project_id}:{self.resource_id}"
```
kacpermuda:
```suggestion
        return f"{self.resource_type.value}:{self.project_id}:{self.resource_id}"
```

client/python/openlineage/client/naming/dataset.py
mobuchowski: Can you link to previous discussion?

client/python/openlineage/client/dataset.py
mobuchowski: IMO specific per namespace/class

client/python/openlineage/client/naming/dataset.py
kacpermuda: Don't remember where it was unfortunately, maybe we lost it on Slack after 90 days.

client/python/openlineage/client/naming/dataset.py
kacpermuda: I think in Airflow we always do file://

client/python/openlineage/client/naming/dataset.py
ddebowczyk92: nope, added

client/python/openlineage/client/naming/dataset.py
ddebowczyk92: I have followed [given](https://openlineage.io/docs/spec/naming/#dataset-naming) naming conventions

client/python/openlineage/client/naming/dataset.py
ddebowczyk92: same as [here](https://github.com/OpenLineage/OpenLineage/pull/3816#discussion_r2200257652)

client/python/openlineage/client/naming/dataset.py
ddebowczyk92: same as [here ](https://github.com/OpenLineage/OpenLineage/pull/3816#discussion_r2200257652)

client/python/openlineage/client/naming/dataset.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def get_name(self) -> str:
+        return f"{self.catalog}.{self.database}.{self.table}"
+
+
+@attr.define
+class AWSGlue(DatasetNaming):
+    """Naming implementation for AWS Glue."""
+
+    region: str = attr.field(validator=_check_not_empty)
```
JDarDagran: nit: picking up comment from Copilot. I think this should be changed in https://openlineage.io/docs/spec/naming/#dataset-naming as well
> [nitpick] Consider renaming region to region_name for the AWSGlue class to match the region_name parameter ?in Athena, improving consistency across naming implementations.
```suggestion
    region_name: str = attr.field(validator=_check_not_empty)
```

JDarDagran:
One more thing around naming formalization - I'm missing some documentation around it, even if it's just a mention that these exist.

JDarDagran:
One more thing around naming formalization - I'm missing some documentation around it, even if it's just a mention that these exist.

PR #3779, opened 2025-06-17
[SPARK] Fine-grained timeout config

pawel-big-lebowski:
Extend timeout feature available in circuit breaker to enable timeouts on building input/output datasets, job and run facets. Current approach results in no events emitted, when timeout occurs. 

Fine-grained timeouts will allow incomplete events with debug facet containing an information about the incomplete lineage.

Within this change, a new circuit breaker `TimeoutCircuitBreaker` is introduced. Additionally, Spark config allows two extra parameters: 
 * `spark.openlineage.timeout.buildDatasetsTimePercentage` - If a timeout is set within a circuit breaker, this configures a percentage of the configured  timeout that can be spent on building datasets.
 * `spark.openlineage.timeout.facetsBuildingTimePercentage` - If a timeout is set within a circuit breaker, this configures a percentage of the configured timeout that can be spent on building facets which includes job facets, run facets, and dataset  facets. This timeout applies effectively on everything besides event serialization and transport.

PR #3817, opened 2025-06-27
name threads used in Java client

mobuchowski:
This helps with debugging thread dumps - for example, when application gets stuck, I want to know if stuck thread

```
"pool-35-thread-1" #109 prio=5 os_prio=0 cpu=307.27ms elapsed=573.64s tid=0x0000f6c71cc622e0 nid=0x177 waiting on condition  [0x0000f6c6f9c61000]
   java.lang.Thread.State: WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@17.0.14/Native Method)
	- parking to wait for  <0x0000000607a52cb8> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
	at java.util.concurrent.locks.LockSupport.park(java.base@17.0.14/LockSupport.java:341)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionNode.block(java.base@17.0.14/AbstractQueuedSynchronizer.java:506)
	at java.util.concurrent.ForkJoinPool.unmanagedBlock(java.base@17.0.14/ForkJoinPool.java:3465)
	at java.util.concurrent.ForkJoinPool.managedBlock(java.base@17.0.14/ForkJoinPool.java:3436)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(java.base@17.0.14/AbstractQueuedSynchronizer.java:1630)
	at java.util.concurrent.LinkedBlockingQueue.take(java.base@17.0.14/LinkedBlockingQueue.java:435)
	at java.util.concurrent.ThreadPoolExecutor.getTask(java.base@17.0.14/ThreadPoolExecutor.java:1062)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@17.0.14/ThreadPoolExecutor.java:1122)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(java.base@17.0.14/ThreadPoolExecutor.java:635)
	at java.lang.Thread.run(java.base@17.0.14/Thread.java:840)

   Locked ownable synchronizers:
	- None
```

comes from OpenLineage

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/circuitBreaker/TaskQueueCircuitBreaker.java
+++ b/client/java/src/main/java/io/openlineage/client/circuitBreaker/TaskQueueCircuitBreaker.java
@@ -7,6 +7,7 @@ package io.openlineage.client.circuitBreaker;
 
 import io.micrometer.common.lang.NonNull;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.openlineage.client.OpenLineageClientUtils;
 import io.openlineage.client.metrics.MicrometerProvider;
 import java.util.List;
 import java.util.Optional;
@@ -48,7 +49,12 @@ public class TaskQueueCircuitBreaker implements CircuitBreaker {
     eventQueue = new ArrayBlockingQueue<>(config.getQueueSize());
     eventProcessingExecutor =
         new ThreadPoolExecutor(
-            config.getThreadCount(), config.getThreadCount(), 60L, TimeUnit.SECONDS, eventQueue);
+            config.getThreadCount(),
+            config.getThreadCount(),
+            60L,
+            TimeUnit.SECONDS,
+            eventQueue,
+            new OpenLineageClientUtils.ExecutorThreadFactory("openlineage-taskqueue"));
   }
 
   @Override
@@ -77,7 +83,7 @@ public class TaskQueueCircuitBreaker implements CircuitBreaker {
           .ifPresent(
               m ->
                   log.info(
-                      "Openlineage async stats: dropped={}, timeout={}, queueDepth={}, failed={}",
+                      "OpenLineage async stats: dropped={}, timeout={}, queueDepth={}, failed={}",
                       m.counter(DROPPED_METRIC).count(),
                       m.counter(TIMED_OUT_METRIC).count(),
                       getPendingTasks(),
--- a/client/java/src/main/java/io/openlineage/client/OpenLineageClientUtils.java
+++ b/client/java/src/main/java/io/openlineage/client/OpenLineageClientUtils.java
@@ -38,6 +38,8 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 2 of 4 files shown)


client/java/src/main/java/io/openlineage/client/utils/ThreadFactory.java
```diff
@@ -0,0 +1,52 @@
...
+
+package io.openlineage.client.utils;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A ThreadFactory implementation that creates named threads for use in ThreadPoolExecutor. This
+ * factory creates threads with a specified name prefix followed by a sequential number.
+ */
+public class ThreadFactory implements java.util.concurrent.ThreadFactory {
```
pawel-big-lebowski: I've done some other executor service related changes in other PR -> https://github.com/OpenLineage/OpenLineage/pull/3779

I like the thread naming idea, but:
 * I would love to have a single ExecutorService for all OL related threads
 * Close the service in OpenLineage.close method
 
I think we should avoid public `ThreadFactory` class.

client/java/src/main/java/io/openlineage/client/utils/ThreadFactory.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /**
+   * Creates a ThreadFactory with the specified name prefix and daemon flag. Threads will be named
+   * as "{namePrefix}-{number}".
+   *
+   * @param namePrefix the prefix for thread names
+   * @param daemon whether threads should be daemon threads
+   */
+  public ThreadFactory(String namePrefix, boolean daemon) {
```
pawel-big-lebowski: Why do we need daemon threads?

mobuchowski: Removed the ability to pass the daemon to match the behavior of [DefaultThreadFactory](https://github.com/openjdk/jdk/blob/12196baf6700d00c244747cfa22767e532a4a963/src/java.base/share/classes/java/util/concurrent/Executors.java#L609)

client/java/src/main/java/io/openlineage/client/utils/ThreadFactory.java
mobuchowski: You need ThreadFactory class to pass to the single executor service though.

pawel-big-lebowski: @mobuchowski do you agree that there should be a single `ExecutorService` for all the OpenLineage related threads? 

If so, perhaps you could add some singleton and avoid `public` ThreadFactory. 
Passing the `namePrefix` to the thread factory does not make sense for a single `ExecutorService` either.

client/java/src/main/java/io/openlineage/client/circuitBreaker/TaskQueueCircuitBreaker.java
```diff
@@ -46,9 +46,7 @@ public TaskQueueCircuitBreaker(@NonNull TaskQueueCircuitBreakerConfig config) {
     this.blockingTimeInSeconds = config.getBlockingTimeInSeconds();
     this.shutdownTimeoutSeconds = config.getShutdownTimeoutSeconds();
     eventQueue = new ArrayBlockingQueue<>(config.getQueueSize());
-    eventProcessingExecutor =
-        new ThreadPoolExecutor(
-            config.getThreadCount(), config.getThreadCount(), 60L, TimeUnit.SECONDS, eventQueue);
+    eventProcessingExecutor = OpenLineageClientUtils.getOrCreateExecutor();
```
mobuchowski: @pawel-big-lebowski have you left the threadpoolexecutor here on purpose? 

If yes, then we need non-static thread factory then

pawel-big-lebowski: The purpose of leaving this was that it has custom `close` logic and I didn't want to interfere with that. `Close` logic is not covered with tests and I didn't want to break the behaviour.

mobuchowski: Ok, I will revert back to custom threadpool executor, but will have to expose thread factory or some other constructor for this.

pawel-big-lebowski (approved):

PR #3819, opened 2025-06-27
spec: SqlJobFacet now contains dialect

mobuchowski:
In some cases, like with dbt SQL events in structured logs integration, we include a SQL facet, but not have any information that would allow us to determine the dialect of the SQL in the SQL job facet. 

Optional `dialect` field solves that issue.


integration/spark/app/src/test/java/io/openlineage/spark/agent/OpenLineageRunEventTest.java
```diff
@@ -65,7 +65,7 @@ void testSerializeRunEvent() throws IOException, URISyntaxException {
             .url(URI.create("https://github.com/apache/spark"))
             .tag("v1.0.0")
             .build();
-    OpenLineage.SQLJobFacet sqlJobFacet = ol.newSQLJobFacet("SELECT * FROM test");
+    OpenLineage.SQLJobFacet sqlJobFacet = ol.newSQLJobFacet("SELECT * FROM test", "bigquery");
```
pawel-big-lebowski: I get this is just a test. But, this should be `spark`, right?

pawel-big-lebowski (approved):
According to the [Spark doc:](https://spark.apache.org/docs/latest/sql-ref-ansi-compliance.html):
`By default, spark.sql.ansi.enabled is true and Spark SQL uses an ANSI compliant dialect instead of being Hive compliant.`

So, specifying a `spark` dialect may be cumbersome. But, I think it's a corner case and this facet brings significant value in other integrations like `dbt`. So, feel free to decide if you want to keep it as is, not fill the `dialect` for Spark, or add some logic to extract spark dialect based on config provided.

kacpermuda (approved):

mobuchowski:
@pawel-big-lebowski looking at the doc, I think this is more about the runtime behavior of the SQL code - which is not the thing we're really concerned with when optionally parsing the query :)

PR #3825, opened 2025-06-30
fix intellij reload

mobuchowski:


pawel-big-lebowski:
Look like a hacky way to hint IntelliJ reload. 
Yes, this is working and solves the problem we've heard about so many times 💯  

```
implementation(project(path: ":shared"), project(path: ":shared", configuration: activeRuntimeElementsConfiguration))
```
Would it make sense to encapsulate the hack in a single method so that we can remove it once IntelliJ gets smarter.

mobuchowski:
@pawel-big-lebowski IDK if IntelliJ will get smarter. This is due to a fact that we use custom configurations, and IntelliJ uses main for code completion/dependencies. I tried to do it in some other way by modifying IntelliJ artifacts, but nothing worked for me.


Screenshot 2025-06-30 at 17 18 06

mobuchowski:
I've validated the change using tool in https://github.com/OpenLineage/OpenLineage/pull/3852

The only different thing is timestamp:


```
================================================================================
JAR COMPARISON REPORT
================================================================================
JAR 1: build/libs/openlineage-spark_2.12-1.35.0-SNAPSHOT.jar
JAR 2: openlineage-spark_2.12-1.35.0-SNAPSHOT-reload.jar

SUMMARY:
  Files in JAR 1: 11061
  Files in JAR 2: 11061
  Identical files: 11060
  Different files: 1
  Only in JAR 1: 0
  Only in JAR 2: 0

DIFFERENT FILES:

--- io/openlineage/spark/agent/version.properties (TEXT_DIFF) ---
--- version.properties
+++ version.properties
@@ -1,2 +1,2 @@
-#Thu Jul 03 14:48:31 CEST 2025

+#Thu Jul 03 14:40:45 CEST 2025

 version=1.35.0-SNAPSHOT
```

pawel-big-lebowski (approved):

PR #3829, opened 2025-07-01
feat: Add continue_on_success and sorting transport in CompositeTransport

kacpermuda:
### Problem
There was no mechanism to control the behavior of continuing emission after a successful transport. In case users wanted to only deliver to a single backend, but have an option to "fallback" to another one, when first emission failed, it was hard to do.

Also, the current implementation of reading config from env variables uses a default descending alphabetical sorting of variable names. This sorting behavior is not clearly documented, which can cause confusion on which transport will emit first when defined as separate env variables. 

### Solution

- Introduced the `continue_on_success` attribute to the `CompositeConfig` class. This flag, when set to `True`, ensures that events continue to be emitted to all transports even after a successful emission to one of them. If set to `False`, it stops further emission once the first successful transport is encountered.
- Added an option to sort transports in composite transport, by newly added transport's `priority` field. Priority defaults to 0 and can be passed in transport's config. The higher priority, the earlier the transport will try to emit event. This allows the user to control the sorting behavior, even when env variables are used to define transports.

Using those two solutions together, you can achieve almost any logic.

Also, when running tests with pytest some deprecation warnings were raised from two deprecated modules that we still use internally to keep compatibility, moved them under warning silencer.

#### One-line summary:
Added `continue_on_success` for fallback behavior and clarified sorting mechanism for transport configurations in CompositeTransport.

### Checklist

----

### DIFF
--- a/client/python/openlineage/client/transport/factory.py
+++ b/client/python/openlineage/client/transport/factory.py
@@ -51,6 +51,12 @@ class DefaultTransportFactory(TransportFactory):
             raise TypeError(msg) from None
 
         transport_name = config.pop("name", None)
+        transport_priority = config.pop("priority", 0)
+        try:
+            transport_priority = int(transport_priority)
+        except ValueError as e:
+            msg = f"Error casting priority `{transport_priority}` to int for transport `{transport_type}`"
+            raise ValueError(msg) from e
 
         transport_class_type_or_str = self.transports.get(transport_type, transport_type)
 
@@ -73,4 +79,6 @@ class DefaultTransportFactory(TransportFactory):
         transport: Transport = transport_class(config_class.from_dict(config))  # type: ignore[call-arg]
         if transport_name and not transport.name:
             transport.name = transport_name
+        if transport_priority and transport.priority == 0:
+            transport.priority = transport_priority
         return transport
--- a/website/docs/client/python.md
+++ b/website/docs/client/python.md
@@ -579,12 +579,22 @@ The `CompositeTransport` is designed to combine multiple transports, allowing ev
 - `type` - string, must be "composite". Required.
 - `transports` - a list or a map of transport configurations. Required.
 - `continue_on_failure` - boolean flag, determines if the process should continue even when one of the transports fails. Default is `true`.
+- `continue_on_success` - boolean flag, determines if the process should continue when one of the transports succeeds. Default is `true`.
+- `sort_transports` - boolean flag, determines if transports should be sorted by `priority` before emission. Default is `false`.
 
 #### Behavior
 
 - The configured transports will be initialized and used in sequence to emit OpenLineage events.
 - If `continue_on_failure` is set to `false`, a failure in one transport will stop the event emission process, and an exception will be raised.
 - If `continue_on_failure` is `true`, the failure will be logged and the process will continue allowing the remaining transports to still send the event.
+- If `continue_on_success` is set to `false`, a success of one transport will stop the event emission process. This is useful if you want to deliver events to at most one backend, and only fallback to other backends in case of failure.
+- If `continue_on_success` is set to `true`, the success will be logged and the process will continue allowing the remaining transports to send the event.
+
+#### Transport Priority
+Each transport in the `transports` configuration can include an optional `priority` field (integer). 
+When `sort_transports` is `true`, transports are sorted by priority in descending order (higher priority values are processed first). 
+Transports without a priority field default to priority 0.
+
 
 #### Notes for Multiple Transports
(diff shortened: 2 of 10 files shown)


mobuchowski:
TBH I reflexively don't like the transport sorting idea, because it conflicts with my idea of how things are configured. But on the other hand we have so many ways to configure the transports, that some `runtime modification` might be a good idea... sad that we haven't figured out dynamic env variables early and had to do with several half baked mechanisms.

kacpermuda:
Agree, we probably could have implement several features better 😄  This sorting may not be a perfect solution, but it allows the user to easily change the order of transports and do it only when needed (only for composite transport) and when explicitly configured so I think it's worth it. Happy to change it if you have some other ideas.

mobuchowski:
Maybe instead of that we can add some numerical value (optional, similar to name) that would fill that role? Then sort is not sort_by_field but just sort? It won't work with _everything_ as you can't specify order id via legacy env variables, but with sort by field you can't sort two transports that don't share any field, like http and kafka

`OPENLINEAGE__TRANSPORT__TRANSPORTS__SOME_HTTP_TRANSPORT__ORDER: 7`

kacpermuda:
Thought about it as well, in my draft it was called `priority`, but finally I decided against it. It would be nice to have some order_key in the transport definition itself, but it only really matters in that specific case of composite transport, and only when you are not happy with the order of the provided transport, that you can influence as a user by simply passing the transports config in preferred order.

As for the example of http and kafka: if you name the transports, then you always have a name field to sort on, otherwise you can always use type. 

I'm okay with the order/priority field as well, just chose a different path here. With current, no extra configs are required, you can easily use names. However, specifying order in env variables is also not a big effort, so that'd work as well

kacpermuda:
We discussed on commiter sync, will move forward with order/priority approach. Will adjust the PR shortly

website/docs/client/python.md
```diff
@@ -465,12 +465,16 @@ The `CompositeTransport` is designed to combine multiple transports, allowing ev
...
+- `continue_on_success` - boolean flag, determines if the process should continue when one of the transports succeeds. Default is `true`.
+- `sort_transports` - boolean flag, determines if transports should be sorted by `priority` before emission. Default is `false`.
 
 #### Behavior
 
 - The configured transports will be initialized and used in sequence to emit OpenLineage events.
 - If `continue_on_failure` is set to `false`, a failure in one transport will stop the event emission process, and an exception will be raised.
 - If `continue_on_failure` is `true`, the failure will be logged and the process will continue allowing the remaining transports to still send the event.
+- If `continue_on_success` is set to `false`, a success of one transport will stop the event emission process. This is useful if you want to deliver events to at most one backend, and only fallback to other backends in case of failure.
+- If `continue_on_success` is set to `true`, the success will be logged and the process will continue allowing the remaining transports to send the event.
```
JDarDagran, line 591:
```suggestion
- If `continue_on_success` is set to `true`, the success will be logged and the process will continue allowing the remaining transports to send the event.


#### Transport Priority

Each transport in the `transports` configuration can include an optional `priority` field (integer). When `sort_transports` is `true`, transports are sorted by priority in descending order (higher priority values are processed first). Transports without a priority field default to priority 0.

```

How about some more explanation about above?

client/python/openlineage/client/transport/factory.py
```diff
@@ -73,4 +74,6 @@ def _create_transport(self, config: dict[str, str]) -> Transport:
         transport: Transport = transport_class(config_class.from_dict(config))  # type: ignore[call-arg]
         if transport_name and not transport.name:
             transport.name = transport_name
+        if transport_priority and transport.priority == 0:
```
JDarDagran, line 82: I think it's missing unit test to verify if this is set properly.

client/python/openlineage/client/transport/factory.py
```diff
@@ -51,6 +51,7 @@ def _create_transport(self, config: dict[str, str]) -> Transport:
             raise TypeError(msg) from None
 
         transport_name = config.pop("name", None)
+        transport_priority = int(config.pop("priority", 0))
```
JDarDagran: Should this log any warning if field is not valid?

client/python/openlineage/client/transport/factory.py
kacpermuda: Added.

website/docs/client/python.md
kacpermuda: Added

client/python/openlineage/client/transport/factory.py
kacpermuda: Added try/except with more detailed error message.

JDarDagran (approved):

PR #3832, opened 2025-07-01
Spark - avoid WARN in DatasetVersionDatasetFacetUtils caused by unnecessary method calls

pawel-big-lebowski:
Production workloads with bigqyery connector often produce significant amount of logs:
```
WARN DatasetVersionDatasetFacetUtils: Couldn't find identifier for dataset in plan
```

Although this is a valid warn, we should change the facet building method ordering: try extracting dataset version only when having a dataset identifier. In this case, we avoid calling a method that would result in `WARN` anyway.

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/DataSourceV2RelationDatasetExtractor.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/DataSourceV2RelationDatasetExtractor.java
@@ -32,15 +32,38 @@ import org.jetbrains.annotations.NotNull;
 public class DataSourceV2RelationDatasetExtractor {
 
   public static <D extends OpenLineage.Dataset> List<D> extract(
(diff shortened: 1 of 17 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/utils/DataSourceV2RelationDatasetExtractor.java
```diff
@@ -31,16 +32,39 @@
 @Slf4j
 public class DataSourceV2RelationDatasetExtractor {
 
+  public static <D extends OpenLineage.Dataset> List<D> extract(
+      DatasetFactory<D> datasetFactory,
+      OpenLineageContext context,
+      DataSourceV2Relation relation,
+      boolean includeVersionFacet) {
```
ddebowczyk92, line 38: Looks good overall, wdyt about splitting extract method into two separate methods - `extract` and `extractIncludingVersionFacet`?

PR #3836, opened 2025-07-01
more resiliency for missing dbt nodes

mobuchowski:
There is an intermittent issue in an environment where the dbt manifest gets somehow corrupted or out of sync with currently running job and has missing dbt nodes. It causes the dbt integration to fail with KeyError.

This has a negative impact on the underlying job execution - so we can't really "fix" it, but we should still emit the events about the job - even if the inputs/outputs are incomplete, we still can capture the error.

Additionally, we add "manifest integrity validation" mechanism, where on start we check if the manifest is healthy, and log the results if it's not. This allows users to have upfront information about that type of error.

### DIFF
--- a/integration/common/openlineage/common/provider/dbt/structured_logs.py
+++ b/integration/common/openlineage/common/provider/dbt/structured_logs.py
@@ -7,12 +7,14 @@ import subprocess
 import sys
 import time
 from collections import defaultdict
-from functools import cached_property
-from typing import Dict, Generator, List, Optional, TextIO
+from functools import cached_property, wraps
+from typing import Any, Dict, Generator, List, Optional, TextIO
 
 from openlineage.client.event_v2 import Dataset, RunEvent, RunState
 from openlineage.client.facet_v2 import (
-    data_quality_assertions_dataset,
+    data_quality_assertions_dataset as dq,
+)
+from openlineage.client.facet_v2 import (
     error_message_run,
     job_type_job,
     processing_engine_run,
@@ -48,6 +50,26 @@ from openlineage.common.utils import (
     get_from_nullable_chain,
     has_lines,
 )
+
+
+def handle_keyerror(func):
+    """
+    Handle KeyError exceptions when node_id was not found in the dbt manifest.
+    """
+
+    @wraps(func)
+    def wrapper(self, node_id: str, *args, **kwargs):
+        try:
+            return func(self, node_id, *args, **kwargs)
+        except KeyError:
+            self.logger.warning(
+                "Did not find node_id %s in the dbt manifest - did "
+                "the manifest change while the job was executing?",
+                node_id,
+            )
+            return None
+
+    return wrapper
 
 
 class DbtStructuredLogsProcessor(DbtLocalArtifactProcessor):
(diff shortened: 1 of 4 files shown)


integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -397,10 +403,13 @@ def _get_attached_dataset(self, test_node_id: str) -> Optional[InputDataset]:
             input_dataset = self.node_to_dataset(node=attached_model_node, has_facets=True)
         return input_dataset
 
-    def _get_assertion(self, node_id: str, success: bool) -> data_quality_assertions_dataset.Assertion:
-        manifest_test_node = self.compiled_manifest["nodes"][node_id]
+    def _get_assertion(self, node_id: str, success: bool) -> Optional[dq.Assertion]:
+        try:
+            manifest_test_node = self.compiled_manifest["nodes"][node_id]
+        except KeyError:
+            self.logger.warning("Did not find node %s in manifest - ha", node_id)
```
kacpermuda: `- ha` is this some leftover from sentence `- has ...`?

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -745,6 +769,76 @@ def _get_model_output(self, node_id) -> ModelNode:
...
+        missing_nodes = list(dict.fromkeys(missing_nodes))
+        missing_parents = list(dict.fromkeys(missing_parents))
+        missing_children = list(dict.fromkeys(missing_children))
+
+        if missing_nodes:
+            self.logger.warning(
+                "Manifest integrity check failed: Found %d nodes in parent_map that don't exist "
+                "in manifest node definitions: %s",
+                len(missing_nodes),
+                missing_nodes[:10],  # Log first 10 to avoid overwhelming logs
```
kacpermuda: Let's add the information that only some nodes are displayed in the warning message, I'd expect all of them to be logged so it can probably be confusing for users

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -745,6 +769,76 @@ def _get_model_output(self, node_id) -> ModelNode:
...
+
+        Returns:
+            dict: Validation results containing:
+                - 'is_valid': bool indicating if manifest is valid
+                - 'missing_nodes': list of node IDs that are missing
+                - 'missing_parents': list of parent node IDs that are missing
+                - 'missing_children': list of child node IDs that are missing
+                - 'total_missing': total count of missing nodes
+                - 'parent_map_size': total number of entries in parent_map
+                - 'available_nodes': total number of available nodes (nodes + sources)
```
kacpermuda: Looks like a really well defined structure, maybe let's do a named tuple for it or some other class?

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -731,19 +794,82 @@ def _get_model_inputs(self, node_id) -> List[ModelNode]:
...
+            inputs = [node for n in input_node_ids if (node := self._get_model_node(n)) is not None]
+            self.node_id_to_inputs[node_id] = inputs
+            return inputs
+        except KeyError:
+            self.logger.warning(
+                "Did not find node_id %s in the dbt manifest - did "
+                "the manifest change while the job was executing?",
+                node_id,
+            )
+        return []
```
kacpermuda, line 808: Do we need both @handle_keyerror wrapper an try/except on KeyError here ?

mobuchowski: It's a bit different structure (returning list) and I decided to leave it. Could have a decorator that wraps that up... though idk if that's worth it at this point.

kacpermuda: Yeah, we can leave it as it's a different case.

PR #3839, opened 2025-07-01
[Java] Cover OpenLineageClient.close() with tests

dolfinus:
### Solution

Cover `OpenLineageClient.close()` with unit tests. Also change logic a bit:
* I f transport fails to close, do not left `circuitBreaker`, `meterRegistry` and default ExecutorService unclosed.
* Shared `ExecutorService` can be shut down safely, and next instance of `OpenLineageClient` will use a new one.

### Checklist

----

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/OpenLineageClient.java
+++ b/client/java/src/main/java/io/openlineage/client/OpenLineageClient.java
@@ -153,19 +153,26 @@ public final class OpenLineageClient implements AutoCloseable {
             "openlineage.emit.time", "openlineage.transport", transport.getClass().getName());
   }
 
+  /** Shutdown the underlying transport, waiting for all events to complete. */
+  @Override
+  public void close() throws Exception {
+    try {
+      transport.close();
+    } catch (Exception e) {
+      throw new OpenLineageClientException("Failed to close transport " + transport, e);
+    } finally {
+      circuitBreaker.ifPresent(CircuitBreaker::close);
+      meterRegistry.close();
+      OpenLineageClientUtils.getExecutor().ifPresent(ExecutorService::shutdown);
+    }
+  }
+
   /**
    * @return an new {@link OpenLineageClient.Builder} object for building {@link
    *     OpenLineageClient}s.
    */
   public static Builder builder() {
     return new Builder();
-  }
-
-  @Override
-  public void close() throws Exception {
-    transport.close();
-    meterRegistry.close();
-    OpenLineageClientUtils.getExecutor().ifPresent(ExecutorService::shutdown);
   }
 
   /**
--- a/client/java/src/main/java/io/openlineage/client/OpenLineageClientUtils.java
+++ b/client/java/src/main/java/io/openlineage/client/OpenLineageClientUtils.java
@@ -349,7 +349,7 @@ public final class OpenLineageClientUtils {
   }
 
   public static ExecutorService getOrCreateExecutor() {
-    if (EXECUTOR == null) {
+    if (EXECUTOR == null || EXECUTOR.isShutdown()) {
       EXECUTOR = Executors.newCachedThreadPool();
     }
     return EXECUTOR;
--- a/client/java/src/test/java/io/openlineage/client/OpenLineageClientTest.java
+++ b/client/java/src/test/java/io/openlineage/client/OpenLineageClientTest.java
@@ -5,7 +5,10 @@
 
 package io.openlineage.client;
 
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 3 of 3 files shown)


client/java/src/main/java/io/openlineage/client/OpenLineageClient.java
```diff
@@ -154,20 +154,45 @@ public void initializeMetrics() {
...
+  }
+
+  /**
+   * Shutdown the underlying transport, waiting for all events to complete or until the timeout is
+   * reached.
+   *
+   * @param timeout in seconds. -1 means to block until last event is processed
+   * @return true if all events were processed before transport was closed, false otherwise.
+   */
+  public boolean shutdown(float timeout) throws Exception {
```
pawel-big-lebowski: I think we're not using float timeout at the moment in any other timeouts. It's either integer of Duration. Is there any reason to have this implemented differently? 

Do we need `shutdown(-1)` here? The same behaviour can be obtained with no-arg `shutdown()`.

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
```diff
@@ -132,16 +132,46 @@ private void emit(Transport transport, BaseEvent event) {
...
+  }
+
+  @Override
+  public boolean shutdown(float timeout) throws Exception {
+    // Force shutdown of all transports
+    boolean result = true;
+    Exception latestException = null;
+    for (Transport transport : transports) {
+      try {
+        result = result && transport.shutdown(timeout);
```
pawel-big-lebowski: Running shutdown on composite transport with 10secs timeout, and 2 transports which shutdown in 10 secs each, would lead to 20 seconds shutdown within this method. Is this desired? 

Not sure how to fix this without overcomplicating things.

client/java/src/main/java/io/openlineage/client/transports/Transport.java
```diff
@@ -17,6 +17,47 @@ public abstract class Transport implements AutoCloseable {
...
+    return waitForCompletion(-1);
+  }
+
+  /**
+   * Blocks until all events have been emitted or timeout is reached.
+   *
+   * @param timeout in seconds. -1 means to block until last event is processed
+   * @return true if all events were processed, false otherwise.
+   */
+  public boolean waitForCompletion(float timeout) {
```
pawel-big-lebowski: `Shutdown the transport, waiting for all events to complete or until the timeout is reached.` 
vs `Blocks until all events have been emitted or timeout is reached.` - what's the difference?

pawel-big-lebowski (commented):
Why do we need both `shutdown` and `waitForCompletion` methods for the transports? Both methods make sense for managing `ExecutorService`, but this approach seems less relevant for the transports.

client/java/src/main/java/io/openlineage/client/OpenLineageClient.java
dolfinus: Replaced `float timeout` with `long timeout, TimeUnit unit`

client/java/src/main/java/io/openlineage/client/transports/CompositeTransport.java
dolfinus: That's how #3812 implemented this method. I also don't have ideas how this could be implemented in different way

client/java/src/main/java/io/openlineage/client/transports/Transport.java
dolfinus: One closes transport, another just calls something like `producer.flush()`, and exits

dolfinus:
After some tries and errors, I think that `.close(timeout)` method is actually useless - in 99% cases user wants to send **all** emitted events, not until some timeout is reached.

mobuchowski:
@dolfinus User yes, maintainer of the system/integration not always. We have a years long issue where Airflow kills worker process by default while OL is still running.

dolfinus:
If there is a transport with 1_000 events and each takes 20ms to send, the following situations are just the same:
* `client.close()` blocks for 1_000 * 20ms = 20 seconds. After 10 seconds Airflow kills the worker process, so only half of events are actually send, other half is lost.
* `client.close(timeout=10)` blocks for 10 seconds, and then releases of throws an exception, but still only half of events are send. After method exited, Airflow kills worker, but pending events are lost in any way.

Even if Airflow can pass its internal timeout value into this method, there is _nothing_ that client can do with this information, it is limited by network speed, backend latency and so on.

dolfinus:
I've reverted changes to Transport interface, `void close()` is just enough. CompositeTransport changes are moved to #3841

pawel-big-lebowski (approved):

PR #3840, opened 2025-07-02
python: Remove v1 code - client.run and client.facet

kacpermuda:
### Problem

We've deprecated this code over a year ago, maybe it's time to remove it. If we can't do it without a major release, than I'll revert the deletion of the two modules, but I think the rest can still be merged - deleting all the test cases with those and support in all transports etc.

### Solution
Removed the v1 facets and client.run entities from the code.

#### One-line summary:
python: Remove v1 code - client.run and client.facet

### Checklist

----


JDarDagran:
I think any removal of the code should happen in major release only.

mobuchowski:
>I think any removal of the code should happen in major release only.

Agree... but maybe that's the right time for it :)

kacpermuda:
I think there are more things to change / remove in the major release (we should probably create some label to track this, or some epic issue). If we can't remove this modules without the major release, I'd not do it for now, we can always just remove all support for it and all internal use, but still leave the modules so it does not throw ImportError.

kacpermuda:
Added ol:2.0 label that we can use for issues and PRs. We can discuss if removing support for v1 events, even without removing code, also requires major release. There is no pressure from my side, just wanted to clean up a bit.

mobuchowski:
I would want to specify explicitly here that I just think of this major release as major release of libraries - the spec should be as it is.

kacpermuda:
Sure, I've also added spec:3.0 label for spec related major changes.

kacpermuda:
Will re-visit once we decide to do major release

PR #3843, opened 2025-07-02
[Python] Reuse requests.Session() in sync HttpTransport

dolfinus:
### Problem

For now, Python implementation of `HttpTransport` creates new `requests.Session()` fo every call. This is both slow and doesn't allow auth using session cookies.

#### One-line summary:

[Python] Reuse `requests.Session()` in sync HttpTransport

### Checklist

----

### DIFF
--- a/client/python/openlineage/client/transport/http.py
+++ b/client/python/openlineage/client/transport/http.py
@@ -158,13 +158,11 @@ class HttpTransport(Transport):
                 raise ValueError(msg)
         self.url = url
         self.endpoint = config.endpoint
-        self.session = None
-        if config.session:
-            self.session = config.session
-            self._prepare_session(self.session)
         self.timeout = config.timeout
         self.verify = config.verify
         self.compression = config.compression
+        self._session: Session | None = None
+        self.session = config.session  # type: ignore[assignment]
 
     def emit(self, event: Event) -> Response:
         # If anyone overrides debuglevel manually, we can potentially leak secrets to logs.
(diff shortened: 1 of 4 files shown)


dolfinus:
It is possible that this approach was used because of Airflow integration quirks, like #1959. But I'm not sure, this issue doesn't provide any context how it can be reproduced.

I could add a workaround for this by calling `OpenLineageClient.close()` or `OpenLineageClient.wait_for_completion()` just after `OpenLineageAdapter.emit(event)`

mobuchowski:
I think it was it: https://github.com/OpenLineage/OpenLineage/pull/2025

Though can't recall what exactly was the issue. AFAIK snowlake connector fixed some bugs that hit us in Airflow later, so maybe worth to try using reusable sessions again.

JDarDagran:
How about more concise approach with:
```python
@property
def session(self) -> Session:
    if self._session is None:
        self._session = Session()
        self._prepare_session(self._session)
    return self._session

@session.setter
def session(self, value: Session | None):
    if value is not None:
        self._prepare_session(value)
    self._session = value
```
?

dolfinus:
Ok, done

client/python/openlineage/client/transport/http.py
```diff
@@ -158,13 +158,11 @@ def __init__(self, config: HttpConfig) -> None:
...
         self.endpoint = config.endpoint
-        self.session = None
-        if config.session:
-            self.session = config.session
-            self._prepare_session(self.session)
         self.timeout = config.timeout
         self.verify = config.verify
         self.compression = config.compression
+        self._session: Session | None = config.session
+        self._prepare_session(self.session)
```
JDarDagran: This isn't needed anymore?

client/python/openlineage/client/transport/http.py
```diff
@@ -173,29 +171,37 @@ def emit(self, event: Event) -> Response:
...
+            headers=headers,
+            timeout=self.timeout,
+            verify=self.verify,
+        )
+        resp.close()
         http_client.HTTPConnection.debuglevel = prev_debuglevel
         resp.raise_for_status()
         return resp
 
+    @property
```
kacpermuda, line 186: Maybe?
```suggestion
    @cached_property
```

dolfinus: `cached_property` doesn't handle case then session is closed

client/python/openlineage/client/transport/http.py
dolfinus: Fixed

dolfinus:
CI is failed because of some timeout, could someone restart it, please?

dolfinus:
I see the same errors with Airflow 2.5 in 2.6 integration tests on `main` branch:
image

Maybe something is wrong with CI? It started failing today

mobuchowski (approved):
I'll merge this as CI errors are unrelated

Issue #3846, opened 2025-07-03
[BUG] SparkApplicationStart sometimes emits job.name=unknown


dolfinus:
### OpenLineage integration/client

Spark integration

### OpenLineage version

openlineage-spark==1.34.0

### Technology and package versions

pyspark==3.3.1

### Deployment details

Using PySpark with roughly this config:
```python
from pyspark.sql import SparkSession

spark = (
  SparkSession.builder
    .appName("myapp")
    .master("yarn")
    .config("spark.jars.packages", "io.openlineage:openlineage-spark_2.12:1.34.0")
    .config("spark.extraListeners", "io.openlineage.spark.agent.OpenLineageSparkListener")
    .config("spark.openlineage.transport.type", "console")
    .getOrCreate()
)
```

### Problem details

During SparkApplicationStart event, OpenLineage Spark integration sends START event for application. It should looks like this:
```js
{
  "eventTime": ...,
  "eventType": "START",
  "job": {
    "name": "myapp",
    "namespace": "default",
  },
  "run": {
    "runId": ...,
    "facets": {
      "spark_applicationDetails": {
        "master": "yarn",
        "appName": "myapp",
        "applicationId": "application_123_456",
        ...
      }
    }
  }
  ...
}
```

But instead I sometimes got events like this:
```js
{
  "eventTime": ...,
  "eventType": "START",
  "job": {
    "name": "unknown",
    "namespace": "default",
  },
  "run": {
    "runId": ...,
    "facets": {}
  }
  ...
}
```

And I see this warning in logs:
```
SparkApplicationNameApplicationJobNameProvider: Failed to obtain the application name. Using the default value [unknown]. Set the [spark.app.name] or [spark.openlineage.appName] property.
```

This appears mostly for `master=yarn`, but not for `master=local`.

Probably SparkContext or SparkConf is not available or fully initialized while SparkApplicationStart event arrives to SparkListener, so `SparkApplicationNameApplicationJobNameProvider` cannot extract `spark.app.name` from Spark config.

### What you think should happen instead

Application name is a part of SparkApplicationStart event:
https://spark.apache.org/docs/latest/api/java/org/apache/spark/scheduler/SparkListenerApplicationStart.html

So it always should be available.

### How to reproduce

I'm not sure. Some of my jobs with `master=yarn` are fine. Others with same Spark version and Hadoop cluster are not.

### Do you plan to make this contribution yourself?

- [x] I am interested in doing this work

[...]

PR #3847, opened 2025-07-03
[Spark] Avoid job.name=unknown for SparkApplicationStart event

dolfinus:
### Problem

Partially solves #3846

### Solution

`SparkApplicationStart` event already has `appName` field which is mandatory:
https://spark.apache.org/docs/latest/api/java/org/apache/spark/scheduler/SparkListenerApplicationStart.html

It is already passed to `EventEmitter(conf, appName)`, and then passed to `OpenLineageContext.applicationName` field to be used within `JobNameBuilder`.

This is a bit ugly, so this PR is in draft until I find a better solution.

#### One-line summary:

[Spark] Avoid job.name=unknown for SparkApplicationStart event

### Checklist

----


dolfinus:
Maybe EvemtEmitter should be one of fields inside of OpenLineageContext, instead of duplication these fields in both classes

pawel-big-lebowski:
> Maybe EvemtEmitter should be one of fields inside of OpenLineageContext, instead of duplication these fields in both classes

Once applicationName is in OpenLineageContext, I would verify if it is still required in `EventEmitter`. Usage in `RddExecutionContext` and `SparkSQLExecutionContext` can be replaced with the value from `olContext`.

dolfinus:
EventEmitter populates parentRun facets, so it need Spark application job+runId, and thus have these 2 fields. Moving these fields to another class will require moving facet builders as well.

There is another issue with EventEmitter:
```python
spark = (
  SparkSession.builder.
    .config("spark.app.name", "myapp")
    .config("spark.openlineage.appName", "override")
    .getOrCreate()
)
```

OL will produce events:
* jobType = APPLICATION, name=myapp
* jobType = SQL_JOB, name=override.execute_append.output_table

This is because JobNameBuilder is applied only to SQL and RDD jobs, not APPLICATION.

wslulciuc:
@pawel-big-lebowski Any plans to get @dolfinus fix into the next release? 🙏  We're running into a [similar issue](https://github.com/OpenLineage/OpenLineage/issues/3846#issuecomment-3255766710).

dolfinus:
The fix is not ready yet, and I have no time work on it for next several weeks

integration/spark/app/src/main/java/io/openlineage/spark/agent/EventEmitter.java
```diff
@@ -47,7 +46,6 @@ public EventEmitter(SparkOpenLineageConfig config, String applicationJobName)
     this.rootParentJobName = Optional.ofNullable(config.getRootParentJobName());
     this.rootParentJobNamespace = Optional.ofNullable(config.getRootParentJobNamespace());
     this.rootParentRunId = convertToUUID(config.getRootParentRunId());
-    this.overriddenAppName = Optional.ofNullable(config.getOverriddenAppName());
```
EugeneYushin, line 52: I bet `EventEmitter` should respect `spark.openlineage.appName`, but it looks like it uses appName coming from Spark context (that is set during context init call) with no respect to `spark.openlineage.appName`:
https://github.com/OpenLineage/OpenLineage/blob/main/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java#L351

dolfinus:
Sorry, I don't have enough time to finish this. If anyone is interested on fixing the issue, please open another PR, I'll leave the branch intact.

PR #3873, opened 2025-07-08
feat: add default configuration option to python client

kacpermuda:
**WIP:** Documentation and tests are still in progress. I wanted to share a draft to gather feedback early. I'll probably also update this PR description after comments from the community.

**Important Note:** The feature introduced in this PR is opt-in only. It must be explicitly enabled by users, so it will not affect current users unless they choose to enable it.

We can also extend this to Java client, so that hosted environments like Google Dataproc or Amazon EMR can also take advantage of it.

## TLDR;
This PR introduces an explicit opt-in configuration mechanism for service providers (e.g., Astronomer, Google Composer, possibly some hosted DBT services with OL etc.) to pre-configure OpenLineage clients in a way that’s both easier for users to control (can be disabled with single flag) and safer for complex transport setups. The solution ensures that defaults will not interfere with user-provided configurations, and provides safeguards to prevent errors when service providers configure advanced transports that may not be compatible with all OL client versions.

## Problem
In hosted environments where a service provider manages the OL client, there is a need to provide a pre-configured OpenLineage setup that facilitates the user experience (e.g. service provider pre-configures their own lineage backend as default). However, there is currently no standardized way to do this without causing potential conflicts with user-provided settings.

Different service providers may attempt to set up defaults in various ways (e.g., using `OPENLINEAGE_URL`, `OPENLINEAGE__TRANSPORT`, or `AIRFLOW__OPENLINEAGE__TRANSPORT` in Airflow), but this can create confusion and make it difficult for users to override these configurations. In some cases, users may need to overwrite all pre-set variables to ensure they are only using their own configuration, which is tedious.

Additionally, service providers may want to safely configure more complex transports or settings without the risk of breaking OL initialization. For example, if a service provider configures a composite transport via `OPENLINEAGE__TRANSPORT` but the user is using an older version of OL (< 1.20), this could lead to an error as the transport type is unknown, hence limiting service-providers to using only http transport and configuring it with `OPENLINEAGE_URL` - with no possibility to configure retries, custom headers etc.

## Solution
This PR introduces a way for service providers to configure default settings that are explicitly opt-in. Users can enable these default configurations with a single flag (`OPENLINEAGE__DEFAULT__ENABLED`), which defaults to `False`. If enabled, the default configuration (e.g., facets, filters, tags) will be safely merged with any user-provided settings. If no user transport is configured, the default will be used. As `transport` is a crucial setup, if both user and default are provided, the merging is hidden behind a separate flag, so that users can take advantage of all other defaults configs except the default transport, when providing their own transport setup.

### Key Features:
- **Opt-in feature flag:** `OPENLINEAGE__DEFAULT__ENABLED` controls whether the default configuration is applied, with the default behavior unchanged if the flag is disabled.
- **Safe merging of configurations:** If the defaults are enabled, the configuration is merged with the user’s settings always giving precedence to user provided values. If there both transport configurations are provided, the default transport is ignored unless the user explicitly enables the `OPENLINEAGE__DEFAULT__MERGE_TRANSPORTS` flag to merge them.
- **Validation and logging:** The default transport configuration is validated before use. If it’s invalid or the transport type is unsupported, it will be logged, but it will not cause OL initialization to fail, not breaking user provided transport configuration.
- **Future-proofing:** This solution allows service providers to pre-configure more complex transports, even if they are introduced in future versions of OL. If a transport is not supported by the user's version, it will be safely ignored without crashing the initialization.

This approach provides a cleaner and more controlled way for service providers to configure OL for users while ensuring that the user’s own settings take precedence and remain safe from conflicts or errors.

## Recommended Configuration for Service Providers

Service providers should configure OpenLineage (OL) for their customers in the following way:
1. **Enable default configuration**  
   This flag enables the default configuration setup for the user. If the user disables the default configuration (OPENLINEAGE__DEFAULT__ENABLED=false), none of the default variables will apply.
   ```bash
   OPENLINEAGE__DEFAULT__ENABLED=true  # If set to false, none of the default variables will affect the user's configuration.
    ```

2. Merge default transport with user-provided configuration. If the user provides their own transport configuration, the default transport will be appended:
    ```bash
    OPENLINEAGE__DEFAULT__MERGE_TRANSPORTS=true  # If set to false, the default transport will be ignored.
    ```

3. Configure the transport settings
The service provider can pre-configure the transport settings:
    ```bash
    OPENLINEAGE__DEFAULT__TRANSPORT__TYPE=http
    OPENLINEAGE__DEFAULT__TRANSPORT__URL=https://my-api.com
    ```

A more complex example can be:
```bash
OPENLINEAGE__DEFAULT__ENABLED=true  # If user change it to false, none of the other default variables have any influence on the setup.

OPENLINEAGE__DEFAULT__MERGE_TRANSPORTS=true # If user provides their own transport, the default one will be appended to it. User can always set it to false.

OPENLINEAGE__DEFAULT__TRANSPORT__TYPE=composite
OPENLINEAGE__DEFAULT__TRANSPORT__PRIORITY=-1  # To make sure it's the last transport used, after user-provided ones.
OPENLINEAGE__DEFAULT__TRANSPORT__NAME=default_service_provided_transport
OPENLINEAGE__DEFAULT__TRANSPORT__CONTINUE_ON_FAILURE=true
OPENLINEAGE__DEFAULT__TRANSPORT__CONTINUE_ON_SUCCESS=false
OPENLINEAGE__DEFAULT__TRANSPORT__SORT_TRANSPORTS=true

OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__PRIORITY=1
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__TYPE=http
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__URL=https://my-api.com
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__ENDPOINT=api/v2/lineage
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__AUTH={...}
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__RETRY__TOTAL=2
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__PRIMARY__CUSTOM_HEADERS__...=... // More complex transport options

OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__BACKUP__PRIORITY=0
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__BACKUP__TYPE=http
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__BACKUP__URL=https://backup-api.com
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__BACKUP__ENDPOINT=api/v3/lineage
OPENLINEAGE__DEFAULT__TRANSPORT__TRANSPORTS__BACKUP__AUTH={...}

// Set this as fallback, in case older OL client is used that does not support providing default configuration
OPENLINEAGE_URL=https://my-api.com
OPENLINEAGE_ENDPOINT=api/v2/lineage
OPENLINEAGE_API_KEY=secret_key
```

In this more complex example, the service provider aims to pre-configure OpenLineage to ensure events are delivered even if the primary API fails, using a composite transport. By leveraging the new default configuration mechanism, they can safely provide this advanced setup without the risk of interfering with user configurations (users can easily disable it if desired). Additionally, these environment variables will not affect earlier versions of the OpenLineage client—they will be safely ignored.

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import contextlib
+import copy
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


kacpermuda:
Using OPENLINEAGE_CONFIG that points to some config file provided by service provider has some flaws:
- Impossible to easily accept part of the configuration provided
- Has precedence over env var provided configs - default may overwrite user code
- No safety mechanism, if transport provided in .yaml file does not exist, whole initialization can fail

So while it may work sometimes, I think the proposed solution offers more control and easier config, even thought it may look complex.

The downside of the proposed solution is adding more complexity into how we can configure OL client. I think it's quite clear, but can be confusing for users, so will definitely prepare a good doc on it.

mobuchowski:
I think increasing complexity in our config in a way that does not increase it linearly (new config var adding something) but by an additional dimension (everything can be configured in the new way and every integration has to care about those special variables) is not a good way to solve this problem. 

>Impossible to easily accept part of the configuration provided

I don't get that. Can you explain why would you need that why override would need to be partial? How would you configure it internally and why would you provide override that you don't want to use? My view of hierarchical config is similar to how depth is drawn in computer graphics: https://en.wikipedia.org/wiki/Z-buffering - you always apply all variables or draw all pixels, but they be overriden (or overdrawn) by something else that has higher priority.

>Has precedence over env var provided configs - default may overwrite user code

So, maybe the solution would be... env var that changes this?

>No safety mechanism, if transport provided in .yaml file does not exist, whole initialization can fail

That's an interesting point but IDK if the reverse of that situation is a good choice. With the config complexity we should have is a traceable config mechanism - for each config var config should be able to tell: I've chosen value X, over Y and Z because of hierarchy V that I have. This is an overkill obviously but even current situation leads to examples like this.

It does not help that integrations can pass config vars directly and they do that bypassing mechanism. Of course this is necessary, to be consistent with upstream methods, but maybe the way to do that would be through some kind of config API that would track the source and priority of this k:v pair.

kacpermuda:
I think this may not be required anymore, will reopen if needed.

PR #3881, opened 2025-07-10
[Python] Remove transport.wait_for_completion()

dolfinus:
### Problem

Address comments from https://github.com/OpenLineage/OpenLineage/issues/3771#issuecomment-3026912591 and https://github.com/OpenLineage/OpenLineage/pull/3812#issuecomment-3026760075.

### Solution

* AsyncHttpTransport and KafkaTransport still have `.wait_for_completion(timeout=-1)` and `.close(timeout=-1)` methods as underlying implementation has proper support for them. Other transport types have only `.close()` which closes all user resources, if any.
* `OpenLineageClient` and `CompositeTransport`/`TransformTransport` have only `.close()` method implemented.

This reduces surface area of the client. Transports which uses client without `.flush()` method don't have to implement it.

Also this synchronizes behavior Python and Java client & transports, as Java has only `void close()`, and no `.waitForCompletion()` https://github.com/OpenLineage/OpenLineage/pull/3839.

### Checklist

----

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -192,17 +192,6 @@ class OpenLineageClient:
         """
         return self.transport.close(timeout)
 
-    def wait_for_completion(self, timeout: float = -1.0) -> bool:
-        """
-        Block until all events are processed or timeout is reached.
-        If the transport is fully synchronous, this method should be a no-op and return True.
-        Params:
-          timeout: Timeout in seconds. `-1` means to block until last event is processed, 0 means no timeout.
-        Returns:
-            bool: True if all events were processed, False if some events were not processed.
-        """
-        return self.transport.wait_for_completion(timeout)
-
     @property
     def config(self) -> OpenLineageConfig:
         """
--- a/client/python/openlineage/client/transport/async_http.py
+++ b/client/python/openlineage/client/transport/async_http.py
@@ -504,9 +504,16 @@ class AsyncHttpTransport(Transport):
         self.event_queue.put(request)
         log.debug("Queued event with event_id %s for async emission to %s", event_id, self.config.url)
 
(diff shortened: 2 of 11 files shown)


dolfinus:
I'm okay witl closing this PR if maintainers deside that `.close(timeout)` and `.wait_for_completion(timeout)` are crucial for OpenLineage client & all transport implementations.

But I doubt that these can be useful for end users. In most cases, transport are blocking anyway, and users want all events to be emitted.

mobuchowski:
@dolfinus I'm ok with it not being respected in some particular cases - like with datazone transport, or when it is a no-op in general. However, lack of client-level timeout prevents us from having timeout in case where it makes sense, as we're not able to do it per-transport, since everyone can write their own custom transport.

mobuchowski:
>Transports don't have to implement methods which have no meaning for them. 

IMO this is still possible with default method being no-op. Not sure why you consider it being other way around than my thinking - I think close/wait_for_completion is an option for transport to do something additional rather than hard requirement. If transport is fully sync and/or stateless and leaves nothing to clean up, there is nothing that it needs to do.

dolfinus:
> However, lack of client-level timeout prevents us from having timeout in case where it makes sense, as we're not able to do it per-transport, since everyone can write their own custom transport.

Okay, we can leave `client.close(timeout)`. But is `client.wait_for_completion()` method really needed for end users? It looks like additional retry mechanism build on top of internal transport retry mechanism.

mobuchowski:
This one we can remove - although we'd really just hide it (as impl detail for close) in the cases where we need to close. 
IMO don't think there's a good need to "flush" the data without closing the client.

dolfinus:
Ok, reverted `.close(timeout)`, only `.wait_for_completion(...)` was removed.

PR #3872, opened 2025-07-08
dbt docker integration tests

mobuchowski:
This PR
- adds new dbt integration tests using docker and duckdb
- removes old ones, basing on the deprecated Airflow integration.

In the future, we can add more detailed duckdb tests, as well as tests using other adapters, like BigQuery and Snowflake.
Also, it's possible to simplify the code a lot by standarizing on uv as the build backend - outside of SQL parser. 

Below is AI generated summary:

  🎯 Purpose & Goals

  - Enable comprehensive dbt integration testing with proper OpenLineage event validation
  - Ensure CI compatibility while maintaining local development efficiency
  - Validate structured logs mode with detailed event hierarchy checking

  📋 Other Effects Accomplished

  - Cross-platform SQL wheel building - works consistently across macOS dev machines and Linux CI
  - Enhanced structured logs validation - verifies complete event hierarchy (job → model → SQL levels)
  - Optimized build process - reuses existing wheels when version/platform match
  - Comprehensive event relationship testing - validates parent-child relationships between event levels

  🔍 Key Improvements

  - Reliable test infrastructure that works in both local and CI environments
  - Detailed event validation ensuring proper OpenLineage event structure and relationships
  - Efficient development workflow avoiding unnecessary rebuilds of expensive SQL components
  - Proper separation of concerns with dedicated structured logs testing

The integration tests now provide thorough validation of the dbt OpenLineage integration's event emission and hierarchy while being optimized for both development speed and CI reliability.

PR #3890, opened 2025-07-11
Adding query ID's for dbt

jroachgolf84:
### Problem

In Airflow, operators like the `SqlExecuteQueryOperator` emit rich lineage data, including query ID. This query ID is not available when using the Astronomer Cosmos project in Airflow.

Closes: NONE

### Solution

By parsing the `run_results`, we're able to parse the `query_id` from the `adapter_response`. This is then added to the `externalQuery` in `run_facets`. This provides the same functionality as what is seen in the `SQLExecuteQueryOperator`.

### Checklist

----


mobuchowski:
@jroachgolf84 some other connectors are using different fields, like BigQuery and Glue, could you make this field depend on adapter type?  

https://github.com/dbt-labs/dbt-bigquery/blob/0995665e490cdee9c408d26aac8e1c19fefaebe0/dbt/adapters/bigquery/connections.py#L62C1-L63C1
https://github.com/aws-samples/dbt-glue/blob/main/dbt/adapters/glue/gluedbapi/commons.py#L30

jroachgolf84:
> We may need to add some tests to cover this change. We may have to update the documentation - is the facet externalQuery new or did it exist before?

Agreed! There are very few unit-tests to be found for dbt, and not speficially for the `DbtArtifactProcessor`. Happy to create some! BTW, I should have marked this PR as a draft.

jroachgolf84:
> @jroachgolf84 some other connectors are using different fields, like BigQuery and Glue, could you make this field depend on adapter type?

@mobuchowski - for these adapters, is the field name not `adapter_response.query_id`? Is that only applicable for Snowflake?

mobuchowski:
@jroachgolf84 I've looked at few, and `query_id` works for Snowflake and Databricks, but not BigQuery (job_id) or Glue (I think it's statement.id). Some others don't have any query id, like Athena or Redshift.

jroachgolf84:
@mobuchowski, perfect, thanks for the heads up. I'll go ahead and update accordingly.

Do you have any recommendations on the best way to test this? I'd lean towards writing tests only for the `get_run()` method to start, which would include tests for multiple different connectors. Thoughts? First time contributing to this code-base for me, and I haven't been able to find docs for running unit/integration tests.

mobuchowski:
@jroachgolf84 we've just added new dbt integration tests: https://github.com/OpenLineage/OpenLineage/pull/3872

however they only run duckdb adapter, and it does not include query id in the adapter response. 
Maybe you're willing to add Snowflake ones - the relevant connection details should be in the CI test environment.

jroachgolf84:
@mobuchowski, I will most certainly take a look at this. In the meantime, do you have any docs about actually executing unit-tests that have been written already?

jroachgolf84:
@mobuchowski, I've gone ahead and added unit-tests for my changes to the `DbtArtifactProcessor`.

integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +211,24 @@ def get_schema_version(cls, metadata):
         str_schema_version = get_from_nullable_chain(metadata, ["metadata", "dbt_schema_version"])
         return cls.get_version_number(str_schema_version)
 
+    def get_query_id(self, run_result) -> str | None:
```
kacpermuda:
```suggestion
    def get_query_id(self, run_result: dict[str, Any) -> str | None:
```

integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +211,24 @@ def get_schema_version(cls, metadata):
...
+        # Validate that there is an adapter_response in the run_result
+        if "adapter_response" not in run_result:
+            return None
+
+        query_id_key: str | None
+
+        # Use the adapter type to make sure the correct key is used
+        if self.adapter_type == Adapter.SNOWFLAKE:
```
kacpermuda: I think for [databricks](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/handle.py#L82) it's also query_id, but in can also be "N/A" instead of None, so we should probably check that as well and return None for it.

integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +211,24 @@ def get_schema_version(cls, metadata):
...
+            query_id_key = "query_id"
+        elif self.adapter_type == Adapter.BIGQUERY:
+            query_id_key = "job_id"
+        else:
+            # TODO: Check with Glue
+            query_id_key = None
```
kacpermuda: I think we can remove it for now, it's already initiated with None. Or if we default to query_id as mentioned in the above comment.

integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -686,14 +709,22 @@ def extract_dialect(self) -> str | None:
...
         run_facets = {
             **self.dbt_version_facet(),
             **self.dbt_run_run_facet(),
             **self.processing_engine_facet(),
         }
         if self._dbt_run_metadata:
             run_facets["parent"] = self._dbt_run_metadata.to_openlineage()
+
+        # If the query_id exists, go ahead and create an externalQuery key-value pair.
+        # This matches what is done  with the SQLExecuteQueryOperator and SnowflakeHook.
```
kacpermuda: Let's remove this comment for now, the dbt integration does not have to know about Airflow and the code itself is clear enough.

integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +211,24 @@ def get_schema_version(cls, metadata):
         str_schema_version = get_from_nullable_chain(metadata, ["metadata", "dbt_schema_version"])
         return cls.get_version_number(str_schema_version)
 
+    def get_query_id(self, run_result) -> str | None:
+        # Validate that there is an adapter_response in the run_result
+        if "adapter_response" not in run_result:
+            return None
+
+        query_id_key: str | None
```
kacpermuda: How about we try to always get `query_id` since it's probably the most common one, and only change the key if we know for sure that some adapter exposes it under different name? In that case, we might support some cases we don't know about, if they expose the query_id under default key.

kacpermuda (commented):
Added some comments. It looks good, but maybe we can improve the default path a bit to support more cases.

integration/common/openlineage/common/provider/dbt/processor.py
kacpermuda: https://github.com/dbt-labs/dbt-adapters/blob/main/dbt-adapters/src/dbt/adapters/contracts/connection.py#L44

integration/common/openlineage/common/provider/dbt/processor.py
jroachgolf84: Added in local branch, will push up!

integration/common/openlineage/common/provider/dbt/processor.py
jroachgolf84: Default implemented!

integration/common/openlineage/common/provider/dbt/processor.py
jroachgolf84: Implemented!

integration/common/openlineage/common/provider/dbt/processor.py
jroachgolf84: Good catch, implemented!

jroachgolf84:
@kacpermuda, I've implemented each of the changes you outlined. Do you mind taking another look?

kacpermuda (approved):
LGTM

PR #3893, opened 2025-07-14
migrate python build system/ci to uv

mobuchowski:
AI description:

This PR modernizes OpenLineage's Python development workflow by migrating from setuptools to UV with path-based dependencies. The change transforms how developers work with the repository: instead of managing separate virtual environments for each integration and manually installing dependencies with pip, developers can now use simple `make` commands like `make setup-all` or `make test-client`. UV's dramatically faster dependency resolution and installation, combined with path-based dependencies between projects (e.g., airflow depending on the local client), means changes propagate immediately without reinstalling packages. A new Makefile provides intuitive commands for common tasks, making it much easier for new contributors to get started and for maintainers to work across multiple integrations.

The migration restructures all Python packages to use the modern `src/` layout and consolidates configuration into `pyproject.toml` files, removing the legacy `setup.py` and `setup.cfg` files. Each integration (client, common, airflow, dbt) now has its own UV-managed environment with a lock file for reproducibility, while path dependencies ensure they reference each other locally during development. A root-level `pyproject.toml` provides shared ruff configuration across all projects, eliminating duplication. The repository now includes a comprehensive `uv.lock` at the root level, ensuring consistent dependency resolution across the entire monorepo.

This is purely an internal development workflow improvement with no changes to published packages or public APIs. All CI workflows have been updated to use `uv sync` for faster builds.


.circleci/continue_config.yml
```diff
@@ -135,6 +135,23 @@ commands:
...
+    parameters:
+      py_env:
+        type: string
+        description: "Python env version"
+        default: "3.9"
+    steps:
+      - run:
+          name: Install uv and Python << parameters.py_env >>
+          command: |
+            sudo apt update && sudo apt install libmysqlclient-dev libabsl-dev
```
JDarDagran: Is there any way this could be cached/skipped? I wonder if ever `apt update` does not take too long for every triggered pipeline in some future.

.circleci/continue_config.yml
```diff
@@ -296,8 +308,10 @@ jobs:
       - image: cimg/python:3.9
     steps:
       - *checkout_project_root
-      - run: python -m pip install build
-      - run: python -m build .
+      - install_uv_python
+      - run: |
+          uv sync --extra dev
```
JDarDagran, line 312: AFAIK `uv sync --extra dev` may install many dependencies and can be slow. Maybe think about caching configured for pip/uv caches so CI isn't too slow?

.circleci/continue_config.yml
```diff
@@ -135,6 +135,23 @@ commands:
...
+      py_env:
+        type: string
+        description: "Python env version"
+        default: "3.9"
+    steps:
+      - run:
+          name: Install uv and Python << parameters.py_env >>
+          command: |
+            sudo apt update && sudo apt install libmysqlclient-dev libabsl-dev
+            curl -LsSf https://astral.sh/uv/install.sh | sh
```
JDarDagran: Shall we pin version of the uv somehow?

.circleci/continue_config.yml
mobuchowski: I don't think it's too slow. Takes 1 second :) 
UV is super fast.

https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/17047/workflows/995f64b7-ac4a-4924-b167-23fcabd75f52/jobs/472072

not much different than before:

https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/17014/workflows/c2674c0f-4532-4611-a3fd-abc6399d0bdf/jobs/470639

.circleci/continue_config.yml
mobuchowski: Pinned to latest release via

```
curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.9.4/uv-installer.sh | sh
```

.circleci/continue_config.yml
mobuchowski: Removed this part

PR #3898, opened 2025-07-15
Upgrade mockito

pawel-big-lebowski:
Manual PR as dependabot won't handle it bcz of `mockito-inline` package that can be removed in most cases


pawel-big-lebowski:
ehhh, this won't work as long as java 8 has to be supported.

PR #3904, opened 2025-07-16
Remove Spark 2 support

pawel-big-lebowski:
Recently, Spark 4 has been released which adds additional complexity to Spark connector codebase and will add more. 
This seems to be the right moment to remove Spark 2 support.


mobuchowski:
Let's do this after next release - in 1.37.0 - if there are no convincing arguments otherwise.

mobuchowski (approved):

PR #3906, opened 2025-07-16
Allow disabling OL from within job's code

pawel-big-lebowski:


integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -94,8 +92,7 @@ public class OpenLineageSparkListener extends org.apache.spark.scheduler.SparkLi
 
   public OpenLineageSparkListener(SparkConf conf) {
     super();
-    this.conf = Objects.requireNonNull(conf).clone();
-    isDisabled = checkIfDisabled();
+    this.conf = Objects.requireNonNull(conf);
```
pawel-big-lebowski, line 95: @d-m-h Do you remember why `clone` was added here? Can we remove it?

d-m-h:
If I recall correctly, it was as a defensive copy of the spark
configuration object.

pawel-big-lebowski:
Looks like it's not needed for my case. 

@d-m-h thanks.

PR #3916, opened 2025-07-19
dbt: more telemetry, fix quadratic file reading

mobuchowski:
We still have issue with dbt integration potentially hanging on very large jobs (>40k emitted OL events). This adds telemetry that allows us to debug those kind of situations.

### DIFF
--- a/client/python/openlineage/client/transport/async_http.py
+++ b/client/python/openlineage/client/transport/async_http.py
@@ -224,7 +224,11 @@ class AsyncHttpTransport(Transport):
             "failed": 0,
         }
 
+        # "Hard" exit. Does not wait for completion for events in flight.
         self.should_exit = threading.Event()
+        # "Soft" exit. Does not exit when there are events in the queue and active tasks
+        self.may_exit = threading.Event()
+
         # Maximum number of concurrent requests
         self.max_concurrent = config.max_concurrent_requests
         self.worker_thread = threading.Thread(target=self._setup_loop, daemon=True)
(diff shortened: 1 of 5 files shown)


client/python/openlineage/client/transport/async_http.py
```diff
@@ -320,6 +325,17 @@ async def _event_loop_worker(self) -> None:
                             log.exception("Error in event processing task")
                 # No tasks are active, didn't process any new items - sleep
                 elif not processed_items:
+                    current_time = time.time()
+                    if current_time - last_processed_time >= 10.0:
+                        log.warning(
```
tnazarew, line 330: why warning?

tnazarew, line 339: why warning?

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -718,6 +726,7 @@ def _run_dbt_command(self) -> Generator[str, None, None]:
             while (
                 self._dbt_log_file is not None and not self.received_dbt_command_completed and i < max_loops
             ):
+                i += 1
```
tnazarew, line 729: 😁

client/python/openlineage/client/transport/async_http.py
mobuchowski: It's quite unusual situation that nothing is being processed for 10 seconds in dbt. We can make this info or every 30 seconds if you think it's too much, but I would like to keep that above DEBUG level.

integration/common/openlineage/common/provider/dbt/structured_logs.py
mobuchowski: :)

client/python/openlineage/client/transport/async_http.py
tnazarew: makes sense

tnazarew (approved):

PR #3933, opened 2025-07-25
add an option to override dbt job name

mobuchowski:
Currently, we use dbt project name as the name of the dbt job level event. We also have an option to use the selector as override. 

However, in many use cases, user defined name would be the best option, especially in the cases where the users run a lot of different subsets of models from single, large, dbt project - that have some business definition.


JDarDagran:
Tests are failing due to this change.

JDarDagran (approved):

PR #3946, opened 2025-07-31
Spark - CLL performance improvements

pawel-big-lebowski:
* Improvements on memory consumption: reuse objects
 * Provide hard limit on amount of input fields returned in the facet (100K)
 * Provide hard limit on amount of dependencies processed (1M)

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilder.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilder.java
@@ -9,6 +9,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.openlineage.client.OpenLineage;
 import io.openlineage.client.OpenLineage.ColumnLineageDatasetFacetFields;
+import io.openlineage.client.OpenLineage.ColumnLineageDatasetFacetFieldsAdditionalBuilder;
+import io.openlineage.client.OpenLineage.SchemaDatasetFacetFields;
 import io.openlineage.client.OpenLineageClientUtils;
 import io.openlineage.client.utils.DatasetIdentifier;
 import io.openlineage.spark.api.OpenLineageContext;
@@ -39,6 +41,9 @@ import org.jetbrains.annotations.NotNull;
 @Slf4j
 public class ColumnLevelLineageBuilder {
 
+  private static final Integer COMPUTED_DEPENDENCY_HARD_LIMIT = 1000000;
+  private static final Integer RETURNED_INPUT_FIELD_LIMIT = 100000;
+
   private Map<ExprId, Set<Dependency>> exprDependencies = new HashMap<>();
   private List<ExprId> datasetDependencies = new LinkedList<>();
   @Getter private Map<ExprId, Set<Input>> inputs = new HashMap<>();
@@ -46,6 +51,10 @@ public class ColumnLevelLineageBuilder {
   private Map<ColumnMeta, ExprId> externalExpressionMappings = new HashMap<>();
   private final OpenLineage.SchemaDatasetFacet schema;
   private final OpenLineageContext context;
+
+  private final Map<ExprId, Dependency> commonDependencies = new HashMap<>();
+
+  private int dependenciesAdded;
 
   public ColumnLevelLineageBuilder(
       @NonNull final OpenLineage.SchemaDatasetFacet schema,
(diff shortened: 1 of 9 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilder.java
```diff
@@ -87,16 +96,29 @@ public void addOutput(ExprId exprId, String attributeName) {
...
       ExprId outputExprId, ExprId inputExprId, TransformationInfo transformationInfo) {
-    exprDependencies
-        .computeIfAbsent(outputExprId, k -> new HashSet<>())
-        .add(new Dependency(inputExprId, transformationInfo));
+    dependenciesAdded++;
+
+    if (dependenciesAdded > COMPUTED_DEPENDENCY_HARD_LIMIT) {
+      // do nothing -> hard limit of allowed dependencies reached
+      return;
+    }
```
ddebowczyk92: each time addDependency() will be called, the `dependenciesAdded` value will be incremented - it makes no sense, if statement should be first

pawel-big-lebowski: yeah, no need to increment the counter. Although the main point is just to stop adding new dependencies.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilder.java
```diff
@@ -162,21 +183,36 @@ public ColumnLineageDatasetFacetFields buildFields(boolean datasetLineageEnabled
     List<TransformedInput> datasetDependencyInputs =
         datasetLineageEnabled ? Collections.emptyList() : datasetDependencyInputs();
 
-    schema.getFields().stream()
-        .map(field -> Pair.of(field, getInputsUsedFor(field.getName())))
-        .filter(pair -> !pair.getRight().isEmpty())
+    ColumnLineageDatasetFacetFieldsAdditionalBuilder additionalBuilder =
+        context.getOpenLineage().newColumnLineageDatasetFacetFieldsAdditionalBuilder();
```
mobuchowski: Create this just before you use it, so you don't allocate for no reason

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilder.java
ddebowczyk92: actually, at the point when the counter hits the limit, the loop that calls addDependency() could be terminated

pawel-big-lebowski: thx

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/column/ColumnLevelLineageBuilder.java
pawel-big-lebowski: Ok, I moved it here so that this gets created once. 
Will fix this once the general approach is approved.

ddebowczyk92 (approved):

mobuchowski (approved):

luke-hoffman1:
@pawel-big-lebowski, I don't see any logging that shows whether these limits are reached. If the limits are reached, wouldn't this mean that partial or missing lineage metadata could go unnoticed?

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/ColumnLevelLineageUtils.java
```diff
@@ -57,11 +59,10 @@ public static Optional<OpenLineage.ColumnLineageDatasetFacet> buildColumnLineage
         olContext.getOpenLineage().newColumnLineageDatasetFacetBuilder();
 
     boolean datasetLineageEnabled =
-        context
-            .getOlContext()
-            .getOpenLineageConfig()
-            .getColumnLineageConfig()
-            .isDatasetLineageEnabled();
+        Optional.of(context.getOlContext().getOpenLineageConfig())
```
ddebowczyk92, line 62: Is this Optional  needed since we initialize new config object on get if it's null
```public ColumnLineageConfig getColumnLineageConfig() {
     if (columnLineageConfig == null) {
       columnLineageConfig = new ColumnLineageConfig();```

pawel-big-lebowski: I added it as I saw NPE in some of the tests. Perhaps this was `schemaSizeLimit` being null when only datasetLineage config being set. 

Similarly here, we can have non-nullable `ColumnLineageConfig` with `getDatasetLineageEnabled=null`

PR #3950, opened 2025-08-04
python: add datadog transport

mobuchowski:
The DatadogTransport sends OpenLineage events to Datadog's data collection endpoints around the world (US, EU, Asia-Pacific regions). It needs a Datadog API key and site setting, but can also get these from DD_API_KEY and DD_SITE environment variables if not provided directly.

  The transport uses two different ways to send events: async HTTP for dbt events (found by checking if JobTypeJobFacet.integration="dbt") to handle lots of data transformation events quickly, and regular HTTP for everything else. It automatically sets up gzip compression, retry attempts, and proper authentication for both methods.

  You can configure standard HTTP settings like timeout and retry options, plus async settings like queue size and how many requests to send at once. The transport converts site names to the correct Datadog URLs (like datadoghq.com becomes https://data-obs-intake.datadoghq.com) and works with OPENLINEAGE__TRANSPORT__TYPE=datadog. It properly handles
  shutdown by closing both transport methods, making sure dbt workflows get fast async processing while other data pipelines get reliable regular delivery.

### DIFF
--- a/client/python/openlineage/client/transport/datadog.py
+++ b/client/python/openlineage/client/transport/datadog.py
@@ -1,0 +1,164 @@
+# Copyright 2018-2025 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+from __future__ import annotations
+
+import os
+from typing import TYPE_CHECKING, Any
+from urllib.parse import urlparse
+
+import attr
+from openlineage.client.transport.async_http import AsyncHttpConfig, AsyncHttpTransport
+from openlineage.client.transport.http import HttpConfig, HttpTransport
+from openlineage.client.transport.transport import Config, Transport
+from openlineage.client.utils import get_only_specified_fields
+
+if TYPE_CHECKING:
+    from openlineage.client.client import Event
+
+
+# A mapping from https://docs.datadoghq.com/getting_started/site/ to
+# Datadog endpoint that accepts OL events.
+SITE_MAPPING = {
+    "datadoghq.com": "https://data-obs-intake.datadoghq.com",
+    "us3.datadoghq.com": "https://data-obs-intake.us3.datadoghq.com",
+    "us5.datadoghq.com": "https://data-obs-intake.us5.datadoghq.com",
+    "datadoghq.eu": "https://data-obs-intake.datadoghq.eu",
+    "ddog-gov.com": "https://data-obs-intake.ddog-gov.com",
+    "ap1.datadoghq.com": "https://data-obs-intake.ap1.datadoghq.com",
+    "ap2.datadoghq.com": "https://data-obs-intake.ap2.datadoghq.com",
+    "datad0g.com": "https://data-obs-intake.datad0g.com",
+}
+
+
+def _is_valid_url(url: str) -> bool:
+    try:
+        parsed = urlparse(url)
+        return parsed.scheme in ("http", "https") and bool(parsed.netloc)
+    except Exception:
+        return False
+
+
+@attr.define
+class DatadogConfig(Config):
+    apiKey: str = attr.field()  # noqa: N815
+    site: str = attr.field(default="datadoghq.com")
+    timeout: float = attr.field(default=5.0)
+    retry: dict[str, Any] = attr.field(
+        default={
+            "total": 5,
+            "read": 5,
+            "connect": 5,
+            "backoff_factor": 0.3,
+            "status_forcelist": [500, 502, 503, 504],
+            "allowed_methods": ["HEAD", "POST"],
+        }
+    )
+    max_queue_size: int = attr.field(default=10000)
+    max_concurrent_requests: int = attr.field(default=100)
+    async_transport_rules: dict[str, dict[str, bool]] = attr.field(default={"dbt": {"*": True}})
+
+    @classmethod
+    def from_dict(cls, params: dict[str, Any]) -> DatadogConfig:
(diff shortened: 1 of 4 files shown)


client/python/openlineage/client/transport/datadog.py
```diff
@@ -0,0 +1,134 @@
...
+        # Check if event has JobTypeJobFacet with integration = "dbt"
+        should_use_async = self._should_use_async_transport(event)
+
+        if should_use_async:
+            return self.async_http.emit(event)
+        else:
+            return self.http.emit(event)
+
+    def close(self, timeout: float = -1) -> bool:
+        http_result = self.http.close(timeout)
```
dolfinus, line 127: Why using async transport for all event types, not just run+dbt?

mobuchowski: Most of the other integrations will emit event maybe once (like Airflow worker) or maybe every few seconds, while dbt structured logs can emit 100s or 1000s events per second. In the first variant I think starting the thread and setting up the event loop is overkill.

dolfinus: What about custom integrations? These can emit lots of events as well

mobuchowski: @dolfinus added option to specify custom mapping for which integration/jobtype would use particular transport.

client/python/openlineage/client/transport/datadog.py
```diff
@@ -0,0 +1,158 @@
...
+            return False
+
+        job_facets = event.job.facets
+        if not job_facets:
+            return False
+
+        job_type_facet = job_facets.get("jobType")
+        if not job_type_facet:
+            return False
+
```
kacpermuda: I think we should add a possibility for user to enforce async, maybe some value in config `use_async` ?

client/python/openlineage/client/transport/datadog.py
dolfinus: I'd rather add explicit config option for async than introduce some integration-specific logic into transport itself

client/python/openlineage/client/transport/datadog.py
mobuchowski: Added explicit config option

client/python/openlineage/client/transport/datadog.py
mobuchowski: You can use 
```
        "*": {"*": True}
```
then.

dolfinus: For me, `use_async=True` is much more readable than `async_transport_rules = {"*": {"*": True}}`. The latter looks overcomplicated

mobuchowski: But this allows us to specify and evolve those rules for users without users having to understand configuration too much. With just use_async we lose the option to specify that specifically for the dbt integration. The configurability here is a secondary goal. 
The goal is to simplify onboarding and instructions, similar to existing GCP and AWS transports, as well as turn on options that we know are working for Datadog but might not in general - like compression.

dolfinus: > With just use_async we lose the option to specify that specifically for the dbt integration.

This option could be mentioned as recommended in dbt integration docs

> without users having to understand configuration too much

At the same time, there is a dedicated documentation section specifically for `async_transport_rules`

mobuchowski: Right, added that in the docs.

website/docs/client/python.md
```diff
@@ -405,6 +405,167 @@ client.close()
...
+The `async_transport_rules` configuration allows fine-grained control over which events use asynchronous transport vs synchronous HTTP transport. Rules are defined as a two-level dictionary:
+
+```yaml
+async_transport_rules:
+  <integration>:
+    <jobType>: <boolean>
+```
+
+First-level keys match against the `integration` field in `JobTypeJobFacet` Second-level keys match against the `jobType` field in `JobTypeJobFacet`.
+Valuye `true` uses async transport, `false` or lack of value uses synchronous HTTP transport.
```
kacpermuda:
```suggestion
Value `true` uses async transport, `false` or lack of value uses synchronous HTTP transport.
```

website/docs/client/python.md
```diff
@@ -405,6 +405,167 @@ client.close()
...
+Use `"*"` to match all integrations or job types. All matching is case-insensitive.
+
+When the mapping for some `integration` - `jobType` pair aren't provided, it will use synchronous HTTPTransport. 
+If you want to send all events via async transport, use double wildcard configuration.
+
+```yaml
+async_transport_rules:
+  "*":
+   "*": true
+```
```
kacpermuda: I think it's worth to underline that `*` rules don't apply if integration (job type facet) is not found. Even if you use *, you might end up using sync http. It's mentioned above, but I feel like it can be confusing.

kacpermuda (approved):
Looks good. What @dolfinus mentioned is also valid, configuring async vs sync looks complex, but it's doable so approving the PR.

website/docs/client/python.md
mobuchowski: I think it's better to override that behavior to make double wildstar send the async events. Will fix that.

website/docs/client/python.md
mobuchowski: Fixed

PR #3969, opened 2025-08-14
feat: Introduce OpenlineageClientRunFacet with OL client version

kacpermuda:
### Problem

When processing OpenLineage (OL) events, it can be challenging to determine the version of the OpenLineage client that generated the event. This version information is often important for the consumer’s backend systems.

For direct use of the OL client, the `producer` field in the event provides this information. However, integrations that utilize the Python client, such as Airflow, overwrite this field with their own value, making it difficult to trace the originating client version.

### Solution

To address this, I’ve introduced a new facet called `OpenlineageClientRunFacet`, which stores information about the OpenLineage client used to emit the event. Currently, it holds only the client version, but this can be expanded in the future to include additional metadata, such as transport information, if necessary. This facet is automatically added to all events just before they are emitted, similar to how the environment variable facet and tags facets are handled.

#### Alternative Considered: Expanding `ProcessingEngineRunFacet`

I considered expanding the [`ProcessingEngineRunFacet`](https://openlineage.io/docs/spec/facets/run-facets/processing_engine) to include the `openlineageClientVersion` property, but ultimately this was not a suitable solution for the following reasons:

- **Facet Creation Requires Integration Information**: We cannot create this facet solely within the client, as it requires information about the integration used (including the integration version). If an event doesn't already contain this facet, I wouldn't be able to add it from within the client. My goal is for all OpenLineage events to include this facet, not just those that already have the processing engine run facet.
  
- **Modifying Existing Facets**: To achieve this approach, I would have to modify an already created facet in the event (similar to how we modify tags facets). While this is technically possible, I don't consider it a clean solution, so I decided to avoid it.

- **Client-Specific Information**: Since this information pertains to the client itself, I don't think it should be dependent on the integration creating the processing engine facet. The client already has the necessary information, so it makes more sense to encapsulate the addition of this information within the client library itself.

- **Inconsistent Facet Availability in Integrations**: In integrations like Airflow, the existence of the processing engine run facet is not stable across all events. Not all events include this facet in older versions. For example, in Airflow, the processing engine facet was initially added to task start events, then to DAG start events in a later version, and eventually to other events in subsequent releases. This inconsistency means that, if we relied on modifying the processing engine facet, we'd face gaps in information, especially for users on older versions of OpenLineage (e.g., Airflow 2.9, where upgrading to the latest provider is no longer possible).

#### One-line summary:

feat: Introduce OpenlineageClientRunFacet with OL client version

### Checklist

----


kacpermuda:
Maybe in the OL2.0 or 3.0 spec we can adjust that and somehow expand the producer field in the event to hold that information, but until then I think this facet is a good workaround that is also backward compatible. Added that idea to the issue with ideas for breaking changes.

kacpermuda:
Will try to solve this another way, with #3980

Issue #3096, opened 2024-09-16
[FEATURE] Add support for conditionals: NULLIF, NVL, NVL2, COALESCE


jphalip:
### Motivation

Currently CASE-WHEN and IF statements are covered for conditionals. It'd be good to add support for other UDFs like: NULLIF, NVL, NVL2, COALESCE

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

PR #3972, opened 2025-08-18
spark: Add support for COALESCE transformation

kchledowski:
### Problem

Column level lineage currently has support for `IF` and `CASE WHEN` conditional transformations, but other conditionals (mainly `NULLIF`, `NVL`, `NVL2` and `COALESCE`) are treated as generic transformations.

Closes: #3096

### Solution

Added support for `COALESCE` conditional transformation. There is no need for any specific handling of `NULLIF`, `NVL` and `NVL2`, as those are parsed by Spark into one of the already supported conditionals (`IF`, `CASE WHEN` or `COALESCE`) and that's how they appear in the Logical Plan

#### One-line summary:

Add support for COALESCE transformation

### Checklist

----

### DIFF
--- a/integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/column/ExpressionDependencyCollectorTest.java
+++ b/integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/column/ExpressionDependencyCollectorTest.java
@@ -75,6 +75,7 @@ import scala.collection.immutable.Seq;
 
 class ExpressionDependencyCollectorTest {
 
+  final String ALIAS_NAME = "res";
   final String NAME1 = "name1";
   final String NAME2 = "name2";
   final String NAME3 = "name3";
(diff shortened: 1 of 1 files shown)


dolfinus:
Doesn't `INDIRECT` mean that the entire dataset depends on specified column? In case `SELECT COALESCE(a, b) AS cond` its only `cond` column which depends on `a` and `b` input columns.

kchledowski:
> Doesn't `INDIRECT` mean that the entire dataset depends on specified column? In case `SELECT COALESCE(a, b) AS cond` its only `cond` column which depends on `a` and `b` input columns.

It's true that in most cases where the entire dataset has a dependency (e.g. `FILTER`, `SORT`, `JOIN`), the type would be `INDIRECT`,  but that is not the only usage for it. 
`INDIRECT` is meant to be used when the output depends on the input, but it's not derived from it, which is exactly what happens in the example you've provided with coalesce. 

More details:
* [OpenLineage – Expression dependency collection process](https://openlineage.io/docs/integrations/spark/spark_column_lineage#expression-dependency-collection-process)
* [OpenLineage – Transformation Type](https://openlineage.io/docs/spec/facets/dataset-facets/column_lineage_facet#type)

tnazarew:
@dolfinus we have special field in CLL facet called `datasetDependencies` that has all those dependencies, `INDIRECT` doesn't have those restrcitions, it already works like that for case when and if statements

dolfinus:
IIRC, `datasetDependencies` was introduced to reduce size of CLL (instead of MxN matrix return only N), and it doesn't have semantic implications.

PR #3980, opened 2025-08-20
feat: Add default tags with OL client version

kacpermuda:
Solving the problem described in #3969, in a better way, without introducing new facet.

I also moved the injection of env vars and tags (so all the pre-emit operations) to before we log the event on debug level, so that the debug is more accurate and contains the actual event emitted later on.

Example tag:
```
      "tags": {
        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.38.0/client/python",
        "_schemaURL": "https://openlineage.io/spec/facets/1-0-0/TagsRunFacet.json#/$defs/TagsRunFacet",
        "tags": [
          {
            "key": "openlineage_client_version",
            "source": "OPENLINEAGE_CLIENT",
            "value": "1.38.0"
          }
        ]
      }
```

Not sure how we should name the `key` and the `source`, I proposed the one above. If you have any better ideas lmk.

----

### DIFF
--- a/client/python/openlineage/client/client.py
+++ b/client/python/openlineage/client/client.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, Union, cast
 
 import attr
 import yaml
-from openlineage.client import event_v2
+from openlineage.client import constants, event_v2
 from openlineage.client.facet_v2 import environment_variables_run, tags_job, tags_run
 from openlineage.client.facets import FacetsConfig
 from openlineage.client.filter import Filter, FilterConfig, create_filter
@@ -159,6 +159,9 @@ class OpenLineageClient:
             msg = "`emit` only accepts RunEvent, DatasetEvent, JobEvent classes"
             raise ValueError(msg)
 
+        event = self.add_environment_facets(event)
+        event = self.update_event_tags_facets(event)
+
         if log.isEnabledFor(logging.DEBUG):
             val = Serde.to_json(event).encode("utf-8")
             log.debug("OpenLineageClient will *try* to emit event %s", val)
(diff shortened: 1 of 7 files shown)


mobuchowski (approved):

client/python/openlineage/client/client.py
```diff
@@ -443,29 +443,46 @@ def _collect_environment_variables(self) -> dict[str, str]:
...
+        ]
+        default_job_tags = [
+            tags_job.TagsJobFacetFields(key, value, "OPENLINEAGE_CLIENT")
+            for (key, value) in _default_tags.items()
+        ]
+        return [*user_job_tags, *default_job_tags]
+
+    @property
+    def _run_tags(self) -> list[tags_run.TagsRunFacetFields]:
+        _default_tags: dict[str, str] = {"openlineage_client_version": constants.__version__}
```
JDarDagran, line 460: let's move it top-level maybe?

kacpermuda: Yeah, that was my initial approach, but then I figured out that maybe people may want to add some runtime info here as well? Like transport type or something? I initially even stored those tags in `constants.py`, but it'd require refactor to get runtime info then, so I kept it here.

JDarDagran: Then runtime tags can be added here in the method in addition to the default "constant" tags.

kacpermuda: We only have one default run tag for now, not sure if we need to split static and runtime tags yet. Maybe let's wait for some more tags to be added and split them once it gets harder to track?

JDarDagran: Sure

PR #3985, opened 2025-08-22
Spark: optimize lineage extraction when reading a lot of S3 objects

pawel-big-lebowski:
Currently, `io.openlineage.spark.agent.util.PlanUtils#getDirectoryPath` method checks for each asset read if this is a file or directory. In case of a regular file, it returns its parent directory. This leads to problems when a method is called on hundreds of thousands of files and each and each call requires `getFileStatus` call to S3. 

Actually, these calls are not necessary as all the files come from the same directory. Before running `getFileStatus`, the code can check if the parent dir was already added.

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PlanUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PlanUtils.java
@@ -12,9 +12,11 @@ import io.openlineage.spark.agent.Versions;
 import io.openlineage.spark.api.naming.NameNormalizer;
 import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PlanUtils.java
```diff
@@ -250,7 +252,34 @@ public static OpenLineage.ParentRunFacet parentRunFacet(
...
+    LinkedHashSet<Path> normalizedPaths = new LinkedHashSet<>();
+    for (Path path : paths) {
+      // check if the root path is already contained in normalized Paths
+      boolean isContained = false;
+      for (Path normalizedPath : normalizedPaths) {
+        if (path.toString().startsWith(normalizedPath.toString())) {
+          isContained = true;
+        }
+      }
+      if (!isContained) {
```
dolfinus:
```suggestion
      Path parent = path.getParent();
      if (parent != null && !normalizedPaths.contains(parent)) {
```

1. Iterating 2 collections in nested for loops gives O(n^2) complexity
2. Compairing string representations is tricky because `"/some/path1/file".contains("/some/path") == true`, but file is not inside this directory

pawel-big-lebowski: 1. `O(n^2)` would occur for each file in a separate directory.
2. Making sure `/` is at the  end of `normalizedPath` should help.
3. The solution provided does not support case with `normalized-data-dir` and `normalized-data-dir/path1/path2/`. I don't know if this is even possible, but looks same impossible to me as `1.`

dolfinus: 3. `getParent()` [truncates last slash](https://github.com/apache/hadoop/blob/0d7efae8734f96a64caf7a4b808c27ce845886c5/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/Path.java#L463) from parent directory path. If LogicalRelation contains both `/some/path/file` and `/some/path/`, then HashSet will contain `/some/path` and `/some/path/`, which should be normalized later into `/some/path` [FilesystemDatasetExtractor](https://github.com/OpenLineage/OpenLineage/blob/80b81356512dda99ff6ec584690825698cbde1e5/client/java/src/main/java/io/openlineage/client/utils/filesystem/GenericFilesystemDatasetExtractor.java#L38)

pawel-big-lebowski: @dolfinus cool, let's do that, thx.

ddebowczyk92 (approved):

PR #3987, opened 2025-08-22
update DbtRunRunFacet to contain project name and more

mobuchowski:
With ability to overwrite job name we don't have a place where dbt project name is always passed. This adds few fields that would contain that information and few other relevant info.


integration/common/openlineage/common/provider/dbt/cloud.py
```diff
@@ -32,13 +32,14 @@ def __call__(self, *args, **kwargs):
 class DbtCloudArtifactProcessor(DbtArtifactProcessor):
     should_raise_on_unsupported_command = False
 
-    def __init__(self, manifest, run_result, profile, catalog, *args, **kwargs):
+    def __init__(self, manifest, run_result, profile, catalog, project_name, *args, **kwargs):
```
kacpermuda: I think adding new required arg with no default value will break workflows that already use this. F.e. [DbtCloudRunJobOperator](https://github.com/apache/airflow/blob/main/providers/dbt/cloud/src/airflow/providers/dbt/cloud/utils/openlineage.py#L179). Any way we can make it optional or provide some default value?

integration/common/openlineage/common/schema/dbt-run-run-facet.json
```diff
@@ -14,9 +14,25 @@
...
+              "type": "str",
+              "example": "0.20.1"
+            },
+            "profile_name": {
+              "type": "str",
+              "example": "default"
+            },
+            "dbt_runtime": {
+              "type": "str",
+              "example": "cloud"
```
kacpermuda, line 32: Is this just `core` or `cloud` ? What would be other expected values for this? How many possibilities do we expect here ?

integration/common/openlineage/common/provider/dbt/cloud.py
```diff
@@ -67,3 +68,16 @@ def get_dbt_metadata(self):
...
+        invocation_id = self.run_metadata.get("invocation_id")
+        if not invocation_id:
+            return {}
+        return {
+            "dbt_run": DbtRunRunFacet(
+                invocation_id=invocation_id,
+                project_name=self.project_name,
+                profile_name=self.profile["name"],
+                dbt_runtime="cloud",
+            )
```
kacpermuda, line 82: We have no access to project_version on dbt cloud ? No idea, just double checking.

integration/common/openlineage/common/schema/dbt-run-run-facet.json
mobuchowski: I feel any future distinguishable flavor of dbt could be captured by this field too. With bad experiences with enums in spec previously I deliberately choose to underspecify this field.

integration/common/openlineage/common/provider/dbt/cloud.py
mobuchowski: Right, updated this with default `None`

integration/common/openlineage/common/provider/dbt/cloud.py
mobuchowski: No, there is no project version AFAIK. https://docs.getdbt.com/dbt-cloud/api-v3#/operations/Retrieve%20Project

PR #3995, opened 2025-08-26

pawel-big-lebowski:


pawel-big-lebowski:
continue with another PR: https://github.com/OpenLineage/OpenLineage/pull/3996

PR #3996, opened 2025-08-26

pawel-big-lebowski:
different approach - json schema fields added in a single file


JDarDagran:
@pawel-big-lebowski  your PR is a good playground for me to experiment with code generation. I reverted #3997 and introcuded couple of changes needed to handle type aliasing.

@kacpermuda I could revert the changes as there was bug fix in `datamodel-code-generator` for what I tried to fix.

pp-eyushin:
👍

website/docs/spec/facets/dataset-facets/subset_definition.md
```diff
@@ -0,0 +1,25 @@
...
+# Datasource Facet
+
+Example:
+
+```json
+{
+    ...
+    "inputs": {
+        "facets": {
+            "dataSource": {
```
dolfinus: This probably should be datasetSubset

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/DatasetNameTrimmer.java
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @param name
+   * @return
+   */
+  default String trim(String name) {
+    return canTrim(name) ? removeLastPath(name) : name;
+  }
+
+  /**
+   * Returns the last path of the dataset name.
```
mobuchowski: What is `last path of dataset name`?

client/java/src/main/java/io/openlineage/client/dataset/partition/ReducedDataset.java
```diff
@@ -0,0 +1,108 @@
...
+    this.reducedPaths = new ArrayList<>();
+    if (!Objects.equals(trimmedPath, dataset.getName())) {
+      // reduced paths are different from trimmed path -> add it
+      this.reducedPaths.add(dataset.getName());
+    }
+    this.dataset = dataset;
+  }
+
+  /**
+   * Given another dataset, it reduces and merges the two datasets. If the paths of the two datasets
```
mobuchowski: What do you mean by reducing? What is `trimming` vs `reducing` here? How are those getting merged? I think those terms need to be clearly defined.

website/docs/spec/facets/dataset-facets/subset_definition.md
pawel-big-lebowski: @dolfinus This file shall be removed. I'll create separate PR to document subset definition facets. Thanks for spotting this.

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/DatasetNameTrimmer.java
pawel-big-lebowski: If a dataset name is a path, then the last path should return trailing directory.

client/java/src/main/java/io/openlineage/client/dataset/partition/ReducedDataset.java
pawel-big-lebowski: `trimming` is a string operation and trimms dataset name. `Reducing` works on a collection of datasets. If a collection has dataset with the same trimmed name, and the same facets, those datasets should be merged into a single dataset with a trimmed name. Merging process is a reducer. I'll add more comments in the code.

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/MultiDirTrimmer.java
```diff
@@ -0,0 +1,86 @@
...
+import java.time.LocalDate;
+import java.time.YearMonth;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import java.util.Arrays;
+
+/**
+ * Trims directory if last path is a string represents a date in a format /yyyy/MM or /yyyy/MM/dd
+ */
+public class MultiDirTrimmer implements DatasetNameTrimmer {
```
dolfinus:
```suggestion
public class MultiDirDateTrimmer implements DatasetNameTrimmer {
```
Current name can be misleading

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/DatasetNameTrimmer.java
mobuchowski: Maybe `path` -> `part`? 
That's what Python: https://docs.python.org/3/library/pathlib.html#accessing-individual-parts
and Java https://docs.oracle.com/javase/8/docs/api/java/nio/file/Paths.html
use.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
```diff
@@ -264,6 +265,11 @@ private List<InputDataset> buildInputDatasets(List<Object> nodes) {
                                 .map(((Class<InputDataset>) InputDataset.class)::cast))
                     .orElse(Stream.empty()))
             .collect(Collectors.toList());
+    datasets =
```
mobuchowski, line 268: Not for this PR, but it would be amazing if someday we could refactor the integration to something that has well-defined places where extraction and later processing of individual parts of the eventual run event takes place. Right now we're doing "random" modifications and sticking them in "random" places in the codebase. 

This is confusing to someone who wants to either add/modify handling particular logical plan node, as it's not well defined what will happen to the results of that change, and to someone who wants to apply some more global processing. Also, it's even confusing to someone who has seen this code a decent amount of time (like me). To go down few levels a bit, it would be best if `OpenLineageRunEventBuilder` class would not do any processing, but just assemble the event with given parts.

pawel-big-lebowski: Great suggestion!

pawel-big-lebowski:
@dolfinus @mobuchowski I applied changes from your comments within the second commit.

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/KeyValueTrimmer.java
```diff
@@ -0,0 +1,19 @@
...
+package io.openlineage.client.dataset.partition.trimmer;
+
+/** Normalizes if last part is a string represents a key value pair in an arbitrary format. */
+public class KeyValueTrimmer implements DatasetNameTrimmer {
+
+  private static final char EQUALITY_SIGN = '=';
+
+  @Override
+  public boolean canTrim(String name) {
+    String lastPart = getLastPart(name);
```
dolfinus, line 17: IMHO this trimmer can detect the _first_ part containing `=` symbol, and trim it with the rest of the string, instead of iterating over last path part until everything is trimmed. But current approach works as well

client/java/src/test/java/io/openlineage/client/dataset/partition/trimmer/KeyValueTrimmerTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+  KeyValueTrimmer trimmer = new KeyValueTrimmer();
+
+  @Test
+  void testTrim() {
+    // nothing to normalize
+    assertThat(trimmer.canTrim("/tmp/cat")).isFalse();
+
+    // normalize key=value
+    assertThat(trimmer.canTrim("/tmp/key=value")).isTrue();
```
dolfinus, line 21: Let's add case with multiple `key=value` directories nested

dolfinus: Also, maybe `.trim()` method should be asserted as well

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/KeyValueTrimmer.java
pawel-big-lebowski: The mechanism needs to support the mixed trimming like `/path/key=value/20250930/key2=value2`, so for the sake of simplicity a trimmer has only to trim the last part. At least, I didn't want to implement same thing for all the trimmers.

dolfinus: MultiDirDateTrimmer already works with several parts, not just last one.

My concern is that for path like `/path/key=value/randomvalue/key2=value2` dataset name will be `/path/key=value/randomvalue`, not `/path`. But this is not a real case, IMHO the whole mixed naming conventions for paths (both partition by `key=value` and `yyyy/mm/dd`) is something that can't appear in real paths, it's either one or another. But I may be wrong :)

pawel-big-lebowski: Yes, `/path/key=value/not-date-value/key2=value2` will result in `/path/key=value/not-date-value`. This is intentional as I did want to trim parts a mechanism does not understand. Let's leave it for later ;) 

But, `/path/key=value/20250930/key2=value2` will be trimmed to `/path` - this is implemented within `trimDatasetName` of `io.openlineage.client.dataset.partition.ReducedDataset`

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
```diff
@@ -0,0 +1,188 @@
...
+
+    return reduced.stream()
+        .map(
+            r -> {
+              OutputDataset source = (OutputDataset) r.getDataset();
+              OutputDatasetOutputFacetsBuilder facetsBuilder =
+                  FacetUtils.toBuilder(openLineage, source.getOutputFacets());
+              if (!r.getNonTrimmedNames().isEmpty()) {
+                facetsBuilder.put(
+                    "subset",
```
dolfinus, line 124: nit: One method uses `.subset(...)`, another `.put("subset", ...)`

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
```diff
@@ -0,0 +1,188 @@
...
+                      return l1;
+                    }));
+
+    // Reducing logic is a bit more complex than just creating map based on trimmed name
+    // If there are datasets with the same trimmed name but different facets, we can't reduce them
+    // So, when having a list of dataset with the same trimmed name, we need to try reducing each
+    // one with each other
+    List<ReducedDataset> reducedDatasets = new ArrayList<>();
+    for (List<ReducedDataset> sameNameList : toReduce.values()) {
+      AtomicBoolean reducedSomething = new AtomicBoolean(true);
```
dolfinus: Is atomic variable required? It's not a object or class level attribute, so how other threads can access it?

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/MultiDirDateTrimmer.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @SuppressWarnings("PMD.AvoidLiteralsInIfCondition")
+  private boolean isMultiDirDate(String... dirs) {
+    if (dirs.length < 3) {
+      return false;
+    }
+
+    // get three last parts and verify if they're a date
+    String lastTwoParts = dirs[dirs.length - 3] + dirs[dirs.length - 2] + dirs[dirs.length - 1];
```
dolfinus:
```suggestion
    String lastThreeParts = dirs[dirs.length - 3] + dirs[dirs.length - 2] + dirs[dirs.length - 1];
```

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
```diff
@@ -0,0 +1,188 @@
...
+                        .outputCondition(
+                            openLineage
+                                .newLocationSubsetConditionBuilder()
+                                .locations(r.getNonTrimmedNames())
+                                .build())
+                        .build());
+              }
+              return openLineage
+                  .newOutputDatasetBuilder()
+                  .name(r.getTrimmedDatasetName())
```
dolfinus, line 136: Trimming should probably be applied to columnLineage facet as well, to keep valid references to input dataset. Maybe in another PR

client/java/src/test/java/io/openlineage/client/dataset/partition/trimmer/DateTrimmerTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.jupiter.api.Test;
+
+class DateTrimmerTest {
+
+  DateTrimmer trimmer = new DateTrimmer();
+
+  @Test
+  void testTrim() {
+    assertThat(trimmer.canTrim("/tmp/20250721")).isTrue();
```
dolfinus, line 17: Please add test cases when dataset name doesn't start with `/` (S3 object name) and doesn't contain `/` at all (table names, some may include date as a part of name)

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/OpenLineageRunEventBuilder.java
```diff
@@ -264,6 +265,11 @@ private List<InputDataset> buildInputDatasets(List<Object> nodes) {
...
+        new DatasetReducer(
+                openLineageContext.getOpenLineage(),
+                openLineageContext.getOpenLineageConfig().getDatasetConfig())
+            .reduceInputs(datasets);
```
dolfinus, line 272: PR title and description should be adjusted to mention that not *object* storage are handled here - S3 is an object storage, HDFS is not, both are trimmed.
Trimmers are applied to all dataset names, even if they don't contain any paths, like table name, but in this case dataset name is left intact.

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
pawel-big-lebowski: This is the result of how Java code is generated for `oneOf` in json schema. I don't think there's any good and backward compatible solution for this. At least, not the one I am capable of implementing within this PR.

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
pawel-big-lebowski: That's a great catch. Thank you!

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
dolfinus: Ok, got your point

client/java/src/main/java/io/openlineage/client/dataset/partition/trimmer/KeyValueTrimmer.java
dolfinus: Ok, let's keep current implementation for now

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
pawel-big-lebowski: no, it's not required

client/java/src/main/java/io/openlineage/client/dataset/partition/DatasetReducer.java
pawel-big-lebowski: I'll prepare another PR for that.

client/java/src/test/java/io/openlineage/client/dataset/partition/trimmer/DateTrimmerTest.java
```diff
@@ -0,0 +1,38 @@
...
+    assertThat(trimmer.canTrim("/tmp/20250722T901Z")).isTrue();
+
+    // non date alike patterns
+    assertThat(trimmer.canTrim("/tmp/20122321")).isFalse();
+    assertThat(trimmer.canTrim("/tmp/dataset_20122321")).isFalse();
+  }
+
+  @Test
+  void testTrimNonPath() {
+    assertThat(trimmer.canTrim("tmp/20250721")).isFalse();
```
dolfinus: So trimmers are no applied to object storage paths which not start with `/`, e.g. S3?

pawel-big-lebowski: Oh, I forgot S3 name does not start with `/`. 
I changed the trimmers, so that `canTrim` returns false if trimming resulted in empty string. 
This covers table names which may match trimming conditions.

dolfinus: Ok, but please add tests fro S3 paths which are not trimmed to empty string

pawel-big-lebowski: ok, makes sense. I added that.

dolfinus: Thanks!

mobuchowski (approved):

PR #3997, opened 2025-08-26
Python: fix classes generation for almost identical classes.

JDarDagran:
### Problem

Currently, when defining two classes that use
the same object in the same field, generator would fail.

### Solution

Internals require to import specific objects rather than the entire module when deduplicating models.
### Checklist

----


kacpermuda (approved):
LGTM, simple import change. Why exactly is this needed? What failed before?

JDarDagran:
> LGTM, simple import change. Why exactly is this needed? What failed before?

I didn't go too deep into the code of `datamodel-code-generator` but seems like there's a bug when defining below example:
1. Define 2 classes in JSON schema. Different names but same field name - give it `special`. The classes differ, however, because in one `special` is `SpecialInputField`, in the second `SpecialOutputField`.
2. `datamodel-code-generator` attempts to deduplicate models by hashing classes structure.
3. hashing logic takes all imports needed to define class - in our case this is `from openlineage.client.utils.RedactMixin` and `import attr`. The code may have a bug as it can't hash entire module import for some reason.
4. When I changed `import attr` into `from attr import ...` it works back again for the given example.

JDarDagran:
@pawel-big-lebowski is interested as he experienced the bug.

kacpermuda:
Sounds good. It has no impact on our code so if it helps then we're good to merge.

PR #4003, opened 2025-08-29
spark: Refactor CLL ExpressionDependencyCollector with pluggable visitor pattern

kchledowski:
### Problem

The current `ExpressionDependencyCollector` uses two large `if-else` chains to handle different Logical Plan operators (`Aggregate`, `Join`, etc.) and expressions (`Alias`, `CaseWhen`, etc.). This makes the code hard to read and work with.
It is also inconsistent with the already-existing `NodeVisitor` pattern used for some of the Logical Plan operators.

### Solution

- Apply the same visitor pattern for all remaining operators
- Extract Expression traversal to it's own class (`ExpressionTraverser`), also utilizing visitor pattern
  - Move each expression handler into its own small visitor (`AliasExpressionVisitor`, `CaseWhenExpressionVisitor`, etc.).
  - Replace the monolithic `if-else` chain in `ExpressionDependencyCollector` with a single loop that delegates to the matching visitor.

#### One-line summary:

Refactor CLL ExpressionDependencyCollector with pluggable visitor pattern

### Checklist

----


kchledowski:
Rebased onto main and brought back handling of `Distinct` operator. I initially removed it, because it seemed that it would always appear as `Aggregate` in the optimized logical plan, but that's not the case. 
`Union` in Spark 3.5.X would actually be represented using `Distinct`, this is visible when running the following test:
`SparkIcebergIntegrationTest.testCreateTableAsSelect`
```java
createTempDataset(2).createOrReplaceTempView("temp");

spark.sql("CREATE TABLE source1 USING iceberg AS SELECT * FROM temp");
spark.sql("CREATE TABLE source2 USING iceberg AS SELECT * FROM temp");
spark.sql(
    "CREATE TABLE target USING iceberg AS (SELECT * FROM source1 UNION SELECT * FROM source2)");
```
Resulting optimized logical plan

with Spark 3.3.4
```
CreateTableAsSelect TableSpec(Map(),Some(iceberg),Map(),None,None,None,false), false
:- ResolvedDBObjectName org.apache.iceberg.spark.SparkCatalog@56f1db5f, [default, target]
+- Aggregate [a#4L, b#5L], [a#4L, b#5L]
   +- Union false, false
      :- RelationV2[a#4L, b#5L] spark_catalog.default.source1
      +- RelationV2[a#6L, b#7L] spark_catalog.default.source2
```

with Spark 3.5.6
```
CreateTableAsSelect TableSpec(Map(),Some(iceberg),Map(),None,None,None,false), false, true
   :- ResolvedIdentifier org.apache.iceberg.spark.SparkCatalog@7f680a9a, default.target
   +- Distinct
      +- Union false, false
         :- Project [a#8L, b#9L]
         :  +- SubqueryAlias spark_catalog.default.source1
         :     +- RelationV2[a#8L, b#9L] spark_catalog.default.source1 spark_catalog.default.source1
         +- Project [a#10L, b#11L]
            +- SubqueryAlias spark_catalog.default.source2
               +- RelationV2[a#10L, b#11L] spark_catalog.default.source2 spark_catalog.default.source2
```

pawel-big-lebowski (commented):
Overall I like the change and I think this is a great improvement.

1.  Do you think it's doable to rewrite this PR into two separate PRs: one modifying the code while keeping the tests and intact, and then the other one which modifies the tests? Refactor PR which rewrites lots of code while not modifying any of the tests is a way easier to merge.

2. Would it make sense to implement generic based `isDefinedAt` method as it's already done in `QueryPlanVisitor`. That would remove most of the `isDefinedAt` implementations.

tnazarew (commented):
wouldn't it be better to drop the `Operator` and `Expression` from class names? I think the `JoinVisitor` is better than `JoinOperatorVisitor` and `CaseWhenVisitor` is better than `CaseWhenExpressionVisitor`

besides that we have actually something like `AggregateExpression` so for it to comply with the convention it should have `AggregateExpressionExpressionVisitor` 😄

kchledowski:
>     2. Would it make sense to implement generic based `isDefinedAt` method as it's already done in `QueryPlanVisitor`. That would remove most of the `isDefinedAt` implementations.

@pawel-big-lebowski 
I'd rather not do it to be honest, yes it would remove some code duplication, but I think in the current state the code is more readable (especially to anyone looking at it for the first time). `isDefinedAt` is clearly defined in each class, and you don't need to bother checking parent class(es) to see if there are any more methods there. Another aspect is the current code is just plain simpler than such generic solution.

kchledowski:
>     1. Do you think it's doable to rewrite this PR into two separate PRs: one modifying the code while keeping the tests and intact, and then the other one which modifies the tests? Refactor PR which rewrites lots of code while not modifying any of the tests is a way easier to merge.


I've removed test changes from this PR and created a separate PR with tests (#4058) as a draft for now, once this PR is merged, I will rebase the one with tests onto main and make it ready for review.

pawel-big-lebowski (approved):

PR #4006, opened 2025-09-01
python: add GCP Lineage transport

ddebowczyk92:
The GCPLineageTransport sends OpenLineage events to Google Cloud Data Catalog Lineage API

The transport uses two different ways to send events: async clients for dbt events (found by checking if JobTypeJobFacet.integration="dbt") to handle lots of data transformation events quickly, and regular sync client for everything else.

### DIFF
--- a/client/python/openlineage/client/transport/gcplineage.py
+++ b/client/python/openlineage/client/transport/gcplineage.py
@@ -156,8 +156,6 @@ class GCPLineageTransport(Transport):
 
             event_dict = json.loads(Serde.to_json(event))
 
-            if self.async_client is None:
-                raise RuntimeError("GCP Lineage async client not initialized")
             await self.async_client.process_open_lineage_run_event(
                 parent=self.parent, open_lineage=event_dict
             )
@@ -204,24 +202,20 @@ class GCPLineageTransport(Transport):
         sync_result = True
         async_result = True
 
-        # Close sync client if it exists
-        if self.client:
-            try:
-                self.client.transport.close()
-            except Exception as e:
-                log.warning("Error closing GCP lineage client (sync): %s", e)
-                sync_result = False
+        try:
+            self.client.transport.close()
+        except Exception as e:
+            log.warning("Error closing GCP lineage client (sync): %s", e)
+            sync_result = False
+        # Close async client
+        try:
 
-        # Close async client if it exists
-        if self.async_client:
-            try:
+            async def close_async() -> None:
+                await self.async_client.transport.close()
 
-                async def close_async() -> None:
-                    await self.async_client.transport.close()
-
-                self._run_async_safely(close_async())
-            except Exception as e:
-                log.warning("Error closing GCP lineage client (async): %s", e)
-                async_result = False
+            self._run_async_safely(close_async())
+        except Exception as e:
+            log.warning("Error closing GCP lineage client (async): %s", e)
+            async_result = False
 
         return sync_result and async_result
--- a/client/python/tests/test_gcplineage.py
+++ b/client/python/tests/test_gcplineage.py
@@ -12,7 +12,7 @@ from openlineage.client.transport.gcplineage import GCPLineageConfig, GCPLineage
 from openlineage.client.uuid import generate_new_uuid
 
 
-@pytest.fixture
+@pytest.fixture(autouse=True)
 def mock_gcp_modules():
     """Fixture to mock Google Cloud modules for tests that need GCP transport."""
     with (
@@ -108,7 +108,7 @@ class TestGCPLineageConfig:
 class TestGCPLineageTransportInitialization:
     """Test GCPLineageTransport initialization and client creation."""
 
-    def test_gcplineage_transport_initialization_default_credentials(self, mock_gcp_modules):
+    def test_gcplineage_transport_initialization_default_credentials(self):
         """Test that GCPLineageTransport creates both sync and async clients with default credentials."""
         config = GCPLineageConfig.from_dict({"project_id": "test-project"})
         transport = GCPLineageTransport(config)
@@ -122,7 +122,7 @@ class TestGCPLineageTransportInitialization:
         assert transport.client is not None
         assert transport.async_client is not None
 
-    def test_gcplineage_transport_initialization_with_credentials(self, mock_gcp_modules):
+    def test_gcplineage_transport_initialization_with_credentials(self):
         """Test that GCPLineageTransport creates clients with service account credentials."""
         config = GCPLineageConfig.from_dict(
             {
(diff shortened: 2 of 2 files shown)


client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,214 @@
...
+                # No event loop exists, create one
+                return asyncio.run(self._emit_async(event))
+        else:
+            return self._emit_sync(event)
+
+    def _emit_sync(self, event: Event) -> Any:
+        try:
+            import json
+
+            event_dict = json.loads(Serde.to_json(event))
```
dolfinus: Can this be replaced with `Serde.to_dict()`?

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,214 @@
...
+
+@attr.define
+class GCPLineageConfig(Config):
+    """Configuration for GCP Data Catalog Lineage transport.
+
+    Args:
+        project_id: GCP project ID where the lineage data will be stored
+        location: GCP location (region) for the lineage service (default: "us-central1")
+        credentials_path: Path to service account JSON credentials file (optional)
+        async_transport_rules: Rules for when to use async transport (default: {"dbt": {"*": True}})
```
JDarDagran: I think this requires more explanation how to use it.

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class GCPLineageTransport(Transport):
+    kind = "gcplineage"
+    config_class = GCPLineageConfig
+
+    def __init__(self, config: GCPLineageConfig) -> None:
+        self.config = config
+        self.client: Optional[LineageClient] = None
+        self.async_client: Optional[LineageAsyncClient] = None
```
JDarDagran: I'd change the logic to use cached property and initialize one of these when actually needed.

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,220 @@
...
+        credentials_path: Path to service account JSON credentials file (optional)
+        async_transport_rules: Controls async vs sync transport based on integration and jobType.
+            Format: {"integration": {"jobType": True/False}}. Use "*" wildcards. True=async, False=sync.
+            Default: {"dbt": {"*": True}}
+
+    The transport will use default credentials if credentials_path is not provided.
+    """
+
+    project_id: str = attr.field()
+    location: str = attr.field(default="us-central1")
```
JDarDagran: Please move default value to top-level variable.

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,220 @@
...
+    credentials_path: Union[str, None] = attr.field(default=None)
+    async_transport_rules: dict[str, dict[str, bool]] = attr.field(default={"dbt": {"*": True}})
```
JDarDagran: Using mutable defaults in attr.field can cause shared state issues. Should use factory
```suggestion
    async_transport_rules: dict[str, dict[str, bool]] = attr.field(factory=lambda: {"dbt": {"*": True}})
```

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,220 @@
...
+        async_transport_rules: Controls async vs sync transport based on integration and jobType.
+            Format: {"integration": {"jobType": True/False}}. Use "*" wildcards. True=async, False=sync.
+            Default: {"dbt": {"*": True}}
+
+    The transport will use default credentials if credentials_path is not provided.
+    """
+
+    project_id: str = attr.field()
+    location: str = attr.field(default="us-central1")
+    credentials_path: Union[str, None] = attr.field(default=None)
```
JDarDagran:
```suggestion
    credentials_path: Optional[str] = attr.field(default=None)
```

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _emit_sync(self, event: Event) -> Any:
+        try:
+            event_dict = Serde.to_dict(event)
+
+            if self.client is None:
+                raise RuntimeError("GCP Lineage client not initialized")
+            self.client.process_open_lineage_run_event(parent=self.parent, open_lineage=event_dict)
+
+        except Exception as e:
```
JDarDagran, line 149: Please use `log.exception` instead. Also, could the scope of the exception caught be somehow limited to give more information rather than swallow everything?

client/python/openlineage/client/transport/gcplineage.py
```diff
@@ -0,0 +1,220 @@
...
+                        else:
+                            # Fallback - just close the client
+                            await self.async_client.close()
+
+                try:
+                    loop = asyncio.get_event_loop()
+                    if loop.is_running():
+                        import concurrent.futures
+
+                        with concurrent.futures.ThreadPoolExecutor() as executor:
```
JDarDagran: The async event loop handling is overly complex and potentially problematic.
It uses ThreadPoolExecutor to run asyncio.run() which can cause deadlocks. Also there are multiple nested try-catch blocks for different event loop scenarios.
The same complex pattern is repeated in the close() method.

Perhaps implements some helper method like:
```
    def _run_async_safely(self, coro):
        """Safely run async coroutine in sync context."""
        try:
            asyncio.get_running_loop()
        except RuntimeError:
            # No event loop running
            return asyncio.run(coro)
        else:
            # Event loop is running, use thread pool
            import concurrent.futures

            with concurrent.futures.ThreadPoolExecutor() as executor:
                return executor.submit(asyncio.run, coro).result()
  ```

File: client/python/tests/test_gcplineage.py [hunk 9]
```diff
@@ -0,0 +1,534 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+from openlineage.client.facet import JobTypeJobFacet
+from openlineage.client.run import Job, Run, RunEvent, RunState
+from openlineage.client.transport.gcplineage import GCPLineageConfig, GCPLineageTransport
+from openlineage.client.uuid import generate_new_uuid
+
+# Mock the google cloud modules at module level since they may not be installed
+sys.modules["google"] = MagicMock()
```
JDarDagran: It mocks entire modules at the module level, which can affect other tests and is generally considered an anti-pattern. Better to use context managers or setUp/tearDown methods.

client/python/tests/test_gcplineage.py
```diff
@@ -0,0 +1,498 @@
...
+from unittest.mock import MagicMock, patch
+
+import pytest
+from openlineage.client.facet import JobTypeJobFacet
+from openlineage.client.run import Job, Run, RunEvent, RunState
+from openlineage.client.transport.gcplineage import GCPLineageConfig, GCPLineageTransport
+from openlineage.client.uuid import generate_new_uuid
+
+
+@pytest.fixture
```
JDarDagran:
```suggestion
@pytest.fixture(autouse=True)
```

JDarDagran (approved):
Great job @ddebowczyk92 🚀 

I'm approving the PR but I'd love you to introduce one more change - rename from `gcplineage.py` to `gcp_lineage.py`. I believe this would fit more into Python's PEP 8 convention.

PR #4008, opened 2025-09-02

pawel-big-lebowski:
Part of the #3996 with spec changes only. 
This is easier to review. 
Rest of the #3996 will be a follow-up PR.


client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
```diff
@@ -48,9 +51,14 @@ public Type parse(JsonNode typeJson) {
             JsonNode properties = typeJson.get("properties");
             for (Iterator<Entry<String, JsonNode>> fieldsJson = properties.fields(); fieldsJson.hasNext(); ) {
               Entry<String, JsonNode> field = fieldsJson.next();
-              Type fieldType = parse(field.getValue());
               String description = field.getValue().has("description") ? field.getValue().get("description").asText() : null;
-              fields.add(new Field(field.getKey(), fieldType, description));
+              if (field.getValue().has("const")) {
+                // only String const are supported
+                fields.add(new Field(field.getKey(), new PrimitiveType("string", null), description, field.getValue().get("const").asText()));
+              } else {
```
julienledem: shouldn't this logic be in the parse() method instead?

client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
```diff
@@ -307,6 +315,12 @@ public PrimitiveType(String name, String format) {
       this.format = format;
     }
 
+    public PrimitiveType(String name, String format, String constValue) {
+      super();
+      this.name = name;
+      this.format = format;
+    }
```
julienledem: constValue is not used?

client/java/generator/src/main/java/io/openlineage/client/TypeResolver.java
```diff
@@ -102,7 +103,12 @@ private List<ResolvedField> resolveFields(List<Field> properties) {
           for (Field property : properties) {
             currentProperty = titleCase(property.getName());
             currentName = previousCurrentName + currentProperty;
-            ResolvedField resolvedField = new ResolvedField(property, visit(property.getType()));
+            ResolvedField resolvedField;
+            if (property.getConstValue().isPresent()) {
+              resolvedField = new ResolvedField(property, visit(property.getType()), property.getConstValue().get());
+            } else {
+              resolvedField = new ResolvedField(property, visit(property.getType()));
+            }
```
julienledem: This seems unnecessarily complicated. You are unwrapping an Optional to pass it to a constructor that will rewrap it in an Optional.
Maybe just add a ResolvedField constructor that takes an optional since this is how it stores it.
`ResolveField(Field field, ResolvedType type, Optional<String> constantValue)`

client/java/generator/src/main/java/io/openlineage/client/TypeResolver.java
```diff
@@ -367,6 +382,10 @@ public ResolvedType getType() {
       return type;
     }
 
+    public Optional<String> getConstantValue() {
+      return constantValue;
+    }
+
     @Override
     public String toString() {
       return "ResolvedField{name: " + field.getName() + ", type: " + type +  "}";
```
julienledem, line 381: should be updated.

julienledem: same for equals and hashCode bellow

julienledem (commented):
That makes sense, I have added comments regarding the code generator portion.

client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
pawel-big-lebowski: I think not, bcz `const` shouldn't be returned as a Type (like PrimitiveType of String with constant value) and it should be already put at the `Field` member.

client/java/generator/src/main/java/io/openlineage/client/TypeResolver.java
pawel-big-lebowski: Changing the code so that `constantValue` is stored directly within `field`, so there's no need to change toString, equals, and hashCode for `ResolvedField`. Thanks.

client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
pawel-big-lebowski: Thx, `constValue` should be just a property of a Field instead.

client/java/generator/src/main/java/io/openlineage/client/SchemaParser.java
```diff
@@ -48,9 +49,14 @@ public Type parse(JsonNode typeJson) {
             JsonNode properties = typeJson.get("properties");
             for (Iterator<Entry<String, JsonNode>> fieldsJson = properties.fields(); fieldsJson.hasNext(); ) {
               Entry<String, JsonNode> field = fieldsJson.next();
-              Type fieldType = parse(field.getValue());
               String description = field.getValue().has("description") ? field.getValue().get("description").asText() : null;
-              fields.add(new Field(field.getKey(), fieldType, description));
+              if (field.getValue().has("const")) {
+                // only String const are supported
+                fields.add(new Field(field.getKey(), new PrimitiveType("string", null), description, field.getValue().get("const").asText()));
```
julienledem: to avoid surprises maybe we should still call parse() but check the type is String and throw an exception if not.
With this implementation, one can put whatever in the type and it will be replaced by string.
This could lead to some hair pulling on unexpected behavior.

pawel-big-lebowski: Makes sense, we should definitely avoid hair pulling. [Second commit](https://github.com/OpenLineage/OpenLineage/pull/4008/commits/3eed2aaff83231c65990acbe6b8f8fc400a35395) should resolve that.

mobuchowski (approved):
Approved but would appreciate more comments around the Java generator part - it's less frequently accessed and looked at part of code, so it's a good opportunity to document it more. Thanks.

client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
```diff
@@ -320,6 +327,10 @@ private TypeSpec builderClass(ObjectResolvedType type) {
       // only regular fields are used in the builder
       @Override
       public void onField(ResolvedField f) {
+        if (f.getConstantValue().isPresent()) {
```
julienledem: instead of adding this if statement in every call to onField(), you could add an onConstantField() to the Visitor with the default implementation that does nothing.

julienledem (approved):
I made an additional comment that can simplify a bit the section about handling constant fields.
Otherwise, this looks good.

client/java/generator/src/main/java/io/openlineage/client/JavaPoetGenerator.java
pawel-big-lebowski: Love this—this was exactly what I was missing. Thanks!

PR #4014, opened 2025-09-05
Python: Raise errors if custom token provider cannot be loaded

dolfinus:
### Problem

When using custom token provider for HttpTransport, but there is something wrong with an import string, OpenLineage silently ignores all errors, and fallback to default TokenProvider. This could lead to disabling any authorization for HTTP requests. Also it is hard to debug.

### Solution

If custom `auth: type` cannot be loaded, is not a class or is not a TokenProvider subclass, instead of ignoring it, raise error immediately.

#### One-line summary:

Raise errors if custom token provider cannot be loaded

### Checklist

----


client/python/openlineage/client/transport/http.py
```diff
@@ -57,16 +57,21 @@ def get_bearer(self) -> str | None:
 
 
 def create_token_provider(auth: dict[str, str]) -> TokenProvider:
```
kacpermuda, line 59: Could you add debug logs for each path so it's easier to debug in the future ?

kacpermuda (approved):

client/python/openlineage/client/transport/http.py
dolfinus: Done

PR #4019, opened 2025-09-05
Airflow - disable OL using Airflow Variable

humzahkiani:
### Problem

Currently the only way to toggle the OL integration is by changing the `OPENLINEAGE_DISABLED` environment variable, and restarting airflow. This (changing the env var) is difficult to do without making a code change. Incidents in enterprise production systems often necessitate the toggling to occur without going through the code deployment pipeline, which can take hours and require hoops of bureaucracy to jump through.

### Solution

Add the ability to use the Airflow variable `openlineage.disabled` as a toggle to turn on and off the airflow integration. The existing `OPENLINEAGE_DISABLED` environment variable toggle takes precedence over this. This still requires restarting the airflow instance to take effect. 

### Checklist

----


humzahkiani:
I see the test failures. They seem to hint at a broader potential issue here... accessing the db from module-level code. 
Would appreciate feedback on this approach before I continue.

Another couple ideas I had
- Provide an airflow variable to toggle only the task events using their hooks (since these are able to access the DB more easily from the worker as opposed to the scheduler for the DAGs). Not as optimal for our situation, but curious if you would support it. 
- Schedule-Side cache
    - create an `is_disabled_cache` in `listener.py`
    - Spawn a background thread at startup 
    - Have the thread loop thru the following:
1. Wake up.
2. Create its own, temporary, and safe database session. This is safe because it's happening in a separate thread, not in the Scheduler's main loop.
3. Query the database for the openlineage.disabled variable.
4. Update the shared in-memory variable (_is_disabled_cached) with the result.
5. Go to sleep for a configurable amount of time (e.g., 60 seconds).
   - Now the DAG hook (in the scheduler) can read from the cache instead of having to reach out to the DB (which seems to be unsafe)

mobuchowski:
@humzahkiani `_is_disabled` is evaluated when the OL Plugin is being loaded - deciding whether to load the listener or not. Even if you're able to work around the db issues, it won't work this way. You'd have to add additional checks when the listener event is being called - or somehow remove the active listener?

humzahkiani:
@mobuchowski thanks for the feedback. How do you feel about the other two solutions I described?

humzahkiani:
Tried another implementation @mobuchowski 
Let me know your thoughts

integration/airflow/openlineage/airflow/listener.py
```diff
@@ -22,6 +22,51 @@
...
 from airflow.listeners import hookimpl
 from airflow.utils import timezone
+from airflow.utils.session import create_session
+
+# --- In-Memory Cache for Runtime Disabling ---
+_is_disabled_cached: Optional[bool] = None
+_cache_lock = threading.Lock()
+_stop_event = threading.Event()
+_cache_thread: Optional[threading.Thread] = None
+_CACHE_TTL_SECONDS: int = 60
```
humzahkiani, line 32: I set this as 60 seconds, but I can also make this configurable via an environment variable and default to 60s

kacpermuda:
Few thoughts after a quick glance: 

1. If we pursue this path of introducing a runtime check before every event is emitted and using some cache, we need to hide the entire thing (including cache check, refreshing, spawning a separate thread etc.) behind some config env variable that will default to False. f.e. `OPENLINEAGE_AIRFLOW_ENABLE_RUNTIME_DISABLE` or something like this. We need this so that nothing changes for current users, and so that users have to explicitly opt-in to use this.
2. While I understand the reasoning, I don't really like the idea of refreshing cache every X seconds with some background process. I'd have to think of a better alternative. 

Maybe we should add this to the [OL Airflow provider](https://airflow.apache.org/docs/apache-airflow-providers-openlineage/stable/index.html) first, so that Airflow maintainers can help us come up with a better solution, that will also work well with AF3 (with no direct db access from workers). @humzahkiani as I understand, you're on AF2.5 or 2.6 now and need this feature?

humzahkiani:
@kacpermuda 

1. That makes sense. Happy to make that change. 
2. I would appreciate your thoughts. Originally I started with adding the `is_disable` check to each airflow listener hook to query the airflow variable eagerly. However, I encountered an issue with querying the DB within the DAG hooks that operate on the scheduler, which I learned has a protected connection to the Airflow DB. Which led to me spawning a separate thread with its own connection. This successfully circumvented the aforementioned issues.

My company (Affirm) is currently on Airflow 2.5.1. We are considering upgrading at some point in the near future, so I am open to starting with contributing this change to the OL airflow provider first. I can duplicate this PR there? Unless you have ideas on improvements I should make first.

kacpermuda:
No better idea for now. @JDarDagran any thoughts here?

humzahkiani:
I'll give @JDarDagran a few more days to provide any input. If none, then I'll recreate the PR in the native OL Airflow provider repo so we can move the conversation there.

kacpermuda:
Thanks, most OL folks are on holidays now, hence the delay. They should be back mid next-week, so maybe they'll have a look. I'm also a bit overloaded now, will try to re-visit this once I have a moment.

humzahkiani:
No worries, this can wait until then. Appreciate your help.

mobuchowski:
I must say I don't really like that. It in fact introduces something like feature flags for Airflow, but in backhanded way only for a given provider. I think it's going too far from what we want to do, but I also see the need for that. If you don't want to add a feature like this to Airflow itself, and move to the newest version, I think the best choice for you would be to fork Airflow or this provider and add this feature.

humzahkiani:
@mobuchowski 
To me it seems to be more of a feature flag for the airlflow openlineage integration rather than airflow itself. Without the integration, the `openlineage.disabled` airflow variable functions the same as any other airflow variable, an inert key-value pair that can be retrieved at runtime.  

I will defer to your input of course. We would prefer to not to fork either airflow or the integration. What other components of airflow (or its integrations) would you envision a general airflow feature flagging system to toggle? Or would you prefer to have only a generalized interface to feature flagging, which different concerns (such as this integration) can pull in as needed.

humzahkiani:
Hey @mobuchowski, circling back to see if you have any thoughts. I am going to proceed with creating a PR in the airflow repo to continue the discussion.

mobuchowski:
@humzahkiani I don't think we're going to accept this PR here. We're most likely going to remove those libraries on 01.01.2026, based on the removal of support of those versions of Airflow in all major Airflow providers, and continue to support https://airflow.apache.org/docs/apache-airflow-providers-openlineage/stable/ only.

PR #4022, opened 2025-09-08
capture dbt tags in TagsRunFacet

mobuchowski:
dbt allows tagging models: https://docs.getdbt.com/reference/resource-configs/tags

This PR captures this in TagsRunFacet in model-level events.

### DIFF
--- a/integration/common/openlineage/common/provider/dbt/processor.py
+++ b/integration/common/openlineage/common/provider/dbt/processor.py
@@ -16,6 +16,7 @@ from openlineage.client.facet_v2 import (
     InputDatasetFacet,
     JobFacet,
     OutputDatasetFacet,
+    RunFacet,
     column_lineage_dataset,
     data_quality_assertions_dataset,
     datasource_dataset,
@@ -26,6 +27,7 @@ from openlineage.client.facet_v2 import (
     processing_engine_run,
     schema_dataset,
     sql_job,
+    tags_run,
 )
 from openlineage.client.uuid import generate_new_uuid
 from openlineage.common.provider.dbt.facets import DbtRunRunFacet, DbtVersionRunFacet, ParentRunMetadata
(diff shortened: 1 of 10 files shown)


integration/common/openlineage/common/provider/dbt/processor.py
```diff
@@ -305,6 +307,12 @@ def parse_execution(self, context: DbtRunContext, nodes: Dict) -> DbtEvents:
             if sql:
                 job_facets["sql"] = sql_job.SQLJobFacet(query=sql, dialect=self.extract_dialect())
 
+            run_facets: Dict[str, RunFacet] = {}
+            if tags := output_node.get("tags", None):
+                run_facets["tags"] = tags_run.TagsRunFacet(
+                    tags=[tags_run.TagsRunFacetFields(key=tag, value="true", source="DBT") for tag in tags]
```
kacpermuda, line 313:
```suggestion
                    tags=[tags_run.TagsRunFacetFields(key=tag, value=tag, source="DBT") for tag in tags]
```

Imho, when we only have a key not a key:value it's better to duplicate tag in value than to put generic "true". Do you think we can go this way?

integration/common/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -230,6 +231,7 @@ def _parse_structured_log_event(self, line: str) -> Optional[RunEvent]:
 
         dbt_event_name = dbt_event["info"]["name"]
         self.logger.debug("Parsing dbt event: %s", dbt_event_name)
+        self.logger.debug("dbt event: %s", json.dumps(dbt_event, sort_keys=True))
```
kacpermuda: Is this a leftover or a desired change?

kacpermuda (approved):
Left two small comments, overall LGTM

integration/common/openlineage/common/provider/dbt/processor.py
mobuchowski: What would be the reason? I think the downside is it would break search by value (as in, find all keys for given value) while not improving the key experience at all. IMO it's best that we'd treat all `label` like tags in x:true way.

integration/common/openlineage/common/provider/dbt/structured_logs.py
mobuchowski: Removed, thanks

integration/common/openlineage/common/provider/dbt/processor.py
kacpermuda: My thinking is that users can search by value rather than key. So if we have some tag like key=env, value=prod and then equivalent label=prod that we translate to key=prod, value=true, it can be harder to find. I don't feel like adding a "true" out of nowhere to user tags is valuable and wanted from the user perspective - they might rely on value and not key, so they'll have to debug where did the "true" come from, as they did not set it themselves. Maybe that's just preference, feel free to resolve if you don't agree.

PR #4050, opened 2025-09-22
Implement equals and hashcode method into generated classes

pawel-big-lebowski:
Implement equals and hashcode method into generated classes


client/java/src/test/java/io/openlineage/client/OpenLineageTest.java
```diff
@@ -469,4 +470,25 @@ void factory() throws JsonProcessingException {
...
+                    .nominalStartTime(now)
+                    .nominalEndTime(now)
+                    .build())
+            .build();
+
+    Run run = ol.newRunBuilder().runId(runId).facets(runFacets).build();
+    Run sameRun = ol.newRunBuilder().runId(runId).facets(runFacets).build();
+
+    assertThat(run.hashCode()).isEqualTo(sameRun.hashCode());
+    assertThat(run).isEqualTo(sameRun);
```
julienledem, line 493: nit: I would also add a negative case. something like:
```
Run differentRun = ol.newRunBuilder().runId(UUID.randomUUID()).facets(runFacets).build();
assertThat(run.hashCode()).isNotEqualTo(differentRun.hashCode())
assertThat(run).isNotEqualTo(differentRun);
```

Issue #4051, opened 2025-09-22
[BUG] Spark integration incorrectly identifies Hive/Delta tables as AWS Glue (symlinks facet uses arn:aws:glue)


jsjasonseba:
### OpenLineage integration/client

Spark integration

### OpenLineage version

io.openlineage:openlineage-spark_2.12:1.37.0

### Technology and package versions

Spark: 3.5.6
Storage: Delta Lake on S3
Metastore: Hive Metastore 3.1.3
Runtime: Local (spark.master=local[2])
Language: PySpark Python 3.10

### Environment configuration

.config("spark.extraListeners", "io.openlineage.spark.agent.OpenLineageSparkListener")

```
spark-submit --master local[2] \
    --packages io.openlineage:openlineage-spark_2.12:1.33.0,io.delta:delta-spark_2.12:3.3.1,org.apache.hadoop:hadoop-aws:3.3.4,com.amazonaws:aws-java-sdk-bundle:1.12.774 \
    test_open_lineage.py 
```

### Deployment details

Running Spark 3.5.6 locally
Docker Hive metastore 3.1.3

### Problem details

When reading a Delta Lake table stored in S3 and registered in Hive Metastore, the emitted OpenLineage event contains a Glue ARN inside the symlinks dataset facet.

Observed OpenLineage event:
```
"symlinks": {
  "identifiers": [
    {
      "namespace": "arn:aws:glue:ap-southeast-1:xxxxxxx",
      "name": "table/delta/source_table_2",
      "type": "TABLE"
    }
  ]
}

```

This misidentifies the dataset as an AWS Glue resource, even though Glue is not configured or used. 

It seems like the namespace is set with the account ID configured in the current session. Since the session has an AWS credentials configured, openlineage might mistakenly assumes Glue is being used? In reality AWS credentials is configured only for writing to S3 and uses hive metastore for catalog.

### What you think should happen instead

Symlinks should not identify AWS Glue resource. It shoule refer to hive metastore table.

### How to reproduce

1. Create sample delta table
```
from pyspark.sql import *
from pyspark.sql.functions import *

def main():
    # Create SparkSession
    spark = (
        SparkSession.builder
        .appName("test")
        .config("spark.jars.packages", "io.openlineage:openlineage-spark:1.37.0")
        .config("spark.extraListeners", "io.openlineage.spark.agent.OpenLineageSparkListener")
[...]

PR #4053, opened 2025-09-25
spark: fix false hive glue detection

jsjasonseba:
### Problem

OpenLineage falsely detects glue symlinks when using hive metastore and AWS account is configured.

In the #3695 PR, AwsUtils `getGlueArn` is set to do glue ARN resolution even if it does not satisfies `isHiveUsingGlue` check. This causes non-glue hive catalog to be identified as glue when AWS account ID is configured in the session. This issue specifically happen when AWS account ID is configured in the session because Glue ARN resolution will fallback to configured AWS account ID when other resolution methods fail. 

Closes: #4051

### Checklist

----

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
@@ -32,16 +32,7 @@ public class AwsUtils {
                   getGlueCatalogId(sparkConf, hadoopConf)
                       .map(glueCatalogId -> "arn:aws:glue:" + region + ":" + glueCatalogId));
     } else {
-      try {
-        return awsRegion()
-            .flatMap(
-                region ->
-                    getGlueCatalogId(sparkConf, hadoopConf)
-                        .map(glueCatalogId -> "arn:aws:glue:" + region + ":" + glueCatalogId));
-      } catch (Exception e) {
-        log.error("Failed to retrieve Glue", e);
-        return Optional.empty();
-      }
+      return Optional.empty();
     }
   }
 


integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
```diff
@@ -32,16 +32,7 @@ public static Optional<String> getGlueArn(SparkConf sparkConf, Configuration had
                   getGlueCatalogId(sparkConf, hadoopConf)
                       .map(glueCatalogId -> "arn:aws:glue:" + region + ":" + glueCatalogId));
     } else {
-      try {
-        return awsRegion()
```
pawel-big-lebowski, line 36: Usually, deleting code lines will cause test failures. @mobuchowski do you recall why is this necessary and should not be removed? Some other users are also [sceptical about this behaviour](https://github.com/OpenLineage/OpenLineage/pull/3695/files#r2153849136)

Would it make sense to move the removed lines into `io.openlineage.spark3.agent.lifecycle.plan.catalog.iceberg.GlueCatalogTypeHandler` and make this behaviour configurable if required? `GlueCatalogTypeHandler` can easily contain OpenLineage context and configuration.

jsjasonseba: Yes, it would be great if we can get a confirmation. The original PR does not seem to describe the purpose of those lines and the PR title and description seems unrelated to this change. 

In my opinion, the code logic seems  weird that we attempt to do glue resolution even when `isHiveUsingGlue` is not satisfied. If there are more conditions that warrants glue resolutions then it should have been added in the `isHiveUsingGlue` instead of forcing to always resolve it, unintendedly causing this bug.

I tried running existing tests and found no test failures so I don't think those specific lines are covered. But, I am up to making it configurable if needed.

mobuchowski: Pretty sure I had an use case where it was Glue but not Hive. AFAIK it was Iceberg with Glue: https://iceberg.apache.org/docs/1.5.1/aws/#spark

Unfortunately it's part of way too complex path within https://github.com/OpenLineage/OpenLineage/blob/f6e4968ff60bb697fb04821281fe621af04aabc0/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java#L70 and I had no clear path for adding OSS tests that could reflect that behavior. 

Maybe the condition should also check for
```
spark.sql.catalog.prod_catalog.type=glue
```

jsjasonseba: @mobuchowski I see, I am unfamiliar to glue setup, but is `spark.sql.catalog.prod_catalog.type=glue` a common config for glue setup? 

If it is, I would propose going with adding the condition to the `isHiveUsingGlue` check and remove glue resolution on the else block.

mobuchowski: @jsjasonseba I will merge this and fix up if I need it further later.

PR #4055, opened 2025-09-29
dbt: use alias when naming datasets

mobuchowski:
Right now when users use alias feature: https://docs.getdbt.com/reference/resource-configs/alias we don't take it into account when naming the datasets, which is wrong - the alias is the actual identifier that the table has in the db.

### DIFF
--- a/integration/common/openlineage/common/provider/dbt/processor.py
+++ b/integration/common/openlineage/common/provider/dbt/processor.py
@@ -73,6 +73,9 @@ class UnsupportedDbtCommand(Exception):
 
 @attr.define
 class ModelNode:
+    # in reality Literal["model", "test", "snapshot", "source", "seed"]
+    # but way too much work to convince mypy that it's true
+    type: str
     metadata_node: Dict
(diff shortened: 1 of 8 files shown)


client/python/openlineage/client/transport/composite.py
```diff
@@ -147,7 +147,7 @@ def emit(self, event: Event) -> None:
             )
             raise RuntimeError(msg)
 
-        log.info(
+        log.debug(
```
kacpermuda: Any particular reason behind this change?

mobuchowski: Spam on stdout when running dbt. I can move that to separate PR

kacpermuda: Nah, it's fine, I thought it may be useful on Info, but when emitting lots of events it can spam.

mobuchowski: Actually just done that: https://github.com/OpenLineage/OpenLineage/pull/4056

Issue #4063, opened 2025-10-02
[BUG] Spark Integration Scala 2.13 version in Databricks throws java.lang.ClassNotFoundException: scala.Serializable


n-taylor-pantomath:
### OpenLineage integration/client

Spark integration

### OpenLineage version

- io.openlineage:openlineage-spark_2.13:1.32.0
- io.openlineage:openlineage-spark_2.13:1.37.0

### Technology and package versions

Run against Databricks Job Compute
DBR: 16.4 LTS (includes Apache Spark 3.5.2, Scala 2.13)

### Spark config
```
spark.openlineage.transport.type console
spark.extraListeners io.openlineage.spark.agent.OpenLineageSparkListener,com.databricks.backend.daemon.driver.DBCEventLoggingListener
```

### Environment variables
```
PYSPARK_PYTHON=/databricks/python3/bin/python3
```

### Init script
```
#!/bin/bash

set -euo pipefail

VERSION="2_13"

SOURCE_DIR="/dbfs/FileStore/jars/openlineage/${VERSION}"
TARGET_DIR="/databricks/jars"
mkdir -p "$TARGET_DIR"

echo "Copying contents of $SOURCE_DIR to $TARGET_DIR"
cp -r "${SOURCE_DIR}/"* "$TARGET_DIR"/
echo "Installed OpenLineage files to $TARGET_DIR"
```

The contents of the `/dbfs/FileStore/jars/openlineage/2_13/` directory only include one of the JARs downloaded from the [Maven Repository](https://mvnrepository.com/artifact/io.openlineage/openlineage-spark_2.13/1.38.0). Both versions of OpenLineage listed above were tested, but only one was in the folder at a time.

### Problem details

The job cluster fails to start and the Log4j output includes this error message:
```
25/10/02 13:46:21 ERROR DatabricksMain$DBUncaughtExceptionHandler: Uncaught exception in thread spark-listener-group-shared!
java.lang.NoClassDefFoundError: scala/Serializable
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
[...]

PR #4099, opened 2025-10-21
catch NoClassDefFoundError if buggy impl exist on the classpath

mobuchowski:
Solves https://github.com/OpenLineage/OpenLineage/issues/4063


pawel-big-lebowski (approved):
Could we add a comment for that to make sure this won't get removed in future?

Issue #4101, opened 2025-10-24
Spec: Introduce temporary datasets to support job-to-job lineage


jakub-moravec:
**Purpose:**

The goal is to standardize how to represent job-to-job lineage (for example, individual tasks of an ETL job) via OpenLineage events. 

A more comprehensive background is documented on https://github.com/OpenLineage/OpenLineage/issues/4090.  

**Proposed implementation**

The most suitable (and least invasive for current OL producers/consumers) approach seems to be introducing a concept of "temporary" datasets:

- Job (on any granularity, whether it is an ETL job or an individual task of the ETL job, for example) would not reference other Job directly, but rather reference temporary dataset(s) that it reads from, and temporary dataset(s) that it writes into. 
- [Data Set Type facet](https://openlineage.io/docs/spec/facets/dataset-facets/type) can be used to mark temporary datasets, new "known" values are suggested: 
   - `JOB_OUTPUT` for _datasetType_ (alternative names could be `TRANSFORM_OUTPUT` or `RESULT_SET`, `JOB_OUTPUT` is proposed as it is most generic and works well with the current naming convention) 
   - `TEMPORARY` for _subType_ (there are a couple of alternative names that could be proposed, but works well not only for this use-case, but also for temporary tables, for example)
- A proposed best practice for generating the namespace of the temporary dataset is to use the namespace of the Job that creates it 
- A proposed best practice for generating the name of the temporary dataset is to 
   - base it on the Job name and the name of the output of the parent job (if any), if there is no parent job, the equivalent is the name of the first non-temporary output dataset (if known)
   - base it on the Job name and add an index suffix to accommodate Jobs with multiple outputs  
   - Both approaches ensure that the Job for which the OL event is produced does not need to know what Job succeeds it. But to reference the previous temporary dataset, the Job needs to know which Job precedes it.)

**Example:**

ETL job with 2 source datasets, 3 intermediate tasks, and one output dataset: 

<img width="2101" height="294" alt="Image" src="https://github.com/user-attachments/assets/647a3b71-f8bb-4892-99f6-5cb735d4b242" />

Proposed OL payload: 

```
[
    {
        "eventTime": "...",
        "eventType": "...",
        "producer": "...",
        "schemaURL": "...",
        "job": {
            "namespace": "<job_namespace>",
            "name": "<job_name>.Load task",
            "facets": {
                "jobType": {
                    "processingType": "BATCH",
                    "integration": "example",
                    "jobType": "ingest_data",
[...]

PR #4103, opened 2025-10-24
spec: add support for temporary datasets to support job-to-job lineage

jakub-moravec:
PR for https://github.com/OpenLineage/OpenLineage/issues/4101

### Checklist

----

### DIFF
--- a/client/python/src/openlineage/client/generated/dataset_type_dataset.py
+++ b/client/python/src/openlineage/client/generated/dataset_type_dataset.py
@@ -21,4 +21,4 @@ class DatasetTypeDatasetFacet(DatasetFacet):
 
     @staticmethod
     def _get_schema() -> str:
-        return "https://openlineage.io/spec/facets/1-1-0/DatasetTypeDatasetFacet.json#/$defs/DatasetTypeDatasetFacet"
+        return "https://openlineage.io/spec/facets/1-0-1/DatasetTypeDatasetFacet.json#/$defs/DatasetTypeDatasetFacet"
--- a/spec/facets/DatasetTypeDatasetFacet.json
+++ b/spec/facets/DatasetTypeDatasetFacet.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/facets/1-1-0/DatasetTypeDatasetFacet.json",
+  "$id": "https://openlineage.io/spec/facets/1-0-1/DatasetTypeDatasetFacet.json",
   "$defs": {
     "DatasetTypeDatasetFacet": {
       "allOf": [
--- a/website/static/spec/facets/1-0-1/DatasetTypeDatasetFacet.json
+++ b/website/static/spec/facets/1-0-1/DatasetTypeDatasetFacet.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
-  "$id": "https://openlineage.io/spec/facets/1-1-0/DatasetTypeDatasetFacet.json",
+  "$id": "https://openlineage.io/spec/facets/1-0-1/DatasetTypeDatasetFacet.json",
   "$defs": {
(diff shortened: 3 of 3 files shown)


mobuchowski:
@jakub-moravec technicality: please update the version in spec file when you change it, then pre-commit hooks will use it to generate code.

Otherwise looks good to me - for spec changes would love to have 👍 from more people than just me. 

@kacpermuda @pawel-big-lebowski @harels any comments?

kacpermuda:
Do we need to update version in spec file if just description was changed and not the attribute of the facet?

kacpermuda:
I'm not sure about this way about providing job correlation using datasets, but also it's not really a spec change - it's more creative use of current capabilities. This a non-blocking opinion as I did not read fully into the reasoning and had to leave early from last TSC meeting, so I'm not sure why a simple jobRelationshipJobFacet can not be used to represent the relationship?

The datasets produced by each job are still going to be emitted as usual, only now some of them can be marked as temporary so that they can be easily filtered out by a backend, is that the only change this introduces?

mobuchowski:
@jakub-moravec I though about naming for those datasets more. 

>>- The namespace of the temporary dataset is the same as the namespace of the Job that creates it. 

I think it's not the best solution. Someone could have a job namespace that conflict with 
Instead we could introduce specific namespace for those datasets, like 
```
temp://
```
or actually declare what they are if possible; they might be real files, database tables, that we might want to not show, but there is no reason to actually not tell what they are. 
If they are something in-memory, we could have `inmemory://` namespace. 

Generally mixing job and dataset namespaces isn't the best idea. The former ones are much more free-formed.  

>- The name of the temporary dataset is based on the Job name and the name of the output of the parent job (if any). As a fallback, it can be based on the Job name and an index suffix can be added to accommodate Jobs with multiple outputs.

If there's well-defined name (even if it's randomized uuid), I'd use it instead.
It would make lineage matching easier if we emit those from different processes; it should not change anything for the case where you emit both from the same place (as I understand is your goal). 

Let me know what you think.

jakub-moravec:
@MaciejGajewski the namespaces make sense to me. If it is a file or a database table, the best practice for the namespace is already defined, so the only new one would be `inmemory://`, right? Should it also be documented on https://openlineage.io/docs/spec/naming? I guess that will not hurt. 

For the temporary dataset name, I don't think these will always be named/identifiable, so maybe documenting a couple of options would be best.

Regarding your and @kacpermuda's comments on updating the version, do you think it's needed in the end? That would be the version of the facet, right? So going from `1-0-0` to `1-0-1` or `1-1-0`?

mobuchowski:
>Should it also be documented on https://openlineage.io/docs/spec/naming? I guess that will not hurt.

Yes, definitely. The point would be to have the proper handling of this data on the dataset level, not even by looking at dataset type.

>For the temporary dataset name, I don't think these will always be named/identifiable, so maybe documenting a couple of options would be best.

I think the _name_ could be job related if there's nothing clear. Ideally this is something that followup job can reference, but we don't have to specify this - if that works in your case, this is just what you'll do, whether it's some known name, or even random string.

jakub-moravec:
@mobuchowski sounds good, I documented the `inmemory://` namespace and adjusted the example

kacpermuda:
Looks good, thanks @jakub-moravec ! Is the CI repairable by running pre-commit locally on your side? It looks like it needs the facet version bump but I don't think it's necessary, we should try to workaround it maybe.

mobuchowski:
Unfortunately, I'm pretty sure it does need version bump.

kacpermuda (approved):
Fixed the pre-commit and changed the facet version to 1-0-1 from 1-1-0, hopy you don't mind. Let's wait for another approval as it's a good practice for spec-related PRs

jakub-moravec:
Thank you, @kacpermuda!

mobuchowski (approved):

jakub-moravec:
@kacpermuda, thanks for merging! Should the linked ticket be closed now? https://github.com/OpenLineage/OpenLineage/issues/4101

Issue #4024, opened 2025-09-08
[BUG] Update Snowflake spark lineage event namespace  as per open lineage naming convension which is snowflake://{organization name}-{account name}


mrpalash-amz:
### OpenLineage integration/client

Spark integration

### OpenLineage version

I followed this documentation to setup glue etl job lineage events: https://docs.aws.amazon.com/datazone/latest/userguide/datazone-data-lineage.html

Datazone transport layer package : https://github.com/OpenLineage/OpenLineage/blob/main/client/java/transports-datazone/src/main/java/io/openlineage/client/transports/datazone/AmazonDataZoneTransport.java

### Technology and package versions

I have written a spark job in python which reads input data from existing snowflake tables and create new tables and write transform data in new tables. 

### Environment configuration

Glue verions - glue 5.0 sparks 3.5 scala 2, python 3 

### Problem details

Spark openlineage events for snowflake emits lineage event namespace is snowflake://{organization name}-{account name}.snowflake.computing.com. But open lineage recommended naming convention (https://openlineage.io/docs/spec/naming) is snowflake://{organization name}-{account name} 

### How to reproduce

Create  a spark job in python which reads input data from existing snowflake tables and create new tables in snowflake and write transform data in new tables. This job should emit lineage events which has namespace snowflake://{organization name}-{account name}.snowflake.computing.com. 

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

mrpalash-amz:
Sample lineage event where expected namespace is snowflake://XXXXXX-XXXXXX but namespace received in lineage events i s snowflake://XXXXXX-XXXXXX.snowflakecomputing.com. Please remove suffix snowflakecomputing.com from namespace and follow open lineage naming conventions : https://openlineage.io/docs/spec/naming

```
{
  "eventTime": "2025-08-07T02:36:31.281Z",
  "producer": "https://github.com/OpenLineage/OpenLineage/tree/1.22.0/integration/spark",
  "schemaURL": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/RunEvent",
  "eventType": "COMPLETE",
  "run": {
    "runId": "01988262-ed4d-780c-8e11-828cdab1d3c4",
    "facets": {
      "parent": {
        "_producer": "https://github.com/OpenLineage/OpenLineage/tree/1.22.0/integration/spark",
[...]

PR #4104, opened 2025-10-24
fix snowflake ident parsing

mobuchowski:
Fixes https://github.com/OpenLineage/OpenLineage/issues/4024


integration/spark/vendor/snowflake/src/test/java/io/openlineage/spark/agent/vendor/snowflake/lifecycle/plan/SnowflakeRelationVisitorTest.java
```diff
@@ -60,7 +60,8 @@ public void setUp() {
                 new StructField[] {new StructField("name", StringType$.MODULE$, false, null)})));
 
     when(relation.params().sfFullURL())
-        .thenReturn("https://microsoft_partner.east-us-2.azure.snowflakecomputing.com");
+        .thenReturn(
+            "https://microsoftpartner-eastus2account.east-us-2.azure.snowflakecomputing.com");
```
LewisCowlesMotive: this url looks incorrect. I'm not saying it is, but it looks like it has tonnes of redundancy within. Can you link me to docs suggesting the url structure, or giving similar examples?

client/java/src/test/java/io/openlineage/client/utils/SnowflakeUtilsTest.java
```diff
@@ -0,0 +1,48 @@
...
+  @ParameterizedTest
+  @CsvSource({
+    // Modern organization-account format on Azure
+    "https://myorg-myaccount.east-us-2.azure.snowflakecomputing.com, myorg-myaccount",
+
+    // Modern organization-account format on AWS
+    "https://myorg-myaccount.us-west-2.aws.snowflakecomputing.com, myorg-myaccount",
+
+    // Modern organization-account format on GCP
+    "https://myorg-myaccount.us-central1.gcp.snowflakecomputing.com, myorg-myaccount",
```
kacpermuda: I think the modern format is just `myorg-myaccount.snowflakecomputing.com`

integration/spark/vendor/snowflake/src/test/java/io/openlineage/spark/agent/vendor/snowflake/lifecycle/plan/SnowflakeRelationVisitorTest.java
mobuchowski: It's an example of orgname-accountname identifier setup. https://docs.snowflake.com/en/user-guide/admin-account-identifier

The `eastus2account` part could be easily random, like `ashdf243q52`

LewisCowlesMotive: Do you think you could remove the repetition from this so that it reads more naturally like the other examples?

mobuchowski:
@kacpermuda @LewisCowlesMotive fixed ⬆️ issues.

client/java/src/test/java/io/openlineage/client/utils/SnowflakeUtilsTest.java
```diff
@@ -0,0 +1,35 @@
...
+  @ParameterizedTest
+  @CsvSource({
+    "https://myorg-myaccount.snowflakecomputing.com, myorg-myaccount",
+    "https://myaccount.east-us-2.azure.snowflakecomputing.com, myaccount",
+    "https://myorg-myaccount.us-west-2.aws.snowflakecomputing.com, myorg-myaccount",
+    "https://myorg-myaccount.snowflakecomputing.com, myorg-myaccount",
+    "https://myorg.us-central1.gcp.snowflakecomputing.com, myorg",
+    "https://xy12345.snowflakecomputing.com, xy12345",
+    "https://xy12345.us-west-2.snowflakecomputing.com, xy12345",
+    "http://myorg-myaccount.azure.snowflakecomputing.com, myorg-myaccount",
```
kacpermuda:
```suggestion
    "https://locator.us-central1.gcp.snowflakecomputing.com, locator.us-central1.gcp",
    "https://xy12345.snowflakecomputing.com, xy12345",
    "https://xy12345.us-west-2.snowflakecomputing.com, xy12345.us-west-2",
    "http://myorg-myaccount.azure.snowflakecomputing.com, myorg-myaccount.azure",
```
I think this is how it should behave to be true to how snowflake identifies accounts. We need the whole part of the url pre to `snowflakecomputing.com` to be able to know where the query was executed. Locator itself, without the region is not useful at all, as for different regions it may be used by different organizations.

We should add `snowflake://{locator}.{region}.{cloud}?` to the naming convention and together with that change, it should solve the problem. This also how f.e. Airflow handles snowflake identifiers, the locator in region is still supported.

mobuchowski: Changed that.

integration/spark/vendor/snowflake/src/test/java/io/openlineage/spark/agent/vendor/snowflake/lifecycle/plan/SnowflakeRelationVisitorTest.java
mrpalash-amz: As per Snowflake [documentation](https://docs.snowflake.com/en/user-guide/admin-account-identifier), the `orgname-account_name` combination does **not require** cloud provider or region details.  
We can simplify the URL to:  
`https://microsoft_partner_org-testaccount.snowflakecomputing.com`  

For existing accounts, we should **append region and cloud provider information** to maintain uniqueness in the account identifier.  
Example:  
`https://azureOrg-microsoftpartner_east_us_2_azure.snowflakecomputing.com`  

**Note:** Since the existing account name `microsoftpartner` is already in use, the customer created a new organization name `azureOrg`. They now use the account identifier defined here:  
`https://azureOrg-microsoftpartner_east_us_2_azure.snowflakecomputing.com`

client/java/src/test/java/io/openlineage/client/utils/SnowflakeUtilsTest.java
mrpalash-amz: Supporting the account locator format (`snowflake://{locator}.{region}.{cloud}`) is valid, but it introduces a limitation.  
When customers use this format, all dataset identifiers follow  
`snowflake://{locator}.{region}.{cloud}.{database}.{schema}.{table}`.  

If they later switch to the `orgname-account_name` format, existing lineage nodes will not connect to new ones due to identifier mismatches.  

**Recommendation:**  
Continue using a single consistent identifier format (`orgname-account_name`) to avoid lineage breaks, or clearly document this limitation in OpenLineage so users are aware of the potential impact.

mobuchowski: @mrpalash-amz 

>If they later switch to the orgname-account_name format, existing lineage nodes will not connect to new ones due to identifier mismatches.

That's true. However

> Continue using a single consistent identifier format (orgname-account_name) to avoid lineage breaks

is impossible - there is no way to canonicalize one format to another. Account locator does not give us orgname-accountname format, while the other way around we don't know the cloud locator.

> clearly document this limitation in OpenLineage so users are aware of the potential impact

I've modified the naming spec website page to make that clearer. Thanks.

kacpermuda: Agree with the above from Maciej. It's up to the user to be consistent in their identifier use, as we can't deduce one from another. Good that we now mention it in the docs.

Issue #4025, opened 2025-09-08
[BUG] Remove unnecessary double quotes from lineage events for snowflake datasets like database schema, table names when snowflake tables are in lowercases


mrpalash-amz:
### OpenLineage integration/client

Spark integration

### OpenLineage version

I followed this documentation to setup glue etl job lineage events: https://docs.aws.amazon.com/datazone/latest/userguide/datazone-data-lineage.html

Datazone transport layer package : https://github.com/OpenLineage/OpenLineage/blob/main/client/java/transports-datazone/src/main/java/io/openlineage/client/transports/datazone/AmazonDataZoneTransport.java

### Environment configuration

Glue verions - glue 5.0 sparks 3.5 scala 2, python 3

### Problem details

Create a spark job in python which reads input data from existing snowflake tables and create new tables in snowflake and write transform data in new tables. This job should emit lineage events which has namespace snowflake://{organization name}-{account name}.snowflake.computing.com.

### What you think should happen instead

Showing tables/view/columns in lineage node with double quotes is not a good idea. Because double quotes are needed only to preserve the case in snowflake. Once asset is created. I don't see any point to keep data with double quotes.

### How to reproduce

create snowflake tables in lower cases and then write a  a spark job in python which reads input data from "input_table" snowflake tables and create "output_table" in snowflake and write transform data in output. This job should emit lineage events which input and output datasets name  enclosed by double quotes

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

dolfinus:
Could you please give an example of Snowflake query and produced OpenLineage event?

mrpalash-amz:
Sample GLUE ETL job which produced lineage event with double quotes : 

```
import logging
from awsglue.context import GlueContext
from pyspark.context import SparkContext
from pyspark.sql import functions as F
from pyspark.sql import types as T

logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)

sc = SparkContext()
glueContext = GlueContext(sc)
spark = glueContext.spark_session

sfOptions = {
    "sfURL": "XXXXX-XXXX.snowflakecomputing.com",
    "sfUser": "XXXX",
    "sfPassword": "XXXXXXXX",
    "sfDatabase": "\"poc\"",
    "sfWarehouse": "COMPUTE_WH",
    "sfRole": "ACCOUNTADMIN"
}

def read_snowflake_table(sfSchema, table, sfOptions):
    return spark.read \
[...]

PR #4105, opened 2025-10-24
Snowflake account strip quotes

mobuchowski:
Fixes https://github.com/OpenLineage/OpenLineage/issues/4025
Based on #4104 - need to be merged first.

### DIFF
--- a/integration/spark/vendor/snowflake/src/main/java/io/openlineage/spark/agent/vendor/snowflake/lifecycle/SnowflakeDataset.java
+++ b/integration/spark/vendor/snowflake/src/main/java/io/openlineage/spark/agent/vendor/snowflake/lifecycle/SnowflakeDataset.java
@@ -5,13 +5,17 @@
 
 package io.openlineage.spark.agent.vendor.snowflake.lifecycle;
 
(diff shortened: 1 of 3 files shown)


integration/spark/vendor/snowflake/src/main/java/io/openlineage/spark/agent/vendor/snowflake/lifecycle/SnowflakeDataset.java
```diff
@@ -41,17 +45,90 @@ public static <D extends OpenLineage.Dataset> List<D> getDatasets(
...
+    }
+
+    // Extract account identifier (first segment before the first dot)
+    int firstDotIndex = url.indexOf('.');
+    if (firstDotIndex > 0) {
+      return url.substring(0, firstDotIndex);
+    }
+
+    return url;
+  }
```
kacpermuda: Is this a leftover from another PR? There is also some testing about this, but it's not extracting correct identifier if the locator.region.cloud is used.

mobuchowski: Fixed.

PR #4109, opened 2025-10-28
spec: add a batch API endpoint as part of the core specification

jakub-moravec:
### Problem

Some of the existing consumers and producers can work with batches of OpenLineage events, but the API for this is not standardized. 

### Solution

This PR aims to document the API endpoint, provide necessary documentation, and potentially even SDK support. The first PR draft contains only a subset of the changes to allow for discussion about the proposal and alignment on what should be in the scope of the PR. 

The batch API endpoint should be optional, meaning a clear contract is needed for clients to determine whether or not it is supported (and fallback to the existing single-event API if it is not supported). I would like to discuss which of the following approaches is preferable: 
- if the endpoint returns HTTP 404/501, it should be considered unsupported by the client and the client should fallback to the single-event API
- an additional endpoint (e.g. `/api/v1/options`) could be added, the response would determine if both APIs are supported, or if the implementation supports only the single-event API. While this would be a cleaner design, it comes with a problem of requiring all implementations supporting the new options endpoint so in the end, the proposed change would not be completely opt-in. For that reason, I consider the former approach to be better. 

#### One-line summary:

Add a batch API endpoint as part of the core specification. 

### Checklist

----

### DIFF
--- a/spec/OpenLineage.yml
+++ b/spec/OpenLineage.yml
@@ -9,8 +9,8 @@ info:
 paths:
   /lineage:
     post:
-      summary: Send an event related to the state of a run
-      description: Updates a run state for a job.
+      summary: Send an event related to the state of a run, job, or a dataset
+      description: Updates a state of a run, job, or a dataset.
       operationId: postRunEvent
       tags:
         - OpenLineage
@@ -25,3 +25,79 @@ paths:
       responses:
         "200":
           description: OK
+  /lineage/batch:
+    post:
+      summary: Send a batch of events related to the state of a run, job, or a dataset
+      description: A batch of events with updates of a state of a run, job, or a dataset.
+      operationId: postEventBatch
+      tags:
+        - OpenLineage
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                oneOf:
+                  - $ref: "OpenLineage.json#/$defs/RunEvent"
+                  - $ref: "OpenLineage.json#/$defs/DatasetEvent"
+                  - $ref: "OpenLineage.json#/$defs/JobEvent"
+      responses:
+        "204":
+          description: OK response - batch was received
+        "200":
+          description: OK response - batch was received, response body contains processing details
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  status:
+                    type: string
+                    enum:
+                      - success
+                      - partial_success
+                    description: Indicates that some events failed to process
+                  summary:
+                    type: object
+                    properties:
+                      received:
+                        type: integer
+                        example: 10
+                      successful:
+                        type: integer
+                        example: 6
+                      failed:
+                        type: integer
+                        example: 4
+                      retriable:
+                        type: integer
+                        example: 2
+                      non_retriable:
+                        type: integer
+                        example: 2
+                    required:
+                      - received
+                      - successful
+                      - failed
+                  failed_events:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        index:
+                          type: integer
+                          example: 0
+                        reason:
+                          type: string
+                          description:
+                            "Failure reason (example values: Unsupported event type, Unsupported facets, Server error,
+                            Timeout)"
+                          example: "Unsupported event type, Unsupported facets, Server error, Timeout"
+                        retriable:
+                          type: boolean
+                          example: true
+                required:
+                  - status
+                  - summary


dolfinus:
> if the endpoint returns HTTP 404/501, it should be considered unsupported by the client and the client should fallback to the single-event API

Instead of probing, why not introduce option to HttpTransportConfig which will choose between batch and single endpoint implementations?

jakub-moravec:
@dolfinus might be a good idea to combine the approaches? 
- OpenLineage SDK 
   - HttpTransportConfig determining whether to use batch or single-event endpoint
   - If batch is chosen, but consumer doesn't support it, should we fail or fallback? 
- custom producer 
   - behavior is defined by whoever creates it, but there needs to be some sort of contract based on which the creator of the producer can decide whether they can use the batch API or not

dolfinus:
> If batch is chosen, but consumer doesn't support it, should we fail or fallback?

If user explicitly configured that batch should be used, but it's not supported, then it should fail

kacpermuda:
We could probably also suggest some behavior for consumers on how to act when some of the events in batch have failed like fail the whole batch and return error, or maybe return indexes of failed events and accept the passing ones etc. It will help to write the possible client retry behavior across different consumers, if they behave in the same way.

dolfinus:
Shouldn't that be a transport implementation detail?

kacpermuda:
> Shouldn't that be a transport implementation detail?

But if different consumers would behave differently, how could we implement some retry logic in the transport? It is an implementation detail, but we should probably suggest some way of behavior?

mobuchowski:
There's overall a bunch of issues we need to answer:

- How do we keep this fully additive so no one breaks? We should stick to per-event by default and only batch when explicitly enabled.
- Can we detect support for the batch endpoint on the producer side: for example by OPTIONS on the batch endpoint?
- If detection fails or we get 404/405/501, do we automatically fall back to per-event and emit a clear warning/metric?
- What are we doing for other transports? Seems it's trivial for Kafka and File. Should we do anything for other transports?
- Expose batching as new transports (e.g., HttpBatchTransport) or as methods/config on existing ones?
- Do we allow partial success, and how do we report which items failed?
- What’s the retry/backoff policy for batch failures vs per-event fallback?
- What ordering guarantees do we give inside a batch?
- Are batches atomic, or can servers accept partials?
- Should we have explicit idempotency keys (like run_id + event time?) so retries don’t duplicate events?
- What’s the max events per batch and max payload size?
- How should clients split and retry on something like 413 Payload Too Large?

jakub-moravec:
> Can we detect support for the batch endpoint on the producer side: for example by OPTIONS on the batch endpoint?

I tried to provide my PoV on this in the description. If we require a separate OPTIONS endpoint (either support for the OPTIONS method on the existing endpoint or an entirely new endpoint)... it will either have to be optional (and thus not reliable), or it's a breaking change. So pragmatically, it seems better to me to rely on 404/405/501 responses, as the publishers will need to handle them anyway.

> If detection fails or we get 404/405/501, do we automatically fall back to per-event and emit a clear warning/metric?

I think it would be a nice quality of life improvement, but we could also just fail the batch as @kacpermuda suggested.

> Do we allow partial success, and how do we report which items failed?

This is a good one. I would say partial success makes sense. The question is the semantics of that
- one event in the batch is invalid (doesn't comply with OL schema, for example)
- consumer doesn't support all event types, so they will process Run events correctly, but not dataset events 
... should we distinguish between these? 

Another concern is... our consumer is asynchronous, not sure how common that is. But in our case, we could only confirm that we accepted the batch synchronously, and then report which events failed in a separate response. Does it make sense to design asynchronous interactions? 

- POST batch (return batch ID)
- GET batch/{id} - retrieve status

> What ordering guarantees do we give inside a batch?

While the transport can provide a guarantee within one batch, the consumer doesn't have to. 

> Are batches atomic, or can servers accept partials?

The consumer may not support all event types and just process the ones that are supported. But apart from that, it should try to process all events in the batch, right? 

> Should we have explicit idempotency keys (like run_id + event time?) so retries don’t duplicate events?

That would bring new statefulness requirements. Doesn't seem to be a good tradeoff to me. 

> What’s the max events per batch and max payload size?

Should this be configurable? 

> How should clients split and retry on something like 413 Payload Too Large?

This would be a concern if the max batch size is configurable, right?

dolfinus:
Please consider not to adding any kind of automatic batch support detection using additional `OPTION` request. If the main purpose of this feature is to reduce number of requests send via HTTP, introducing automatic detection will make x2 requests to server. Relying on 4xx/5xx responses is also a bad idea.

> Do we allow partial success, and how do we report which items failed?
> This is a good one. I would say partial success makes sense. The question is the semantics of that
> one event in the batch is invalid (doesn't comply with OL schema, for example)

How event produced by OL client could not comply with OL schema if it accepts only Java/Python classes generated from OL schema?

kacpermuda:
> How event produced by OL client could not comply with OL schema if it accepts only Java/Python classes generated from OL schema?

I think we're discussing it on a spec level here, so we do not assume the events will be emitted only by our clients, there can be many different producers and the consumers must assume that an OL endpoint can be provided with invalid OL events.

dolfinus:
I think that these questions should be discussed in a dedicated issue. Other than batch HTTP API endpoint there can be other transport types which may benefit from batching support (e.g. Kafka  #3771). 

If logic of partial writes/atomic writes, retries, ordering and so on will become a part of OpenLineage (not as transport implementation detail but rather a common mechanism for all transport), when other transports should be included into this discussion. For example, for Kafka using atomic writes requires enabling transaction support which can be a pain to configure and to use.

mobuchowski:
>Please consider not to adding any kind of automatic batch support detection using additional OPTION request. If the main purpose of this feature is to reduce number of requests send via HTTP, introducing automatic detection will make x2 requests to server. 

Only if the connection is used only once. If you have a producer emitting high number of events, the check could be done only once.

For example, there are dbt runs that generate over 40k events. My solution was to add async http transport, which works well too. But the batch transport would drastically reduce the request count, and the single request would not matter in a grand scheme - especially since it does not require backend to do anything, really, other than immediately respond.

However I'm not trying to push that option, if the consensus is to not use this then let's not do it.


> This is a good one. I would say partial success makes sense. The question is the semantics of that one event in the batch is invalid (doesn't comply with OL schema, for example) consumer doesn't support all event types, so they will process Run events correctly, but not dataset events ... should we distinguish between these? 

We did not specify this even for single events, really. If the consumer does not support given event type, facet or something else like this, there is little producer can do. On the other hand, if I'm sending malformed events, I'd want to have 400-level error from the consumer.  


> Another concern is... our consumer is asynchronous, not sure how common that is. But in our case, we could only confirm that we accepted the batch synchronously, and then report which events failed in a separate response. Does it make sense to design asynchronous interactions?

To be honest I think this is outside the scope of this discussion. If you process them asynchronously, you can solve the retry issue on the consumer side, and don't need client level retries for that.

I think we should think of potential issues like 

- part of batch contains malformed events
- part of batch failed to be processed due to transient failure 
- part of batch failed in a way that retry won't help - for example, events are too large

>This would be a concern if the max batch size is configurable, right?

I think it should be, both in event count size, and byte size. And this should not only be a property of particular transport, as it has some complex interactions with CompositeTransport - imagine a CompositeTransport trying to send batch events to two consumers, both of which have different batch configurations.

That's why I think we should have code-level interface that does not guarantee the batches send to the consumer are just directly passed, but that takes collections of events, and single events, and takes care of properly splitting them into batches internally.

jakub-moravec:
@mobuchowski @dolfinus @kacpermuda thank you for the discussion! Would it make sense to split this into several steps/tickets/PRs? 

- one (this) ticket just to define and document the API endpoint => once delivered, custom producers/consumers can start implementing it, but there will not be any support by the OL client SDK 
- one ticket for changes for the existing SDK - for HTTP transport 
- subsequent events if needed for other transports 

Would that make sense? Or do you think all of this needs to be discussed (and delivered) together?

jakub-moravec:
For the API spec, I see the following points mentioned above as the main open-points: 

- How do we report back the result of consuming individual events from the batch
- Sync versus async API - @mobuchowski I think this is part of the API spec discussion, as the API will look differently if optimized for sync vs async communication 

Let's say the response detailing the status of individual events (_batch status body_) can look something like this: 

``` 
{
  "status": "partial_success",
  "summary": {
    "received": 10,
    "processed_successfully": 6,
    "failed": 4,
    "retriable": 2,
    "non_retriable": 2
  },
  "failed_events": [
    {
      "index": 0,
      "reason": "Unsupported event type | Unsupported facets | Server error | Timeout",
      "retriable": true
    }
  ]
}
```

Synchronous interaction

1. Client -- POST batch --> Server 
2. Client  <-- response (HTTP 200): batch status body -- Server 

Asynchronous interaction
1. Client -- `POST batch` --> Server 
2. Client  <-- `response (HTTP 202): batch ID` -- Server 
3. Client -- `GET batch/[ID]/status` --> Server 
4. Client  <-- `response (HTTP 200): batch status body` -- Server 


We could choose one of the approaches, or potentially even support both.

dolfinus:
Asynchronous interaction requires storing raw OL events in some database, tracking each OL event (generating ID/hash, for example), so statuses of parsing these event can be fetched on demand. This may require changing architecture of consumers in a very specific way, which may be incompatible with used technologies.

For example, the project [I'm working on right now](https://github.com/MobileTeleSystems/data-rentgen) stores raw OL events in Kafka topic, and then handles them in a background. Message key == `run:{job.namespace}/{job.name}` for ordering reasons. Multiple OL events will have the same key. If I need to generate individual key for each event, I'll lost ordering of events in partition. Also Kafka doesn't support fetching individual event by key, it requires to fullscan topic to get specific message. Also to detect if some message was handled by background worker or not I have to write each OL event status to some queue/database.

At the same time, receiving a batch of OL events in one HTTP query, and then saving them to Kafka in the same order, is very easy to do - no tracking of individual events, malformed messages can be rejected on the fly using HTTP body data + spec validator.

jakub-moravec:
@dolfinus correct. I think the question is whether to define both sync and async interaction and let each consumer decide which one they'll implement (if any), or to do just the sync one. (Async only doesn't seem to be an option for the reasons you outlined).

dolfinus:
IMHO spec should define the easiest endpoint logic which any consumer could implement. If there's a specialized consumer which have a different API, then it should be implemented as custom/third-party transport.

mobuchowski:
At this point I agree with simplification. Let's just

- define batch endpoint, literally, as accepting json list of events
- the response _optionally_ would include partial success report as described by @jakub-moravec. Build-in transport's retry mechanism will keep retrying the retrieable events. 
- any processing status would be an extension that is not standardized

jakub-moravec:
@dolfinus @mobuchowski sounds good to me - I updated the OpenAPI spec (see changed files), I would propose two OK response codes then - 204 for OK response without any details (works for async consumers too), and 200 for OK response with the retriable items, etc. I think it will help the clients decide whether they need to attempt any re-tries or not. Or would you keep just one response code with the response body optional?

mobuchowski:
@jakub-moravec agree with specifying that via response codes 👍

dolfinus:
I'm also okay with that, thank you!

jakub-moravec:
@dolfinus @mobuchowski thanks! Would it make sense then to merge the spec changes in this PR and create separate one(s) for the client lib changes?

spec/OpenLineage.yml
```diff
@@ -25,3 +25,75 @@ paths:
...
+                      - success
+                      - partial_success
+                    description: Indicates that some events failed to process
+                  summary:
+                    type: object
+                    properties:
+                      received:
+                        type: integer
+                        example: 10
+                      processed_successfully:
```
kacpermuda:
```suggestion
                      successful:
```
to be consistent, we have "failed", "received" etc.

spec/OpenLineage.yml
```diff
@@ -25,3 +25,75 @@ paths:
...
+                          type: integer
+                          example: 0
+                        reason:
+                          type: string
+                          description: Failure reason
+                          enum:
+                            - Unsupported event type
+                            - Unsupported facets
+                            - Server error
+                            - Timeout
```
kacpermuda: Not sure if we should limit ourselves with enum here, we may need different failure reasons without changing the spec file

spec/OpenLineage.yml
```diff
@@ -25,3 +25,75 @@ paths:
...
+                        retriable:
+                          type: boolean
+                          example: true
+                required: []
```
kacpermuda: I understand that for 200 we can't have anything required, but is there any way to say, if any content is present, it must have keys: ["status"] etc. ?

dolfinus: > I understand that for 200 we can't have anything required

According to this comment https://github.com/OpenLineage/OpenLineage/pull/4109?notification_referrer_id=NT_kwDOAEcfHbMxOTk4MDkzMDExNjo0NjYxMDIx#issuecomment-3491527208, 200 response has some non-empty body, so there should be some required keys

jakub-moravec: @kacpermuda I set the most important parts of the response body as required, is this OK or should some other part be required too?

spec/OpenLineage.yml
jakub-moravec: Changed to string with examples

spec/OpenLineage.yml
jakub-moravec: Changed

mobuchowski (approved):

spec/OpenLineage.yml
```diff
@@ -25,3 +25,77 @@ paths:
...
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        index:
+                          type: integer
+                          example: 0
+                        reason:
+                          type: string
+                          description: "Failure reason (example values: Unsupported event type, Unsupported facets, Server error, Timeout)"
```
kacpermuda:
```suggestion
                          description: "Failure reason"
```

kacpermuda:
Good to merge after fixing the pre-commit

Issue #4073, opened 2025-10-06
[PROPOSAL] Generic Control Flow Lineage Facet


luke-hoffman1:
### **Purpose**
Enterprises rely on complex orchestration frameworks to execute data processing workloads. OpenLineage excels at capturing data lineage (i.e., how datasets are produced and consumed), but **control flow lineage**—the direct predecessor and successor relationships between jobs in a workflow—is not a first-class citizen in the spec.

Currently, this concept is only represented by a source-specific facet, `run.facets.airflow.task.upstream_task_ids`. This requires every new framework integration (e.g., Azure Data Factory, Prefect, Dagster) to create its own custom facet to describe its unique orchestration graph, leading to inconsistency and redundant effort. This proposal introduces a **new, generic facet** to model control flow in a standardized way for any source.

---

### **Spec Gaps Addressed**

* **Standardization:** Without a generic facet, every framework integration must invent a new custom facet (e.g., `adf.upstream_activity_ids`), preventing a uniform view of orchestration across different tools.
* **Interoperability:** In modern data stacks, workflows often span multiple frameworks (e.g., an Airflow DAG triggering an Azure Data Factory pipeline). A generic control flow facet would allow for a seamless, end-to-end lineage graph across these tool boundaries.
* **Debuggability:** When a job fails, it's critical to quickly identify its operational blast radius. A generic control flow facet immediately answers the questions: "**what job runs need to succeed for this job to run?**" and "**if this run succeeds, what other jobs will run?**"
* **Clarity:** It distinguishes the *'why'* a job runs (control flow) from the *'what'* it produces (data flow), allowing lineage visualization tools to create richer, more intuitive graphs that show both types of dependencies.

---

### **Use Cases**

* **Multi-level Orchestration:** In combination with the parent facet, provide the ability to model complex workflows within a single framework, such as an Azure Data Factory pipeline that orchestrates other child pipelines.
* **Cross-Framework Lineage:** Natively connect a run from one orchestrator (e.g., Airflow) to a run in another (e.g., a Spark job on Databricks), providing a complete operational picture.
* **Operational Insight:** Immediately identify downstream jobs impacted by an upstream failure, or trace back the chain of execution that led to a specific run, independent of the datasets involved.

---

### **Proposed Implementation**

Add a **new, optional facet** on the existing OpenLineage `Run` object. This approach requires no core schema changes and is fully additive, ensuring backward compatibility.

[...]

PR #4112, opened 2025-10-29
feat(spec): add JobDependenciesRunFacet

kacpermuda:
### Problem

Closes: #4073

### Solution

Introduce JobDependenciesRunFacet

#### One-line summary:

feat(spec): Introduce JobDependenciesRunFacet

### Checklist

----


spec/facets/JobDependenciesRunFacet.json
```diff
@@ -0,0 +1,92 @@
...
+        },
+        "run": {
+          "$ref": "#/$defs/RunIdentifier"
+        },
+        "root": {
+          "$ref": "#/$defs/RootIdentifier"
+        }
+      },
+      "required": ["job"]
+    },
```
kacpermuda, line 59: @mobuchowski Do you think we should add these run/job related objects to a base spec? We're repeating that in parentRunFacet and here. We can do it now or on next spec major release, added that idea to the issue already.

website/docs/spec/facets/run-facets/job_dependencies.md
```diff
@@ -0,0 +1,97 @@
+---
+sidebar_position: 5
+---
+
+# Job Dependencies Facet
+
+# TODO - description
```
kacpermuda: TODO for me

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
```diff
@@ -0,0 +1,92 @@
...
+          "type": "string"
+        },
+        "name": {
+          "description": "The unique name of a job within that namespace",
+          "type": "string"
+        }
+      },
+      "required": ["namespace", "name"]
+    },
+    "RootIdentifier": {
```
dolfinus: Could we add a description what this field is used for?

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
```diff
@@ -0,0 +1,92 @@
...
+        "run": {
+          "$ref": "#/$defs/RunIdentifier"
+        },
+        "job": {
+          "$ref": "#/$defs/JobIdentifier"
+        }
+      },
+      "required": ["run", "job"]
+    },
+    "JobRunIdentifier": {
```
dolfinus:
```suggestion
    "JobDependency": {
```

If in future new fields will be added like `description`, `type`, `mandatory/optional`, model name will not be in sync with its content

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
kacpermuda: Done

dolfinus: I'm a bit confused what is the difference between this root run and parentRun.root

kacpermuda: It's the same. Wanted to keep everything in one spec file here for simplicity. I left a comment above, that maybe we should put some of these in the base spec instead, it looks like we might be needing it in multiple places.

website/docs/spec/facets/run-facets/job_dependencies.md
```diff
@@ -0,0 +1,105 @@
...
+# Job Dependencies Facet
+
+Modern data workflows often involve jobs that depend on the completion of other jobs before they can run. 
+While tools like Apache Airflow track these dependencies internally, understanding them across systems is 
+essential for full lineage visibility.
+
+The ``JobDependenciesRunFacet`` captures direct upstream and downstream relationships between job runs, 
+allowing OpenLineage to represent how execution order flows through a pipeline. It records which job runs 
+must finish before the current run begins, and which job runs are waiting on this one to complete. It can 
+also specify the trigger rule — for example, whether all upstream jobs must succeed before this one starts, or only one.
```
jakub-moravec: Would it make sense to add a disclaimer saying that when this facet is used to document a dependency between two jobs, it doesn't imply that data is being passed from one job to another? So that the difference between runtime dependency (control flow) and data lineage (data flow) is more obvious?

kacpermuda: That would definitely improve clarity. Would you like to take a shot at it and propose an updated version of the whole description after making the changes?

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
dolfinus: I mean, what's the use case for this? To annotate that Airflow Task2 depends in Task1 created by DagRun1?

kacpermuda: Right now in most cases in Airflow it will be the same as root in parentRunFacet, but in other systems it does not have to be, so we should allow passing it as optional here. Also for Airflow: the task root does not have to be a dagrun1, it can go further than that - I'm working on extending it right now, so that it can propagate through airflow parent-child processes (e.g. triggerDagRunOperator) and also allow passing parent information to dagrun when triggering by API - then the root can be anything.

dolfinus: What do you thinkg about having a `type` field of JobDependency which will describe how current job/run depends on upstream (`TRIGGER`, `FLOW`)?

kacpermuda: That's an interesting idea. It would possibly interfere with parent run facet a bit, but maybe it's not a bad thing.
cc @jakub-moravec @mobuchowski WDYT?

jakub-moravec: @dolfinus 

What would be the semantics of `FLOW`? A data movement? 

Another type could be smth like `RUNS_AFTER`? Job2 needs to run after Job1, otherwise it doesn't have up-to-date inputs, but at the same time Job1 is not triggering Job2 directly?

website/docs/spec/facets/run-facets/job_dependencies.md
jakub-moravec: @kacpermuda sure, happy to. I guess it's better to wait until the other review threads are resolved

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
dolfinus: Not a data movement, just logical dependency, task1 -> task2 in Airflow DAG. Yes, naming should be adjusted to avoid confusion

kacpermuda: What is the proposed path forward then? Adding a string field named `type` in JobDependency that will allow to specify the relationship type? What would be the final examples we can provide to the user? `TRIGGER`  and `RUNS_AFTER`? Is trigger implying parent/child relationship and would be a duplicate of parent run facet info?

dolfinus: > Is trigger implying parent/child relationship

It's not. Using Airflow as example:
* dag1 -(parent)-> dag1.task1
* dag2 -(parent)-> dag2.task2
* dag2 -(parent)-> dag2.task2
* dag1.task1 -(triggers)-> dag2.task2 using [TriggerDagRunOperator](https://airflow.apache.org/docs/apache-airflow-providers-standard/stable/operators/trigger_dag_run.html)
* dag2.task2 -(dependency)-> dag2.task3

Currently OL sends only:
* dag1 -(parent)-> dag1.task1
* dag2 -(parent)-> dag2.task2
* dag2 -(parent)-> dag2.task2

kacpermuda: I still don fully follow how a job can trigger another job and not be its parent (regardless if it's currently supported by OL Airflow or not).

From your example:  dag1.task1 triggers dag2.task2. Let's assume it's triggering a dag2, as it can't trigger the task itself. So dag1.task1 triggers dag2, it means parent run of dag2 is dag1.task1 and after [this](https://github.com/apache/airflow/pull/57809) and the following change to triggerdagrunoperator is merged, we will put that in DagRun's parent. I agree that currently the OL does not send this relationship explicitly, but it can and will and I'm more concerned about: what would be the difference between TRIGGER relationship in this facet vs parent-child relationship present in parentRunFacet (apart from that in this new facet we can have many "TRIGGER" jobs).

dolfinus: Another example I can imagine.

I have an application which under the hood starts Spark session, and then can run some actions (`df = read(); df.write()`, `df.count()` and so on) based on API requests. Currently OL graph looks like this:
* `My App instance -(parent)> Spark session`
* `Spark session -(parent)-> Spark SQLJob`

And there is no way to express:
* `My app HTTP request -(triggers)-> Spark SQLJob`

Spark SQLJob is both a part of Spark session and triggered by external request, not just one parent.

kacpermuda: Ok, two issues with that:

1. `Spark SQLJob` is not actually triggered by `My app`, the `Spark session` is, so I'm not sure if this new facet should include child of a child job (etc.). 
2. We have root part in parentRunFacet, to track the root job, so in that case, `Spark SQLJob` could have that reference to `My app`.

It's a bit more clear to me now what the possible use case is for type=TRIGGER in JobDependency, I'm just not sure if we should actually pursue it, as it can be confusing and contradict the info from parentRunFacet in some cases.

So if we had:
- Airflow DAG -(parent)> My App instance
- My App instance -(parent)> Spark session
- Spark session -(parent)-> Spark SQLJob

The Spark SQLJob could have: root=Airflow DAG, parent=Spark session and then in the jobdependencyfacet it could have upstream({type=TRIGGER, job=My App instance}). Is that not confusing or am I still not getting the point here?

mobuchowski: I think it makes sense to express that dependency. Right now we don't have a good way to represent pull-based pipelines - parent run facet explicitely describes top-down dependencies.

spec/facets/JobDependenciesRunFacet.json
mobuchowski: Yeah, the resulting event is the same. However, our code generators would create a different code structure, potentially breaking users code.

I think it makes sense at some point, and it wouldn't be "major OL release", but a major "OL library release" if that makes sense. The consumers should be not impacted, and old producers would continue to work with new consumers too.

kacpermuda: Ok, leaving it in the issue with ideas for major release and not pursuing it now. Hopefully, there will be no more facets needing definition of run, job and roots as we're getting out of possible names

jakub-moravec:
(Edit: this doesn't take in consideration other comments from today, I missed those)

@dolfinus @kacpermuda 

- I added a draft of the dependency `type`, please review the enum constants and adjust or add more. From my PoV, I don't mind that it overlaps a little bit with the parent-run facet, but it might be worth specifying and documenting using examples which one should be used when. My take is
   - Parent Run facet says that taskA run within Job1 
   - Job Dependencies facet says that taskA was explicitly triggered by Job1 (DIRECT_INVOCATION dependency), and that taskB can be run after taskA finishes (IMPLICIT_DEPENDENCY)
- Side not: we would be supporting both use-cases described here than https://github.com/OpenLineage/OpenLineage/issues/4073#issuecomment-3452071005   
- What is weird to me, though, after thinking about this more, is having the `root` as part of `JobDependency` - this straight up duplicates what is defined in Parent run facet, doesn't it? 
- I also added more fine-grained conditions that could be used to document in more detail when/under which conditions the downstream job was/can be executed. That's something I missed in the proposed facet when going through it in more detail, but at the same time, this might make more sense for design lineage (only Job is specified, no Run), so I'm wondering what you'll think about that... but then again, I guess the same applies for `trigger_rule`

I prepared the changes here https://github.com/OpenLineage/OpenLineage/compare/feat-spec-job-dependencies-facet...jakub-moravec:OpenLineage:feat-spec-job-dependencies-facet - please let me know if I should create a PR to merge them to your working branch

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
dolfinus: > Spark SQLJob is not actually triggered by My app, the Spark session is

If no `My app HTTP request` received, then `Spark SQLJob` will not start. SQLJob is created only then some _action_ is performed using Spark DataFrame API or Spark SQL. It _requires_ running Spark session, and cannot be done without it. Also I can't tell that Spark session _triggered_ this SQLJob, it was a user request.

> We have root part in parentRunFacet, to track the root job, so in that case, Spark SQLJob could have that reference to My app.

`parentRunFacet.root` is an arbitrary starting point of the chain of runs. It can be just anything - as you've already mentioned, it can be `My app` or it can be `Airflow DAG`, whatever. This is not something reliable.

I'd prefer not to break `Spark session -(parent)-> Spark SQLJob` or `My app -(parent)> My app HTTP request` bonds. Instead we can introduce _different_ types of these bonds, maybe even arbitrary ones.

This discussion can be moved into an issue, to discuss more scenarios.

kacpermuda:
Thanks, I can apply those changes manually, it's not a lot of code.

> What is weird to me, though, after thinking about this more, is having the root as part of JobDependency - this straight up duplicates what is defined in Parent run facet, doesn't it?

Yes, I think it's a bit of duplication here. My thinking was to easily expose root information, but maybe that's not optimal solution. The root info is about the job dependency's root, so having that in the facet allows us to easily track root for each job dependency, but maybe we can always assume, that if somebody wants root info, they will go the event for that job and look at parent run facet.

website/static/spec/facets/1-0-0/JobDependenciesRunFacet.json
jakub-moravec: "I still don fully follow how a job can trigger another job and not be its parent (regardless if it's currently supported by OL Airflow or not)."

This can be as simple as one shell script calling another

spec/facets/JobDependenciesRunFacet.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        "dependency_type": {
+          "type": "string",
+          "description": "Used to describe whether the upstream job directly triggers the downstream job, or whether the dependency is implicit (e.g. time-based).",
+          "example": "DIRECT_INVOCATION|IMPLICIT_DEPENDENCY"
+        },
+        "sequence_trigger_rule": {
+          "type": "string",
+          "description": "Used to describe the exact sequence condition on which the downstream job can be executed (FINISH_TO_START - downstream job can start when upstream finished; FINISH_TO_FINISH - job executions can overlap, but need to finish in specified order; START_TO_START - jobs need to start at the same time in parallel).",
+          "example": "FINISH_TO_START|FINISH_TO_FINISH|START_TO_START"
```
dolfinus, line 50: Can you please provide an example of `FINISH_TO_FINISH`? I didn't get how it is possible, it looks like a parent -> child relation

kacpermuda:
Added proposed impl from @jakub-moravec (with str instead of enums to be more flexible in the future) and removed the root. Still reading into it, but wanted to have this code in the PR so it's easier to discuss

spec/facets/JobDependenciesRunFacet.json
kacpermuda: ^ @jakub-moravec

jakub-moravec: @dolfinus I think the main use case is when you have jobs that periodically move that, streams, etc. One job writing to a table, another reading from it and moving the increments elsewhere. They can run in parallel, but the upstream job needs to finish first, so that all increments are moved.

dolfinus: Ok, thanks

mobuchowski (approved):

PR #4116, opened 2025-11-03
spec: add ordinal position to SchemaDatasetFacet

mobuchowski:
Small addition to spec: ordinal_position that describes the position of the field in schema.
It's 1 indexed, because most of the databases are doing it that way.


kacpermuda:
We should probably bump the facet version?

website/static/spec/facets/1-1-1/SchemaDatasetFacet.json
kacpermuda, line 1: This is a leftover from previous commit, where version was not bumped.

pawel-big-lebowski (approved):

website/static/spec/facets/1-1-1/SchemaDatasetFacet.json
mobuchowski: Removed.

PR #4124, opened 2025-11-07
spark: Add schema and CLL facets for Snowflake writes

kchledowski:
Add schema and column-level lineage support for Snowflake.

### Meaningful description
The OpenLineage Spark integration currently lacks schema and column-level lineage metadata for datasets when Spark uses the Snowflake connector to read from or write to Snowflake.

This PR enhances the Spark integration by adding schema and column lineage facets to lineage events for Snowflake datasets. It parses the SQL query executed via the Spark-Snowflake connector to extract input dependencies and column-level transformations.  The logic is shared between JDBC and Snowflake via a new SqlCollector class, ensuring consistent lineage capture.


integration/spark/vendor/snowflake/build.gradle
```diff
@@ -53,6 +53,7 @@ dependencies {
         exclude group: "org.apache.spark:spark-catalyst_${scala}"
     }
 
+    testImplementation(project(path: ":spark3"))
```
tnazarew, line 56: just checking if it's not added by accident

kchledowski: not by accident, it's needed to invoke `ColumnLevelLineageUtils.buildColumnLineageDatasetFacet` in test

tnazarew (approved):

PR #4145, opened 2025-11-17
website: deploy only last 10 versions

mobuchowski:
This decreases memory pressure for website build process. Most likely we'd need to deploy older versions as separate static sites if we want more history.


harels:
Why 10 specifically?

mobuchowski:
@harels no reason for particular number - it's less than current number of versions so should build successfully, but greater than 2-3 - should cover most of the recent users.

PR #4151, opened 2025-11-18
dbt: add DbtNodeJobFacet

mobuchowski:
Dbt manifest contains additional information that is useful for OL consumers. 
This adds a facet that contains few of those, and is a container which we can use later to add more.

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/facets.py
+++ b/integration/common/src/openlineage/common/provider/dbt/facets.py
@@ -63,6 +63,16 @@ class DbtRunRunFacet(BaseFacet):
 
 @attr.define
 class DbtNodeJobFacet(JobFacet):
+    """Job facet containing dbt node metadata.
+
+    This facet embeds information from the dbt manifest schema, specifically from the node
+    properties as defined in the dbt manifest specification:
+    https://schemas.getdbt.com/dbt/manifest/v12/index.html#nodes_additionalProperties
+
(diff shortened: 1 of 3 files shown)


integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -58,3 +59,16 @@ class DbtRunRunFacet(BaseFacet):
...
         return GITHUB_LOCATION + "dbt-run-run-facet.json"
+
+
+@attr.define
+class DbtNodeJobFacet(JobFacet):
+    original_file_path: Optional[str] = attr.field(default=None)
+    database: Optional[str] = attr.field(default=None)
+    schema: Optional[str] = attr.field(default=None)
+    alias: Optional[str] = attr.field(default=None)
+    unique_id: Optional[str] = attr.field(default=None)
```
kacpermuda, line 80: Can you add some more context on this facet and its content as docstring? Maybe some links to dbt docs?

Also, this is information about any node? Should we add some node_type here as well? Will it have different information based on node type?

Not a dbt expert, just some questions that popped in my head.

mobuchowski: This facet is more about explicitely embedding some information from dbt node. 

We know the node type already via JobTypeJob facet, didn't think duplicating this information is needed. But even if you want it, you can infer it via 

```
nodeType = unique_id.split(".")[0]
```

Updated the docstring with explanation.

Issue #3973, opened 2025-08-19
[BUG] SQL Lineage parse causes SigInt


mattcasters:
### OpenLineage integration/client

Java client

### OpenLineage version

OpenLineage libraries from Maven, version 1.37.0:

`io.openlineage:openlineage-java`
`io.openlineage:openlineage-sql-java`

### Technology and package versions

Java 17, minimal test of 

### Environment configuration

Java 17 on RHEL8:

```
java --version
openjdk 17.0.16 2025-07-15 LTS
OpenJDK Runtime Environment (Red_Hat-17.0.16.0.8-1) (build 17.0.16+8-LTS)
OpenJDK 64-Bit Server VM (Red_Hat-17.0.16.0.8-1) (build 17.0.16+8-LTS, mixed mode, sharing)
` ``

### Deployment details

_No response_

### Problem details

Consider the following SQL query (simplified as much as possible):

```
WITH w AS ( SELECT tw.id AS id FROM tableW AS tw )
SELECT th.id AS id
FROM ( ( tableH AS th JOIN w ON ( th.id = w.id ) ) )
```

With this query in String `sql`, we can now simply run the following Java code:

`OpenLineageSql.parse(List.of(sql), "mysql");

This causes the JVM to exit with the following on the console:

`Process finished with exit code 139 (interrupted by signal 11:SIGSEGV)`


### What you think should happen instead

Instead of crashing, explain what goes wrong in `SqlMeta.errors()`


### How to reproduce

Run the following Java class:

```
import io.openlineage.sql.OpenLineageSql;

import java.util.List;

public class Repro {
    public static void main(String[] args){
        String sql = """
WITH w AS ( SELECT tw.id AS id FROM tableW AS tw )
SELECT th.id AS id
FROM ( ( tableH AS th JOIN w ON ( th.id = w.id ) ) )
""";
        OpenLineageSql.parse(List.of(sql), "mysql");
    }
}
```

### Additional details

_No response_

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

mattcasters:
First analyses results for https://github.com/apache/hop/issues/2946

PR #4153, opened 2025-11-18
sql: Fix Seg Faults when parsing complex queries

kchledowski:
### Problem

SQL Parser is not able to parse some of the queries, mainly more complex ones. Apparently it runs out of stack memory. 

Given a troublesome SQL query:
- With 1 MB stack size (default on Linux x64) the parsing fails with 139 error code.
- With 2 MB stack size (default on macOS ARM64) the parsing is successful.

Closes: #3973

### Solution

OpenLineage is currently using sqlparser (https://docs.rs/sqlparser) v0.50.0 from a custom fork. 
Testing showed that starting with v0.52.0, the issue is gone.
The solution is to migrate to the latest version of sqlparser, which is v0.59.0

#### One-line summary:

sql: Fix Seg Faults when parsing complex queries

### Checklist

----

### DIFF
--- a/integration/sql/impl/Cargo.toml
+++ b/integration/sql/impl/Cargo.toml
@@ -14,4 +14,4 @@ crate-type = ["rlib"]
 [dependencies]
 anyhow = {workspace = true}
 
-sqlparser = {git = "https://github.com/OpenLineage/sqlparser-rs/", branch = "main"}
+sqlparser = "=0.59.0"
--- a/integration/sql/impl/src/visitor.rs
+++ b/integration/sql/impl/src/visitor.rs
@@ -6,9 +6,11 @@ use crate::lineage::*;
 
 use anyhow::{anyhow, Result};
 use sqlparser::ast::{
-    AlterTableOperation, Expr, FromTable, Function, FunctionArg, FunctionArgExpr,
-    FunctionArguments, Ident, ObjectName, Query, Select, SelectItem, SetExpr, Statement, Table,
-    TableFactor, TableFunctionArgs, Use, Value, WindowSpec, WindowType, With,
+    AccessExpr, AlterTableOperation, CreateTableLikeKind, Expr, FromTable, Function, FunctionArg,
+    FunctionArgExpr, FunctionArguments, Ident, ObjectName, ObjectNamePart, Query,
+    RenameTableNameKind, Select, SelectItem, SetExpr, Statement, Subscript, Table, TableFactor,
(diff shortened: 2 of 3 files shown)


integration/sql/impl/Cargo.toml
```diff
@@ -14,4 +14,4 @@ crate-type = ["rlib"]
 [dependencies]
 anyhow = {workspace = true}
 
-sqlparser = {git = "https://github.com/OpenLineage/sqlparser-rs/", branch = "main"}
```
kacpermuda, line 17: What we've done before is just updating our fork's main and then use the latest feature. If you feel like we should stop using our fork and use sqlparser directly, we can also do that but we should probably discuss it on slack first. I'm not against, but it's probably been done for a reason 😄

kchledowski: We've already discussed it to some degree with @mobuchowski ,
there are some API changes in sqlparser that had to be addressed in OL, so there are two options to make this upgrade smooth:
- we push a new version of sqlparser to the fork, but to a separate branch, not main and we use that
- we switch back to the upstream library

I believe the fork was needed for some custom additions that were not there in the upstream, but that is no longer required, so I feel it would be better to switch back to upstream.
If needed in the future we can always use a fork again

kacpermuda: Makes sense, I'll post a message on slack just to that others are aware.

mobuchowski: Yeah @kacpermuda afaik the features we want in the library are in the sqlparser now.

JDarDagran: https://github.com/apache/datafusion-sqlparser-rs/compare/main...OpenLineage:sqlparser-rs:main
doesn't show any significant changes, only readme file.

JDarDagran:
Would it be possible to add some more complex example in tests that fails now and is fixed with this PR?

kchledowski:
> Would it be possible to add some more complex example in tests that fails now and is fixed with this PR?

To make such test reliable, it would have to be forced to run with 1 MB stack size somehow. I guess it should be possible to spawn a new thread with 1 MB stack size and run a test in it, I'll look into it

mobuchowski:
IDK if it's worth spending time doing that. Bump of dependency is good enough reason for me to merge that :)

JDarDagran:
That's totally fine. I was rather asking for it hoping it's a small task to do.

Issue #3160, opened 2024-10-16
[INTEGRATION] Support Apache Gravitino data catalog pluggability 


lisancao:
### What is the target system?

Apache Gravitino 

### What kind of integration is this?

- [ ] Produces OpenLineage metadata
- [X] Consumes OpenLineage metadata
- [ ] Something else

### How should this integration be implemented?

Apache Gravitino- like Polaris or Unity Catalog OS, is an open source data catalog with arguably a larger scope to support end to end view. I think it makes sense for an open source data catalog/metadata lake to be able to consume data directly from OpenLineage and sync the metadata across its instances, and then use it as part of the governance strategy. 

I am curious how the catalog story is evolving here and OpenLineage's strategy for integrating into them all, since it makes a lot of sense for the open source catalogs to leverage OpenLineage. 

### Where should this integration be implemented?

- [ ] In the target system
- [X] In the OpenLineage repo
- [ ] Somewhere else

### Do you plan to make this contribution yourself?

- [ ] I am interested in doing this work

harels:
Hey @lisancao , thanks for creating this issue!
I fully agree with you, and this totally makes sense.
I'm not familiar enough with Gravitino to provide an opinion yet, but would be happy to collaborate and figure out the best approach!

lisancao:
> Hey @lisancao , thanks for creating this issue! I fully agree with you, and this totally makes sense. I'm not familiar enough with Gravitino to provide an opinion yet, but would be happy to collaborate and figure out the best approach!

Very happy to hear it- and love that you are on board :) Would it be best to do a quick meet or design doc to figure out the best approaches and get both community members' sign offs?

harels:
If it works for you, maybe we can have a quick discussion and coordinate over slack in the OpenLineage account? Happy to do it another way if that doesn't work for you.
https://join.slack.com/t/openlineage/shared_invite/zt-253bh1nqo-LqEauJHjLXoD7t4isJjbyg

mobuchowski:
OSS data catalog support is definitely something that we'd love to have 🚀

lisancao:
Sounds good, I will join the slack and invite some folks for us to chat over :)

pawel-big-lebowski:
It's kind of not clear to me what is planned to achieve: the issue says it's a consumer integration and that the integration should be implemented in the OpenLineage repo. If you want to consume OL events, then no code changes on OL repo should be required.

lisancao:
[...]

PR #4179, opened 2025-12-01
Add Apache Gravitino catalog support in Spark module

shaofengshi:
related: #3160

Add Apache Gravitino catalog support for Spark integration with configurable identifier mapping and GVFS filesystem support

### Meaningful description
Problem: 

- OpenLineage lacked support for Apache Gravitino, a unified metadata management system that provides a single entry for multi-catalog data assets. Users working with Gravitino-managed catalogs (Iceberg, JDBC, Hive) couldn't capture proper lineage with Gravitino's namespace hierarchy.

Solution: Implemented Gravitino integration through:

- Catalog Handlers: Added GravitinoHandler, GravitinoIcebergHandler, and GravitinoJDBCHandler to detect and process Gravitino-managed catalogs
- Configuration Provider: Created GravitinoInfoProviderImpl with SparkGravitinoInfoProvider to read Gravitino settings from Spark configuration
- GVFS Support: Implemented GVFSFilesystemDatasetExtractor and GVFSUtils for Gravitino Virtual File System paths
- Identifier Mapping: Added GravitinoUtils to convert Spark identifiers to Gravitino's catalog.schema.table format
- Custom Facets: Created GravitinoFacets.LocationDatasetFacet to capture fileset location metadata

Configuration options:

- spark.sql.gravitino.metalake: Metalake name for Gravitino connector
- spark.hadoop.fs.gravitino.client.metalake: Metalake name for filesystem operations
- spark.sql.gravitino.useGravitinoIdentifier: Toggle Gravitino identifier format (default: true)
- spark.sql.gravitino.catalogMappings: Map Spark catalog names to Gravitino catalogs (e.g., "catalog1:gravitino1,catalog2:gravitino2")

Alternatives Considered:

- External post-processing of lineage events: Rejected due to loss of runtime context and increased complexity
- Gravitino-specific OpenLineage fork: Rejected to maintain compatibility with upstream OpenLineage ecosystem

### DIFF
--- a/integration/spark/GRAVITINO_QUICKSTART.md
+++ b/integration/spark/GRAVITINO_QUICKSTART.md
@@ -2,6 +2,8 @@
 
 Get started with OpenLineage and Apache Gravitino in 5 minutes.
 
+> **Important:** To use Gravitino integration with OpenLineage, you **must** set `spark.openlineage.dataset.useGravitinoIdentifier=true`. Without this configuration, Gravitino identifiers will not be used in lineage events (default is `false`).
+
 ## Step 1: Add Dependencies
 
 Add the required JARs to your Spark application:
@@ -32,6 +34,9 @@ spark.sql.catalog.gravitino_catalog.gravitino.catalog=iceberg_catalog
 spark.extraListeners=io.openlineage.spark.agent.OpenLineageSparkListener
 spark.openlineage.transport.type=http
 spark.openlineage.transport.url=http://openlineage-backend:5000
+
+# Enable Gravitino dataset identifiers (REQUIRED for Gravitino integration)
+spark.openlineage.dataset.useGravitinoIdentifier=true
 ```
 
 ## Step 3: Run Your Spark Job
@@ -92,6 +97,7 @@ Expected output structure:
 
 ```properties
 spark.sql.gravitino.metalake=production
+spark.openlineage.dataset.useGravitinoIdentifier=true
 
 # Iceberg catalog
 spark.sql.catalog.iceberg_prod=org.apache.gravitino.spark.connector.catalog.GravitinoCatalog
@@ -106,6 +112,7 @@ spark.sql.catalog.postgres_prod.gravitino.catalog=postgres
 
 ```properties
 spark.sql.gravitino.metalake=production
+spark.openlineage.dataset.useGravitinoIdentifier=true
 spark.hadoop.fs.gravitino.client.metalake=production
 spark.hadoop.fs.gravitino.client.uri=http://gravitino-server:8090
 spark.hadoop.fs.gvfs.impl=org.apache.gravitino.filesystem.hadoop.GravitinoVirtualFileSystem
@@ -120,6 +127,7 @@ val df = spark.read.parquet("gvfs://fileset/catalog/schema/my_fileset/")
 
 ```properties
 spark.sql.gravitino.metalake=production
+spark.openlineage.dataset.useGravitinoIdentifier=true
 spark.sql.gravitino.catalogMappings=dev_iceberg:prod_iceberg,dev_postgres:prod_postgres
 ```
 
@@ -131,6 +139,7 @@ spark.sql.gravitino.catalogMappings=dev_iceberg:prod_iceberg,dev_postgres:prod_p
 1. OpenLineage listener is registered: `spark.extraListeners=io.openlineage.spark.agent.OpenLineageSparkListener`
 2. Transport URL is correct: `spark.openlineage.transport.url=http://openlineage-backend:5000`
 3. Gravitino metalake is configured: `spark.sql.gravitino.metalake=my_metalake`
+4. **Gravitino identifiers are enabled:** `spark.openlineage.dataset.useGravitinoIdentifier=true` (REQUIRED)
 
 ### Issue: "Gravitino metalake configuration not found"
 
--- a/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/SparkGravitinoInfoProviderTest.java
+++ b/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/SparkGravitinoInfoProviderTest.java
@@ -15,7 +15,11 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-public class SparkGravitinoInfoProviderTest {
+class SparkGravitinoInfoProviderTest {
+
+  private static final String LOCAL_MASTER = "local[*]";
+  private static final String TEST_APP_NAME = "test";
+  private static final String CATALOG3 = "catalog3";
 
   @BeforeAll
   @SneakyThrows
@@ -33,12 +37,12 @@ public class SparkGravitinoInfoProviderTest {
 
   @SneakyThrows
   @Test
-  public void testGetMetalakeName() {
+  void testGetMetalakeName() {
     GravitinoInfoProviderImpl provider = GravitinoInfoProviderImpl.newInstanceForTest();
 
     SparkSession.builder()
-        .master("local[*]")
-        .appName("test")
+        .master(LOCAL_MASTER)
+        .appName(TEST_APP_NAME)
         .config(SparkGravitinoInfoProvider.metalakeConfigKeyForConnector, "metalake_name")
         .getOrCreate();
     Assertions.assertEquals("metalake_name", provider.getMetalakeName());
@@ -46,8 +50,8 @@ public class SparkGravitinoInfoProviderTest {
     provider = GravitinoInfoProviderImpl.newInstanceForTest();
     cleanUpExistingSession();
     SparkSession.builder()
-        .master("local[*]")
-        .appName("test")
+        .master(LOCAL_MASTER)
+        .appName(TEST_APP_NAME)
         .config(SparkGravitinoInfoProvider.metalakeConfigKeyForFS, "metalake_name2")
         .getOrCreate();
     Assertions.assertEquals("metalake_name2", provider.getMetalakeName());
@@ -55,8 +59,8 @@ public class SparkGravitinoInfoProviderTest {
     provider = GravitinoInfoProviderImpl.newInstanceForTest();
     cleanUpExistingSession();
     SparkSession.builder()
-        .master("local[*]")
-        .appName("test")
+        .master(LOCAL_MASTER)
+        .appName(TEST_APP_NAME)
         .config(SparkGravitinoInfoProvider.metalakeConfigKeyForConnector, "metalake_name3")
         .config(SparkGravitinoInfoProvider.metalakeConfigKeyForFS, "metalake_name4")
         .getOrCreate();
@@ -64,18 +68,18 @@ public class SparkGravitinoInfoProviderTest {
 
     GravitinoInfoProviderImpl provider2 = GravitinoInfoProviderImpl.newInstanceForTest();
     cleanUpExistingSession();
-    SparkSession.builder().master("local[*]").appName("test").getOrCreate();
+    SparkSession.builder().master(LOCAL_MASTER).appName(TEST_APP_NAME).getOrCreate();
     Assertions.assertThrowsExactly(RuntimeException.class, () -> provider2.getMetalakeName());
   }
 
   @SneakyThrows
   @Test
-  public void testGetCatalogName() {
+  void testGetCatalogName() {
     GravitinoInfoProviderImpl provider = GravitinoInfoProviderImpl.newInstanceForTest();
 
     SparkSession.builder()
-        .master("local[*]")
-        .appName("test")
+        .master(LOCAL_MASTER)
+        .appName(TEST_APP_NAME)
         .config(
             SparkGravitinoInfoProvider.catalogMappingConfigKey,
             "catalog1:gravitino1,catalog2:gravitino2")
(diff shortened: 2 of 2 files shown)


client/java/src/main/java/io/openlineage/client/utils/gravitino/SparkGravitinoInfoProvider.java
```diff
@@ -0,0 +1,205 @@
...
+   */
+  @SneakyThrows
+  private String getSparkConfigValue(String configKey) {
+    log.trace("Reading Spark configuration key: {}", configKey);
+
+    Class<?> sparkSessionClass = Class.forName(SPARK_SESSION_CLASS_NAME);
+
+    // SparkSession s = SparkSession.active()
+    Method activeMethod = sparkSessionClass.getMethod("active");
+    Object sparkSessionInstance = activeMethod.invoke(null);
```
dolfinus, line 191: Why not passing `SparkSession`/`SparkConf` via method args?

client/java/src/main/java/io/openlineage/client/utils/jdbc/JdbcDatasetUtils.java
```diff
@@ -61,6 +63,10 @@ public static DatasetIdentifier getDatasetIdentifier(
     try {
       JdbcExtractor extractor = getExtractor(uri);
       JdbcLocation location = extractor.extract(uri, properties);
+      GravitinoInfoProviderImpl provider = GravitinoInfoProviderImpl.getInstance();
+      if (provider.useGravitinoIdentifier()) {
+        return GravitinoUtils.getGravitinoDatasetIdentifierFromJdbc(location.toName(parts));
```
dolfinus, line 68: Why this approach is used instead of `GravitinoExtractor implements JdbcExtractor`?

integration/spark/GRAVITINO_QUICKSTART.md
```diff
@@ -0,0 +1,166 @@
...
+spark.sql.catalog.gravitino_catalog.gravitino.metalake=my_metalake
+spark.sql.catalog.gravitino_catalog.gravitino.catalog=iceberg_catalog
+
+# OpenLineage Configuration
+spark.extraListeners=io.openlineage.spark.agent.OpenLineageSparkListener
+spark.openlineage.transport.type=http
+spark.openlineage.transport.url=http://openlineage-backend:5000
+
+# Enable Gravitino dataset identifiers (REQUIRED for Gravitino integration)
+spark.openlineage.dataset.useGravitinoIdentifier=true
```
dolfinus, line 39: Why this can't be automatic, e.g. by checking a type of relation, like we already do for Iceberg and others?

client/java/src/main/java/io/openlineage/client/utils/filesystem/gvfs/GVFSUtils.java
```diff
@@ -0,0 +1,129 @@
...
+  private static final String PATH_SEPARATOR = "/";
+
+  /** Minimum number of path components required: catalog, schema, fileset */
+  private static final int MIN_GVFS_PATH_PARTS = 3;
+
+  public static boolean isGVFS(URI uri) {
+    return GVFSFilesystemDatasetExtractor.SCHEME.equalsIgnoreCase(uri.getScheme());
+  }
+
+  public static void injectGVFSFacets(
```
dolfinus, line 33: Prefer using a DatasetSymlink instead, like [Iceberg](https://github.com/OpenLineage/OpenLineage/blob/65669d726d7ee44c439ed9efd6a6e6cceb74091c/integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandlerTest.java#L142-L178) already do

client/java/src/main/java/io/openlineage/client/utils/GravitinoUtils.java
```diff
@@ -0,0 +1,23 @@
...
+*/
+
+package io.openlineage.client.utils;
+
+import io.openlineage.client.utils.gravitino.GravitinoInfoProviderImpl;
+import java.util.Arrays;
+
+public class GravitinoUtils {
+
+  private static final String DEFAULT_CATALOG_NAME = "spark_catalog";
```
dolfinus, line 13: If this is the name of Spark config key `spark.sql.catalog.${catalog_name}`, then it just an arbitrary user-defined alias for Spark entity, not a name of Gravitino catalog. It should not be used as dataset name

client/java/src/main/java/io/openlineage/client/utils/GravitinoUtils.java
```diff
@@ -0,0 +1,23 @@
...
+
+  // For datasource v1 jdbc relation
+  public static DatasetIdentifier getGravitinoDatasetIdentifierFromJdbc(String jdbcName) {
+    GravitinoInfoProviderImpl provider = GravitinoInfoProviderImpl.getInstance();
+    String metalake = provider.getMetalakeName();
```
dolfinus, line 18: Let's use [Gravitino URL as a namespace](https://github.com/OpenLineage/OpenLineage/blob/65669d726d7ee44c439ed9efd6a6e6cceb74091c/integration/spark/spark3/src/test/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/IcebergHandlerTest.java#L142-L178), to make dataset identifier unique (different Gravitino instances can have the same metalakes/catalogs/datasets). Regarding metalake name, IMHO it should be used as a prefix for dataset name, just like a catalog name

client/java/src/main/java/io/openlineage/client/utils/filesystem/gvfs/GVFSUtils.java
shaofengshi: OK, I will check it

client/java/src/main/java/io/openlineage/client/utils/gravitino/SparkGravitinoInfoProvider.java
shaofengshi: Yeah, good question; this has several considerations:
1) This class is located in the client/java module, which is designed to be a lightweight client library without direct Spark dependencies. Adding SparkSession or SparkConf as method parameters would break the clean separation between client utilities and Spark-specific code, and potentially increasing the client library's footprint and complexity.
2) This class implements the GravitinoInfoProvider interface, which is designed to be parameter-less, allowing for different implementations (potentially non-Spark ones in the future). Changing the interface to accept Spark objects would break this abstraction.

Your suggestion could be implemented, but it would require major API redesign for GravitinoInfoProvider interface and its implementations, and change the module dependency (add Spark dependencies to the client module) which introduces a tighter coupling between client and Spark. 

So, current implementation is not perfect, but is a compromised solution based on several considerations.

dolfinus: > which is designed to be a lightweight client library without direct Spark dependencies

But there is a class called `SparkGravitinoInfoProvider` which cannot be used without Spark. Maybe interface implementation should be moved into integrations/spark, and SparkSession/SparkConf object will be passed as class constructor?

integration/spark/GRAVITINO_QUICKSTART.md
```diff
@@ -0,0 +1,166 @@
+# Gravitino Integration - Quick Start Guide
```
mobuchowski, line 1: Can we have this as a website doc - https://openlineage.io/docs/ - I believe the md file in subdirectory will be buried and never read by users.

Website code docs exist in website/ directory in the root of the repository.

shaofengshi:
Hi Maxim, after checking your comments, we think the current implemetation is something over-design, which no just cover the case 1) that the user uses Gravitino catalog in Spark, but also cover the case 2) that directly configure a Spark DSv2 catalog (jdbc, iceberg, etc) instead of using Gravitino.

The support for case 2) is unnecessary, as it needs user to configure "useGravitinoIdentifier=true" and a catalog mapping property, which introduces complexity and is error-prone.

After some internal discussion, we decide to refactor the current implementation:
- support case 1) is good enough now; will remove the code for 2);
- use the DatasetSymlink way as you suggested;
- keep the code clean and simple.

@dolfinus Maxim, do you prefer to a new PR, or update on this PR? Thanks for your feedback!

integration/spark/GRAVITINO_QUICKSTART.md
mobuchowski: I have the same question :)

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/GravitinoIcebergHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import io.openlineage.spark.api.OpenLineageContext;
+import io.openlineage.spark3.agent.lifecycle.plan.catalog.iceberg.IcebergHandler;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+
+@Slf4j
+public class GravitinoIcebergHandler extends IcebergHandler {
```
mobuchowski, line 19: I'd rather not use inheritance here. If this can be an instance of `BaseCatalogTypeHandler` - then great, otherwise it would be better if this was it's own `CatalogHandler`.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/GravitinoJDBCHandler.java
```diff
@@ -0,0 +1,53 @@
...
+import io.openlineage.spark.agent.util.GravitinoUtils;
+import io.openlineage.spark.api.OpenLineageContext;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.connector.catalog.Identifier;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+
+@Slf4j
+public class GravitinoJDBCHandler extends JdbcHandler {
```
mobuchowski, line 18: The same comment as for `GravitinoIcebergHandler`

integration/spark/GRAVITINO_QUICKSTART.md
```diff
@@ -0,0 +1,166 @@
...
+### Issue: Catalog not found
+
+**Check:**
+1. Gravitino server is running and accessible
+2. Catalog exists in Gravitino: `curl http://gravitino-server:8090/api/metalakes/my_metalake/catalogs`
+3. Catalog configuration matches Gravitino catalog name
+
+## Next Steps
+
+- Read the [full Gravitino integration guide](GRAVITINO.md)
```
mobuchowski, line 160: I don't think this file exists?

integration/spark/GRAVITINO_QUICKSTART.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+curl http://openlineage-backend:5000/api/v1/lineage
+```
+
+Expected output structure:
+```json
+{
+  "namespace": "my_metalake",
+  "name": "iceberg_catalog.db.users",
```
mobuchowski, line 81: Can those be symlinks, with original name/namespace pointing at the object storage location?
This would fit right in with other Iceberg/catalog implementations. 
Note, in some time we might change those as catalog names are usually more clear to users. If we do that, we'd rather be able to do this via single configuration that we'd be able to default, rather than rewrite all the code that has different assumptions all over the place.

For naming of those, can we use something like:
namespace: identify Gravitino instance with scheme uniquely identifying Gravitino datasets
name: unique identifier within the space?

client/java/src/main/java/io/openlineage/client/utils/gravitino/SparkGravitinoInfoProvider.java
```diff
@@ -0,0 +1,205 @@
...
+   */
+  public static final String catalogMappingConfigKey = "spark.sql.gravitino.catalogMappings";
+
+  private static final int CATALOG_MAPPING_PARTS = 2;
+
+  @Override
+  public boolean isAvailable() {
+    try {
+      SparkGravitinoInfoProvider.class.getClassLoader().loadClass(SPARK_SESSION_CLASS_NAME);
+      log.debug("SparkSession class found, Gravitino info provider is available");
```
mobuchowski, line 50: I don't understand this - why would `SparkSession` existence mean that Gravitino is available?

client/java/src/main/java/io/openlineage/client/utils/gravitino/GravitinoInfoProviderImpl.java
```diff
@@ -0,0 +1,155 @@
...
+*/
+package io.openlineage.client.utils.gravitino;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class GravitinoInfoProviderImpl {
```
mobuchowski, line 13: Why is `GravitinoInfoProviderImpl` not implementing `GravitinoInfoProvider`? Do we need this abstraction at all?

integration/spark/GRAVITINO_QUICKSTART.md
shaofengshi: OK，will move the doc there.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/GravitinoIcebergHandler.java
shaofengshi: I will remove the GravitinoIcebergHandler.java and GravitinoJDBCHandler.java, as it overwrites the default IcebergHandler and JDBCHandler 's behavior, which is not elegant. (That's also why we had the "useGravitinoIdentifier" configuration). Gravitino should only handle the data sources that configured via Gravitino catalog, we will optimize this PR.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/GravitinoJDBCHandler.java
shaofengshi: The same as above.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/GravitinoIcebergHandler.java
mobuchowski: Thanks!

integration/spark/GRAVITINO_QUICKSTART.md
shaofengshi: yeah, the reason is that we want to use our style Identifier ('catalog'.'schema'.'table') as the dataset name for Iceberg and JDBC tables. so this is the switch config. Now we will try to use symlink way, so this switch will be removed in the new PR I assume.

client/java/src/main/java/io/openlineage/client/utils/gravitino/SparkGravitinoInfoProvider.java
shaofengshi: You're correct, this can be optimized; not just check SparkSession, but also check the availability of Gravitino related configurations. Let me refine it.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -7,6 +7,8 @@
 
 import io.openlineage.client.utils.DatasetIdentifier;
 import io.openlineage.client.utils.filesystem.FilesystemDatasetUtils;
+import io.openlineage.client.utils.filesystem.gvfs.GVFSUtils;
```
pawel-big-lebowski, line 10: Imports to the gravitino classes are error-prone for users which don't have Gravitino on classpath. 
Whenever possible, gravitino classes should be moved to gravitino vendor submodule and separate from the rest of project.

PR #4182, opened 2025-12-02
Spec job parametrs facet

jakub-moravec:
Adding a facet to capture input parameters of jobs for each run. 

### Meaningful description

The JobRunParametersFacet is designed to capture the input parameters supplied to a job at the time of execution. By recording these key–value inputs in a structured format, downstream consumers such as lineage visualizers, observability tools, workflow UIs, and audit systems can surface exactly how a job was configured when it ran. This enables reproducibility, debugging, change tracking, and richer lineage context by showing not just what jobs ran and what data they touched, but also why they behaved the way they did based on the parameters they received.

### DIFF
--- a/website/static/spec/facets/1-0-0/ExecutionParametersRunFacet.json
+++ b/website/static/spec/facets/1-0-0/ExecutionParametersRunFacet.json
@@ -1,0 +1,59 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/ExecutionParametersRunFacet.json",
+  "$defs": {
+    "ExecutionParameter": {
+      "type": "object",
+      "properties": {
+        "key": {
+          "type": "string",
+          "description": "Unique identifier of the property.",
+          "example": "task.timeout"
+        },
+        "name": {
+          "type": "string",
+          "description": "Human-readable name of the property.",
+          "example": "Task timeout"
+        },
+        "description": {
+          "type": "string",
+          "description": "Human-readable description of the property.",
+          "example": "How long (in seconds) can individual task run without timing out."
+        },
+        "value": {
+          "type": "string",
+          "description": "Value of the property.",
+          "example": 3600
+        }
+      },
+      "required": ["key"],
+      "additionalProperties": false
+    },
+    "ExecutionParametersRunFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/RunFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "parameters": {
+              "description": "The parameters passed to the Job at runtime",
+              "type": "array",
+              "items": {
+                "$ref": "#/$defs/ExecutionParameter"
+              }
+            }
+          }
+        }
+      ],
+      "type": "object"
+    }
+  },
+  "type": "object",
+  "properties": {
+    "executionParameters": {
+      "$ref": "#/$defs/ExecutionParametersRunFacet"
+    }
+  }
+}


client/python/src/openlineage/client/generated/job_run_parameters.py
```diff
@@ -0,0 +1,49 @@
...
+    name: str | None = attr.field(default=None)
+    """
+    Human-readable name of the property.
+
+    Example: External trigger
+    """
+    description: str | None = attr.field(default=None)
+    """Detailed explanation of the property."""
+
+    value: str | float | bool | dict[str, Any] | list[Any] | None = attr.field(factory=list)
```
kacpermuda: Should we make value required? I think the python generator should not use `factory=list`, probably some edge case here.

client/python/src/openlineage/client/generated/job_run_parameters.py
```diff
@@ -0,0 +1,49 @@
...
+
+    @staticmethod
+    def _get_schema() -> str:
+        return (
+            "https://openlineage.io/spec/facets/1-0-0/JobRunParametersFacet.json#/$defs/JobRunParameterFields"
+        )
+
+
+@attr.define
+class JobRunParametersFacet(RunFacet):
```
kacpermuda: Not a big fan of the name, can we be more specific what differs this from a tag run facet? Examples that came to my mind: RuntimeParametersRunFacet, InputParametersRunFacet, ExecutionParametersRunFacet, RunConfigurationRunFacet, RuntimeConfigurationRunFacet

website/docs/spec/facets/run-facets/job_run_parameters.md
```diff
@@ -0,0 +1,37 @@
+---
+sidebar_position: 8
+---
+
+# Job Run Parameters Facet
+
+The facet contains the input parameters and configurations passed to the Job at runtime.
```
kacpermuda: Can we add a short sentence what differs this from TagsRunFacet, and when one should use one over the other ?

website/docs/spec/facets/run-facets/job_run_parameters.md
```diff
@@ -0,0 +1,37 @@
...
+        "facets": {
+            "jobRunParameters": { 
+                "_producer": "https://some.producer.com/version/1.0",
+                "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/JobRunParametersFacet.json",
+                "parameters": [
+                    {
+                        "key": "airflow.dagRun.external_trigger",
+                        "name": "External trigger", 
+                        "description": "Determines whether the DAG run was triggered externally.",
+                        "value": false
```
kacpermuda: nit: can we come up with different example? Something that is clearly an input param to a job, in Airflow the external_trigger is not an input, so it can confuse people 😄

client/python/src/openlineage/client/generated/job_run_parameters.py
```diff
@@ -0,0 +1,49 @@
...
+    parameters: list[JobRunParameterFields] | None = attr.field(factory=list)
+    """The parameters passed to the Job at runtime"""
```
kacpermuda: Maybe we should generalize it to store attributes of a run? Not sure if that makes sense, but we can have some input/config values, but also something we set during execution and is also saved as a custom attribute of a run. Just dropping an idea.

client/python/src/openlineage/client/generated/job_run_parameters.py
jakub-moravec: @kacpermuda I don't think value should be required. In generic, input parameters (in programming languages, parameter files, etc) can be nullable. If we force providing a value, producer would need to explicitely remove input parameters that don't have value

client/python/src/openlineage/client/generated/job_run_parameters.py
jakub-moravec: @kacpermuda `InputParametersRunFacet` sounds good to me. `ExecutionParametersRunFacet` too ... `RuntimeParametersRunFacet` might be represented as params provided during runtime, which might not be the case. Which one do you like the most?

website/docs/spec/facets/run-facets/job_run_parameters.md
jakub-moravec: @kacpermuda do you have a good example from Airflow? I could provide something generic, but it would be nice to have a real example

client/python/src/openlineage/client/generated/job_run_parameters.py
jakub-moravec: @kacpermuda yes, that makes sense to me. What exactly would you change? Removing the mention of "runtime"?

website/docs/spec/facets/run-facets/job_run_parameters.md
jakub-moravec: @kacpermuda I wonder if we should instead clarify in https://openlineage.io/docs/spec/facets/run-facets/tag-facet what is the tag facet for (I could do that within this PR, if we agree on the content). In general, I understand Tags as something I want to use to categorize or filter on. But I might not have enough context on how the existing producers are using the tag facet. Can you provide some insights here? (cc @mobuchowski)

client/python/src/openlineage/client/generated/job_run_parameters.py
kacpermuda: Makes sense, thanks ! I'm also wondering about a type here, should we maybe change this to `str | None` and allow only json strings? For more complex types, one can always put them as json str. WDYT? If not, then can you apply something like below to our python code generator? It will set default to None for the `value` field instead of list, which makes more sense. This will work if in the future if we'll also allow any JSON type for some field.
```
--- a/client/python/src/openlineage/client/generator/templates/dataclass.jinja2
+++ b/client/python/src/openlineage/client/generator/templates/dataclass.jinja2
@@ -79,8 +79,11 @@ class {{ class_name }}:
     {#- attrs requires field() for validators -#}
     {%- if 'True' == validators.validator_needed(field.type_hint) %} = attr.field() {%- endif -%}
     {%- if not (field.required or (field.represented_default == 'None' and field.strip_default_none)) -%}
+        {#- use None as default for a specific JSON type hint -#}
+        {%- if type_hint == 'str | float | bool | dict[str, Any] | list[Any] | None' -%}
+            {%- set default_value = 'default=None' -%}
         {#- use factories for list and dict types -#}
-        {%- if 'list' in type_hint -%}
+        {%- elif 'list' in type_hint -%}
             {%- set default_value = 'factory=list' -%}
         {%- elif 'dict' in type_hint -%}
```

client/python/src/openlineage/client/generated/job_run_parameters.py
kacpermuda: Maybe `ExecutionParametersRunFacet` ? I feel like this can hold both input and runtime attributes without a confusing name.

website/docs/spec/facets/run-facets/job_run_parameters.md
kacpermuda: Yeah, we can add explanation there, but maybe we can also cross link between those two, saying something like " if you're looking for tagging your run, see TagRun Facet" and here do something like " if you're interested in providing input parameters of a job, see ... ". How is that?

website/docs/spec/facets/run-facets/job_run_parameters.md
kacpermuda: Maybe something more trivial like task max_retries, or task timeout - this is usually run specific and passed as input param ? Or maybe something like `bash_command`, example from BashOperator's input args.

client/python/src/openlineage/client/generated/job_run_parameters.py
kacpermuda: Yes, together with a change of name that we discussed above, I think we can say here something like: 
```suggestion
    parameters: list[JobRunParameterFields] | None = attr.field(factory=list)
    """The parameters provided to the Job either as input or during runtime."""
```

client/python/src/openlineage/client/generated/job_run_parameters.py
jakub-moravec: Yeah, I think it might be simpler to just allow strings/nulls, it doesn't add too much complexity on the consumer if you wanted to type the values.

jakub-moravec:
@kacpermuda I incorporated your comments, can you have a look when you have time?

client/python/uv.lock
kacpermuda, line 1: Hmm, this file is in `.gitignore`, not sure why it's not ignored. Did you rebase ?

spec/facets/ExecutionParametersRunFacet.json
```diff
@@ -0,0 +1,59 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/JobRunParametersFacet.json",
+  "$defs": {
+    "ExecutionParameterFields": {
```
kacpermuda:
```suggestion
    "ExecutionParameter": {
```

It's describing a single parameter, so maybe this name is better?

spec/facets/ExecutionParametersRunFacet.json
```diff
@@ -0,0 +1,59 @@
...
+            }
+          }
+        }
+      ],
+      "type": "object"
+    }
+  },
+  "type": "object",
+  "properties": {
+    "jobRunParameters": {
```
kacpermuda:
```suggestion
    "executionParameters": {
```
Leftover from previous name?

spec/facets/ExecutionParametersRunFacet.json
```diff
@@ -0,0 +1,59 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/JobRunParametersFacet.json",
```
kacpermuda:
```suggestion
  "$id": "https://openlineage.io/spec/facets/1-0-0/ExecutionParametersRunFacet.json",
```

website/docs/spec/facets/run-facets/execution_parameters.md
```diff
@@ -0,0 +1,39 @@
...
+                ]
+            }
+        }
+    }
+    ...
+}
+```
+
+
+The facet specification can be found [here](https://openlineage.io/spec/facets/1-0-0/JobRunParametersFacet.json)
```
kacpermuda:
```suggestion
The facet specification can be found [here](https://openlineage.io/spec/facets/1-0-0/ExecutionParametersRunFacet.json)
```

website/docs/spec/facets/run-facets/execution_parameters.md
```diff
@@ -0,0 +1,39 @@
...
+Example:
+
+```json
+{
+    ...
+    "run": {
+        "facets": {
+            "executionParameters": { 
+                "_producer": "https://some.producer.com/version/1.0",
+                "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/JobRunParametersFacet.json",
```
kacpermuda:
```suggestion
                "_schemaURL": "https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/ExecutionParametersRunFacet.json",
```

website/static/ExecutionParametersRunFacet.json
kacpermuda, line 1: Same comments here as in the spec, some leftovers from previous naming

website/static/JobRunParametersFacet.json
kacpermuda, line 1: This should be deleted, it was generated from previous spec file with old name, there is no automated clearing of files.

client/python/uv.lock
jakub-moravec: Yes, I did, not sure why .gitignore is not applying, let me check

kacpermuda: Yeah, weird, there is a second one in client/python, but `git check-ignore -v client/python/uv.lock` shows it should be correctly ignored

jakub-moravec:
> Thanks @jakub-moravec , I spotted some leftovers from previous naming, after we clear them up we should be good to merge.

Should be better now @kacpermuda

kacpermuda:
Can you also check pre-commit? It shows some untracked file: 
>  Change detected in spec/facets/ExecutionParametersRunFacet.json: website/static/spec/facets/1-0-0/ExecutionParametersRunFacet.json is untracked

Maybe some leftover from previous scripts? Feel free to remove all website files and run prek locally again, it will re-generate the,

kacpermuda:
CI should pass now, accepted the Pr, will wait for another accept as it's a spec change and then we can merge. Thanks !

client/python/src/openlineage/client/generated/job_run_parameters.py
dolfinus: > input parameters (in programming languages, parameter files, etc) can be nullable

@jakub-moravec Having a nullable parameter (can be `null`) and having optional parameter (could be omitted) is not the same thing, actually. First can be `key=some, value=null`, but for the latter having `key=some` doesn't make sense for me - why not skipping the entire item?

dolfinus:
I'd like to continue this discussion before merging:
https://github.com/OpenLineage/OpenLineage/pull/4182#discussion_r2603275829

client/python/src/openlineage/client/generated/job_run_parameters.py
jakub-moravec: @dolfinus as we agreed with @kacpermuda to change the field to string, it is optional (not nullable) now. Does that work for you?

dolfinus: > it is optional (not nullable) now

It's optional (omittable) with default value `None`. Not passing a `value` is the same as passing `value=None`, and vice versa. My question what is the case of passing only `key=something` without any value as execution param

jakub-moravec: @dolfinus I tried to comment on that here https://github.com/OpenLineage/OpenLineage/pull/4182#discussion_r2606036496

Imagine a simple example with a parameter .properties file 

key1=valueA
key2= 

In order to be able to process this .properties file generically and and capture all parameters in the facet, you need to be able not to specify a value. 

Going further, specifying `null`/`None` as a parameter doesn't have to always be equivalent to omitting the parameter - it depends on the default behaviour of software that is being parametrized. The software can have a default behaviour that you are overriding by providing a `null` in the configuration. 

These might seem as edge cases, but my engineering backround is in Java/Spring and I've encountered these cases too many times in the past :)

dolfinus: So having a default `value=None` is just a Python client artifact, and is not intended to use for non-optional nullable values at all?

mobuchowski: Optional means nullable in Python. It's not Java, where there is three-value logic - Optional.empty, Optional.of and null.

dolfinus: Ok

client/python/src/openlineage/client/generated/execution_parameters_run.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Example: Task timeout
+    """
+    description: str | None = attr.field(default=None)
+    """
+    Human-readable description of the property.
+
+    Example: How long (in seconds) can individual task run without timing out.
+    """
+    value: str | None = attr.field(default=None)
```
dolfinus, line 31: nit: please consider using an `Any` for values. Storing complex values, like maps/arrays, in JSON serialized field inside actual JSON document is very inconvenient

jakub-moravec: I don't have a strong preference here. @kacpermuda was suggesting String (and I agree that will support big majority of cases even without the need to encode anything in serialized JSON. 
@mobuchowski do you have an opinion on this?

mobuchowski: In general think it's better to represent nested configuration as actual object.

I just would make sure we're not introducing ambiguity here: for example for nested configuration like Spark's
```
spark.sql.config.option=something
spark.sql.config.otheroption=somethingelse
```

you could represent it as those pair values, or nested json like
```
{
  "spark":  {
    "sql": {
      "config": {
         "option": "something",
         "otheroption": "somethingelse"
       }
    }
  }
}
```

jakub-moravec: @mobuchowski we can document the expected representation, but I don't think there is anything on the level of spec that could prevent it, unless you have some ideas

dolfinus: > you could represent it as those pair values, or nested json like

From Spark's perspective, configuration is a `Map<String, String>`, and nothing more. IMHO parameters should have exactly the same values as specific tool/engine uses, not something artificial

PR #4185, opened 2025-12-05
feat: Add debug mode to file transport

kacpermuda:
feat: Add debug mode to file transport, that adds more information to file paths

### Meaningful description
I use file transport a lot for debugging, and it would be nice to have a job name in the file path, and also prettified jsons already so it's easier to read. It's False by default, so no impact for current users. We can extend it to include even more details in the future if needed.

### DIFF
--- a/client/python/src/openlineage/client/transport/file.py
+++ b/client/python/src/openlineage/client/transport/file.py
@@ -47,6 +47,13 @@ class FileConfig(Config):
             - Factory function path (e.g., "mymodule.get_filesystem")
             - Instance attribute path (e.g., "mymodule.fs_instance")
         fs_kwargs (dict): Keyword arguments for constructing/calling the filesystem provider.
+        debug_mode (bool): Enable debug mode (default: False). Only effective when append=False. When enabled:
+            - Filenames include short timestamp (2-digit year), job/dataset name sanitized, and event type.
+              For RunEvent: uses job name and event type (e.g., START, COMPLETE).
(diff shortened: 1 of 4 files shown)


mobuchowski (approved):

client/python/src/openlineage/client/transport/file.py
```diff
@@ -195,23 +204,45 @@ def _create_file_handler(self) -> FileHandler:
...
         """Get the file path, adding timestamp if not in append mode."""
         if self.config.append:
             return self.config.log_file_path
-        else:
+        if not self.config.debug_mode:
             time_str = datetime.now().strftime("%Y%m%d-%H%M%S.%f")
             return f"{self.config.log_file_path}-{time_str}.json"
 
+        # Debug mode
+        time_str = datetime.now().strftime("%y%m%dT%H%M%S.%f")
```
dolfinus, line 216: Why having different timestamp formats for debug and non-debug? The difference is only in year component 2025 -> 20, and using "T" as date-time delimiter

kacpermuda: It's just shorter (year) and better for readability (T instead of `-`). I did not want to change the current implementation when not using debug mode in case somebody relies on it, so I only adjusted it for the debug mode.

mobuchowski: As long as we agree that "debug mode" is truly debug mode and nobody should depend on it in any production context, I don't think we need consistency here.

PR #4189, opened 2025-12-09
feat: add arbitrary extra info to JobDependency in JobDependenciesRunFacet

kacpermuda:
feat: Add arbitrary extra info to JobDependency in JobDependenciesRunFacet

### Meaningful description
In some cases, a job dependency may include supplementary details not covered by the OL job name/namespace or OL runId, such as identifiers from external systems. These can be stored in a general-purpose extra field, as the core information is already represented by the structured fields.

cc @jakub-moravec , lmk if you have a better idea for it, I know you're probably already using this facet


jakub-moravec:
@kacpermuda I don't have any input based on usage of this facet. 

Generally
 - I would expect the _important_ info to be ideally captured in the namespace/name of the referenced job
 - If you need to add more, isn't it sufficient to just allow `additionalProperties` on the referenced job? Or do you think having the `extra` (name could be better :D) object makes it more structured?

kacpermuda:
> If you need to add more, isn't it sufficient to just allow additionalProperties on the referenced job? Or do you think having the extra (name could be better :D) object makes it more structured?

That makes sense, just allowing additionalProperties on the JobDependency should be fine, I could then just add `airflow` map with airflow specific info I need.

kacpermuda:
@jakub-moravec So I've tried the above approach, but it's not ideal - in python impl, we'd have to create a JobDependency like this when using additional properties:
```
job1 = JobDependency(job=JobIdentifier(namespace="ns", name="name")).with_additional_properties(airflow={"a": 1, "b": 2})
```

So I think it's better to use an extra dict here, for easier creation:
```
job1 = JobDependency(job=JobIdentifier(namespace="ns", name="name"), extra={"airflow": {"a": 1, "b": 2}})
```

Let me know if you have a better naming idea to use instead of `extra`, I borrowed this one from Airflow naming that is used in Assets and other places.

dolfinus:
IMHO, the additionalProperties is designed exactly for this purpose, and there's no need to introduce another property for this. Otherwise why having additionalPropeties in the first place...

kacpermuda:
Yeah, I kinda agree. It's a limitation of our python implementation that causes "need" for this in spec, which is not an ideal cause. If we go with the additionalProperties approach, I'd also document this somewhere with an example of how to use it in Python. Let me think about it more and ping some more folks to hear what they have to say.

PR #4197, opened 2025-12-12
Python: fix .with_additional_properties() annotation

dolfinus:
Python: fix `.with_additional_properties()` annotation

### Meaningful description

Currently this code fails on mypy:
```bash
from openlineage.client.generated.base import BaseFacet

facet = BaseFacet().with_additional_properties(a="b")
```

```
123.py:3: error: Argument "a" to "with_additional_properties" of "BaseFacet" has incompatible type "str"; expected "dict[str, Any]"  [arg-type]
    facet = BaseFacet().with_additional_properties(a="b")
                                                     ^~~
Found 1 error in 1 file (checked 1 source file)
```

This is because `def func(**kwargs: dict[str, Any])` expect _every item_ of `kwargs` map to be a `dict[str, Any]`.

### DIFF
--- a/client/python/src/openlineage/client/generated/base.py
+++ b/client/python/src/openlineage/client/generated/base.py
@@ -86,7 +86,7 @@ class BaseFacet(RedactMixin):
     def skip_redact(self) -> list[str]:
         return self._base_skip_redact + self._additional_skip_redact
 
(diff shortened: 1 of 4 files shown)


dolfinus:
Failed `unit-test-integration-common` is unrelated

kacpermuda:
Failing SQL parser test is unrelated.

PR #4218, opened 2025-12-17
DataZone transport: Add cross-region support

RohithKayathi:
### Summary:

Add cross-region support for datazone transport

### Description:
The data transformation jobs (for eg: spark jobs) can be run in different region from that of DataZone domain. This change enables support for post lineage events to datazone domain in cross-region.

### Testing:
1. Added necessary unit tests
2. Tested cross region functionality by using a sample python script


website/docs/client/python.md
```diff
@@ -1454,6 +1454,7 @@ The `AmazonDataZoneTransport` requires `boto3` package to be additionally instal
 
 - `type` - string, must be `"amazon_datazone_api"`. Required.
 - `domainId` - string, specifies the DataZone / SageMaker Unified Studio domain id. The lineage events will be then sent to the following domain. Required.
+- `region` - string. When provided, the DataZone client will be configured to use this specific region. Optional, default: None (uses AWS SDK default region resolution).
```
kacpermuda: Can we also underline that this will not be used even when provided as long as `endpointOverride` is also provided? (in both python and java docs)

RohithKayathi: Done

PR #4252, opened 2026-01-14
spec: more details about data quality assertions

mobuchowski:
Severity levels and failures are common enough (dbt, soda, GE) that it can be standarized. 
Also, add explicit property field to capture additional interesting fields - like dbt ones. 

AI description:

This PR extends the DataQualityAssertionsDatasetFacet schema (v1.1.0) to capture additional metadata about data quality test assertions. Three new optional fields are added to the assertion object: severity (string) indicates the configured severity level of the test, typically "error" or "warn"; failures (integer) captures the actual count of rows that failed the assertion; and properties (object) provides an extension point for tool-specific assertion configuration.

These additions enable richer data quality lineage without breaking backwards compatibility. The severity and failures fields are generic concepts applicable across data quality tools, while the properties object allows integrations like dbt to include tool-specific configuration such as threshold expressions (warnIf, errorIf) and failure calculation methods (failCalc). All new fields are optional, so existing producers and consumers remain unaffected.


spec/facets/DataQualityAssertionsDatasetFacet.json
```diff
@@ -29,6 +29,20 @@
...
+                  },
+                  "severity": {
+                    "type": "string",
+                    "description": "The configured severity level of the assertion. Common values are 'error' (test failure blocks pipeline) or 'warn' (test failure produces warning only).",
+                    "example": "error"
+                  },
+                  "failures": {
+                    "type": "integer",
+                    "description": "The actual number of rows that failed the assertion test."
+                  },
```
kacpermuda, line 41: Maybe instead of having just failures count we should also track warnings count (since we allow warning as severity). So maybe a more generic count property ? OR at least failures_count and warnings_count?

spec/facets/DataQualityAssertionsDatasetFacet.json
```diff
@@ -29,6 +29,20 @@
                     "type": "string",
                     "description": "Column that expectation is testing. It should match the name provided in SchemaDatasetFacet. If column field is empty, then expectation refers to whole dataset.",
                     "example": "id"
+                  },
+                  "severity": {
+                    "type": "string",
+                    "description": "The configured severity level of the assertion. Common values are 'error' (test failure blocks pipeline) or 'warn' (test failure produces warning only).",
+                    "example": "error"
+                  },
+                  "failures": {
```
jakub-moravec: Would it also be useful to have a capture which rows failed (using indexes/rowids?) the assertion test? Not sure if all tools can provide this info, but some can

spec/facets/DataQualityAssertionsDatasetFacet.json
mobuchowski: The warning/error count is not separate - those tests basically execute the SQL and evaluate the return values vs warning/error threshold. I think `failures` is a good name - because it describes how many rows "failed" the predicate that is the given test.

kacpermuda (approved):

kacpermuda:
Let's not merge it yet, I'm also working on some DQ related facet changes, we might have to adjust/expand this information. Will come back with some more feedback shortly

PR #4263, opened 2026-01-20
spark: Add support for DataSourceRDD with Iceberg

kchledowski:
Add support for DataSourceRDD with Iceberg

### Meaningful description
When reading from Iceberg data source and mixing RDD actions with Dataframe actions (example below, but this is also what can happen when working with DynamicFrames on AWS Glue), OL is not able to extract data paths and build input datasets. This PR adds support for extracting paths from DataSourceRDD, which makes it possible.

```java
spark.sql("CREATE TABLE append_source USING iceberg AS SELECT a FROM temp");
spark.sql("CREATE TABLE append_table (a long) USING iceberg");
JavaRDD<Row> inputRDD = spark.read().table("append_source").toJavaRDD();
JavaRDD<Row> filteredRDD = inputRDD.filter(row -> row.getLong(row.fieldIndex("a")) > 1);

StructType schema =
    new StructType(
        new StructField[] {new StructField("a", LongType$.MODULE$, false, Metadata.empty())});

Dataset<Row> processedDataFrame = spark.createDataFrame(filteredRDD, schema);
Dataset<Row> finalDataFrame = processedDataFrame.select("a");
finalDataFrame.writeTo("append_table").append();
```

### DIFF
--- a/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkGenericIntegrationTest.java
+++ b/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkGenericIntegrationTest.java
@@ -56,7 +56,7 @@ class SparkGenericIntegrationTest {
   @SuppressWarnings("PMD")
   private static final String LOCAL_IP = "127.0.0.1";
 
-  private static final int MOCK_SERVER_PORT = 1083;
+  private static final int MOCK_SERVER_PORT = 1088;
   private static SparkSession spark;
   private static ClientAndServer mockServer;
 
--- a/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkIcebergIntegrationTest.java
+++ b/integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkIcebergIntegrationTest.java
@@ -62,7 +62,7 @@ import org.mockserver.model.HttpRequest;
 @Tag("iceberg")
 @Slf4j
 class SparkIcebergIntegrationTest {
-  private static final int MOCK_SERVER_PORT = 1084;
+  private static final int MOCK_SERVER_PORT = 1089;
 
   @SuppressWarnings("PMD")
   private static final String LOCAL_IP = "127.0.0.1";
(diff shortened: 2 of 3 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/RddPathUtils.java
```diff
@@ -206,6 +217,101 @@ public Stream<Path> extract(ParallelCollectionRDD rdd) {
...
+    public boolean isDefinedAt(Object rdd) {
+      return rdd instanceof DataSourceRDD;
+    }
+
+    @Override
+    public Stream<Path> extract(DataSourceRDD rdd) {
+      return extractInputPartitions(rdd).stream()
+          .flatMap(
+              ip ->
+                  inputPartitionPathExtractorFactory.createInputPartitionPathExtractors().stream()
```
mobuchowski: Maybe we can do this once instead of repeating building the extractors for each input partition.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/RddPathUtils.java
```diff
@@ -206,6 +217,101 @@ public Stream<Path> extract(ParallelCollectionRDD rdd) {
...
+        }
+      } catch (ClassNotFoundException
+          | InstantiationException
+          | IllegalAccessException
+          | InvocationTargetException
+          | NoSuchMethodException e) {
+        if (log.isDebugEnabled()) {
+          log.debug(
+              "{} is not on classpath: {}", ICEBERG_INPUT_PARTITION_PATH_EXTRACTOR, e.getMessage());
+        }
```
mobuchowski, line 311: Can we also catch NoClassDefFoundError?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/RddPathUtils.java
kchledowski: Good point, I'll change that

kchledowski:
@mobuchowski  I've addressed your suggestions (https://github.com/OpenLineage/OpenLineage/pull/4263/commits/75698fad9629d97998abfd75d03f595dcda91c76), I've also noticed some flakiness in tests when I run them locally (even though they were working fine in CI), so I took the chance to resolve that in https://github.com/OpenLineage/OpenLineage/pull/4263/changes/096d7b0ae34c9b7355b846c41384bcf90cb23608 and https://github.com/OpenLineage/OpenLineage/pull/4263/changes/e4896b929dbe3f1ad9ded60aaa12306b3e35553d

mobuchowski (approved):
Thanks @kchledowski

Issue #4183, opened 2025-12-03
[BUG] [SPARK] misleading spark job statuses and event statuses


tnazarew:
### OpenLineage integration/client

Spark integration

### OpenLineage version

all spark versions, all OL versions 

### Problem details

The spark application always ends with `SparkListenerApplicationEnd` that only has a `time` field. That means no matter what happens inside the spark application we will get eventType `COMPLETE` from OL events on application level. 

That means we can only get the information about failures from job level events but here it gets even worse.

If the failure happens not while transforming the data but e.g. when we connect to source database, we will not receive any `START` or `COMPLETE` events for this job.  Below are some examples when it can be a problem

### Only application events
lets assume that the spark app reads from Postgres and writes to some file. 
#### Success
If everything is ok we will get:
- 1 Application `START` event
- 1 Job `START` event
- 1 Job `COMPLETE` event
- 1 Application `COMPLETE` event

#### Job level Failure
If failure happens during data transformations we will get
- 1 Application `START` event
- 1 Job `START` event
- 1 Job `FAILURE` event
- 1 Application `COMPLETE` event

Which is quite bad because it means consumer has to deduce from job FAILURE + app COMPLETE that the app failed. 

#### App level Failure
Even worse, if there is failure while connecting to Postgres (e.g. wrong credentials) we will get 
- 1 Application `START` event
- 1 Application `COMPLETE` event

Which is identical to java or python script that is run by spark-submit but has no actions and is successful (in theory, I doubt that anyone actually does that). 

### Not enough events
Now the spark app has 3 jobs, 
#### Success
if everything works we have 
- 1 Application `START` event
- 3 Job `START` events
- 3 Job `COMPLETE` events
- 1 Application `COMPLETE` event

#### App level Failure
but if we have app level failure in the second job we will get
- 1 Application `START` event
- 1 Job `START` event
- 1 Job `COMPLETE` event
- 1 Application `COMPLETE` event

Which looks like successful spark application with only one job.

### What you think should happen instead

in case of any failure that fails the application, we should emit application FAILURE event so there is no ambiguity regarding the status of the application. We would still miss the job events of jobs that didn't start but it would be much better then current situation. 

### How to reproduce

[...]

PR #4265, opened 2026-01-21
emit application level FAIL event instead of COMPLETE when at least o…

tnazarew:
…ne job from the application has FAIL status

related: #4183 

Spark: Emit `FAIL` application event if any jobs inside it failed

### Meaningful description

Current behaviour is: application end event always has a `COMPLETE` type. 
Implemented functionality to collect the statuses of spark jobs and in case of any failure, emit application end `FAIL` event.

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
@@ -5,6 +5,8 @@
 
 package io.openlineage.spark.agent;
 
+import static io.openlineage.client.OpenLineage.RunEvent.EventType.COMPLETE;
+import static io.openlineage.client.OpenLineage.RunEvent.EventType.FAIL;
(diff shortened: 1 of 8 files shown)


dolfinus:
Well, this actually depends on a code which starts and stops Spark session. For example in this case:
```python
spark = SparkSession.builder.getOrCreate()

with spark:
  df = spark.createDataFrame(...)
  df.saveAsTable("my.table")  # raises PermissionDeniedError, so command event is emitted with FAIL 
  df.saveAsTable("another.table")

# should emit FAIL for application event
```

```python
spark = SparkSession.builder.getOrCreate()

with spark:
  df = spark.createDataFrame(...)
  try:
    df.saveAsTable("my.table")  # raises PermissionDeniedError, so command event is emitted with FAIL
  except Exception:
    print("The show must go on!")

  df.saveAsTable("another.table")  # executes successfully

# should emit COMPLETE for application event
```

I'm not sure that there is any way to distinct between case 1 and case2. Maybe if last sql/rdd event was failed?

PR #4275, opened 2026-01-25
Website: restyle animated wall

merobi-hub:
Restyled wall background and text for improved contrast.

### Meaningful description
The animated wall lacks sufficient contrast for easy legibility in light mode. This adds a dark bkgd to the wall to create a dark-style wall regardless of the mode, so that the higher-contrast styles are always applied.


merobi-hub:
Screenshot_20260124_210608
Screenshot_20260124_210537

PR #4282, opened 2026-01-27
Java: Send Content-Encoding header for compressed requests

Lukas-Riedel:
Send `Content-Encoding` header for compressed requests from the Java client

### Meaningful description
This PR ensures that the Java client sends the `Content-Encoding` header whenever request body compression is enabled, so the server can correctly decode the incoming payload. This behavior already exists in the Python client and is now consistently applied in the Java client.


jakub-moravec:
@mobuchowski could this be added in the next OL release - we need this to be able to support compression on the consumer side

mobuchowski:
@jakub-moravec sure.

PR #4283, opened 2026-01-27
spark: Extract input symlinks from DataSourceRDD

kchledowski:

Extract input symlinks from DataSourceRDD

### Meaningful description

https://github.com/OpenLineage/OpenLineage/pull/4263 added support for input dataset extraction based on DataSourceRDD, specifically for Iceberg.
This is an extension to that, providing ability to extract symlinks from DataSourceRDD (and potentially other RDDs).

The core of the change is that instead of getting only a `Path` from an RDD, we are now getting an `DatasetIdentifier`, which may or may not contain symlinks.

Additionally, metadata extraction from DataSourceRDD with Iceberg is now based on the table information instead of each separate file being read, which makes it possible to figure out the symlinks and also makes it more consistent with the behaviour of Iceberg handler for Dataframes

### DIFF
--- a/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/RddPathUtilsTest.java
+++ b/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/RddPathUtilsTest.java
@@ -44,6 +44,7 @@ import scala.collection.JavaConverters;
 import scala.collection.Seq;
 import scala.collection.mutable.ArrayBuffer;
 
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
 class RddPathUtilsTest {
 
   private static final String DATA_FIELD_NAME = "data";


integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java
```diff
@@ -82,7 +82,7 @@ class RddExecutionContext implements ExecutionContext {
   private final OpenLineageRunEventBuilder runEventBuilder;
   private final UUID runId = UUIDUtils.generateNewUUID();
   private final OpenLineageContext olContext;
-  private List<URI> inputs = Collections.emptyList();
+  private List<DatasetIdentifier> inputs = Collections.emptyList();
   private List<URI> outputs = Collections.emptyList();
```
mobuchowski, line 86: Should we do the same with outputs?

kchledowski: We could for just the sake of being consistent, because I don't think it would give us any other benefits.

If I'll have some spare time, I can prepare a PR with that

Issue #4253, opened 2026-01-14
[BUG] DBT Fusion support in `DbtLocalArtifactProcessor()`


ah12068:
### OpenLineage integration/client

DBT integration

### OpenLineage version

openlineage-python=1.14.0

### Technology and package versions

See: https://github.com/astronomer/astronomer-cosmos/issues/2263

### Environment configuration

See: https://github.com/astronomer/astronomer-cosmos/issues/2263

### Deployment details

See: https://github.com/astronomer/astronomer-cosmos/issues/2263

### Problem details

When using DBT Fusion the function: `DbtLocalArtifactProcessor()` returns a `KeyError: 'profiles_dir'` because the file `run_results.json` is completely different to the contents of a regular DBT run.

Example:
In DBT Fusion:
```
{
    "metadata": {
        "dbt_schema_version": "https://schemas.getdbt.com/dbt/run-results/v6.json",
        "dbt_version": "2.0.0-preview.92",
        "generated_at": "YYYY-MM-DDT11:08:54.601858Z",
        "invocation_id": "00000000-0000-0000-0000-000000000000",
        "env": {}
    },
    "results": [],
    "args": {
        "command": "run",
        "which": "run",
        "__other__": {
            "command": "run",
            "which": "run"
        }
    }
```
In DBT:
```
{
  "metadata": {
    "dbt_schema_version": "https://schemas.getdbt.com/dbt/run-results/v6.json",
    "dbt_version": "1.10.15",
    "generated_at": "YYYY-MM-DDT11:41:37.499928Z",
    "invocation_id": "00000000-0000-0000-0000-000000000000",
    "invocation_started_at": "YYYY-MM-DDT11:41:28.260832Z",
    "env": {}
  },
  "results": [],
  "elapsed_time": 6.718767166137695,
  "args": {
    "profiles_dir": "path_to_profiles",
    ...
  }
```

### What you think should happen instead

Why i think behaviour is erroneous: 

The output should be a list of events - but an error is returned. In downstream packages like [cosmos](https://github.com/astronomer/astronomer-cosmos) they use this to generate Airflow Assets which enable asset-aware scheduling.

What i think is the expected outcome:

The output should be a list of events, but if the `run_results.json` do not have the necessary properties then we'd need to either skip/source this same information elsewhere

### How to reproduce

[...]

PR #4302, opened 2026-01-30
dbt: Better error handling for finding the directory of `profiles.yml`

ah12068:

closes: https://github.com/OpenLineage/OpenLineage/issues/4253

Implements edge case handling for where `run_results["profiles_dir"]` does not exist, that is key doesn't exist, and proceeds to find `profiles.yml` in a default fashion [outlined by dbt](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles#advanced-customizing-a-profile-directory) and using the [existing method in the codebase](https://github.com/OpenLineage/OpenLineage/blob/837ad5a3dc395e9e40e1cd111d799328e085fba4/integration/common/src/openlineage/common/provider/dbt/local.py#L282)

### Meaningful description

Description of the problem: #4253 

Solution: 
Thinking about #4253, a simpler change would be to handle the error in the event that `run_results["profiles_dir"]` does not exist. Therefore, defaulting to the existing code as a means to handle this situation. 

Note: Rather than focusing on DBT Fusion, the idea is that the code handles errors a bit better which consequently supports current DBT Fusion progress. Though the only options that support both DBT Core and Fusion are `--profiles-dir`, `~/.dbt/` directory.

Alternatives considered:
1. use a `try` but if that fails enforce `project_dir` to be some string that enforces a `FileNotFound` error which proceeds to use the backup methods to find `profiles.yml` not in `run_results.json`.

### DIFF
--- a/integration/flink/shared/src/main/java/io/openlineage/flink/utils/ClassUtils.java
+++ b/integration/flink/shared/src/main/java/io/openlineage/flink/utils/ClassUtils.java
@@ -49,11 +49,16 @@ public class ClassUtils {
     return false;
   }
 
+  /**
+   * Checks if Flink 2.x-specific classes are present on the classpath. Uses
+   * JobStatusChangedListenerFactory as it is a Flink 2.x-only class that is not backported through
+   * connectors (unlike LineageGraph which can appear in Flink 1.x with modern connectors).
+   */
   public static boolean hasFlink2Classes() {
     try {
       ClassUtils.class
           .getClassLoader()
-          .loadClass("org.apache.flink.streaming.api.lineage.LineageGraph");
+          .loadClass("org.apache.flink.core.execution.JobStatusChangedListenerFactory");
       return true;
     } catch (Exception e) {
       // swallow- we don't care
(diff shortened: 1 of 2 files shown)


integration/common/src/openlineage/common/provider/dbt/local.py
```diff
@@ -263,11 +263,10 @@ def get_dbt_metadata(
         except FileNotFoundError:
             catalog = None
 
-        profile_dir = run_result["args"]["profiles_dir"]
-
         try:
+            profile_dir = run_result["args"]["profiles_dir"]
             profile = self.load_yaml_with_jinja(os.path.join(profile_dir, "profiles.yml"))[self.profile_name]
-        except FileNotFoundError:
+        except KeyError or FileNotFoundError:
```
kacpermuda: This is not working as you'd expect, It should be a tuple of error classes to work.
```suggestion
        except (KeyError, FileNotFoundError):
```

kacpermuda: Can we also add a debug log here, something like "profiles dir" could not be extracted from run result, falling back to default behavior" ?

kacpermuda: This should also fix failing tests, then please fix the DCO so we can merge this, click the failing DCO test for the guide on how to fix it.

ah12068: Thank you for the feedback - highly appreciate it and your patience as i'm fairly new to the open source scene 🙏 

I'll aim to get this done sometime this week.

kacpermuda: No problem, thanks for the contribution ! Looking forward to reviewing the PR again after changes.

ah12068:
Closing this to make a clean PR and avoid git force pushes to solve DCO error

PR #4313, opened 2026-02-02
clint: Adding support for JWT authentication

jakub-moravec:
Adding the JWT authenticator support for Java and Python clients

### Meaningful description
Adding JWT authenticator so that customers do not need to implement a custom authenticator when OL client is configured to send requests to a JWT authenticated consumer API

### DIFF
--- a/client/java/src/test/java/io/openlineage/client/transports/JwtTokenProviderTest.java
+++ b/client/java/src/test/java/io/openlineage/client/transports/JwtTokenProviderTest.java
@@ -65,7 +65,7 @@ class JwtTokenProviderTest {
   }
 
   /** Helper method to clear environment variables using reflection. */
-  @SuppressWarnings("unchecked")
+  @SuppressWarnings({"unchecked", "PMD"})
   private void clearEnvironmentVariables(Set<String> keys) throws Exception {
     Class<?> classOfMap = System.getenv().getClass();
     Field field = classOfMap.getDeclaredField("m");


client/python/src/openlineage/client/transport/http.py
```diff
@@ -53,6 +57,197 @@ def get_bearer(self) -> str | None:
...
+                self._token_expiry = self._extract_expiry_from_jwt(token)
+
+            log.debug("Successfully fetched JWT token, expires at: %s", self._token_expiry)
+
+        except (requests.RequestException, ValueError) as e:
+            msg = f"Failed to fetch JWT token from {self.token_endpoint}: {e}"
+            raise RuntimeError(msg) from e
+        except Exception as e:
+            msg = f"Failed to fetch JWT token from {self.token_endpoint}: {e}"
+            raise RuntimeError(msg) from e
```
mobuchowski: We can merge those two cases if we're handling them the same way

client/java/src/test/java/io/openlineage/client/transports/JwtTokenProviderTest.java
```diff
@@ -0,0 +1,466 @@
...
+            invocation -> {
+              HttpClientResponseHandler handler = invocation.getArgument(1);
+              return handler.handleResponse(mockHttpResponse);
+            });
+
+    // First call
+    provider.getToken();
+
+    // Wait for token to expire (plus buffer time)
+    Thread.sleep(2000);
```
mobuchowski: Maybe we can mock the Instant.now() to not add 2 seconds of test runtime every time.

website/docs/client/python/configuration.md
```diff
@@ -321,6 +329,143 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+- `tokenFields` - list of strings, JSON field names to search for the token in the response. The provider tries each field in order. Optional, default: `["token", "access_token"]`.
+- `expiresInField` - string, JSON field name containing the token expiration time in seconds. Optional, default: `"expires_in"`.
+- `grantType` - string, OAuth grant type parameter sent in the token request. Optional, default: `"urn:ietf:params:oauth:grant-type:jwt-bearer"`.
+- `responseType` - string, OAuth response type parameter sent in the token request. Optional, default: `"token"`.
+
+##### Behavior
+
+- The provider sends a POST request with URL-encoded form data containing the API key and OAuth parameters.
+- The response is expected to be JSON containing the JWT token and optionally an expiration time.
+- Tokens are cached and automatically refreshed 60 seconds before expiration.
```
mobuchowski: Can we make that configurable? Also, 60s by default sounds low - would it make sense to make 5-15 minutes a default?

client/python/tests/test_jwt_token_provider.py
```diff
@@ -0,0 +1,407 @@
...
+        mock_post.return_value = mock_response
+
+        provider = JwtTokenProvider({"apiKey": "test-key", "tokenEndpoint": "https://auth.example.com/token"})
+
+        # First call
+        bearer1 = provider.get_bearer()
+        assert bearer1 == "Bearer jwt-token-1"
+
+        # Wait for token to expire (plus buffer)
+        time.sleep(2)
```
kacpermuda: Same here as Maciej mentioned already, let's mock time not to wait 2s for test to finish.

website/docs/client/python/configuration.md
```diff
@@ -321,6 +329,143 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+<TabItem value="env-vars" label="Environment Variables">
+
+Standard OAuth configuration:
+
+```sh
+OPENLINEAGE__TRANSPORT__TYPE=http
+OPENLINEAGE__TRANSPORT__URL=https://backend:5000
+OPENLINEAGE__TRANSPORT__AUTH__TYPE=jwt
+OPENLINEAGE__TRANSPORT__AUTH__APIKEY=your-api-key
+OPENLINEAGE__TRANSPORT__AUTH__TOKENENDPOINT=https://auth.example.com/token
```
kacpermuda:
```suggestion
OPENLINEAGE__TRANSPORT__AUTH__TOKEN_ENDPOINT=https://auth.example.com/token
```
I think we need to provide an underscore everywhere, as env vars are read as lowercase. In client config we do not read "tokenendpoint", we only check "token_endpoint". This is to all examples and all variables, they will be read as lowercase and not camelCase, so it will not work. Please also add tests for that, since this example would not work

website/docs/client/python/configuration.md
```diff
@@ -321,6 +329,143 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+<Tabs groupId="integrations">
+<TabItem value="env-vars" label="Environment Variables">
+
+Standard OAuth configuration:
+
+```sh
+OPENLINEAGE__TRANSPORT__TYPE=http
+OPENLINEAGE__TRANSPORT__URL=https://backend:5000
+OPENLINEAGE__TRANSPORT__AUTH__TYPE=jwt
+OPENLINEAGE__TRANSPORT__AUTH__APIKEY=your-api-key
```
kacpermuda: Same as below, even though `apikey` is accepted, let's use `api_key` in env var examples, it was accepted for longer time, the `apikey` support was added more recently specially for env vars.

website/docs/client/python/configuration.md
kacpermuda: +1 this must be configurable by the user.

website/docs/client/python/configuration.md
mobuchowski: Good catch :)

jakub-moravec:
@mobuchowski @kacpermuda thank you both very much for the review, I added a second commit addressing your comments

client/python/src/openlineage/client/transport/http.py
```diff
@@ -53,6 +57,221 @@ def get_bearer(self) -> str | None:
...
+        super().__init__(config)
+        self.api_key = config.get("apiKey") or config.get("apikey") or config.get("api_key")
+        if not self.api_key:
+            msg = "apiKey is required for JWT token provider."
+            raise KeyError(msg)
+
+        # Support multiple naming conventions for backwards compatibility
+        # Preferred: token_endpoint (from TOKEN_ENDPOINT env var)
+        # Also support: tokenEndpoint, tokenendpoint
+        token_endpoint = config.get("token_endpoint") or config.get("tokenEndpoint")
```
kacpermuda, line 106:
```suggestion
        token_endpoint = config.get("tokenEndpoint") or config.get("token_endpoint")
```

This is a bit misleading, it says it supports `tokenendpoint` when it doesn't check for it. I think no change was needed here, we can keep the same logic as everywhere else (camelCase first, then snake_case). My previous comment was only about examples in python docs, that would not work - docs needed adjusting, not code.

What we need to do (can be a spearate PR) is clearly specify that when configuring OL client with env variables, one must use snake case configs, as env vars keys are read as lowercase.

client/python/src/openlineage/client/transport/http.py
```diff
@@ -53,6 +57,221 @@ def get_bearer(self) -> str | None:
...
+        if token_refresh_buffer:
+            try:
+                self.token_refresh_buffer = int(token_refresh_buffer)
+            except ValueError:
+                log.warning(
+                    "Invalid tokenRefreshBuffer value %s, using default: %s instead",
+                    token_refresh_buffer,
+                    self.TOKEN_REFRESH_BUFFER_SECONDS,
+                )
+                self.token_refresh_buffer = self.TOKEN_REFRESH_BUFFER_SECONDS
```
kacpermuda: I think we should not silence this error, is user provided wrong config - let's raise.

website/docs/client/python/configuration.md
```diff
@@ -321,6 +329,144 @@ client = OpenLineageClient(transport=HttpTransport(http_config))
...
+When using JWT authentication with HTTP transport, configure the `auth` section as follows:
+
+- `type` - string, must be `"jwt"`. Required.
+- `apiKey` - string, the API key used to obtain the JWT token. Required.
+- `tokenEndpoint` - string, the URL endpoint for token generation. Required.
+- `tokenFields` - list of strings, JSON field names to search for the token in the response. The provider tries each field in order. Optional, default: `["token", "access_token"]`.
+- `expiresInField` - string, JSON field name containing the token expiration time in seconds. Optional, default: `"expires_in"`.
+- `grantType` - string, OAuth grant type parameter sent in the token request. Optional, default: `"urn:ietf:params:oauth:grant-type:jwt-bearer"`.
+- `responseType` - string, OAuth response type parameter sent in the token request. Optional, default: `"token"`.
+- `tokenRefreshBuffer` - integer, number of seconds before token expiry to trigger a refresh. Optional, default: `120`.
```
kacpermuda, line 341: This is missing from the above list of options, for http transport. Maybe we should not list them twice, but simply reference this paragraph from the http transport jwk auth type bullet? It's easy for those two to go out of sync, as in the above example.

website/docs/client/python/configuration.md
```diff
@@ -335,8 +481,16 @@ Async transport API is experimental, and can change over the next few releases.
 - `timeout` - float specifying timeout (in seconds) value used while connecting to server. Optional, default: `5`.
 - `verify` - boolean specifying whether the client should verify TLS certificates from the backend. Optional, default: `true`.
 - `auth` - dictionary specifying authentication options. Optional, by default no authorization is used. If set, requires the `type` property.
-  - `type` - string specifying the "api_key" or the fully qualified class name of your TokenProvider. Required if `auth` is provided.
-  - `apiKey` - string setting the Authentication HTTP header as the Bearer. Required if `type` is `api_key`.
+  - `type` - string specifying value for one of the out-of-the-box available authentication methods (`api_key` or `jwt`), or the fully qualified class name of your TokenProvider. Required if `auth` is provided.
+  - Configuration options for `api_key` authentication:
+    - `apiKey` - string setting the Authentication HTTP header as the Bearer. Required if `type` is `api_key`.
+  - Configuration options for `jwt` authentication:
```
kacpermuda: Same here, let's just reference the JWT token paragraph, so that all these jwt configs are in one place only.

kacpermuda: Probably the same case for java docs.

client/python/tests/test_jwt_token_provider.py
```diff
@@ -0,0 +1,503 @@
...
+        assert config.auth.response_type == "cloud_iam"
+
+    @patch.dict(
+        "os.environ",
+        {
+            "OPENLINEAGE__TRANSPORT__TYPE": "http",
+            "OPENLINEAGE__TRANSPORT__URL": "http://backend:5000",
+            "OPENLINEAGE__TRANSPORT__AUTH__TYPE": "jwt",
+            "OPENLINEAGE__TRANSPORT__AUTH__API_KEY": "test-env-api-key-with-underscores",
+            "OPENLINEAGE__TRANSPORT__AUTH__TOKEN_ENDPOINT": "https://auth.example.com/token/with/underscores",
```
kacpermuda, line 470: Thanks, this test is exactly what I had in mind. Can you also add the opposite test, make sure if `OPENLINEAGE__TRANSPORT__AUTH__TOKENENDPOINT` is set (notice lack of snake case in tokenendpoint) the code will crash as this is required config.

client/java/src/test/java/io/openlineage/client/transports/JwtTokenProviderTest.java
```diff
@@ -0,0 +1,603 @@
...
+        (Map<String, String>) field.get(System.getenv());
+    writeableEnvironmentVariables.putAll(newEnv);
+  }
+
+  /** Helper method to clear environment variables using reflection. */
+  @SuppressWarnings("unchecked")
+  private void clearEnvironmentVariables(Set<String> keys) throws Exception {
+    Class<?> classOfMap = System.getenv().getClass();
+    Field field = classOfMap.getDeclaredField("m");
+    field.setAccessible(true);
```
mobuchowski, line 72: Those setAccessible cause PMD failures, either disable those checks or try using something else for config.

```
> Task :pmdTest
/home/circleci/openlineage/client/java/src/test/java/io/openlineage/client/transports/JwtTokenProviderTest.java:61:     AvoidAccessibilityAlteration:   You should not modify visibility of constructors, methods or fields using setAccessible()
/home/circleci/openlineage/client/java/src/test/java/io/openlineage/client/transports/JwtTokenProviderTest.java:72:     AvoidAccessibilityAlteration:   You should not modify visibility of constructors, methods or fields using setAccessible()
```

client/python/src/openlineage/client/transport/http.py
jakub-moravec: sounds good, removing error handling

jakub-moravec:
Thanks @mobuchowski @kacpermuda, I added a new commit addressing your comments. Please let me know if you have some other feedback. If not, what's the best way to prepare the PR for merge - should I force-push all changes in 1 commit after your review?

mobuchowski:
@jakub-moravec looks to me that the feedback has been addressed, thanks! 
Can you fix the PMD issue that causes the tests to fail? https://app.circleci.com/jobs/github/OpenLineage/OpenLineage/509808

Then we can merge it.

jakub-moravec:
Apologies, I missed the second occurrence. It should be fixed now.

mobuchowski (approved):

Issue #3220, opened 2024-11-04
DatabricksEventFilter check on isDisabled blocked threads


mzheng-plaid:
**Details:**
We run Spark on EMR with OpenLineage and we saw logs of:
```
pool-37-thread-1" #106 prio=5 os_prio=0 tid=0x0000ffffa1073000 nid=0xc8 waiting for monitor entry [0x0000fff71f7d2000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.apache.spark.sql.execution.QueryExecution.optimizedPlan$lzycompute(QueryExecution.scala:166)
	- waiting to lock <0x0000fff7cd838f90> (a org.apache.spark.sql.execution.QueryExecution)
	at org.apache.spark.sql.execution.QueryExecution.optimizedPlan(QueryExecution.scala:166)
	at io.openlineage.spark.agent.filters.EventFilterUtils.lambda$getLogicalPlan$2(EventFilterUtils.java:44)
	at io.openlineage.spark.agent.filters.EventFilterUtils$$Lambda$3250/2059122199.apply(Unknown Source)
	at java.util.Optional.map(Optional.java:215)
	at io.openlineage.spark.agent.filters.EventFilterUtils.getLogicalPlan(EventFilterUtils.java:44)
	at io.openlineage.spark.agent.filters.DatabricksEventFilter.isSerializeFromObject(DatabricksEventFilter.java:67)
	at io.openlineage.spark.agent.filters.DatabricksEventFilter.isDisabled(DatabricksEventFilter.java:41)
	at io.openlineage.spark.agent.filters.EventFilterUtils.lambda$isDisabled$0(EventFilterUtils.java:35)
	at io.openlineage.spark.agent.filters.EventFilterUtils$$Lambda$3242/21183944.test(Unknown Source)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
	at java.util.Spliterators$ArraySpliterator.tryAdvance(Spliterators.java:958)
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:126)
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:499)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:486)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
[...]

PR #4315, opened 2026-02-02
fix(spark): Early return in DatabricksEventFilter for non-Databricks platforms (#3220)

mobuchowski:
AI description:
The DatabricksEventFilter performs expensive checks on every Spark event to filter out internal Databricks operations, but these checks are unnecessary when running outside the Databricks platform. This causes performance overhead for users running Spark on other platforms like EMR, Dataproc, or standalone clusters, as the filter still evaluates node names and query plans for events that will never match Databricks patterns.

This fix adds an early return in isDisabled() (by DatabricksUtils.isRunOnDatabricksPlatform) that quickly checks if the job is running on Databricks by first looking for the DATABRICKS_RUNTIME_VERSION environment variable (a fast operation), then falling back to checking SparkConf for the workspace URL. Non-Databricks environments now exit the filter immediately without performing any further checks, significantly reducing overhead for the majority of OpenLineage users.

### DIFF
--- a/integration/spark/shared/src/test/java/io/openlineage/spark/agent/filters/DatabricksEventFilterTest.java
+++ b/integration/spark/shared/src/test/java/io/openlineage/spark/agent/filters/DatabricksEventFilterTest.java
@@ -27,6 +27,8 @@ import org.junitpioneer.jupiter.SetEnvironmentVariable;
 
 class DatabricksEventFilterTest {
 
+  private static final String DATABRICKS_RUNTIME_13_3 = "13.3";
+
   SparkListenerEvent event = mock(SparkListenerEvent.class);
   OpenLineageContext context = mock(OpenLineageContext.class);
   DatabricksEventFilter filter = new DatabricksEventFilter(context);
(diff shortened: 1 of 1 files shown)


pawel-big-lebowski:
> The DatabricksEventFilter performs expensive checks on every Spark event to filter out internal Databricks operations, but these checks are unnecessary when running outside the Databricks platform. 

@mobuchowski not sure if this desc reflects what's in the code.

mobuchowski:
@pawel-big-lebowski 
`DatabricksEventFilter` is using that:
```
  @Override
    if (!DatabricksUtils.isRunOnDatabricksPlatform(context)
        || !context.getQueryExecution().isPresent()) {
      return false;
    }

    return isWriteIntoDeltaCommand() || isDisabledDatabricksPlan(event);
  }
```
which should be now faster with `DATABRICKS_RUNTIME_VERSION` check. Sorry if that was unclear.

pawel-big-lebowski (approved):

Issue #2544, opened 2024-03-28
[SPARK][DATABRICKS] DropTableVisitor throwing an exception


pawel-big-lebowski:
`DropTableVisitor` assumes a child of Spark logical plan node is `ResolvedTable`:
```
ResolvedTable resolvedTable = ((ResolvedTable) (dropTable).child());
```
I observed that on Databricks runtime `14.2` (spark 3.5), this assumption no longer holds and `ResolvedIdentifier` is returned instead.

PR #4316, opened 2026-02-02
fix(spark): Handle ResolvedIdentifier in DropTableVisitor (#2544)

mobuchowski:
AI description:

When users run DROP TABLE commands on Databricks Runtime 14.2+, OpenLineage throws a ClassCastException because DropTableVisitor assumes the child of a DropTable node is always a ResolvedTable. However, newer Spark versions (3.4+) and Databricks runtimes wrap table references in ResolvedIdentifier, causing the visitor to fail and preventing lineage extraction for drop operations.

This fix adds handling for ResolvedIdentifier alongside the existing ResolvedTable support. When ResolvedIdentifier is detected (via reflection to maintain backwards compatibility with older Spark versions), the visitor extracts the underlying ResolvedTable from the identifier before processing. This allows DROP TABLE lineage to work correctly across all supported Spark and Databricks versions.


pawel-big-lebowski:
@mobuchowski: for those type of issues, running full Spark tests is recommended to cover several Spark versions

mobuchowski:
@pawel-big-lebowski it's 🟢

pawel-big-lebowski (approved):

PR #4320, opened 2026-02-03
dbt: Better error handling for finding the directory of `profiles.yml`

ah12068:

closes: https://github.com/OpenLineage/OpenLineage/issues/4253

Implements edge case handling for where `run_results["profiles_dir"]` does not exist, that is key doesn't exist, and proceeds to find `profiles.yml` in a default fashion [outlined by dbt](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles#advanced-customizing-a-profile-directory) and using the [existing method in the codebase](https://github.com/OpenLineage/OpenLineage/blob/837ad5a3dc395e9e40e1cd111d799328e085fba4/integration/common/src/openlineage/common/provider/dbt/local.py#L282)

### Meaningful description

Description of the problem: #4253 

I've opened a new PR in order to avoid git issues and therefore pass DCO checks + actioning on your feedback.

Original PR is at #4302 with feedback

Solution: 
Thinking about #4253, a simpler change would be to handle the error in the event that `run_results["profiles_dir"]` does not exist. Therefore, defaulting to the existing code as a means to handle this situation. 

Note: Rather than focusing on DBT Fusion, the idea is that the code handles errors a bit better which consequently supports current DBT Fusion progress. Though the only options that support both DBT Core and Fusion are `--profiles-dir`, `~/.dbt/` directory.

Alternatives considered:
1. use a `try` but if that fails enforce `project_dir` to be some string that enforces a `FileNotFound` error which proceeds to use the backup methods to find `profiles.yml` not in `run_results.json`.

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/local.py
+++ b/integration/common/src/openlineage/common/provider/dbt/local.py
@@ -268,7 +268,7 @@ class DbtLocalArtifactProcessor(DbtArtifactProcessor):
             profile = self.load_yaml_with_jinja(os.path.join(profile_dir, "profiles.yml"))[self.profile_name]
         except (KeyError, FileNotFoundError):
             self.logger.warning(
-                "Could not load profile directory from run_results.json. " 
+                "Could not load profile directory from run_results.json. "
                 "Falling back to default profiles.yml location."
             )
             profile = self.load_yaml_with_jinja(


ah12068:
Apologies for being messy @kacpermuda with the closed PRs

integration/common/src/openlineage/common/provider/dbt/local.py
```diff
@@ -263,11 +263,15 @@ def get_dbt_metadata(
...
-
         try:
+            profile_dir = run_result["args"]["profiles_dir"]
             profile = self.load_yaml_with_jinja(os.path.join(profile_dir, "profiles.yml"))[self.profile_name]
-        except FileNotFoundError:
+        except (KeyError, FileNotFoundError):
+            self.logger.warning(
+                "Could not load profile directory from run_results.json. "
+                "Falling back to default profiles.yml location.",
+                exc_info=True
```
ah12068: added this line so that end users can see what exactly is the error when being warned:
image

kacpermuda: Not sure if that's needed, as it's not actionable from the user perspective, right? Logging message is pretty clear, user can easily find it in the code, and the traceback info here can be messy and suggest there was an actual error that was not handled. Imho we can remove the `exc_info=True`

kacpermuda:
No problem, can you also add some tests for that change to make sure it works as expected?

integration/common/tests/dbt/test_dbt_local.py
ah12068, line 1: The approach i did was to overwrite some of the existing values that are loaded into `DbtLocalArtifactProcess().get_dbt_metadata()` to enforce the actual errors as a way to test. Some feedback here would be appreciated

mobuchowski (approved):

PR #4342, opened 2026-02-16
Website: enhance dark mode and remove Typography

merobi-hub:
Enhance dark mode and remove Typography

### Meaningful description
Dark mode was not implemented fully on the Community and Ecosystem pages due to issues with applying custom colors to Typography elements.

This removes the Typography library, which was overkill and only in use on these pages, and adds CSS as needed for replacement of Typography styling and full implementation of dark mode. An added benefit of the change is that fonts are now consistent across the landing pages.


merobi-hub:
Screenshot_20260216_163215
Screenshot_20260216_163149
Screenshot_20260216_163104

jakub-moravec:
Hi @merobi-hub , I noticed other dark mode problem: 

image

vs 

image

- notice tags/source code formatting not having visible highlighting in dark mode.

Would it be worth addressing this in the same PR? I can try to chip in with the fix if you want.

merobi-hub:
@jakub-moravec yes, I think it makes sense to add that fix to the PR, and I would be happy for you to do it if you'd like. Thanks!

website/src/css/custom.css
jakub-moravec:
```suggestion
    background-color: #3a3a3a;
    border: 1px solid #555555;
```

jakub-moravec: Image

PR #4345, opened 2026-02-18
dbt: Use get method for dbt project version retrieval

zagoodman:

Uses `.get()` for optional dbt project's `version` key.

### Meaningful description

Per dbt's [docs](https://docs.getdbt.com/reference/project-configs/version#:~:text=Starting%20in%20dbt%20version%201.5%2C%20version%20in%20the%20dbt_project.yml%20is%20an%20optional%20parameter):
> Starting in dbt version 1.5, version in the dbt_project.yml is an optional parameter.

However, the existing `dbt_project["version"]` throws an error if the `version` tag is omitted from `dbt_project.yml` rather than return `None`, which is fixed by this patch.

### Checklist

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/local.py
+++ b/integration/common/src/openlineage/common/provider/dbt/local.py
@@ -106,7 +106,7 @@ class DbtLocalArtifactProcessor(DbtArtifactProcessor):
 
         self.target = target
         self.project_name = dbt_project["name"]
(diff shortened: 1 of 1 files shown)


mobuchowski:
Thanks for the contribution @zagoodman

zagoodman:
> Thanks for the contribution @zagoodman

Happy to help, thanks for the review @mobuchowski!

PR #4351, opened 2026-02-20
fix(spark): Emit FAIL event for application end when jobs fail

### LINKED ISSUE #4350: [BUG] Spark integration always emits COMPLETE for application end event, even when jobs fail

Spark integration

### OpenLineage version

`io.openlineage:openlineage-spark_2.12:1.44.0`, `io.openlineage:openlineage-spark_2.12:1.43.0`

- OpenLineage: 1.44.0
- Apache Spark: 3.1.1 (AWS Glue 3.0), 3.5.x (AWS Glue 5.0)
- Python: 3
- Java: 17

### Environment configuration

**Environment**: AWS Glue 5.0 (Spark 3.5.4, Scala 2.12)

**OpenLineage JAR**: `openlineage-spark_2.12-1.44.0.jar` loaded via `--extra-jars` with `--user-jars-first=true` (overriding Glue 5's bundled v1.34.0)

**Spark configuration** (passed via Glue job `--conf` argument):
```
spark.extraListeners=io.openlineage.spark.agent.OpenLineageSparkListener
spark.openlineage.namespace=aws_glue_test
spark.openlineage.transport.type=http
spark.openlineage.transport.url=https://<host>
spark.openlineage.transport.endpoint=/api/v1/lineage
spark.openlineage.transport.auth.type=api_key
spark.openlineage.transport.auth.apiKey=<key>
```

### Deployment details

AWS Glue 5.0 and 3.0

### Problem details

The Spark integration always emits a COMPLETE event for the application-level end event (`SparkListenerApplicationEnd`) and also for sql job level end event, even when individual Spark jobs have failed. This makes it impossible to detect failed applications from OpenLineage events alone.

Additionally, SQL execution end events never include an `errorMessage` run facet with failure details, even when the execution failure information is available (Spark 3.4+ via `executionFailure()` on `SparkListenerSQLExecutionEnd`).

The application-level end event should emit FAIL when any job within the application has failed. SQL execution end events should emit FAIL with an `errorMessage` run facet when `executionFailure()` is present (e.g., output committer failures).

### How to reproduce

1. Run any PySpark job that fails (e.g., write to a non-existent S3 bucket):
     `df.write.mode("overwrite").parquet("s3://non-existent-bucket/path/")`
2. Observe the OpenLineage events emitted.
3. The SQL job level end event has `eventType`: COMPLETE instead of FAIL.
4. The application-level end event has `eventType`: COMPLETE instead of FAIL.

### Additional details

_No response_

tstrilka:
closes: #4350

Emit FAIL event for application-level end when Spark jobs fail, instead of always emitting COMPLETE.

### Meaningful description
**Problem:** The application-level end event was always emitted as COMPLETE, even when individual Spark jobs failed. Also SQL job level end event was always emitted as COMPLETE. This made it impossible to detect failed applications from OpenLineage events alone. Additionally, SQL execution end events never included an `errorMessage` run facet with failure details.

**Solution:**
- Track job failures via `JobFailed` in `onJobEnd` (works on all Spark versions) and propagate the failure to the application end event.
- On Spark 3.4+ (SPARK-40834), also check `executionFailure()` on `SparkListenerSQLExecutionEnd` to detect output committer failures that occur after the Spark job completes successfully. When detected, emit a FAIL event with an `errorMessage` run facet at the SQL execution level.
- Add `end(applicationEnd, eventType)` overload to `ExecutionContext` interface.

**Tested on:** AWS Glue 3.0 (Spark 3.1) and AWS Glue 5.0 (Spark 3.5).

### Checklist


dolfinus:
This looks similar to #4265

tstrilka:
> This looks similar to #4265

This supersedes #4265. Key differences:
1. Last-event-wins instead of sticky boolean — as you suggested in #4265, we track the most recent job/SQL execution outcome (lastExecutionEventType) rather than a sticky failedJobsFound flag. This correctly handles the try/except scenario: if a failure is caught and subsequent operations succeed, the app emits COMPLETE.
2. No NPE — #4265 called context.getStatus() on a potentially null context from rddExecutionRegistry.remove(). This PR avoids that entirely by tracking failure state on the listener itself, independent of execution contexts.
3. SQL execution failure detection - also picks up output committer failures via executionFailure() (Spark 3.4+, SPARK-40834).

Tested with the shadow JAR in AWS Glue 5.0 (Spark 3.5.4) — no NPE, FAIL/COMPLETE events emitted correctly.

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -170,6 +182,39 @@ public void end(SparkListenerSQLExecutionEnd endEvent) {
     eventEmitter.emit(event);
   }
 
+  private Throwable getExecutionFailure(SparkListenerSQLExecutionEnd endEvent) {
+    try {
+      scala.Option<?> failure = endEvent.executionFailure();
```
dolfinus: This will compile only with Spark 3.4, but if I understand correctly, project `integration/spark` is compiled with spark.version==3.3.4:
https://github.com/OpenLineage/OpenLineage/blob/d36fc6a74a7b6ad9fea8b9f89700c35126fab650/integration/spark/gradle.properties#L4

This probably should use Reflection API instead

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/ExecutionContext.java
```diff
@@ -38,6 +39,10 @@ public interface ExecutionContext {
 
   void end(SparkListenerApplicationEnd applicationEnd);
 
+  default void end(SparkListenerApplicationEnd applicationEnd, EventType eventType) {
+    end(applicationEnd);
+  }
```
dolfinus: This looks a bit awkward, eventType is lost by default

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -135,30 +133,44 @@ public void end(SparkListenerSQLExecutionEnd endEvent) {
...
+            .jobBuilder(buildJob())
+            .jobFacetsBuilder(getJobFacetsBuilder(olContext.getQueryExecution().get()));
+
+    if (runFacetsBuilder != null) {
+      contextBuilder.runFacetsBuilder(runFacetsBuilder);
+    }
+
+    RunEvent event = runEventBuilder.buildRun(contextBuilder.build());
 
     if (eventType.equals(EventType.COMPLETE)) {
```
dolfinus:
```suggestion
    if (eventType.equals(EventType.COMPLETE) || eventType.equals(EventType.FAIL)) {
```

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -171,6 +191,14 @@ private void sparkSQLExecEnd(SparkListenerSQLExecutionEnd endEvent) {
                         return null;
                       }));
     }
+    // Check executionFailure directly (Spark 3.4+, SPARK-40834) to detect output committer
+    // failures that occur after the Spark job completes successfully.
+    try {
+      scala.Option<?> failure = endEvent.executionFailure();
+      lastExecutionEventType = (failure != null && failure.isDefined()) ? FAIL : COMPLETE;
+    } catch (NoSuchMethodError e) {
```
dolfinus: Same note about reflection

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
tstrilka: fixed - switched to reflection

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/ExecutionContext.java
tstrilka: fixed - switched the relationship so end(applicationEnd, eventType) is the abstract method now

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
tstrilka: added

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
tstrilka: fixed - using reflection

dolfinus (approved):

tnazarew:
his is similar to the #4265 and both have issue with applications that write handling for job failures. 
as @dolfinus brought up the case where the sticky boolean fails, this approach would have issue with the example below

```
spark = SparkSession.builder.getOrCreate()

with spark:
  df = spark.createDataFrame(...)
  df.saveAsTable("important.table")

try:
    df.saveAsTable("unimportant.table")  # raises PermissionDeniedError, so command event is emitted with FAIL
  except Exception:
    print("Writing to table failed but it's not a priority so we don't fail on it")
```

That is probably more of a corner case but still the main issue seems to be "we can't assume that failed job means failed application when job failures can simply be caught in the code"

dolfinus:
I think that detecting that last nested job failure within app fails means the entire app fails is good enough approximation. But this can be covered with some kind of feature flag, if users know that this produces false positives

tstrilka:
> I think that reacting other last nested job failure within app fails means the entire app fails is good enough approximation. But this can be covered with some kind of feature flag, if users know that this produces false positives

I can add the feature flag in this PR or as a follow-up — what do you prefer?

dolfinus:
Let's wait what @tnazarew thinks about this solution first

tnazarew:
@dolfinus @tstrilka 

I don't really like how this will impact the meaning of application status. 
Now it's basically - appCompete -> application ended the status will be always `COMPLETE` so it's only to convey message like "the application is no longer running, anything more you have to figure out separately, the status is set to `COMPLETE` because it has to be something..."

If the application status is not constant, it implies that it _**means**_ something.

Even with reliable mechanism to extrapolate the app status from job statuses, the best case scenario is
- `FAIL` -> there was an error in the job that caused the processing to fail
- `COMPLETE` -> no jobs failed... or there was an error in the application code before job that would fail but didn't get the chance, who knows

when this mechanism is unreliable we have
- `FAIL` -> there was an error in the job that might have caused processing to fail, but not necessarily. Did the processing fail? We don't know
- `COMPLETE` -> no jobs failed... well at least not the last one, or there was an error in the application code before job that would fail but didn't get the chance, who knows

So while in the first scenario we would at least have reliable meaning of fail, in the second we don't even have that. 

Even the best case scenario is far from optimal and I was hesitant to implement it as changes app status from carrying no information to carrying potential misinformation.

dolfinus:
That makes sense, thanks @tnazarew

tstrilka:
Fair point about caught exceptions. However, the current behavior (without this fix) has the opposite problem - when AWS Glue job actually fails, all OpenLineage events (meaning final state ones) still report COMPLETE. So there is already misinformation - real failures go undetected :-(

tnazarew:
@tstrilka we probably need to emphasise it more in the docs for the spark integration but the application event `COMPLETE` as it is now is just information that there will be no more events emited by this application, maybe also execution time is in there but that's it. 

It doesn't carry any information about the status of the application. So it's not misinformation but lack of it. 

There are ways to detect application failure but they are outside of spark integration.

tstrilka:
@tnazarew thanks for your feedback, I think you make a fair point about the application level FAIL from JobFailed being unreliable. I'm ok with dropping that part.

There are two separate things in this PR and I want to make sure the second one is still ok, meaning the SQL execution level FAIL via executionFailure() (Spark 3.4+). This is Spark telling us directly that the execution failed. The SQL execution context already emits FAIL on JobFailed so this just covers an additional failure path that JobFailed misses (like output committer failures).

Would you be okay with keeping that part while dropping the application level logic?

Also, you mentioned there are ways to detect application failure outside of the Spark integration, could you point me in the right direction? To understand the intended approach better? Thanks.

tnazarew:
@tstrilka 

> Would you be okay with keeping that part while dropping the application level logic?

yes, if it improves failure detection then I see no issue 
 
> Also, you mentioned there are ways to detect application failure outside of the Spark integration, could you point me in the right direction? To understand the intended approach better? Thanks.

There are no ready to use solutions, at least not publicly available. You'd have to implement it and they have their own pitfalls.

How could be done is e.g.
Wrapper script for spark submit which would:
- take all the arguments for spark submit 
- emit OL start event using transport config from properties (store the run id, job name, job namespace)
- execute the actual spark submit but inject stored information to parent properties so it will be used in the integration
- after spark is done, get the application status
- if the status is 0 (success) emit OL complete event, else emit OL fail event

You'd get additional layer 

| Layer   | Event                                                                    |
|---------|--------------------------------------------------------------------------|
| wrapper | {type: `START`    , id: `X`,  name: `Y`},                                |
| app     | {type: `START`    , id: `A`,  name: `B`, parent:  {id: `X`, name: `Y`} } |
| job     | {type: `START`    , id: `A1`, name: `B1`, parent: {id: `A`, name: `B`} } |
| job     | {type: `COMPLETE` , id: `A1`, name: `B1`, parent: {id: `A`, name: `B`} } |
| job     | {type: `START`    , id: `A2`, name: `B2`, parent: {id: `A`, name: `B`} } |
| job     | {type: `COMPLETE` , id: `A2`, name: `B2`, parent: {id: `A`, name: `B`} } |
| app     | {type: `COMPLETE` , id: `A`,  name: `B`, parent:  {id: `X`, name: `Y`} } |
| wrapper | {type: `COMPLETE` , id: `X`,  name: `Y`}                                 |

Something similar is done in airflow spark_submit operator.

tstrilka:
@tnazarew thank you for the detailed explanation of the wrapper approach. I appreciate it.
                                                                                                      
Please note that I updated the PR based on your feedback by:
- removing all application-level FAIL logic (`lastExecutionEventType`, `JobFailed` tracking) - application end events now always emit COMPLETE
- keeping the SQL execution-level FAIL via `executionFailure()` with `errorMessage` run facet - this is a reliable signal from Spark

I tested this on AWS Glue 5.0 - application-level events are always COMPLETE, SQL execution level events correctly emit FAIL or COMPLETE depending on the outcome.

tstrilka:
@tnazarew Regarding the wrapper approach - it works well for self-managed Spark (e.g., standalone, YARN), but doesn't apply to managed services like AWS Glue, EMR Serverless, or Databricks where we don't control the spark-submit invocation. For those environments, detecting failures within the integration (like the SQL execution-level FAIL via `executionFailure()`) is the only option. With this implemented approach (on signals) it is reliable.

tstrilka:
Hi @tnazarew, kindly asking if the current implementation (just SQL execution level FAIL via executionFailure() (Spark 3.4+)) is o.k. for approval? Thanks.

PR #4358, opened 2026-02-24
Go client

tnazarew:
add OpenLineage Go client

### Meaningful description
Added OpenLineage Go client
- class generation
- transport implementation
- pre-commit config
- CI steps configured

Note:
Some unconventional changes:
- `DataQualityMetricsInputDatasetFacet` is not generated, it's 1-1 copy of `DataQualityMetricsDatasetFacet` and it caused some conflicts while generating classes so I skip it. 
- transport method Emit() returns map string -> string so the target system can return some value if needed

based on the @ThijsKoot implementation from https://github.com/ThijsKoot/OpenLineage/tree/go-client
with great help from Claude 

### Checklist

### DIFF
--- a/client/go/pkg/openlineage/client.go
+++ b/client/go/pkg/openlineage/client.go
@@ -8,6 +8,7 @@ package openlineage
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/google/uuid"
@@ -29,6 +30,10 @@ type ClientConfig struct {
 // NewClient creates a new OpenLineage client.
 // producer is a URI identifying the producer of this metadata (e.g., "https://github.com/OpenLineage/OpenLineage/tree/1.23.0/integration/spark")
 func NewClient(producer string, cfg *ClientConfig) (*Client, error) {
+	if cfg == nil {
+		return nil, errors.New("config must not be nil")
+	}
+
 	if cfg.Disabled {
 		return &Client{
 			disabled: true,
--- a/client/go/pkg/openlineage/context.go
+++ b/client/go/pkg/openlineage/context.go
@@ -7,6 +7,7 @@ package openlineage
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-stack/stack"
 	"github.com/google/uuid"
@@ -63,24 +64,16 @@ type Run interface {
 	// The returned map contains any metadata from the consumer; it may be nil.
 	Emit(context.Context, Emittable) (map[string]string, error)
 
-	// Finish emits a COMPLETE event if no error has occurred, or a FAIL event otherwise.
-	// If an error is passed it is recorded via RecordError before finishing — nil is ignored.
+	// Finish emits a COMPLETE event if err is nil, or a FAIL event with an ErrorMessage facet otherwise.
 	// This allows the common pattern:
 	//
-	//	defer run.Finish()
+	//	defer run.Finish(nil)
 	//
 	// as well as the single-call error pattern:
 	//
 	//	err := doWork()
 	//	run.Finish(err)
-	Finish(errs ...error)
-
-	// Returns true if RecordError was called for this Run.
-	HasFailed() bool
-
-	// RecordError emits an OTHER event with an ErrorMessage facet.
-	// Once this is called, the run is considered to have failed.
-	RecordError(error)
+	Finish(err error)
 }
 
 type run struct {
@@ -88,9 +81,7 @@ type run struct {
 	runID        uuid.UUID
 	jobName      string
 	jobNamespace string
-
-	hasFailed bool
-	client    *Client
+	client       *Client
 }
 
 // JobName implements RunContext.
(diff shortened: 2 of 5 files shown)


client/go/pkg/openlineage/client.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return olc.transport.Emit(ctx, event.AsEmittable())
+}
+
+// NewRun creates a Run and sets it as the active Run in ctx.
+// If ctx already contains a RunContext, it set as the parent.
+func (c *Client) NewRun(ctx context.Context, job string) (context.Context, Run) {
+	r := run{
+		client:       c,
+		runID:        uuid.New(),
```
dolfinus:
```suggestion
		runID:        uuid.NewV7(),
```

It is highly recommended to use UUID v7 in OpenLineage implementations.

client/go/README.md
```diff
@@ -0,0 +1,444 @@
...
+
+OpenLineage defines three types of events:
+
+- **RunEvent**: Represents a job execution (run) with START, RUNNING, COMPLETE, ABORT, FAIL, or OTHER states
+- **JobEvent**: Static job metadata without run information
+- **DatasetEvent**: Static dataset metadata
+
+### Runs
+
+The `Run` interface provides an ergonomic way to instrument your code:
```
mobuchowski: I see we're going for trace-like interface. 
I like it, but would also like to have an option just to build the run manually and emit at my own pace.

client/go/README.md
```diff
@@ -0,0 +1,444 @@
...
+
+run.RecordOutputs(
+    ol.NewOutputElement("user_orders", "s3://bucket/warehouse"),
+)
+
+// If something goes wrong
+if err := doWork(); err != nil {
+    run.RecordError(err) // Marks run as failed, Finish() will emit FAIL
+    return err
+}
```
mobuchowski, line 110: We can make it look similar to https://pkg.go.dev/github.com/DataDog/dd-trace-go/v2/ddtrace#example-package-Datadog - obviously I'm biased, but I think it's a bit more ergonomic with single return
```
	// Perform an operation.
	_, err := os.ReadFile("~/test.json")

	// We may finish the child span using the returned error. If it's
	// nil, it will be disregarded.
	child.Finish(tracer.WithError(err))
	if err != nil {
		log.Fatal(err)
	}
```

client/go/LOCAL_DEVELOPMENT.md
```diff
@@ -0,0 +1,173 @@
+debu# Using the OpenLineage Go Client Locally
```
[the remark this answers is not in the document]
tnazarew: file was redundant so I removed it

mobuchowski:
Thanks for the awesome work on making Go first-class client in OpenLineage ecosystem @tnazarew
The direction we're going in looks roughly right. That does not mean it's complete. there are things missing compared to Java and Python clients.
However, the PR is already massive and useful in it's state. Nothing starts perfect, and I'd accept the PR as it is. 
But I think it's a good idea to document the gaps we have in this PR and have some plan to implement solutions for some of them - where it makes sense.

Some generated information about those gaps:

Transports aren't up to parity:

```
| Transport | Python | Java | Go |
|-----------|--------|------|----|
| HTTP | ✅ | ✅ | ✅ |
| Console | ✅ | ✅ | ✅ |
| GCP Lineage | ✅ | ✅ | ✅ |
| Kafka | ✅ | ✅ | ❌ |
| File | ✅ | ✅ | ❌ |
| Noop/Disabled | ✅ (NoopTransport) | ✅ (NoopTransport) | ❌ (no noop transport) |
| Composite | ✅ | ✅ | ❌ |
| Async HTTP | ✅ | ❌ | ❌ |
| Amazon DataZone | ✅ | ✅ | ❌ |
| Datadog | ✅ | ❌ | ❌ |
```

Of course I don't expect you to implement Amazon or Datadog transports.

But I think Composite is extremely useful.


Some other feature gaps:
```
| Feature | Python | Java | Go |
|---------|--------|------|----|
| YAML config file (`openlineage.yml`) | ✅ | ✅ | ❌ |
| Env var config (`OPENLINEAGE_CONFIG`) | ✅ | ✅ | ❌ |
| `OPENLINEAGE_DISABLED` env var | ✅ | ✅ | ❌ (only `ClientConfig.Disabled` field) |
| Event filtering (regex/exact match) | ✅ | ✅ | ❌ |
| Facet disabling | ✅ | ✅ | ❌ |
| Circuit breaker | ❌ | ✅ | ❌ |
| Metrics (Micrometer) | ❌ | ✅ | ❌ |
| Composite transport | ✅ | ✅ | ❌ |
| Parent-child run facets | ✅ (manual) | ✅ (builder) | ✅ (context-based, automatic) |
| Custom token providers | ✅ (OAuth2/JWT refresh) | ✅ (JWT refresh) | ❌ (static tokens only) |
```

Circuit breaker is not very relevant to Go though.
I think yaml/env var conf, ability to disable both emission and facets, and custom token providers would be very useful.

In addition to that, things I'd like to see - again, probably not in this PR:
- Some integration tests, like with docker. We have precedent in Python client.
- Update the website with documentation of a Go client. The readme could be here too, but it's a developer documentation at best, not the end user one, as it's buried deep in subdirectory.

PR #4370, opened 2026-03-03
add spark streaming micro-batch throttle mechanism

mobuchowski:
Spark Structured Streaming emits a START + COMPLETE OpenLineage event pair for every micro-batch. At a 10-second trigger interval that is roughly 360 events per hour per streaming query, indefinitely, creating unnecessary load on the OpenLineage backend.

## Solution

A new StreamingMicroBatchThrottler reduces event volume via two optional modes: count-based (spark.openlineage.filter.streaming.microbatchThrottle=N — emit every Nth batch) and time-based (spark.openlineage.filter.streaming.microbatchThrottleMinutes=M — emit at most once every M minutes). When both are set, OR semantics apply; when neither is set all micro-batches continue to emit (fully backward-compatible). The throttle is checked once per micro-batch at SparkListenerSQLExecutionStart; if suppressed, all subsequent emit calls for that batch (jobStart, SQLExecutionEnd, jobEnd) are skipped atomically. Batch (non-streaming) queries are never affected.

### Checklist
- [*] AI was used in creating this PR: Claude Code

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
@@ -62,8 +62,12 @@ class SparkSQLExecutionContext implements ExecutionContext {
   private boolean emittedOnJobEnd = false;
   private Integer activeJobId;
   private AtomicBoolean finished = new AtomicBoolean(false);
-  /** Set to true when this micro-batch is throttled; all subsequent emit calls are no-ops. */
-  private boolean throttled = false;
+  /**
+   * Set to true when this micro-batch is throttled; all subsequent emit calls are no-ops. Declared
+   * volatile for the same reason {@link #finished} uses AtomicBoolean: lifecycle events may arrive
+   * on different threads.
+   */
+  private volatile boolean throttled = false;
 
   private SparkSQLQueryParser sqlRecorder = new SparkSQLQueryParser();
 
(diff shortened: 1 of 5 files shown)


tnazarew (commented):
few scenarios that I think are not handled here:
- we configure max 1 event every hour but there is a failure after 30 minutes, as I understand we just don't emit it with this throttling
- for each micro-batch we have X outputs defined (so X different events for a batch) and we set throttle to emit every Xth event, only one job would be visible instead of X different ones.
- I think we could define a spark app in which after stopping the stream we do something else, then this other job would also not be visible

there should be some throttle id based on the job name and namespace, that way we could define emitting only Nth event with a job of this type

mobuchowski:
@tnazarew updated it:
- we never throttle failures
- we have now per-job throttle state - each streaming query now has it's own counter/timestamp, so multiple outputs don't interfere.
- throttling throttles only streaming jobs - checks if it's streaming

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java
```diff
@@ -303,6 +323,10 @@ public void start(SparkListenerJobStart jobStart) {
   @Override
   public void end(SparkListenerJobEnd jobEnd) {
     log.debug("SparkListenerJobEnd - executionId: {}", executionId);
+    // Always emit failure events even when this micro-batch was throttled, so that job failures
+    // are never silently dropped.
+    boolean isFailed = jobEnd.jobResult() instanceof JobFailed;
```
tnazarew, line 345: this would mean we will emit FAIL event without Start event but I guess we can live with that

Issue #4162, opened 2025-11-20
[BUG] Iceberg handler broken for glue catalog because of missing aws identifier


usamakunwar:
### OpenLineage integration/client

Spark integration

### OpenLineage version

io.openlineage:openlineage-spark_2.12:1.39.0

### Problem details

For iceberg datasets with glue catalog the handler raises a null pointer exception as a result of getGlueArn returning none. It appears to be because of this change https://github.com/OpenLineage/OpenLineage/pull/4053.
I think this means no iceberg datasets are created

```
06:51:20,981 ERROR io.openlineage.spark.agent.util.PlanUtils                     - Apply failed:
java.lang.NullPointerException: null
	at java.util.Objects.requireNonNull(Objects.java:209) ~[?:?]
	at java.util.Optional.of(Optional.java:113) ~[?:?]
	at io.openlineage.spark3.agent.lifecycle.plan.catalog.iceberg.IcebergHandler.getDatasetIdentifier(IcebergHandler.java:173) ~[openlineage-spark128.jar:?]
	at io.openlineage.spark3.agent.lifecycle.plan.catalog.CatalogUtils3.lambda$getDatasetIdentifier$2(CatalogUtils3.java:63) ~[openlineage-spark128.jar:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179) ~[?:?]
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:179) ~[?:?]
	at java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1602) ~[?:?]
	at java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129) ~[?:?]
	at java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
[...]

PR #4384, opened 2026-03-09
spark: fix spark iceberg glue catalog symlink

usamakunwar:
Fixes dataset symlinks when using glue catalog for iceberg. 
closes: #4162 

### Meaningful description

[getGlueArn](https://github.com/OpenLineage/OpenLineage/blob/291532fd436babd68edef682d6e07d56d0fecd31/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java#L31) currently returns null if not hive, [this](https://github.com/OpenLineage/OpenLineage/commit/d5db993253568d53c43ed9f8850788b6064b27d2) fixed the NPE, however if the ARN is null it looks like the identifier won't be [created](https://github.com/OpenLineage/OpenLineage/blob/291532fd436babd68edef682d6e07d56d0fecd31/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/iceberg/GlueCatalogTypeHandler.java#L37) at all.

Adding a function **isIcebergUsingGlue** that checks if the catalog used is glue, and if so return an ARN. 
Glue is determined by spark conf prefix "spark.sql.catalog."
e.g. 
spark.sql.catalog.iceberg.catalog-impl=org.apache.iceberg.aws.glue.GlueCatalog

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/AwsUtils.java
@@ -9,6 +9,7 @@ import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.Optional;
 import lombok.SneakyThrows;
 import lombok.experimental.UtilityClass;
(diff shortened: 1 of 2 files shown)


mobuchowski:
Thanks for the contribution @usamakunwar.

PR #4398, opened 2026-03-16
dbt: add sourceCodeLocation facet with repo URL override and git autodetect

himakolavennu:
Add `sourceCodeLocation` job facets to dbt OpenLineage events using explicit repo URL input or git-origin autodetection, normalized to `host/org/repo`.

### Meaningful description
dbt Core events currently do not include source repository location metadata, which makes it harder to link lineage to source code in downstream systems.
This PR adds `sourceCodeLocation` facet emission for dbt job events across both local-artifact and structured-log processing paths.
The repo URL can now be provided explicitly via `--openlineage-repo-url` or `OPENLINEAGE_REPO_URL`, with fallback autodetection using `git remote get-url origin`. URLs are normalized to `host/org/repo` (including SSH forms) to match downstream lineage processor expectations.

### Checklist

### DIFF
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -30,7 +30,13 @@ from openlineage.client.transport import (
 )
 from openlineage.client.transport.http import HttpConfig, HttpTransport
 from openlineage.client.transport.noop import NoopConfig, NoopTransport
-from openlineage.client.utils import deep_merge_dicts, get_git_repo_url
+from openlineage.client.utils import (
+    deep_merge_dicts,
+    get_git_branch,
+    get_git_repo_url,
+    get_git_tag,
+    get_git_version,
+)
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore", DeprecationWarning)
@@ -533,8 +539,14 @@ class OpenLineageClient:
         return tags_facet
 
     @cached_property
-    def _source_code_location_url(self) -> str | None:
-        return get_git_repo_url(self.config.facets.source_code_location.repo_url)
+    def _source_code_location(self) -> dict[str, str | None]:
+        scl = self.config.facets.source_code_location
+        return {
+            "url": get_git_repo_url(scl.repo_url),
+            "version": get_git_version(scl.version),
+            "branch": get_git_branch(scl.branch),
+            "tag": get_git_tag(scl.tag),
+        }
 
     def add_source_code_location_facet(self, event: Event) -> Event:
         """Adds sourceCodeLocation job facet if not already present and not disabled."""
@@ -543,15 +555,18 @@ class OpenLineageClient:
         if not isinstance(event, (RunEvent, event_v2.RunEvent)):
             return event
 
-        url = self._source_code_location_url
-        if not url:
+        scl = self._source_code_location
+        if not scl["url"]:
             return event
 
         event.job.facets = event.job.facets or {}
         if "sourceCodeLocation" not in event.job.facets:
             event.job.facets["sourceCodeLocation"] = source_code_location_job.SourceCodeLocationJobFacet(
                 type="git",
-                url=url,
-                repoUrl=url,
+                url=scl["url"],
+                repoUrl=scl["url"],
+                version=scl["version"],
+                branch=scl["branch"],
+                tag=scl["tag"],
             )
         return event
--- a/client/python/src/openlineage/client/facets.py
+++ b/client/python/src/openlineage/client/facets.py
@@ -9,6 +9,9 @@ import attr
 class SourceCodeLocationConfig:
     disabled: bool = False
     repo_url: str | None = None
+    version: str | None = None
+    tag: str | None = None
+    branch: str | None = None
 
 
 @attr.define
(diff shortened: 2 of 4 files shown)


integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -177,6 +184,63 @@ def dbt_run_metadata(self):
...
+                return f"{host}/{path}"
+
+        return url
+
+    def _get_repo_url(self) -> Optional[str]:
+        """Get normalized repo URL: explicit override first, then git autodetect."""
+        raw_url = self.repo_url
+        if not raw_url and self.project_dir:
+            try:
+                result = subprocess.run(
```
mobuchowski: Let's add timeout=5 to this invocation, in rare occasions git can hang

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -177,6 +184,63 @@ def dbt_run_metadata(self):
...
+        parsed = urlparse(url)
+        if parsed.scheme in ("https", "http", "ssh"):
+            host = parsed.hostname or ""
+            path = parsed.path.strip("/")
+            if host and path:
+                return f"{host}/{path}"
+
+        return url
+
+    def _get_repo_url(self) -> Optional[str]:
```
mobuchowski: Can we add a `functools.cached_property` on this method, so we won't call git N times for every dbt model and test?

dolfinus:
I like the idea. Part of implementation (git autodetection, `OPENLINEAGE_REPO_URL` env variable) can be reused in other integrations

kacpermuda:
Maybe this should live in a client, and the facet can be added by default (if not already present + with option to disable this automatic behavior), so that all integrations can use it?

himakolavennu:
> Maybe this should live in a client, and the facet can be added by default (if not already present + with option to disable this automatic behavior), so that all integrations can use it?

Good idea! Can do that in a separate PR.

kacpermuda:
> Can do that in a separate PR.

Let's discuss and choose a single spot for this implementation, no need to implement it here just to move it to client in next PR. 

@mobuchowski @dolfinus WDYT is the best place for that?

mobuchowski:
I would move it to python client's utils - as it's not pulling any new dependencies, can't see anything wrong with that.

client/python/src/openlineage/client/utils.py
```diff
@@ -55,6 +57,30 @@ def deep_merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any,
     return merged
 
 
+def get_git_repo_url(repo_url: str | None = None, timeout: int = 5) -> str | None:
```
dolfinus: [Source Code Location Facet](https://openlineage.io/docs/spec/facets/job-facets/source-code-location) also contains fields like version, tag and branch, which can probably be autodetected as well, or passed explicitly via config

client/python/src/openlineage/client/client.py
```diff
@@ -517,3 +537,36 @@ def _update_tag_facet(
...
+
+    @cached_property
+    def _source_code_location(self) -> dict[str, str | None]:
+        scl = self.config.facets.source_code_location
+        return {
+            "url": get_git_repo_url(scl.repo_url),
+            "version": get_git_version(scl.version),
+            "branch": get_git_branch(scl.branch),
+            "tag": get_git_tag(scl.tag),
+        }
```
kacpermuda, line 549: This can take up to 20 seconds (4 subprocesses with 5 seconds timeout each)? We definitely need to adjust that

kacpermuda: @himakolavennu Any way we can just get all the git information in a single subprocess with a single command (possibly some parsing of output will be needed), but then if git is not there, we will not be running 4 subprocesses for nothing, one will do and will fail fast.

client/python/src/openlineage/client/facets.py
```diff
@@ -5,6 +5,16 @@
 import attr
 
 
+@attr.define
+class SourceCodeLocationConfig:
+    disabled: bool = False
```
kacpermuda, line 10: I think we should disable it by default, users probably do not expect OpenLineage to run git commands. We should have an option to do that (and to pass explicitly, as this Pr does), but I don't feel like adding this and enabling by default is a good choice.

client/python/src/openlineage/client/client.py
kacpermuda: Or maybe we can read all that information from .git directory without using subprocess at all? We can then skip the entire check if not git directory is accessible.

client/python/src/openlineage/client/facets.py
kacpermuda: This probably contradicts with what i said earlier:

> the facet can be added by default (if not already present + with option to disable this automatic behavior)

But now that I think about it, the git calls should not be made by default. If we switch the approach to some dir / files check, then we can probably leave it on by default, this will not be expensive.

client/python/src/openlineage/client/client.py
dolfinus: At least we can skip running those processes if there is no `.git` folder in current directory or its parents

PR #4400, opened 2026-03-17
provide data quality assertions for singular tests

mobuchowski:
AI description:
dbt singular tests are custom SQL test files placed in the `tests/` directory that validate data quality using arbitrary SQL queries. Unlike generic tests (e.g. `not_null`, `unique`), singular tests have no `attached_node` in the dbt manifest — they reference models via `ref()` and `source()` calls directly in SQL. PR #4372 added support so that singular tests no longer crash during OpenLineage event emission, but their test results were silently dropped and never attached to any dataset as `DataQualityAssertionsDatasetFacet`.

This PR fixes the structured logs processor to correctly resolve the parent datasets for singular tests by using dbt's `parent_map`, which is pre-populated with all upstream models and sources referenced via `ref()`/`source()`. The method `_get_attached_dataset()` has been renamed to `_get_attached_datasets()` and now returns a list: for generic tests it returns the single `attached_node` dataset as before, while for singular and source tests (both have `attached_node: null`) it traverses `parent_map` to find all referenced datasets. As a side effect, source tests — which also had `attached_node: null` — now correctly attach assertions to their source datasets as well. An integration test using `dbt-ol --consume-structured-logs` has been added to validate the end-to-end behavior against a real DuckDB database.

### Checklist
- [x] AI was used in creating this PR - yes, Claude Code

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/structured_logs.py
+++ b/integration/common/src/openlineage/common/provider/dbt/structured_logs.py
@@ -474,13 +474,11 @@ class DbtStructuredLogsProcessor(DbtLocalArtifactProcessor):
                 assertion_facet = dq.DataQualityAssertionsDatasetFacet(assertions=[assertion])
                 inputs = []
                 for attached_dataset in self._get_attached_datasets(node_unique_id):
-                    dataset_facets = {**attached_dataset.facets}
-                    dataset_facets["dataQualityAssertions"] = assertion_facet
                     inputs.append(
                         Dataset(
                             name=attached_dataset.name,
                             namespace=attached_dataset.namespace,
-                            facets=dataset_facets,
+                            facets={**attached_dataset.facets, "dataQualityAssertions": assertion_facet},  # type: ignore[dict-item]
                         )
                     )
 
--- a/integration/dbt/tests/integration/test_structured_logs.py
+++ b/integration/dbt/tests/integration/test_structured_logs.py
@@ -197,12 +197,13 @@ class TestStructuredLogs:
                 f"got empty inputs. Event: {event.get('job', {}).get('name')}"
             )
 
-            # At least one input should have dataQualityAssertions facet
+            # All inputs should have dataQualityAssertions facet
             inputs_with_assertions = [
                 inp for inp in inputs if "dataQualityAssertions" in inp.get("facets", {})
             ]
-            assert len(inputs_with_assertions) >= 1, (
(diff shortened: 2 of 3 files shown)


integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -77,6 +77,7 @@ class DbtNodeJobFacet(JobFacet):
     schema: str | None = attr.field(default=None)
     alias: str | None = attr.field(default=None)
     unique_id: str | None = attr.field(default=None)
+    test_type: str | None = attr.field(default=None)
```
kacpermuda, line 80: Let's add this to spec file as well

integration/common/src/openlineage/common/provider/dbt/structured_logs.py
```diff
@@ -464,16 +473,16 @@ def parse_node_finished_event(self, event) -> RunEvent:
...
                     dataset_facets["dataQualityAssertions"] = assertion_facet
-                    inputs = [
+                    inputs.append(
                         Dataset(
                             name=attached_dataset.name,
                             namespace=attached_dataset.namespace,
                             facets=dataset_facets,
                         )
-                    ]
+                    )
```
kacpermuda:
```suggestion
                    inputs.append(
Dataset(
name=attached_dataset.name,
namespace=attached_dataset.namespace,
facets={**attached_dataset.facets, "dataQualityAssertions": assertion_facet},
)
                    )
```

integration/dbt/tests/integration/test_structured_logs.py
```diff
@@ -153,3 +153,136 @@ def test_structured_logs_event_hierarchy(self, dbt_runner, reset_test_server):
...
+                f"got empty inputs. Event: {event.get('job', {}).get('name')}"
+            )
+
+            # At least one input should have dataQualityAssertions facet
+            inputs_with_assertions = [
+                inp for inp in inputs if "dataQualityAssertions" in inp.get("facets", {})
+            ]
+            assert len(inputs_with_assertions) >= 1, (
+                f"Expected at least one input with dataQualityAssertions facet, got none. Inputs: {inputs}"
+            )
```
kacpermuda, line 207: Each event should have inputs and but at least one of them should have dataQualityAssertions facet? Maybe let's check explicitly if all inputs have that facet

kacpermuda (approved):

PR #4401, opened 2026-03-17
feat(spark): Add MongoDB, Iceberg, and Kinesis streaming support for Spark 4.0

anzararshad:
## Summary

This PR adds Spark 4.0 streaming support for **MongoDB reads**, **Iceberg writes**, and **Kinesis reads** by implementing the new constructor pattern introduced in Spark 4.0.

## Motivation

**Problem:** OpenLineage Spark integration doesn't track streaming reads from MongoDB/Kinesis or streaming writes to Iceberg in Spark 4.0, resulting in incomplete lineage graphs for modern lakehouse architectures.

**Common Use Case:**
```
MongoDB Collections (operational data)
    ↓ [Spark 4.0 Streaming]
Iceberg Tables (data lake)
    ↓ [Analytics/ML]
```

Without this PR, the critical streaming ETL step is invisible in lineage tracking.

### 1. MongoDB Streaming Read Support

**File:** `MongoMicroBatchStreamStrategy.java`

Added Spark 4.0 constructor matching the signature pattern used by `KafkaMicroBatchStreamStrategy`:

```java
public MongoMicroBatchStreamStrategy(
    DatasetFactory<OpenLineage.InputDataset> inputDatasetDatasetFactory,
    StructType schema,
    SparkDataStream stream,
    Optional<Offset> offsetOption) {
  super(inputDatasetDatasetFactory, schema, stream, offsetOption);
}
```

**Enhanced Database Extraction:** Extracts database name from MongoDB connection URI when not explicitly provided:

```java
// Extracts "mydb" from "mongodb://user:pass@host:27017/mydb?params"
private String extractDatabaseFromURI(String uri) { ... }
```

### 2. Iceberg Streaming Write Support

**File:** `WriteToDataSourceV2Visitor.java`

Added handler for `org.apache.iceberg.spark.source.SparkWrite.StreamingAppend`:

```java
private static final String ICEBERG_STREAMING_WRITE_CLASS_NAME =
    "org.apache.iceberg.spark.source.SparkWrite.StreamingAppend";
```

**Key Implementation:** Handles Iceberg's inner class structure by accessing the outer `SparkWrite` class via the `this$0` synthetic field to retrieve table and schema information.

### 3. Kinesis Streaming Read Support

**File:** `KinesisMicroBatchStreamStrategy.java`

Added Spark 4.0 constructor following the same pattern as MongoDB and Kafka for consistency.

**Note:** While our production use case focuses on MongoDB → Iceberg, we noticed Kinesis had the same Spark 3.x-only limitation and applied the fix for consistency. Happy to split this into a separate PR if preferred.

### 4. Spark 4.0 Integration

**File:** `StreamingDataSourceV2ScanRelationDatasetBuilder.java`

Registered MongoDB and Kinesis strategies alongside Kafka and Delta.

### Unit Tests (3 new test files)
- ✅ `MongoMicroBatchStreamStrategyTest.java` - Verifies Spark 4.0 constructor
- ✅ `KinesisMicroBatchStreamStrategyTest.java` - Verifies Spark 4.0 constructor
- ✅ `WriteToDataSourceV2VisitorTest.java` - Verifies visitor initialization

### Production Validation

These changes have been running in production tracking lineage for:
- **10+ MongoDB collections** → Iceberg tables
- **Spark 4.0 streaming jobs** with 2-minute microbatches
- **OpenMetadata 1.11.5** integration
- **No performance degradation** observed

## Compatibility

✅ **Spark 4.0** - Primary target (tested)
✅ **Spark 3.x** - Backward compatible (existing constructors preserved)
✅ **MongoDB Connector 10.x** - Tested
✅ **Iceberg 1.4.x** - Tested

## Related Work

- Similar pattern used in existing `KafkaMicroBatchStreamStrategy` for Kafka streaming
- Follows existing `IcebergDataSourceV2RelationDatasetBuilder` patterns for batch Iceberg

## Questions for Reviewers

1. **Kinesis inclusion:** Happy to split Kinesis support into a separate PR if preferred
2. **Additional connectors:** Similar pattern could be applied to other streaming sources if valuable


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/MongoMicroBatchStreamStrategy.java
```diff
@@ -31,18 +44,65 @@ public List<OpenLineage.InputDataset> getInputDatasets() {
...
+      return new ArrayList<>(); // Database is required
+    }
+
+    OpenLineage.InputDataset dataset =
+        datasetFactory.getDataset(collectionName, connectionURL + "/" + databaseName, schema);
+
+    return Arrays.asList(dataset);
+  }
+
+  private String extractDatabaseFromURI(String uri) {
```
mobuchowski, line 149: This breaks for URIs that don't have a database path component.

For a URI like `mongodb://user:pass@host:27017` (no `/mydb` at the end), `lastIndexOf('/')` finds the slash inside `://`, so `afterSlash` becomes `user:pass@host:27017`. That string then gets used as the database name, and the final namespace ends up as:

```
mongodb://user:pass@host:27017/user:pass@host:27017
```

```java
int schemeEnd = uri.indexOf("://");
if (schemeEnd == -1) return null;
int dbSlash = uri.indexOf('/', schemeEnd + 3); // first slash after ://
if (dbSlash == -1) return null;
String afterSlash = uri.substring(dbSlash + 1);
// strip query params as before
```

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/MongoMicroBatchStreamStrategy.java
```diff
@@ -13,15 +13,28 @@
...
 public class MongoMicroBatchStreamStrategy extends StreamStrategy {
+  // Constructor for Spark 3.x
   public MongoMicroBatchStreamStrategy(
       DatasetFactory<OpenLineage.InputDataset> inputDatasetDatasetFactory,
       StreamingDataSourceV2Relation relation) {
     super(inputDatasetDatasetFactory, relation.schema(), relation.stream(), Optional.empty());
   }
 
+  // Constructor for Spark 4.0 (matches KafkaMicroBatchStreamStrategy signature)
+  public MongoMicroBatchStreamStrategy(
```
mobuchowski, line 32: Can we add some logs when it fails to find lineage on the short circuit paths?

mobuchowski (changes requested):

dolfinus:
Why mixin everything into one PR?

anzararshad:
> Why mixin everything into one PR?

Hi @dolfinus,

Good point — I grouped them since they all follow the same Spark 4.0 streaming constructor pattern, but splitting them does make reviews easier.

I can break this into three PRs if you prefer:
- MongoDB streaming support
- Iceberg streaming writes
- Kinesis streaming support

Let me know what works best for you 👍

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/WriteToDataSourceV2Visitor.java
```diff
@@ -101,6 +105,79 @@ public List<OutputDataset> apply(LogicalPlan plan) {
...
+
+    String tableName = tableNameOpt.get();
+    // Extract namespace and name from table identifier
+    // e.g., "hive.bronze.application" -> namespace="hive.bronze", name="application"
+    String[] parts = tableName.split("\\.");
+    String namespace =
+        parts.length > 1
+            ? String.join(".", java.util.Arrays.copyOfRange(parts, 0, parts.length - 1))
+            : "default";
+    String name = parts[parts.length - 1];
```
mobuchowski, line 125: Can we use existing `CatalogUtils3` and/or `SparkTableLineageUtils` to derive the table name/namespace?

anzararshad: @mobuchowski I looked into both options:

 - **`CatalogUtils3`** — This lives in the `spark3` module, which depends on `shared`. Using it from `shared` would create a circular dependency, so it's not directly accessible here.
 - **`SparkTableLineageUtils`** — I couldn't find this class in the codebase — did you mean a different utility?

 As an alternative, I can extract the parsing logic into a private helper method (similar to the pattern in `SqlUtils.java` in the `shared` module), which would clean up the inline string splitting. Would that work, or did you have another approach in mind?

PR #4403, opened 2026-03-20
client/python: make sourceCodeLocation opt-in and reduce subprocess overhead

himakolavennu:
Python client: make `sourceCodeLocation` facet opt-in and reduce git subprocess calls from 4 to 2.

### Meaningful description
Follow-up to [#4398](https://github.com/OpenLineage/OpenLineage/pull/4398) addressing reviewer comments that were left unresolved at merge time.

**`disabled=True` by default** — git autodetect is now opt-in. Users who want the `sourceCodeLocation` facet populated automatically must explicitly set `disabled: false` in their config.
  This avoids surprising behavior for users who don't expect subprocess calls during event emission.

  **Reduce subprocess calls from 4 → 2** — the previous implementation spawned up to 4 separate git subprocesses (one each for SHA, branch, tag, and URL). The new implementation:
  - First checks for a `.git` directory by walking up from CWD (`_find_git_root`). If not in a git repo, no subprocesses are spawned at all.
  - Consolidates SHA + branch + tag into a single `git log -1 --format=%H|%D` call (`_get_git_snapshot`), parsing ref decorations to extract all three values.
  - Keeps a separate `git remote get-url origin` call for the URL (still needed when not explicitly configured).
  - Maximum subprocess count is now 2 (down from 4), and 0 when all values are explicitly configured or when running outside a git repo.

### Checklist


client/python/src/openlineage/client/utils.py
```diff
@@ -57,6 +58,52 @@ def deep_merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any,
...
+
+    Returns the first directory that contains ``.git``, or ``None`` if no git
+    repository is found before reaching the filesystem root.  Checking before
+    spawning any subprocesses lets callers short-circuit entirely when not
+    running inside a git repository.
+    """
+    path = pathlib.Path(start or os.getcwd()).resolve()
+    while True:
+        if (path / ".git").exists():
+            return str(path)
```
dolfinus:
```suggestion
            return os.fspath(path)
```

client/python/src/openlineage/client/client.py
```diff
@@ -541,11 +540,31 @@ def _update_tag_facet(
...
+                branch = auto_branch
+            if tag is None:
+                tag = auto_tag
+
         return {
-            "url": get_git_repo_url(scl.repo_url),
-            "version": get_git_version(scl.version),
-            "branch": get_git_branch(scl.branch),
-            "tag": get_git_tag(scl.tag),
+            "url": get_git_repo_url(scl.repo_url) if (scl.repo_url or in_git_repo) else None,
```
dolfinus: This looks complicated. IMHO, there should be an early return `if not in_git_repo`, and this block can be rewritten as:

```suggestion
            "url": scl.repo_url or get_git_repo_url(),
```

client/python/src/openlineage/client/client.py
```diff
@@ -541,11 +540,31 @@ def _update_tag_facet(
...
+        branch: str | None = scl.branch
+        tag: str | None = scl.tag
+
+        # Cache the git root check so we don't call _find_git_root twice.
+        in_git_repo = _find_git_root() is not None
+
+        # Auto-detect sha/branch/tag in a single subprocess only when inside a
+        # git repo and at least one field still needs to be resolved.
+        if (sha is None or branch is None or tag is None) and in_git_repo:
+            auto_sha, auto_branch, auto_tag = _get_git_snapshot()
```
kacpermuda: Can we please make timeouts configurable as well, since we already have some configs in `self.config.facets.source_code_location`? Default 5 secs may be to much for users, and they could customize it if needed.

client/python/src/openlineage/client/utils.py
```diff
@@ -57,6 +58,52 @@ def deep_merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any,
...
+
+def _get_git_snapshot(timeout: int = 5) -> tuple[str | None, str | None, str | None]:
+    """Return ``(sha, branch, tag)`` from a single ``git log`` call.
+
+    Uses ``%H|%D`` format to get the full commit SHA and all ref decorations in
+    one subprocess instead of three separate calls.  Branch is extracted from
+    ``HEAD -> <name>`` and tag from ``tag: <name>`` in the decoration string.
+    Returns ``(None, None, None)`` on any failure.
+    """
+    raw = _run_git_command(["git", "log", "-1", "--format=%H|%D"], timeout)
```
kacpermuda: Just to confirm, there is no reliable way to get that information from git directory? Do we need to run subprocesses? We run this only when in git_repo, so maybe let's just use that as a source?

client/python/src/openlineage/client/client.py
```diff
@@ -541,11 +540,31 @@ def _update_tag_facet(
     @cached_property
     def _source_code_location(self) -> dict[str, str | None]:
         scl = self.config.facets.source_code_location
```
kacpermuda, line 544: Can you please document that change in python client docs? Similar to what we have [here](https://openlineage.io/docs/client/python/configuration#attaching-environment-variables) for env var facet, maybe we should rename it to "facets" section and then have those two there.

client/python/src/openlineage/client/utils.py
```diff
@@ -85,35 +132,6 @@ def get_git_repo_url(repo_url: str | None = None, timeout: int = 5) -> str | Non
     return _run_git_command(["git", "remote", "get-url", "origin"], timeout)
```
kacpermuda: Same here, no way to get that from git repo files alone, without running git command?

client/python/src/openlineage/client/client.py
mobuchowski: @kacpermuda Can we just give it 0.5 second timeout by default?

kacpermuda: I have a feeling that adding any non-configurable timeout in the client will eventually bite us, but definitely decreasing a timeout is a right move here, we can always make it configurable as a followup if needed

client/python/src/openlineage/client/utils.py
```diff
@@ -57,61 +57,152 @@ def deep_merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any,
...
-            args,
-            capture_output=True,
-            text=True,
-            check=True,
-            cwd=os.getcwd(),
-            timeout=timeout,
-        )
-        return result.stdout.strip() or None
-    except Exception:
+def _find_git_dir(cwd: str | None = None) -> Path | None:
```
dolfinus: These new functions probably should be moved somewhere to `client/python/src/openlineage/client/git.py`

client/python/src/openlineage/client/client.py
```diff
@@ -539,14 +540,30 @@ def _update_tag_facet(
...
+                sha = get_git_version(git_dir)
+            if branch is None:
+                branch = get_git_branch(git_dir)
+            if tag is None:
+                tag = get_git_tag(git_dir)
+
+        if url is None:
+            return None
+
+        return {"url": url, "version": sha, "branch": branch, "tag": tag}
```
kacpermuda, line 568: This looks way better now, without subprocesses, thanks ! Do you mind also adding a try/except on this (here or even better - below in `add_source_code_location_facet`), so it does not break event emission when f.e. we have no permission to read git directory or there is other file reading error? We can do a broad except Exception with a warning level log to the user. Thanks !

client/python/tests/test_source_code_location.py
```diff
@@ -21,111 +31,316 @@ def _no_git_autodetect():
...
-# get_git_version / get_git_branch / get_git_tag
+# get_git_branch
 # ---------------------------------------------------------------------------
-class TestGetGitVersion:
-    def test_explicit_value_returned_as_is(self):
-        assert get_git_version(version="deadbeef") == "deadbeef"
+class TestGetGitBranch:
+    def test_symbolic_ref(self, tmp_path):
+        git_dir = tmp_path / ".git"
+        git_dir.mkdir()
```
dolfinus, line 177: nit: creating a fake git repo can be moved to a fixture

PR #4430, opened 2026-04-01
Add Paimon dataset naming support

jsingh-yelp:
Add Paimon dataset naming support, this follows the naming and namespace convention from native lineage support provided by paimon/flink.

### DIFF
--- a/client/python/src/openlineage/client/naming/dataset.py
+++ b/client/python/src/openlineage/client/naming/dataset.py
@@ -484,6 +484,22 @@ class WASBS(DatasetNaming):
         return self.object_key
 
 
+@attr.define
+class Paimon(DatasetNaming):
+    """Naming implementation for Paimon."""
+
+    warehouse_path: str = attr.field(validator=_check_not_empty)
+    catalog: str = attr.field(validator=_check_not_empty)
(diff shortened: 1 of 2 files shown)


client/python/src/openlineage/client/naming/dataset.py
```diff
@@ -484,6 +484,22 @@ def get_name(self) -> str:
         return self.object_key
 
 
+@attr.define
+class Paimon(DatasetNaming):
+    """Naming implementation for Paimon."""
+
+    warehouse_path: str = attr.field(validator=_check_not_empty)
+    catalog: str = attr.field(validator=_check_not_empty)
```
dolfinus, line 492: According to documentation:
https://paimon.apache.org/docs/master/concepts/catalog/

catalog is just an arbitrary name for the Hive/REST/JDBC/Filesystem location which can be used by Flink/Spark/etc. Different applications can use different names but point to the same location, so it's not an identifying information.

Also warehouse is a storage for data files, and catalog is a storage for metadata (table schema, table name, properties and so on). In terms of OpenLineage these are dataset symlinks, just like Hive or Iceberg tables.

Issue #4385, opened 2026-03-10
[FEATURE] Bring Your Own OpenLineage + terraform


tnazarew:
## BYOL

There is an idea called Bring Your Own Lineage (BYOL).
The purpose of it is to enable the users to declare lineage entities and supply them to their Catalogs.
That fills the gap left by the components which don't have runtime lineage support.

Databricks already has something like that [here](https://docs.databricks.com/aws/en/data-governance/unity-catalog/external-lineage).

In case of OpenLineage it would greatly lower the barrier of entry, users would not be limited to components integrated OpenLineage.
That + OpenLineages portability, would enable user to define data once in OpenLineage format and use it across different catalogs.

From implementation perspective, it would be more or less what we already have (client API and transports) but with validation layer to restrict what can be added.

## Terraform BYOOL Provider

I would like to create the `Bring Your Own OpenLineage` with Terraform. Terraform is a popular tool for
controlling the state of the environment. Each terraform resource is a managed entity, which
means that terraform keeps track of its state and can create, update or delete it.
Terraform has an API for creating custom providers and resources so lineage entities could be defined and managed as such.

## OpenLineage Terraform Provider

-------------------------------------

NOTE: 
I already created a working version of [OpenLineage Terraform Dataplex provider](https://github.com/tnazarew/openlineage-terraform-dataplex-provider). The structure of config is simplified for readability and the fields not used by the Dataplex were skipped but more details on how *exactly* I might implement the feature are here. 

-------------------------------------

## Overview

BYOOL provider would implement 3 types of resources for static lineage

1. dataset events
2. job events
3. run-ish events (if the consumer doesn't support static lineage events, we can put job event in the run wrapper without any runtime facets)

When changes are applied, the provider would read the configuration, validate it, construct the OpenLineage event and emit it. There are some complications, however.

## Issues

For each resource Terraform has few states, lets call them:

- `config` is the desired state of the resource defined in the terraform configuration file
- `previous` is the state of the resource stored in the terraform state file
- `current` is the state of the resource in the environment

When `apply` is executed, Terraform compares `current` and `previous` to check for external changes, then compares `config` and `previous` to see what needs to be modified. 

In cases of some components all the states have 1-1 structure, it would not be a case for OpenLineage as state of the consumer can't be 1-1 mapped to OL.
[...]

PR #4437, opened 2026-04-03
Terraform base resource

tnazarew:

Create base for BYOOL Terraform implementation.

### Meaningful description

Create base for BYOOL Terraform implementation. Module will consist of:
1. wrapper for defining terraform schema for OL resources config
2. models for OL part of terraform resource state
3. event generation logic
4. `Capability` interface to define facets handled by the consumer 

This will provide a generic API for creating OpenLineage resources in terraform providers. The configuration created that way will be reusable between different implementations. The `Capacity` interface will enable user to enable or disable certain facets depending on if consumer is handling them. That way if the facet is enabled, it's value will be read from config as optional, used in the event and stored in the internal terraform state. In case it's disabled, it will be ignored by the terraform. That way there will be no unused fields stored in the state but users should be able to seamlessly move between OL-based resources without changes to the config.

The logic of sending the events to consumer and handling the actual resource state will be on the consumer-implementation side.

Added `terraform-provider-openlineage-file` as example implementation of the provider in PR #4438
related issue: #4385 

### Checklist


byool/terraform/openlineage-base-resource/ol/schema_generator.go
```diff
@@ -0,0 +1,682 @@
...
+			},
+		},
+	}
+}
+
+// ── Job facet blocks ──────────────────────────────────────────────────────────
+
+func jobTypeBlock() schema.SingleNestedBlock {
+	return schema.SingleNestedBlock{
+		Description: "Job type classification (facets.JobType)",
```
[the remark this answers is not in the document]
tnazarew: there is no `Optional` for `SingleNestedBlock` or `ListNestedBlock`

byool/terraform/openlineage-base-resource/ol/schema_generator.go
```diff
@@ -0,0 +1,682 @@
...
+// stubListBlock returns a copy of b where every leaf attribute in its nested
+// object is Optional+Computed and every nested block is also stubbed.
+func stubListBlock(b schema.ListNestedBlock) schema.ListNestedBlock {
+	return schema.ListNestedBlock{
+		Description: b.Description,
+		NestedObject: schema.NestedBlockObject{
+			Attributes: stubAttributes(b.NestedObject.Attributes),
+			Blocks:     stubBlocks(b.NestedObject.Blocks),
+		},
+	}
```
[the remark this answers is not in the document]
tnazarew: same as above, no `Optional` here

byool/terraform/openlineage-base-resource/ol/ol_models.go
```diff
@@ -0,0 +1,293 @@
...
+// Wraps the list of dataset tags, mirroring TagsJobFacetModel.
+type TagsDatasetFacetModel struct {
+	Tag []TagsDatasetModel `tfsdk:"tag"`
+}
+
+// SymlinksDatasetFacetModel — facets.SymlinksDatasetFacet
+// Wraps the list of alternate dataset identifiers. Uses a nested "symlink" list
+// so the outer block name ("symlinks") and the element name ("symlink") are distinct.
+type SymlinksDatasetFacetModel struct {
+	Identifier []IdentifierModel `tfsdk:"identifiers"`
```
kacpermuda, line 283: The `tfsdk` tag on `Identifier` here is `"identifiers"` (plural), but the schema declares the corresponding block as `"identifier"` (singular) at `schema_generator.go:316`, and the README architecture table (`README.md:70`) claims it's `"symlink"`. **Three different names**, so the framework will fail to bind state to this model.

Pick one and align all three sites. Suggest `"identifier"` since the element model is `IdentifierModel` and the schema block already uses that.

byool/terraform/openlineage-base-resource/ol/ol_models.go
```diff
@@ -0,0 +1,293 @@
...
+type ColumnLineageDatasetModel struct {
+	Fields  []ColumnLineageFieldModel          `tfsdk:"fields"`  // field-level mappings
+	Dataset []ColumnLineageDatasetElementModel `tfsdk:"dataset"` // dataset-level mappings
+}
+
+// ColumnLineageFieldModel — facets.FieldValue (keyed by output column name)
+// Maps a single output column to its contributing input fields.
+type ColumnLineageFieldModel struct {
+	Name        types.String      `tfsdk:"name"`         // output column name — becomes the map key
+	InputFields []InputFieldModel `tfsdk:"input_fields"` // which input columns feed this output column
```
kacpermuda, line 244: Tag here is `tfsdk:"input_fields"` (plural), but the schema declares the corresponding block as `"input_field"` (singular) at `schema_generator.go:543`. These must match — the framework binds state to model fields by `tfsdk` tag, so this will fail at provider startup.

byool/terraform/openlineage-base-resource/ol/ol_models.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// InputFieldModel — facets.DatasetElement (inside FieldValue)
+// A single input column contributing to an output column.
+type InputFieldModel struct {
+	Namespace       types.String          `tfsdk:"namespace"`
+	Name            types.String          `tfsdk:"name"`
+	Field           types.String          `tfsdk:"field"`
+	Transformations []TransformationModel `tfsdk:"transformations"`
```
kacpermuda, line 253: Tag here is `tfsdk:"transformations"` (plural) but the schema block is `"transformation"` (singular) at `schema_generator.go:553`. Same plural/singular drift as the `InputFields` issue above.

Note that `ColumnLineageDatasetElementModel.Transformations` on line 262 uses singular `tfsdk:"transformation"` and *does* match its block — so the inconsistency is just on `InputFieldModel` here. Worth picking one form everywhere.

byool/terraform/openlineage-base-resource/ol/schema_generator.go
```diff
@@ -0,0 +1,713 @@
...
+							"field":     optionalString("Input column name"),
+						},
+						Blocks: map[string]schema.Block{
+							"transformation": transformationListBlock(),
+						},
+						Validators: requireFields("namespace", "name", "field"),
+					},
+				},
+			},
+			Validators: requireFields("name", "input_fields"),
```
kacpermuda, line 559: The block above is named `"input_field"` (singular, line 543), but this validator references `"input_fields"` (plural). `objectvalidator.AlsoRequires` will either error during schema validation or silently pass on a name that doesn't exist — neither is the intended behavior.

Same fix as the model tag mismatch in `ol_models.go:244`: pick one spelling and use it everywhere.

byool/terraform/openlineage-base-resource/ol/schema_generator.go
```diff
@@ -0,0 +1,713 @@
...
+			Attributes: map[string]schema.Attribute{
+				"namespace": optionalString("Input dataset namespace"),
+				"name":      optionalString("Input dataset name"),
+				"field":     optionalString("Output field this dataset contributes to"),
+			},
+			Blocks: map[string]schema.Block{
+				"transformation": transformationListBlock(),
+			},
+			Validators: requireFields(
+				"namespace", "name", "field", "transformations"),
```
kacpermuda, line 578: The block above is named `"transformation"` (singular, line 575), but this validator references `"transformations"` (plural). Same drift as `columnLineageFieldsBlock` above.

byool/terraform/openlineage-base-resource/ol/schema_generator.go
```diff
@@ -0,0 +1,713 @@
...
+			"type":          optionalString("Catalog type e.g. hive"),
+			"name":          optionalString("Catalog name"),
+			"metadata_uri":  optionalString("e.g. hive://localhost:9083"),
+			"warehouse_uri": optionalString("e.g. hdfs://localhost/warehouse"),
+			"source":        optionalString("Source system e.g. spark"),
+			"catalog_properties": schema.MapAttribute{
+				Optional:    true,
+				Description: "Additional catalog-specific properties as key-value pairs",
+				ElementType: types.StringType,
+				Default:     mapdefault.StaticValue(types.MapNull(types.StringType)),
```
kacpermuda, line 487: **This will reject the provider at startup.** `Default` requires `Computed: true` in `terraform-plugin-framework` v1 — the framework's own schema validation (`(*resource.Resource).ValidateImplementation`) returns `Schema Using Attribute Default For Non-Computed Attribute` when this combination is detected.

Nothing currently catches it because no test exercises framework schema binding (see separate comment on `schema_generator_test.go`).

```suggestion
			"catalog_properties": schema.MapAttribute{
				Optional:    true,
				Computed:    true,
				Description: "Additional catalog-specific properties as key-value pairs",
				ElementType: types.StringType,
				Default:     mapdefault.StaticValue(types.MapNull(types.StringType)),
```

byool/terraform/openlineage-base-resource/ol/schema_generator.go
```diff
@@ -0,0 +1,713 @@
...
+		Validators:  []validator.List{listvalidator.SizeAtLeast(1)},
+		NestedObject: schema.NestedBlockObject{
+			Attributes: map[string]schema.Attribute{
+				"type":        optionalString("DIRECT or INDIRECT"),
+				"subtype":     optionalString("e.g. IDENTITY, AGGREGATION, FILTER"),
+				"description": optionalString("Human-readable transformation description"),
+				"masking": schema.BoolAttribute{
+					Optional:    true,
+					Description: "True if this transformation masks/anonymises data",
+					Default:     booldefault.StaticBool(false),
```
kacpermuda, line 595: Same problem as `catalog_properties`: `Default` requires `Computed: true` per the plugin-framework v1 rules. Provider startup will reject the schema.

```suggestion
				"masking": schema.BoolAttribute{
					Optional:    true,
					Computed:    true,
					Description: "True if this transformation masks/anonymises data",
					Default:     booldefault.StaticBool(false),
				},
```

byool/terraform/openlineage-base-resource/ol/schema_generator_test.go
```diff
@@ -0,0 +1,451 @@
...
+ * Copyright 2018-2026 contributors to the OpenLineage project
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package ol
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
```
kacpermuda, line 11: **Important test gap.** None of the tests in this package exercise the actual `terraform-plugin-framework` schema binding. They construct Go structs directly and verify field shapes; they don't call:

- `(*resource.Resource).ValidateImplementation(...)` — would catch the `Default`-without-`Computed` issues on `catalog_properties` and `masking`,
- model decoding against the generated schema — would catch the `tfsdk` tag vs schema block name mismatches in `ol_models.go:244,253,283`,
- or any acceptance-style test that loads the schema with a sample config.

A single test that does e.g. `(&BaseJobResource{}).Schema(ctx, ...)` and runs the framework's schema validation would surface most of the bugs flagged on this PR as fast feedback. Right now they only manifest at provider startup against a real Terraform run, which is a long debug loop for downstream provider authors.

byool/terraform/openlineage-base-resource/ol/event_builder.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// producer is the URI that identifies this provider as the source of OL events.
+// In OpenLineage, every event carries a "producer" field so consumers know
+// which system generated the lineage data.
+const producer = "https://github.com/OpenLineage/openlineage/byool/terraform"
```
kacpermuda, line 22: Hard-coded for everyone. Per the [OL spec](https://openlineage.io/spec/), `producer` should identify the *specific* emitter — but every provider built on this base will report `https://github.com/OpenLineage/openlineage/byool/terraform`, indistinguishable in lineage data.

Suggest making it a constructor parameter:

```go
func NewJobEventBuilder(diags *diag.Diagnostics, cap JobCapability, producer string) *OpenLineageEventBuilder { ... }
```

so a Marquez-targeted provider can self-identify as e.g. `terraform-provider-openlineage-marquez/v1.0.0`. This also lets you drop the package-level constant entirely.

byool/terraform/openlineage-base-resource/ol/event_builder.go
```diff
@@ -0,0 +1,233 @@
...
+	return &OpenLineageEventBuilder{Diagnostics: diags, cap: cap.capability}
+}
+
+// ── Package-level convenience wrappers ───────────────────────────────────────
+
+// BuildRunEvent is a package-level convenience wrapper around
+// NewJobEventBuilder(...).BuildRunEvent(...).
+func BuildRunEvent(data *JobResourceModel, cap JobCapability) *openlineage.RunEvent {
+	var diags diag.Diagnostics
+	return NewJobEventBuilder(&diags, cap).BuildRunEvent(data)
```
kacpermuda, line 48: These wrappers exist alongside the proper `New*EventBuilder` constructors, are shorter to type, and **silently discard diagnostics**. The README warns about this, but in practice consumers will reach for the obvious symbol — and the required-field errors from `requireString` end up in `/dev/null` while a half-built event with empty strings still gets emitted.

Two cleaner options:
1. Delete these wrappers entirely; force consumers through the builder constructor (only two extra lines).
2. Rename to make the danger obvious (`BuildRunEventDiscardingErrors`, etc.).

Option 1 is my preference — the constructor pattern is barely longer and surfaces real errors.

byool/terraform/openlineage-base-resource/ol/event_builder.go
```diff
@@ -0,0 +1,233 @@
...
+	var diags diag.Diagnostics
+	return NewDatasetEventBuilder(&diags, cap).BuildDatasetEvent(data)
+}
+
+// ── Top-level event builders ──────────────────────────────────────────────────
+
+// BuildRunEvent wraps a JobEvent with run-specific fields (event type + run ID).
+func (e *OpenLineageEventBuilder) BuildRunEvent(data *JobResourceModel) *openlineage.RunEvent {
+	jobEvent := e.BuildJobEvent(data)
+	runID := uuid.New()
```
kacpermuda, line 63: Each call generates a brand-new `runID` via `uuid.New()`, and `resourceBase.Update` re-emits via `ConsumerEmit` on every `terraform apply`. So a single declarative resource modified twice over three `apply` runs looks downstream like **three distinct runs of the same job** — `TestBuildRunEvent_GeneratesUniqueRunIDs` confirms this is by design.

That's a notable lineage-semantics choice that probably surprises downstream consumers (Marquez, Atlas, Datakin, etc.) which treat distinct `runId`s as distinct executions. Worth documenting loudly or revisiting:

- Persist `runID` in state and reuse it across operations (resource-as-run model), or
- Emit `EventTypeStart` on Create and `EventTypeComplete` on Update so each apply produces a paired event lifecycle, not an unrelated complete-only event.

Either is more in line with how producers like Spark/dbt model runs.

byool/terraform/openlineage-base-resource/ol/base_resource.go
```diff
@@ -0,0 +1,187 @@
...
+	// Returns false if it no longer exists (triggers re-create).
+	ConsumerRead(ctx context.Context, model any) (exists bool, diags diag.Diagnostics)
+
+	// ConsumerDelete removes the entity from the consumer.
+	ConsumerDelete(ctx context.Context, model any) diag.Diagnostics
+
+	// BaseSchema returns the capability-driven base schema for this resource type.
+	// Implemented by BaseJobResource (GenerateJobSchema) and BaseDatasetResource
+	// (GenerateDatasetSchema) — consumers inherit it through embedding.
+	BaseSchema() schema.Schema
```
kacpermuda, line 55: `BaseSchema()` is part of the public `ResourceBackend` interface, but the README at `README.md:39` says it's *"**not** implemented by consumers"* — it's promoted automatically when the consumer embeds `BaseJobResource` or `BaseDatasetResource`.

A new contributor reading just this interface has no way to know they'd be sabotaging the design by implementing it directly. Suggest a stronger doc comment that says so explicitly:

```suggestion
	// BaseSchema returns the capability-driven base schema for this resource type.
	// Consumers MUST NOT implement this directly. Embedding BaseJobResource or
	// BaseDatasetResource provides the correct implementation through method promotion
	// (GenerateJobSchema / GenerateDatasetSchema, dispatched via Capability()).
	BaseSchema() schema.Schema
```

byool/terraform/openlineage-base-resource/ol/capability.go
```diff
@@ -0,0 +1,155 @@
...
+func EmptyJobCapability() JobCapability {
+	return JobCapability{capability{
+		jobEnabled:     map[JobFacet]bool{},
+		datasetEnabled: map[DatasetFacet]bool{},
+	}}
+}
+
+// WithFacetEnabled returns a new JobCapability with the given job facets enabled.
+// To control which dataset facets are emitted on inputs/outputs, use WithDatasetFacetEnabled.
+func (c JobCapability) WithFacetEnabled(facets ...JobFacet) JobCapability {
```
kacpermuda, line 110: Stylistic, but `cap` is a Go built-in — every use of `cap` as a parameter name shadows it inside the function. The pattern repeats throughout `schema_generator.go` (`GenerateJobSchema`, `addBlock`, `inputsBlock`, `outputsBlock`, etc.) and the README's example code.

Subtle hazard if anyone ever wants `cap(slice)` in the same scope. Suggest `c` or `caps`.

byool/terraform/openlineage-base-resource/README.md
```diff
@@ -0,0 +1,349 @@
...
+- `DatasetModel` — dataset identity + all dataset facet blocks (shared by inputs, outputs, and standalone dataset resources)
+- individual facet structs (job facets, dataset facets, column lineage)
+
+**Facet wrapper models** — three facets use an outer `SingleNestedBlock` wrapping an inner list, consistent with the `OwnershipJobModel / owners` pattern:
+
+| Model | Outer block | Inner list key | Element model |
+|---|---|---|---|
+| `TagsJobFacetModel` | `tags` | `tag` | `TagsJobModel` |
+| `TagsDatasetFacetModel` | `tags` | `tag` | `TagsDatasetModel` |
+| `SymlinksDatasetFacetModel` | `symlinks` | `symlink` | `IdentifierModel` |
```
kacpermuda, line 70: Per the comment on `ol_models.go:283`, this row is out of sync with the code: the schema declares the inner block as `"identifier"`, the model tag says `"identifiers"`, and this table claims `"symlink"`.

After the symlinks tag/block mismatch is fixed, this table should reflect whichever name wins so docs and code agree.

kacpermuda (commented):
Testing out reviewing with claude, so treat this as possibly wrong advice, but figured out it might be a good PR to verify how it handles it.

PR #4457, opened 2026-04-08
feat: Extend DataQualityAssertionsDatasetFacet with new information

kacpermuda:
feat: Extend DataQualityAssertionsDatasetFacet with new information

### Meaningful description
After recent introduction of TestRunFacet in #4448, we can extend the information gathered by DataQualityAssertionsDatasetFacet, which can be seen as the same facet but on Dataset level.

### Checklist


kacpermuda:
@mobuchowski lmk your thoughts, I treat `assertion` at dataset facet as an equivalent to `type` from TestRunFacet, but we can also treat it as `name`, and add `type` separately, then it would match to TestRunFacet almost 1:1. Now the `name` and `status` is required in TestRunFacet but `assertion` (type of test) and `success` required for DQDatasetFacet (so treating `assertion` as `name` instead of `type` make sense, current producers/consumers may use `assertion` for something, but not sure if the contract was that it's going to be a type and not an arbitrary name)

mobuchowski (approved):

PR #4465, opened 2026-04-13
python: Add support for dataset normalization in Python Client

kchledowski:

Add support for dataset normalization in Python Client

### Meaningful description

Implementation of dataset normalization (see #3996) in the Python client.
In Java client this is implemented as a kind of util that can be used in different integrations (currently used only in Spark integration).
Here the approach is for the client to make changes to the event before emitting it.
By default, the normalization is disabled. It has to be enabled with config to take effect.

### Checklist

### DIFF
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -150,7 +150,7 @@ class OpenLineageClient:
             if _filter:
                 self._filters.append(_filter)
 
-        self._dataset_reducer = DatasetReducer(self.config.dataset)
+        self._dataset_reducer: DatasetReducer | None = None
 
     @classmethod
     def from_environment(cls: type[_T]) -> _T:
@@ -612,6 +612,9 @@ class OpenLineageClient:
         return event
 
     def reduce_datasets(self, event: Event) -> Event:
+        if not self._dataset_reducer:
+            self._dataset_reducer = DatasetReducer(self.config.dataset)
+
         if not isinstance(event, event_v2.RunEvent):
             return event
         if event.inputs:
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast
 import attr
 import yaml
 from openlineage.client import constants, event_v2
+from openlineage.client.dataset import DatasetConfig, DatasetNormalizer
 from openlineage.client.facet_v2 import (
     environment_variables_run,
     source_code_location_job,
@@ -68,6 +69,7 @@ class OpenLineageConfig:
     facets: FacetsConfig = attr.field(factory=FacetsConfig)
     filters: list[FilterConfig] = attr.field(factory=list)
     tags: TagsConfig = attr.field(factory=TagsConfig)
+    dataset: DatasetConfig = attr.field(factory=DatasetConfig)
 
     @classmethod
     def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
@@ -90,6 +92,18 @@ class OpenLineageConfig:
                 job=params["tags"].get("job", {}),
                 run=params["tags"].get("run", {}),
             )
+        if "dataset" in params:
+
+            def split_into_list(value: str | list[str]) -> list[str]:
+                if isinstance(value, list):
+                    return value
+                return [item.strip() for item in value.split(";") if item.strip()]
+
+            ds = params["dataset"]
+            ds["disabled_trimmers"] = split_into_list(ds.get("disabled_trimmers", []))
+            ds["extra_trimmers"] = split_into_list(ds.get("extra_trimmers", []))
+            config.dataset = DatasetConfig(**ds)
+
         return config
 
 
@@ -142,6 +156,8 @@ class OpenLineageClient:
             if _filter:
                 self._filters.append(_filter)
 
+        self._dataset_normalizer = DatasetNormalizer(self.config.dataset)
+
     @classmethod
     def from_environment(cls: type[_T]) -> _T:
         warnings.warn(
@@ -183,6 +199,8 @@ class OpenLineageClient:
         event = self.add_environment_facets(event)
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
+        if self.config.dataset.normalization_enabled:
+            event = self.normalize_datasets(event)
 
         if log.isEnabledFor(logging.DEBUG):
             val = Serde.to_json(event).encode("utf-8")
@@ -594,4 +612,13 @@ class OpenLineageClient:
                 tag=scl["tag"],
                 pullRequestNumber=scl["pullRequestNumber"],
             )
+        return event
+
+    def normalize_datasets(self, event: Event) -> Event:
+        if not isinstance(event, (RunEvent, event_v2.RunEvent)):
+            return event
+        if event.inputs:
+            event.inputs[:] = self._dataset_normalizer.normalize_inputs(event.inputs)
+        if event.outputs:
+            event.outputs[:] = self._dataset_normalizer.normalize_outputs(event.outputs)
         return event
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -39,7 +39,7 @@ from openlineage.client.transport import (
 )
 from openlineage.client.transport.http import HttpConfig, HttpTransport
 from openlineage.client.transport.noop import NoopConfig, NoopTransport
-from openlineage.client.utils import deep_merge_dicts
+from openlineage.client.utils import deep_merge_dicts, split_into_list
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore", DeprecationWarning)
@@ -93,13 +93,7 @@ class OpenLineageConfig:
                 run=params["tags"].get("run", {}),
             )
         if "dataset" in params:
-
-            def split_into_list(value: str | list[str]) -> list[str]:
-                if isinstance(value, list):
-                    return value
-                return [item.strip() for item in value.split(";") if item.strip()]
-
-            ds = params["dataset"]
+            ds = dict(params["dataset"])
             ds["disabled_trimmers"] = split_into_list(ds.get("disabled_trimmers", []))
             ds["extra_trimmers"] = split_into_list(ds.get("extra_trimmers", []))
             config.dataset = DatasetConfig(**ds)
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -194,6 +194,9 @@ class OpenLineageClient:
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
         if self.config.dataset.reducing_enabled:
+            # Reduce datasets to their canonical form (e.g. "/data/events" instead of
+            # "/data/events/dt=2025-09-01"). All subsequent processing (e.g. event filtering)
+            # will work on these modified values.
             event = self.reduce_datasets(event)
 
         if log.isEnabledFor(logging.DEBUG):
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast
 import attr
 import yaml
 from openlineage.client import constants, event_v2
-from openlineage.client.dataset import DatasetConfig, DatasetNormalizer
+from openlineage.client.dataset import DatasetConfig, DatasetReducer
 from openlineage.client.facet_v2 import (
     environment_variables_run,
     source_code_location_job,
@@ -150,7 +150,7 @@ class OpenLineageClient:
             if _filter:
                 self._filters.append(_filter)
 
-        self._dataset_normalizer = DatasetNormalizer(self.config.dataset)
+        self._dataset_reducer = DatasetReducer(self.config.dataset)
 
     @classmethod
     def from_environment(cls: type[_T]) -> _T:
@@ -193,8 +193,8 @@ class OpenLineageClient:
         event = self.add_environment_facets(event)
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
-        if self.config.dataset.normalization_enabled:
-            event = self.normalize_datasets(event)
+        if self.config.dataset.reducing_enabled:
+            event = self.reduce_datasets(event)
 
         if log.isEnabledFor(logging.DEBUG):
             val = Serde.to_json(event).encode("utf-8")
@@ -608,11 +608,11 @@ class OpenLineageClient:
             )
         return event
 
-    def normalize_datasets(self, event: Event) -> Event:
+    def reduce_datasets(self, event: Event) -> Event:
         if not isinstance(event, event_v2.RunEvent):
             return event
         if event.inputs:
-            event.inputs[:] = self._dataset_normalizer.normalize_inputs(event.inputs)
+            event.inputs[:] = self._dataset_reducer.reduce_inputs(event.inputs)
         if event.outputs:
-            event.outputs[:] = self._dataset_normalizer.normalize_outputs(event.outputs)
+            event.outputs[:] = self._dataset_reducer.reduce_outputs(event.outputs)
         return event
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -615,7 +615,7 @@ class OpenLineageClient:
         return event
 
     def normalize_datasets(self, event: Event) -> Event:
-        if not isinstance(event, (RunEvent, event_v2.RunEvent)):
+        if not isinstance(event, event_v2.RunEvent):
             return event
         if event.inputs:
             event.inputs[:] = self._dataset_normalizer.normalize_inputs(event.inputs)
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -617,8 +617,7 @@ class OpenLineageClient:
 
         if not isinstance(event, event_v2.RunEvent):
             return event
-        if event.inputs:
-            event.inputs[:] = self._dataset_reducer.reduce_inputs(event.inputs)
-        if event.outputs:
-            event.outputs[:] = self._dataset_reducer.reduce_outputs(event.outputs)
-        return event
+
+        new_inputs = self._dataset_reducer.reduce_inputs(event.inputs) if event.inputs else event.inputs
+        new_outputs = self._dataset_reducer.reduce_outputs(event.outputs) if event.outputs else event.outputs
+        return attr.evolve(event, inputs=new_inputs, outputs=new_outputs)
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -654,8 +654,7 @@ class OpenLineageClient:
 
         if not isinstance(event, event_v2.RunEvent):
             return event
-        if event.inputs:
-            event.inputs[:] = self._dataset_reducer.reduce_inputs(event.inputs)
-        if event.outputs:
-            event.outputs[:] = self._dataset_reducer.reduce_outputs(event.outputs)
-        return event
+
+        new_inputs = self._dataset_reducer.reduce_inputs(event.inputs) if event.inputs else event.inputs
+        new_outputs = self._dataset_reducer.reduce_outputs(event.outputs) if event.outputs else event.outputs
+        return attr.evolve(event, inputs=new_inputs, outputs=new_outputs)
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -149,7 +149,7 @@ class OpenLineageClient:
             if _filter:
                 self._filters.append(_filter)
 
-        self._dataset_reducer = DatasetReducer(self.config.dataset)
+        self._dataset_reducer: DatasetReducer | None = None
 
     @classmethod
     def from_environment(cls: type[_T]) -> _T:
@@ -649,6 +649,9 @@ class OpenLineageClient:
         return event
 
     def reduce_datasets(self, event: Event) -> Event:
+        if not self._dataset_reducer:
+            self._dataset_reducer = DatasetReducer(self.config.dataset)
+
         if not isinstance(event, event_v2.RunEvent):
             return event
         if event.inputs:
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -193,6 +193,9 @@ class OpenLineageClient:
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
         if self.config.dataset.reducing_enabled:
+            # Reduce datasets to their canonical form (e.g. "/data/events" instead of
+            # "/data/events/dt=2025-09-01"). All subsequent processing (e.g. event filtering)
+            # will work on these modified values.
             event = self.reduce_datasets(event)
 
         if log.isEnabledFor(logging.DEBUG):
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast
 import attr
 import yaml
 from openlineage.client import constants, event_v2
-from openlineage.client.dataset import DatasetConfig, DatasetNormalizer
+from openlineage.client.dataset import DatasetConfig, DatasetReducer
 from openlineage.client.facet_v2 import (
     environment_variables_run,
     source_code_location_job,
@@ -149,7 +149,7 @@ class OpenLineageClient:
             if _filter:
                 self._filters.append(_filter)
 
-        self._dataset_normalizer = DatasetNormalizer(self.config.dataset)
+        self._dataset_reducer = DatasetReducer(self.config.dataset)
 
     @classmethod
     def from_environment(cls: type[_T]) -> _T:
@@ -192,8 +192,8 @@ class OpenLineageClient:
         event = self.add_environment_facets(event)
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
-        if self.config.dataset.normalization_enabled:
-            event = self.normalize_datasets(event)
+        if self.config.dataset.reducing_enabled:
+            event = self.reduce_datasets(event)
 
         if log.isEnabledFor(logging.DEBUG):
             val = Serde.to_json(event).encode("utf-8")
@@ -645,11 +645,11 @@ class OpenLineageClient:
             )
         return event
 
-    def normalize_datasets(self, event: Event) -> Event:
+    def reduce_datasets(self, event: Event) -> Event:
         if not isinstance(event, event_v2.RunEvent):
             return event
         if event.inputs:
-            event.inputs[:] = self._dataset_normalizer.normalize_inputs(event.inputs)
+            event.inputs[:] = self._dataset_reducer.reduce_inputs(event.inputs)
         if event.outputs:
-            event.outputs[:] = self._dataset_normalizer.normalize_outputs(event.outputs)
+            event.outputs[:] = self._dataset_reducer.reduce_outputs(event.outputs)
         return event
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -652,7 +652,7 @@ class OpenLineageClient:
         return event
 
     def normalize_datasets(self, event: Event) -> Event:
-        if not isinstance(event, (RunEvent, event_v2.RunEvent)):
+        if not isinstance(event, event_v2.RunEvent):
             return event
         if event.inputs:
             event.inputs[:] = self._dataset_normalizer.normalize_inputs(event.inputs)
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING, Any, TypeVar, cast
 import attr
 import yaml
 from openlineage.client import constants, event_v2
+from openlineage.client.dataset import DatasetConfig, DatasetNormalizer
 from openlineage.client.facet_v2 import (
     environment_variables_run,
     source_code_location_job,
@@ -67,6 +68,7 @@ class OpenLineageConfig:
     facets: FacetsConfig = attr.field(factory=FacetsConfig)
     filters: list[FilterConfig] = attr.field(factory=list)
     tags: TagsConfig = attr.field(factory=TagsConfig)
+    dataset: DatasetConfig = attr.field(factory=DatasetConfig)
 
     @classmethod
     def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
@@ -89,6 +91,18 @@ class OpenLineageConfig:
                 job=params["tags"].get("job", {}),
                 run=params["tags"].get("run", {}),
             )
+        if "dataset" in params:
+
+            def split_into_list(value: str | list[str]) -> list[str]:
+                if isinstance(value, list):
+                    return value
+                return [item.strip() for item in value.split(";") if item.strip()]
+
+            ds = params["dataset"]
+            ds["disabled_trimmers"] = split_into_list(ds.get("disabled_trimmers", []))
+            ds["extra_trimmers"] = split_into_list(ds.get("extra_trimmers", []))
+            config.dataset = DatasetConfig(**ds)
+
         return config
 
 
@@ -141,6 +155,8 @@ class OpenLineageClient:
             if _filter:
                 self._filters.append(_filter)
 
+        self._dataset_normalizer = DatasetNormalizer(self.config.dataset)
+
     @classmethod
     def from_environment(cls: type[_T]) -> _T:
         warnings.warn(
@@ -182,6 +198,8 @@ class OpenLineageClient:
         event = self.add_environment_facets(event)
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
+        if self.config.dataset.normalization_enabled:
+            event = self.normalize_datasets(event)
 
         if log.isEnabledFor(logging.DEBUG):
             val = Serde.to_json(event).encode("utf-8")
@@ -631,4 +649,13 @@ class OpenLineageClient:
                 tag=scl["tag"],
                 pullRequestNumber=scl["pullRequestNumber"],
             )
+        return event
+
+    def normalize_datasets(self, event: Event) -> Event:
+        if not isinstance(event, (RunEvent, event_v2.RunEvent)):
+            return event
+        if event.inputs:
+            event.inputs[:] = self._dataset_normalizer.normalize_inputs(event.inputs)
+        if event.outputs:
+            event.outputs[:] = self._dataset_normalizer.normalize_outputs(event.outputs)
         return event
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -38,7 +38,7 @@ from openlineage.client.transport import (
 )
 from openlineage.client.transport.http import HttpConfig, HttpTransport
 from openlineage.client.transport.noop import NoopConfig, NoopTransport
-from openlineage.client.utils import deep_merge_dicts
+from openlineage.client.utils import deep_merge_dicts, split_into_list
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore", DeprecationWarning)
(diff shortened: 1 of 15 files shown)


kacpermuda:
Thanks for implementing this, will take a look on Monday !

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+    KeyValueTrimmer,
+    DateTrimmer,
+    MultiDirDateTrimmer,
+    YearMonthTrimmer,
+]
+
+
+@attr.define
+class DatasetConfig:
+    normalization_enabled: bool = False
```
kacpermuda, line 34: Can we call it something other than normalization? In my mind at least, normalization would be replacing something in the name for a normalized version f.e. `date=2024-01-01` to `date={DATE}`. This is only about trimming, so if we decide to do an actual normalization in the future, we'll avoid naming confusion.

```suggestion
    trimming_enabled: bool = False
```

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+        condition = base_subset_dataset.LocationSubsetCondition(locations=sorted(self.original_names))
+
+        if isinstance(dataset, InputDataset):
+            i_facets = dataset.inputFacets if dataset.inputFacets is not None else {}
+            i_facets["subset"] = base_subset_dataset.InputSubsetInputDatasetFacet(inputCondition=condition)
+            dataset.inputFacets = i_facets
+        elif isinstance(dataset, OutputDataset):
+            o_facets = dataset.outputFacets if dataset.outputFacets is not None else {}
+            o_facets["subset"] = base_subset_dataset.OutputSubsetOutputDatasetFacet(outputCondition=condition)
+            dataset.outputFacets = o_facets
```
kacpermuda, line 92: What do we do here when it's just Dataset? Not Input/Output Dataset. Do we just ignore the subset facet entirely? Shouldn't we also do something?

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+    def normalize_outputs(self, datasets: list[Any]) -> list[Any]:
+        output_datasets = [
+            OutputDataset(
+                outputFacets={},
+                **attr.asdict(ds, recurse=False),
+            )
+            if not hasattr(ds, "outputFacets")
+            else ds
+            for ds in datasets
+        ]
```
kacpermuda, line 138: I do not understand the need for this filtering, most of the libraries do not use output/input facets, they use just facets.

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+        if self.dataset.facets != other_dataset.facets:
+            return False
+        if getattr(self.dataset, "inputFacets", None) != getattr(other_dataset, "inputFacets", None):
+            return False
+        if getattr(self.dataset, "outputFacets", None) != getattr(other_dataset, "outputFacets", None):
+            return False
+        return True
+
+
+class DatasetNormalizer:
```
kacpermuda, line 112: General note, any of the trimmers can fail, we should handle it and log warning / then skip entire normalization. Currently there is no exception handling here

client/python/src/openlineage/client/dataset/trimmers.py
```diff
@@ -0,0 +1,181 @@
...
+    DatasetNameTrimmer interface for dataset name trimmers in OpenLineage events.
+
+    A trimmer receives a dataset name (string) and returns
+    the trimmed version of that name.
+    """
+
+    SEPARATOR = "/"
+
+    def trim(self, name: str) -> str:
+        raise NotImplementedError
```
kacpermuda, line 27: Should we mark this method as abstract to make sure it's overwritten? Or at least use something like `typing.Protocol`

client/python/src/openlineage/client/dataset/dataset.py
kacpermuda: Maybe good idea is to let the child trimmers implement `_trim` and then call `trim` that would be just try/except call on `_trim`?

client/python/src/openlineage/client/dataset/trimmers.py
```diff
@@ -0,0 +1,181 @@
...
+
+    EQUALITY_SIGN = "="
+
+    def trim(self, name: str) -> str:
+        if not self._has_multiple_directories(name):
+            return name
+
+        last = self._get_last_part(name)
+
+        if last.count("=") != 1:
```
kacpermuda, line 80:
```suggestion
        if last.count(self.EQUALITY_SIGN) != 1:
```

client/python/src/openlineage/client/dataset/trimmers.py
```diff
@@ -0,0 +1,181 @@
...
+        "/20250721"           -> "/20250721"
+    """
+
+    DATE_PATTERNS = [
+        ("%Y-%m-%d", re.compile(r"\d{4}-\d{2}-\d{2}")),
+        ("%d.%m.%Y", re.compile(r"\d{2}\.\d{2}\.\d{4}")),
+        ("%Y%m%d", re.compile(r"\d{8}")),
+    ]
+
+    NOISE_REGEX = re.compile(r"[TZ\s:.\-\d]*")
```
kacpermuda, line 112: Why do we need this? Please add some explanation

client/python/src/openlineage/client/dataset/trimmers.py
```diff
@@ -0,0 +1,181 @@
...
+                return self._remove_last_part(name, parts_to_remove=3)
+
+            y, m = parts[-2:]
+            if _valid_date(f"{y}/{m}", "%Y/%m"):
+                return self._remove_last_part(name, parts_to_remove=2)
+
+        return name
+
+
+class YearMonthTrimmer(DatasetNameTrimmer):
```
kacpermuda, line 161: Shouldn't that use DateTrimmer as a base? It looks similar

client/python/src/openlineage/client/client.py
```diff
@@ -90,6 +92,18 @@ def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
                 job=params["tags"].get("job", {}),
                 run=params["tags"].get("run", {}),
             )
+        if "dataset" in params:
+
+            def split_into_list(value: str | list[str]) -> list[str]:
+                if isinstance(value, list):
+                    return value
+                return [item.strip() for item in value.split(";") if item.strip()]
```
kacpermuda, line 100: Let's put that in general utils, may be useful for future.

client/python/src/openlineage/client/client.py
```diff
@@ -90,6 +92,18 @@ def from_dict(cls, params: dict[str, Any]) -> OpenLineageConfig:
...
+
+            ds = params["dataset"]
```
kacpermuda, line 102: Let's make sure it's a dict or we raise a type error that is clear, otherwise error message from below may not be clear enough for the users.

client/python/src/openlineage/client/client.py
```diff
@@ -595,3 +613,12 @@ def add_source_code_location_facet(self, event: Event) -> Event:
                 pullRequestNumber=scl["pullRequestNumber"],
             )
         return event
+
+    def normalize_datasets(self, event: Event) -> Event:
+        if not isinstance(event, (RunEvent, event_v2.RunEvent)):
```
kacpermuda, line 618: We can just use v2.RunEvent here, it's a new feature, no need to support legacy RunEvent.

client/python/src/openlineage/client/client.py
```diff
@@ -595,3 +613,12 @@ def add_source_code_location_facet(self, event: Event) -> Event:
                 pullRequestNumber=scl["pullRequestNumber"],
             )
         return event
+
+    def normalize_datasets(self, event: Event) -> Event:
```
kacpermuda, line 617: Mentioned it somewhere above as well, we need to be clear about what happens if a trimming operation fails. Do we fail to emit event? Do we emit original event? Do we let the user choose? I would say we emit event unchanged, as the trimming never happened. And let's document that behavior. It's a low risk for built-in trimmers, but people can add their own custom trimmers, that can have a very complex logic and be failing a lot, and we need to be prepared for that.

website/docs/client/python/configuration.md
```diff
@@ -1971,3 +1972,148 @@ client = OpenLineageClient(config=config)
 </TabItem>
 </Tabs>
 
+
+## Dataset Name Normalization
```
kacpermuda, line 1976: Can you add some information that the trimming is performed until no known trimmers change the name? I'm not sure if it's clear here, that the order of the trimmers does not matter, and that they will not only trim the last part, but they will run in sequence, in random (or not?) order and trim until there is nothing known left. This may be crucial to understand for people trying to add their own trimmers.

kacpermuda (commented):
Left some comments

client/python/src/openlineage/client/dataset/dataset.py
kchledowski: I've actually wanted to avoid calling it "trimming", because apart from trimming it also does grouping/reducing the datasets by name. I've come up with "normalization" as a better word for it, but I'm happy to hear any other suggestions

client/python/src/openlineage/client/dataset/dataset.py
kchledowski: subset facet extends `InputSubsetInputDatasetFacet` or `OutputSubsetOutputDatasetFacet`, meaning, as a concept, it is always part of either `InputDataset` or `OutputDataset`.

Technically, such dataset can "arrive" in the normalizer represented as `Dataset`, that's why depending on whether this is input or output normalization, first thing the normalizer does is constructing either `InputDataset` or `OutputDataset` for each dataset, e.g. for input normalization:
```
        input_datasets = [
            InputDataset(
                inputFacets={},
                **attr.asdict(ds, recurse=False),
            )
            if not hasattr(ds, "inputFacets")
            else ds
            for ds in datasets
        ]
```

In this code what decides that the `InputDataset` should be constructed is the fact that it is executed within the `normalize_inputs` method.

This additional check `if not hasattr(ds, "inputFacets")`, was just put there to make sure "inputFacets" attribute is not duplicated in the constructor of `InputFacets`.

Now that I think of it, it might be better to do `if not isinstance(ds, InputDataset)` instead of `if not hasattr(ds, "inputFacets")`

kchledowski: I made changes to make it more explicit about what happens.

Generally we know whether it is `InputDataset` or `OutputDataset` from the fact that either `normalize_inputs` or `normalize_outputs` was called. Subset facet is added accordingly to either `inputFacets` or `outputFacets`, even if the class was the plain/base `Dataset`. The proper child class (`InputDataset`/`OutputDataset`) is constructed to make sure any further serialization of the Dataset includes the `inputFacets`/`outputFacets`

client/python/src/openlineage/client/dataset/trimmers.py
kchledowski: I'd rather not abuse inheritance for this, keeping each trimmer separate makes it easier to change if needed

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+                for input_field in field.inputFields or ():
+                    input_field.name, _ = self._apply_trimmers(input_field.name)
+
+        for input_field in cll.dataset or ():
+            input_field.name, _ = self._apply_trimmers(input_field.name)
+
+    def _apply_trimmers(self, name: str) -> tuple[str, bool]:
+        was_trimmed = False
+
+        while True:
```
kacpermuda: **Unbounded loop on user-supplied trimmers.** This `while True:` only terminates when a full pass over `self.trimmers` is a fixed point. That assumption is safe for the four built-in trimmers because each is monotonically *shortening* — but `extra_trimmers` accept arbitrary `DatasetNameTrimmer` subclasses, and a user-supplied trimmer that returns a longer string, alternates between two values, or just changes the path without shortening it (think a simple `name + "/"` bug) makes this loop run forever and freeze the calling thread on every emit.

Two cheap defenses, ideally combined:
- An iteration cap (e.g. `for _ in range(MAX_TRIMMER_PASSES):` with `MAX_TRIMMER_PASSES = 16`); on hitting it, log a warning naming the offending name and break.
- A simple cycle detector — track names seen in the current name's trim chain in a small set; bail out as soon as a name reappears.

The documentation at `configuration.md:2008` ("in an undefined order until no additional segments of the path can be removed") implicitly assumes monotonic trimmers without saying so.

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def _reduce_datasets(
+        self, datasets: list[Dataset], dataset_type: Literal["input", "output"]
+    ) -> list[Dataset]:
+        reduced_datasets: list[ReducedDataset] = []
+
+        for ds in datasets:
+            original_name = ds.name
+            normalized_name, name_was_normalized = self._apply_trimmers(ds.name)
+            ds.name = normalized_name
```
kacpermuda: **In-place mutation of the caller's dataset.** `ds.name = normalized_name` rewrites the original dataset object the user passed in — the mutation survives even after `_reduce_datasets` returns, because `event.inputs[:] = ...` in `client.normalize_datasets` only swaps the *list contents*, not the elements.

Concretely: if a user holds a reference to the event for retry/logging, or constructs the event once and emits twice (e.g. with two clients in different normalization configs), the second emit sees already-trimmed names and may double-trim or no-op unexpectedly.

Either:
- `ds = copy.copy(ds)` upfront (shallow, since we only rewrite `.name`), or
- document explicitly that `normalize_inputs` / `normalize_outputs` mutate their arguments and the public `client.emit()` contract is "event is mutated".

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+            return
+
+        cll = dataset.facets.get("columnLineage")
+        if not isinstance(cll, column_lineage_dataset.ColumnLineageDatasetFacet):
+            return
+
+        if cll.fields:
+            for field in cll.fields.values():
+                for input_field in field.inputFields or ():
+                    input_field.name, _ = self._apply_trimmers(input_field.name)
```
kacpermuda: Same in-place mutation concern as `_reduce_datasets` — `input_field.name = ...` rewrites the user-owned `InputField` object. With column lineage, this is more dangerous: it's not unusual for one input field reference to be shared across multiple `Fields` entries (or between input and output column-lineage facets). A single trim then cascades to every reference. Worth either deepcopying the modified inputField or — better — building new `InputField` instances with the trimmed name and reassigning.

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+    def _apply_trimmers(self, name: str) -> tuple[str, bool]:
+        was_trimmed = False
+
+        while True:
+            new_name = name
+            for trimmer in self.trimmers:
+                try:
+                    new_name = trimmer.trim(new_name)
+                except Exception as e:
+                    log.debug(
```
kacpermuda: Thanks for adding try/except here. `log.debug` may be too quiet here. In production, default Python logging levels are usually WARNING or INFO; a misbehaving custom trimmer would silently produce wrong lineage with no visible signal. Let's do at least `log.warning` on the *first* failure of a given trimmer class (track seen classes on the normalizer to avoid spamming on every event).

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+    disabled_trimmers: list[str] = attr.field(factory=list)
+    extra_trimmers: list[str] = attr.field(factory=list)
+
+    def get_dataset_name_trimmers(self) -> list[DatasetNameTrimmer]:
+        active: list[DatasetNameTrimmer] = []
+        seen_classes: set[type[DatasetNameTrimmer]] = set()
+
+        for cls in BUILTIN_TRIMMERS:
+            fqn = f"{cls.__module__}.{cls.__name__}"
+            if fqn in self.disabled_trimmers:
```
kacpermuda, line 56: Typos in `disabled_trimmers` are silently ignored — an entry like `"openlineage.client.dataset.trimmers.KevValueTrimmer"` (typo) doesn't match any built-in FQN, so nothing is disabled and the user has no signal that their config is wrong.

Suggest validating that every entry in `disabled_trimmers` matches a built-in trimmer FQN, and `log.warning` for entries that don't (mirror what `extra_trimmers` does at line 55-58 for non-`DatasetNameTrimmer` classes).

client/python/src/openlineage/client/client.py
```diff
@@ -183,6 +193,8 @@ def emit(self, event: Event) -> None:
         event = self.add_environment_facets(event)
         event = self.update_event_tags_facets(event)
         event = self.add_source_code_location_facet(event)
+        if self.config.dataset.normalization_enabled:
```
kacpermuda: Subtle ordering choice that deserves an explicit comment: filters at line 209 are applied *after* normalization. So a filter that matches on dataset name (e.g. `match: "/data/events/dt=2025-09-01"`) will not match the normalized event (where the name is `/data/events`). This may be the right call (filter the canonical name, not the partition path), but it's silent — please add a one-liner here explaining the choice and add some note to the documentation - when exactly this dataset reduction happens.

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,201 @@
...
+class ReducedDataset:
+    dataset: Dataset
+    original_names: set[str]
+    dataset_type: Literal["input", "output"]
+
+    def as_dataset(self) -> Dataset:
+        if not self.original_names:
+            return self.dataset
+
+        dataset = copy.deepcopy(self.dataset)
```
kacpermuda: `copy.deepcopy(self.dataset)` is heavy: for a dataset with a large `schema` facet (hundreds of fields) or large `columnLineage`, this clones the entire facet tree just to attach a `subset` facet to the input/output-facets dict. We're not modifying anything inside `facets` — only the outer `inputFacets` / `outputFacets` dict. Maybe we should consider shallow clone here? This avoids cloning the schema/CLL payload entirely and makes the per-emit cost predictable.

website/docs/client/python/configuration.md
```diff
@@ -1971,3 +1972,153 @@ client = OpenLineageClient(config=config)
...
+2. They share identical facets.
+
+Dataset name normalization is **disabled by default**. It must be explicitly enabled via the `normalization_enabled` flag.
+
+The list of enabled trimmers can be managed using `disabled_trimmers` and `extra_trimmers` configuration parameters.
+
+In most cases, trimmers work on the last directory segment of the dataset name.
+
+The trimming process runs repeatedly (in a loop), applying trimmers one by one
+in an undefined order until no additional segments of the path can be removed.
```
kacpermuda, line 2020: "in an undefined order" is only true when every trimmer is a *monotonically shortening* function — i.e. `len(trimmer.trim(x)) <= len(x)` with strict inequality on a fixed point. The four built-ins satisfy this; user-supplied `extra_trimmers` are not constrained to.

A custom trimmer that grows or cycles the name will hang the loop in `_apply_trimmers` (see comment on `dataset.py:176`). Once that's bounded with an iteration cap, this paragraph should also document the contract for custom trimmers: "the `trim` method MUST be idempotent and never return a name longer than its input" (or equivalent).

client/python/tests/dataset/test_normalizer.py
```diff
@@ -0,0 +1,470 @@
...
+                    namespace="ns", name="/data/table/day=1", inputFacets={"inputStatistics": stats}
+                ),
+            ]
+        )
+
+        assert len(result) == 1
+        assert result[0].name == "/data/table"
+        assert result[0].inputFacets.get("inputStatistics") == stats
+
+    def test_failing_trimmer_does_not_prevent_normalization(self):
```
kacpermuda: Test name is misleading. `test_failing_trimmer_does_not_prevent_normalization` reads as "if a trimmer fails, the whole normalization step still runs" — but what's actually being tested is "a *failing extra trimmer* doesn't stop the *built-in* trimmers from running" (the failing trimmer raises on every input; the result is still trimmed because `KeyValueTrimmer` does the work).

Something like `test_failing_extra_trimmer_does_not_block_other_trimmers` is closer to what's verified, and a complementary test where the *only* registered trimmer fails (so the name should pass through untouched) would exercise the more interesting failure path.

Same note on line 446 for the outputs version.

client/python/src/openlineage/client/dataset/dataset.py
kacpermuda: Maybe `reducing`? I'm not happy with normalization, but given that it's already named like this in java integration, and "we may want to introduce something similar in the future" is not the most solid reason for spending too much time on this now, we can go on with that name if "reducing" does not sound right. I think it would be more in place here.

kacpermuda: Other ideas: `path_collapsing`, `path_canonicalization`.

website/docs/client/python/configuration.md
```diff
@@ -1971,3 +1972,153 @@ client = OpenLineageClient(config=config)
 </TabItem>
 </Tabs>
 
+
+## Dataset Name Normalization
+
```
kacpermuda, line 1989: One more important thing to add here, that only RunEvent is supported for now. Maybe we can underline it somewhere at the beginning of this section?

client/python/src/openlineage/client/dataset/trimmers.py
kacpermuda: The two-line comment helps, thanks. Could you push it a bit further and explain why this regex exists in the algorithm — i.e. that we strip known noise around a date match so we can recognize dates embedded in segments like dt=2025-07-22T09:01Z? As written, the comment describes the character class but not the role it plays in _looks_like_date (we can do that in docstring of `_looks_like_date`, or make this regex local and document it there entirely). A reader landing on this cold still has to reverse-engineer the algorithm.

kacpermuda (commented):
Thanks for the changes, leaving some more comments, some are AI supported but I've found them to be interesting.

client/python/src/openlineage/client/client.py
kchledowski: The filters work on job names, not dataset names, so it wouldn't matter, but still I think it makes sense to have a comment in the code, warning that the dataset names are changed

client/python/src/openlineage/client/dataset/dataset.py
kchledowski: I've used deepcopy here, to make the `as_dataset` method idempotent, but yeah, it does come with a lot of unnecessary load. 
I'll replace it with a shallow copy and than a copy of input/output facets, it will still be idempotent, but without a need for deepcopy

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,259 @@
...
+
+                if cls not in seen_classes:
+                    active.append(cls())
+                    seen_classes.add(cls)
+            except Exception as e:
+                log.warning(
+                    "Failed to load trimmer '%s': %s. Skipping.",
+                    fqn,
+                    e,
+                )
```
dolfinus, line 82:
```suggestion
                log.exception(
                    "Failed to load trimmer '%s'. Skipping.",
                    fqn,
                )
```
This shows exception traceback

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,259 @@
...
+    def _reduce_datasets(
+        self, datasets: list[Dataset], dataset_type: Literal["input", "output"]
+    ) -> list[Dataset]:
+        reduced_datasets: list[ReducedDataset] = []
+
+        for ds in datasets:
+            ds = copy.copy(ds)
+            original_name = ds.name
+            trimmed_name, name_was_trimmed = self._trim(ds.name)
+            ds.name = trimmed_name
```
dolfinus: Why not using `attr.evolve` here?

client/python/tests/dataset/test_config.py
```diff
@@ -0,0 +1,70 @@
...
+def test_added_trimmers_must_be_actual_trimmers():
+    config = DatasetConfig(extra_trimmers=["tests.dataset.test_config.SomeClass"])
+
+    assert DEFAULT_TRIMMERS == [type(t) for t in config.get_dataset_name_trimmers()]
+
+
+def test_no_duplicate_trimmers():
+    config = DatasetConfig(extra_trimmers=["openlineage.client.dataset.trimmers.KeyValueTrimmer"])
+
+    assert [*DEFAULT_TRIMMERS] == [type(t) for t in config.get_dataset_name_trimmers()]
```
dolfinus:
```suggestion
    assert DEFAULT_TRIMMERS == [type(t) for t in config.get_dataset_name_trimmers()]
```

dolfinus (approved):

client/python/src/openlineage/client/dataset/dataset.py
kchledowski: I don't think a full stack trace is needed here, just an error message saying why the trimmer loading failed.
Also, I'd rather be cautious with logging on the ERROR level, especially in cases where it doesn't prevent the event from being emitted. In this case a WARNING log saying the custom trimmer, that user provided, can not be loaded is enough

client/python/src/openlineage/client/dataset/dataset.py
kchledowski: Good point, to be honest I've just learned what `attr.evolve` does (I don't have much experience with Python) and somehow didn't think to use it here. But let's be consistent, I'll do it here, since I already use it in other places

client/python/src/openlineage/client/dataset/dataset.py
kacpermuda: Makes sense, let's leave it on warning - it's enough

dolfinus: The reason why I recommend to show traceback here is because `import_string` will throw something like `ImportError("Cannot import some.Some")` without actual reason why import failed. You can use `logger.warning(..., exc_info=True)` to show traceback for warning level. Or you can add explicit `exc_info=logger.isEnabledFor(logging.DEBUG)`

kacpermuda: We can to logging.debug("Exception details:", exc_info=True) after the current warning log. We've done that before in other places, where we log just error message on warning leve and full traceback on debug level. When someone wants to have a full traceback, they must enable debug logging.

kchledowski: Stack trace would give some additional details, but I wonder if maybe the information about import failure is enough for the user, is signals the user they probably need to fix the path to the module if it can not be imported.

To me, when I see a full stack trace, I think that something must have gone really wrong somewhere, I'm not sure if we should show that to the user, but maybe I am wrong

kchledowski: `exc_info=logger.isEnabledFor(logging.DEBUG)` this seems like a reasonable option, I think we can go with that

dolfinus (approved):

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,259 @@
...
+    def _apply_trimmers(self, name: str) -> str:
+        for trimmer in self.trimmers:
+            try:
+                name = trimmer.trim(name)
+            except Exception as e:
+                log_once.warning(
+                    "Skipping trimmer '%s' due to an error when trimming dataset name '%s': %s.",
+                    f"{trimmer.__class__.__module__}.{trimmer.__class__.__name__}",
+                    name,
+                    e,
```
dolfinus, line 258: Let's add exc_info trick here as well

client/python/src/openlineage/client/utils.py
```diff
@@ -55,9 +56,48 @@ def deep_merge_dicts(dict1: dict[Any, Any], dict2: dict[Any, Any]) -> dict[Any,
...
+    message. First occurrence is logged at the original level; subsequent
+    occurrences are demoted to DEBUG.
+
+    Deduplication scope is bounded using an LRU cache to prevent unbounded
+    memory growth.
+    """
+
+    def __init__(self, max_messages: int = 5_000) -> None:
+        super().__init__()
+        self._seen: OrderedDict[str, None] = OrderedDict()
```
kacpermuda, line 169: Why ordered dict instead of set?
`self._seen[message] = None` this seems like a golang hack to lack of sets translated into python 😄 Also why do we move to end in `self._seen.move_to_end(message)`? First time seeing this kind of behavior.

client/python/src/openlineage/client/client.py
```diff
@@ -142,6 +150,8 @@ def __init__(  # noqa: PLR0913
             if _filter:
                 self._filters.append(_filter)
 
+        self._dataset_reducer = DatasetReducer(self.config.dataset)
```
kacpermuda: Just a side note, this will always try to load trimmers, even when it's disabled. Probably a low-cost / no-op, but we can easily add an `if self.config.dataset.reducing_enabled` here, or just simply make it None, and initialize it in self.reduce_datasets, if not initialized:

```
def reduce_datasets(self, event: Event) -> Event:
    if not self.__dataset_reducer:
        self.__dataset_reducer = DatasetReducer(self.config.dataset)
    <current code>
```

client/python/src/openlineage/client/dataset/dataset.py
kacpermuda: I think this is still the case for ds.name right? We pass ds.name to trimmers, so they do not work on name copy, they actually modify the dataset, so later attr.evolve is a no-op? Maybe I'm wrong at this point, ngl I'm a bit confused now. Let's just double check.

kacpermuda (approved):
Thanks for adding this, and applying all the suggestions. Left two more comments, one that can actually make it a real lazy load. Approving, but would appreciate another look from @mobuchowski or @JDarDagran, since I've looked at this code for a while, may have missed something.

client/python/src/openlineage/client/utils.py
kchledowski: This is made so that it would be possible to track log message insertion (to cache) order, so that the oldest message could be evicted first
```
A, B, C → full
A again → reorder → B, C, A
D → evict B
```

rahul-madaan:
Hi @mobuchowski or @JDarDagran, requesting your review on this PR as a final step.

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,260 @@
...
+            o_facets = dict(getattr(dataset, "outputFacets", None) or {})
+            o_facets["subset"] = base_subset_dataset.OutputSubsetOutputDatasetFacet(outputCondition=condition)
+
+            base_dict = attr.asdict(dataset, recurse=False)
+            base_dict.pop("outputFacets", None)
+            return OutputDataset(**base_dict, outputFacets=o_facets)
+        return dataset
+
+    def matches(self, other_dataset: Dataset) -> bool:
+        return self._name_matches(other_dataset) and self._facets_match(other_dataset)
```
mobuchowski: This should look at namespace too, right?

client/python/src/openlineage/client/client.py
```diff
@@ -595,3 +610,15 @@ def add_source_code_location_facet(self, event: Event) -> Event:
...
+        if not self._dataset_reducer:
+            self._dataset_reducer = DatasetReducer(self.config.dataset)
+
+        if not isinstance(event, event_v2.RunEvent):
+            return event
+        if event.inputs:
+            event.inputs[:] = self._dataset_reducer.reduce_inputs(event.inputs)
+        if event.outputs:
+            event.outputs[:] = self._dataset_reducer.reduce_outputs(event.outputs)
+        return event
```
mobuchowski: Is this supposed to mutate the passed dataset? 

Maybe we need to construct a new one with reduced datasets, rather than mutating.

client/python/src/openlineage/client/dataset/dataset.py
```diff
@@ -0,0 +1,260 @@
...
+        return self.dataset.name == other_dataset.name
+
+    def _facets_match(self, other_dataset: Dataset) -> bool:
+        if self.dataset.facets != other_dataset.facets:
+            return False
+        if getattr(self.dataset, "inputFacets", None) != getattr(other_dataset, "inputFacets", None):
+            return False
+        if getattr(self.dataset, "outputFacets", None) != getattr(other_dataset, "outputFacets", None):
+            return False
+        return True
```
mobuchowski, line 139: I think we need to treat `{}` and `None` as a same thing - 

```
ds1 = InputDataset(namespace="ns", name="/data/table/day=1"). # default {} here
ds2 = InputDataset(namespace="ns", name="/data/table/day=2", facets=None)
```

They should reduce, but won't with current code.

mobuchowski (changes requested):

client/python/src/openlineage/client/dataset/dataset.py
kchledowski: Great spot! It definitely should consider the namespace when reducing. I'll fix it here and I will also raise a new PR with the same change for the Java client, where the reducer also does not check the namespace.

client/python/src/openlineage/client/client.py
kchledowski: It's not mutating the datasets, it does construct new ones. But it is mutating the lists (inputs/outputs) containing the datasets. I will change it to construct new lists to avoid confusion.

chuenchen309:
I ran this branch locally while going through the open Python-side PRs. Short version: the three review comments from Jun 9 are all addressed in the head commit, and what's left before a re-review looks like a rebase.

Each comment has a commit from the same day:

- *"This should look at namespace too"* → `7431ccd5`: `_name_matches` became `_identity_matches` and now compares `namespace` as well as `name`.
- *"we need to treat `{}` and `None` as a same thing"* → `0a3b6f5b`: `_facets_match` normalizes with `or {}` on `facets`, `inputFacets` and `outputFacets`.
- *"Maybe we need to construct a new one rather than mutating"* → `1eec7d1a`: `reduce_datasets` dropped the `event.inputs[:] = ...` slice assignment and returns `attr.evolve(event, inputs=..., outputs=...)`.

I checked those by running rather than by reading — the same script on `60a216a7` (the commit before the three fixes) and on `7431ccd5` (head):

| check | `60a216a7` | `7431ccd5` |
|---|---|---|
| two datasets, different namespaces, names trim to the same value | reduced to 1 ❌ | stays 2 ✅ |
| same namespace, same trimmed name (control) | reduced to 1 ✅ | reduced to 1 ✅ |
| `facets={}` vs `facets=None` | stays 2 ❌ | reduced to 1 ✅ |
| caller's `inputs` list after `reduce_datasets` | rewritten in place ❌ | untouched ✅ |
| `event.inputs` on the event that was passed in | 1 element ❌ | still 2 ✅ |
| returned event's `inputs` | same list object ❌ | new list, 1 element ✅ |

<details><summary>script</summary>

```python
from openlineage.client.client import OpenLineageClient
from openlineage.client.dataset import DatasetConfig, DatasetReducer
from openlineage.client.event_v2 import InputDataset, Job, Run, RunEvent, RunState
from openlineage.client.uuid import generate_new_uuid

r = lambda ds: DatasetReducer(DatasetConfig()).reduce_inputs(ds)

print("namespace:", len(r([InputDataset(namespace="ns_a", name="/data/table/day=1"),
                          InputDataset(namespace="ns_b", name="/data/table/day=2")])), "expect 2")
print("control:  ", len(r([InputDataset(namespace="ns", name="/data/table/day=1"),
                           InputDataset(namespace="ns", name="/data/table/day=2")])), "expect 1")
print("{} v None:", len(r([InputDataset(namespace="ns", name="/data/table/day=1"),
                           InputDataset(namespace="ns", name="/data/table/day=2", facets=None)])), "expect 1")

inputs = [InputDataset(namespace="ns", name="/data/table/day=1"),
          InputDataset(namespace="ns", name="/data/table/day=2")]
event = RunEvent(eventType=RunState.START, eventTime="2026-08-06T00:00:00.000000Z",
                 run=Run(runId=str(generate_new_uuid())), job=Job(namespace="ns", name="job"),
                 inputs=inputs)
new_event = OpenLineageClient(config={"dataset": {"reducing_enabled": True}},
                              transport=None).reduce_datasets(event)
print("caller list:", [d.name for d in inputs], "expect both, untrimmed")
print("event.inputs:", len(event.inputs), "expect 2")
print("returned:   ", len(new_event.inputs), new_event.inputs is not inputs, "expect 1 True")
```
</details>

Full `client/python` suite on this branch: 656 passed, against 606 on `main` — both excluding `tests/test_http_integration.py`, which needs docker-compose and errors out in my environment either way.

The part that does need someone: GitHub now reports the branch as conflicting with `main`. `client.py`, `utils.py` and `tests/test_client.py` have each moved a few times since Jun 9. The single red check, `integration-test-integration-hive`, is the Jun 9 CircleCI run; that job passes on recent PRs (#4778, #4801), so a rebase should take care of it as well.

kchledowski:
Rebased onto main.
The CI failed with something unrelated, could someone re run it?
image

mobuchowski (approved):

Issue #4359, opened 2026-02-25
[PROPOSAL] Introduce explicit lineage to support complex ETL-like jobs


nantunes:
Following up on the discussion in issue https://github.com/OpenLineage/OpenLineage/issues/3458, because we are facing the same questions and challenges while adopting OpenLineage in our tools, and given the expressed interest from the community in understanding where the "parent run" approach might fall short for certain workloads, we have put together this proposal for discussion.

## Summary

Currently OpenLineage relies on **implicit dataset level lineage**, i.e. if a Run has `inputs=[A]` and `outputs=[B]`, an edge `A -> B` is assumed. However, this way the number of links will be the cartesian product of inputs by outputs, with all inputs influencing all outputs in the same Run.

While this works well for lazy-evaluation engines like Apache Spark (where an "Action" tightly couples specific inputs to just one specific output), it creates significant semantic distortion for eager-execution engines like traditional ETL tools (e.g. Pentaho).

The specification also supports **explicit column level lineage** via the dataset's *Column Level Lineage* facet, but this solution is insufficient for many use cases. The facet solves the problem at the "microscopic" level (column-to-column) but leaves a gap at the "macroscopic" level (dataset-to-dataset). Integrators often know that "Table A derived Table B" without having the granularity to map specific columns. Currently, if an integrator cannot provide column level details, they end up back to the mentioned cartesian product inference.

We propose making lineage a **first-class citizen** by introducing a mechanism to define **explicit complete lineage**. This would separate the audit of what was accessed (inputs, outputs) from the definition of what and how was used (lineage), and allow integrators to report accurate lineage for complex control flows without artificially breaking jobs into fictitious sub-runs.

## The Problem

The spec doesn't distinguish between **IO (Input/Output)** and the **flow of data**.

- **IO event:** "File A was read by Job X."
  - This is a fact useful for audit trails and security (who touched the data?).
- **Flow of data event:** "Column B was calculated by Job X using Column A."
  - This is a fact useful for impact analysis and debugging.

By inferring the data flow from IO events, the system is somewhat "guessing" the lineage. The standard should allow producers to be more precise.

In Spark, the unit of work is defined by an Action (`.write()`, `.collect()`, `.save()`). Transformations accumulate in a logical plan and only materialize (lazy evaluation) when the action is invoked. As such, a Run maps to a specific optimized plan writing to a single destination.

[...]

PR #4468, opened 2026-04-15
Explicit lineage proposal

mobuchowski:
## Problem
OpenLineage infers dataset lineage from the cartesian product of a run's inputs × outputs. A job reading {A,B} and writing {C,D} produces 4 edges — even when the real flow is A→C and B→D only. The spurious edges aren't always a problem at small scale, but they compound on bulk ETL jobs where many independent tables are processed together, making impact analysis and compliance audits harder to rely on.

```mermaid
flowchart LR
    subgraph Current["Current spec (cartesian product)"]
      direction LR
      A1[Table A] --> C1[Table C]
      A1 -.->|false positive| D1[Table D]
      B1[Table B] -.->|false positive| C1
      B1 --> D1
    end
    subgraph Proposed["Explicit lineage"]
      direction LR
      A2[Table A] --> C2[Table C]
      B2[Table B] --> D2[Table D]
    end
```

`ColumnLineageDatasetFacet` solves this at column granularity, but leaves a gap at dataset granularity, and the current model can't express dataset-to-dataset derivation (views, aliases) or mixed granularity.

## Proposal
Adds two documents under `proposals/explicit_lineage_facet/`:

- **ExplicitLineageProposal.md** — full spec for three new facets, one per event type:

```mermaid
flowchart LR
    RE[RunEvent] --> RF["run.facets.lineage<br/>LineageRunFacet<br/><i>observed flow during run</i>"]
    JE[JobEvent] --> JF["job.facets.lineage<br/>LineageJobFacet<br/><i>declared flow for job</i>"]
    DE[DatasetEvent] --> DF["dataset.facets.lineage<br/>LineageDatasetFacet<br/><i>structural derivation</i>"]
```

Each entry explicitly maps a target entity to its source inputs at entity and/or column granularity, via a discriminated union on `type`:

```mermaid
classDiagram
    class LineageEntry { <<oneOf>> }
    class LineageDatasetEntry {
      namespace, name
      type = DATASET
      inputs: LineageInput[]
      fields: map~string, LineageFieldEntry~
    }
    class LineageJobEntry {
      namespace, name
      type = JOB
      runId?
      inputs: LineageInput[]
    }
    class LineageInput { <<oneOf>> }
    class LineageDatasetInput {
      namespace, name, type=DATASET
      field?
      transformations[]
    }
    class LineageJobInput {
      namespace, name, type=JOB
      runId?
    }
    LineageEntry <|-- LineageDatasetEntry
    LineageEntry <|-- LineageJobEntry
    LineageDatasetEntry --> LineageInput
    LineageJobEntry --> LineageInput
    LineageInput <|-- LineageDatasetInput
    LineageInput <|-- LineageJobInput
```

Limited `type: JOB` support for sink/generator cases now; extensible to full job-to-job later. Intended to eventually supersede `ColumnLineageDatasetFacet`.

- **ExplicitLineageAdoptionGuideline.md** — producer/consumer adoption guide.

## Compatibility & Migration
New `openlineage.lineage.compatibility = both` setting auto-generates the legacy `inputs`/`outputs`/CLL from the new facets, so existing consumers keep working. No coordination required, no core model changes (facets only).

```mermaid
flowchart LR
    P1["Phase 1<br/>Producers emit both<br/>(compatibility = both)"] --> P2["Phase 2<br/>Consumers add<br/>new-facet support"] --> P3["Phase 3<br/>Deprecate old<br/>format + CLL"]
```

## Related
Tracks #4359.

### DIFF
--- a/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
+++ b/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
@@ -93,6 +93,7 @@ This example shows how the new facets can replace ColumnLevelLineageDatasetFacet
 ```
 
 This single entry combines:
+
 - **Entity-level**: output_X comes from input_A
 - **Dataset-wide op**: group_by_col affects entire output
 - **Column mapping**: col_a -> col1
@@ -373,3 +374,4 @@ Reading specific columns from a structured table and writing to an unstructured 
   ]
 }
 ```
+
--- a/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
+++ b/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
@@ -57,7 +57,7 @@ This example shows how the new facets can replace ColumnLevelLineageDatasetFacet
             "total": {
               "inputs": [
                 { "namespace": "postgresql://analytics:5432", "name": "input", "type": "DATASET", "field": "amount",
-                  "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION" }] }
+                  "transformations": [{ "type": "DIRECT", "subtype": "AGGREGATION" }] }
               ]
             }
           }
@@ -107,7 +107,6 @@ A VIEW derives from base tables. No job involved. LineageDatasetFacet on the dat
 
 ```json
 {
-  "eventType": "COMPLETE",
   "eventTime": "2026-03-25T10:00:00.000Z",
   "dataset": {
     "namespace": "postgresql://warehouse:5432",
@@ -130,7 +129,7 @@ A VIEW derives from base tables. No job involved. LineageDatasetFacet on the dat
 
 ## 5. Static job lineage (on JobEvent)
 
-A catalog declares what a job is designed to do. LineageJobFacet on the job:
+A catalog declares what a job is designed to do. This is treated as design/static lineage because the event does not contain a run section. LineageJobFacet on the job:
 
 ```json
 {
@@ -232,7 +231,7 @@ The per-field detail names a `LineageJobInput` (`type: JOB` with no `namespace`/
             "fields": {
               "total": {
                 "inputs": [
-                  { "type": "JOB", "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION", "description": "sum()" }] }
+                  { "type": "JOB", "transformations": [{ "type": "DIRECT", "subtype": "AGGREGATION", "description": "sum()" }] }
                 ]
               }
             }
@@ -256,7 +255,6 @@ A stored procedure `dag_b.task_1` reads its input from upstream `dag_a.task_3` v
 
 ```json
 {
-  "eventType": "COMPLETE",
   "job": {
     "namespace": "airflow://prod",
     "name": "dag_b",
@@ -380,4 +378,3 @@ Reading specific columns from a structured table and writing to an unstructured 
   ]
 }
 ```
-
--- a/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
+++ b/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
@@ -2,7 +2,7 @@
 
 This file contains examples for [ExplicitLineageProposal.md](ExplicitLineageProposal.md). See [ExplicitLineageSchema.md](ExplicitLineageSchema.md) for the proposed schema.
 
-## 1. Dataset-level lineage on RunEvent (solves #4359)
+## 1. Dataset-level lineage on RunEvent
 
 ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. LineageRunFacet on the run:
 
@@ -38,7 +38,9 @@ ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. LineageRunF
 
 `inputs`/`outputs` remain as the carrier for dataset facets, as the fallback for older consumers, and as the event boundary for the job's direct input/output participation.
 
-## 2. Column-level lineage on RunEvent (subsumes CLL)
+## 2. Column-level lineage on RunEvent
+
+This example shows how the new facets can replace ColumnLevelLineageDatasetFacet.
 
 ```json
 "run": {
@@ -204,7 +206,47 @@ extract_task (JOB) ──> extracted_data (DATASET)
 
 That edge is implicit in the RunEvent boundary, not repeated with a `LineageJobInput`.
 
-## 8. Job-to-job — declared chain on JobEvent
+### 8. Enriched generator — field-level transformation detail, job still implicit
+
+The same extract task as in previous example, but the producer knows that `total` was derived via a `sum()` aggregation over the (untracked) external API response. 
+There is still no tracked upstream dataset, so the entry's top-level `inputs` is `[]`. 
+The per-field detail names an `LineageJobInput` (`type: JOB` with no `namespace`/`name`), which resolves to the event's own job, and carries the transformation:
+
+```json
+{
+  "eventType": "COMPLETE",
+  "run": {
+    "runId": "abc-789",
+    "facets": {
+      "lineage": {
+        "_producer": "https://example.com/extract",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "entries": [
+          { "namespace": "postgres://prod", "name": "extracted_data", "type": "DATASET",
+            "inputs": [],
+            "fields": {
+              "total": {
+                "inputs": [
+                  { "type": "JOB", "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION", "description": "sum()" }] }
+                ]
+              }
+            }
+          }
+        ]
+      }
+    }
+  },
+  "job": { "namespace": "airflow://prod", "name": "data_pipeline.extract_task" },
+  "inputs": [],
+  "outputs": [
+    { "namespace": "postgres://prod", "name": "extracted_data" }
+  ]
+}
+```
+
+The implicit edge `extract_task (JOB) ──> extracted_data (DATASET)` is unchanged — the identity-less `LineageJobInput` only adds the column-level transformation that event membership could not express.
+
+## 9. Job-to-job — declared chain on JobEvent
 
 A stored procedure `dag_b.task_1` reads its input from upstream `dag_a.task_3` via a non-materialized handoff. No intermediate dataset exists. The catalog declares the design-time data flow on the JobEvent for `dag_b`:
 
@@ -235,7 +277,7 @@ A stored procedure `dag_b.task_1` reads its input from upstream `dag_a.task_3` v
 
 Note that `runId` is omitted on both ends — declared lineage describes the job definition, not a specific execution. The same shape would apply to a non-materialized view consumed by a downstream job, or to an Airflow task receiving data via XCom from an upstream task.
 
-## 9. Job-to-job — observed chain on RunEvent
+## 10. Job-to-job — observed chain on RunEvent
 
 The same data flow as Example 8, observed at runtime by an integration that can identify both the upstream and downstream runs. `runId` is populated on each end, binding the chain to specific executions:
 
@@ -266,7 +308,7 @@ The same data flow as Example 8, observed at runtime by an integration that can 
 }
 ```
 
-## 10. Job-to-job — cross-namespace chain
+## 11. Job-to-job — cross-namespace chain
 
 A streaming job in a Flink cluster feeds an in-memory topic consumed by a job in a separate Beam pipeline. Neither job is in the namespace of the other, and the topic is not modeled as a dataset. The Beam pipeline emits the cross-namespace chain on its RunEvent:
 
@@ -298,7 +340,7 @@ A streaming job in a Flink cluster feeds an in-memory topic consumed by a job in
 
 The source job (`flink://ingest/event_normalizer`) is in a different namespace from both the event's job and the target entry. Without explicit `(namespace, name)` on both ends, the chain could not be expressed.
 
-## 11. Structured-to-unstructured — known source columns, no target schema
+## 12. Structured-to-unstructured — known source columns, no target schema
 
 Reading specific columns from a structured table and writing to an unstructured file (e.g., a PDF). The target has no schema, so there is no `fields` map — but we can still express that the output came from specific source columns:
 
--- a/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
+++ b/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
@@ -4,17 +4,19 @@ This file contains examples for [ExplicitLineageProposal.md](ExplicitLineageProp
 
 ## 1. Dataset-level lineage on RunEvent
 
-ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. LineageRunFacet on the run:
+ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. `LineageJobFacet` on the job:
 
 ```json
 {
   "eventType": "COMPLETE",
-  "run": {
-    "runId": "abc-123",
+  "run": { "runId": "abc-123" },
+  "job": {
+    "namespace": "http://etl-server:8080",
+    "name": "bulk_etl",
     "facets": {
       "lineage": {
         "_producer": "https://example.com/etl",
-        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
         "entries": [
           { "namespace": "postgresql://warehouse:5432", "name": "table_c", "type": "DATASET",
             "inputs": [{ "namespace": "postgresql://warehouse:5432", "name": "table_a", "type": "DATASET" }] },
@@ -24,7 +26,6 @@ ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. LineageRunF
       }
     }
   },
-  "job": { "namespace": "http://etl-server:8080", "name": "bulk_etl" },
   "inputs": [
     { "namespace": "postgresql://warehouse:5432", "name": "table_a" },
     { "namespace": "postgresql://warehouse:5432", "name": "table_b" }
@@ -43,12 +44,13 @@ ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. LineageRunF
 This example shows how the new facets can replace ColumnLevelLineageDatasetFacet.
 
 ```json
-"run": {
-  "runId": "abc-123",
+"job": {
+  "namespace": "http://etl-server:8080",
+  "name": "analytics_etl",
   "facets": {
     "lineage": {
       "_producer": "...",
-      "_schemaURL": "...",
+      "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
       "entries": [
         { "namespace": "postgresql://analytics:5432", "name": "output", "type": "DATASET",
           "fields": {
@@ -178,8 +180,8 @@ Conceptually, the event still establishes:
 transactions (DATASET) ──> fraud_detector (JOB)
 ```
 
-That edge is implicit in the RunEvent boundary, not repeated in `LineageRunFacet`. 
-OpenLineage consumer may use that to visualize the relationship.
+That edge is implicit in the RunEvent boundary, not repeated in a lineage facet.
+An OpenLineage consumer may use it to visualise the relationship.
 
 ## 7. Data generator — job produces data without tracked input
 
@@ -211,17 +213,19 @@ That edge is implicit in the RunEvent boundary, not repeated with a `LineageJobI
 
 The same extract task as in previous example, but the producer knows that `total` was derived via a `sum()` aggregation over the (untracked) external API response. 
 There is still no tracked upstream dataset, so the entry's top-level `inputs` is `[]`. 
-The per-field detail names an `LineageJobInput` (`type: JOB` with no `namespace`/`name`), which resolves to the event's own job, and carries the transformation:
+The per-field detail names a `LineageJobInput` (`type: JOB` with no `namespace`/`name`), which resolves to the event's own job, and carries the transformation:
 
 ```json
 {
   "eventType": "COMPLETE",
-  "run": {
-    "runId": "abc-789",
+  "run": { "runId": "abc-789" },
+  "job": {
+    "namespace": "airflow://prod",
+    "name": "data_pipeline.extract_task",
     "facets": {
       "lineage": {
         "_producer": "https://example.com/extract",
-        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
         "entries": [
           { "namespace": "postgres://prod", "name": "extracted_data", "type": "DATASET",
             "inputs": [],
@@ -237,7 +241,6 @@ The per-field detail names an `LineageJobInput` (`type: JOB` with no `namespace`
       }
     }
   },
-  "job": { "namespace": "airflow://prod", "name": "data_pipeline.extract_task" },
   "inputs": [],
   "outputs": [
     { "namespace": "postgres://prod", "name": "extracted_data" }
@@ -280,17 +283,19 @@ Note that `runId` is omitted on both ends — declared lineage describes the job
 
 ## 10. Job-to-job — observed chain on RunEvent
 
-The same data flow as Example 8, observed at runtime by an integration that can identify both the upstream and downstream runs. `runId` is populated on each end, binding the chain to specific executions:
+The same data flow as Example 9, observed at runtime by an integration that can identify both the upstream and downstream runs. `runId` is populated on each end, binding the chain to specific executions:
 
 ```json
 {
   "eventType": "COMPLETE",
-  "run": {
-    "runId": "run-b-001",
+  "run": { "runId": "run-b-001" },
+  "job": {
+    "namespace": "airflow://prod",
+    "name": "dag_b.task_1",
     "facets": {
       "lineage": {
         "_producer": "https://example.com/airflow",
-        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
         "entries": [
           { "namespace": "airflow://prod", "name": "dag_b.task_1", "type": "JOB",
             "runId": "run-b-001",
@@ -303,7 +308,6 @@ The same data flow as Example 8, observed at runtime by an integration that can 
       }
     }
   },
-  "job": { "namespace": "airflow://prod", "name": "dag_b.task_1" },
   "inputs": [],
   "outputs": []
 }
@@ -316,12 +320,14 @@ A streaming job in a Flink cluster feeds an in-memory topic consumed by a job in
 ```json
 {
   "eventType": "COMPLETE",
-  "run": {
-    "runId": "beam-run-77",
+  "run": { "runId": "beam-run-77" },
+  "job": {
+    "namespace": "beam://analytics",
+    "name": "enrichment_pipeline",
     "facets": {
       "lineage": {
         "_producer": "https://example.com/beam",
-        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
         "entries": [
           { "namespace": "beam://analytics", "name": "enrichment_pipeline", "type": "JOB",
             "runId": "beam-run-77",
@@ -333,7 +339,6 @@ A streaming job in a Flink cluster feeds an in-memory topic consumed by a job in
       }
     }
   },
-  "job": { "namespace": "beam://analytics", "name": "enrichment_pipeline" },
   "inputs": [],
   "outputs": []
 }
@@ -348,12 +353,14 @@ Reading specific columns from a structured table and writing to an unstructured 
 ```json
 {
   "eventType": "COMPLETE",
-  "run": {
-    "runId": "abc-012",
+  "run": { "runId": "abc-012" },
+  "job": {
+    "namespace": "analytics",
+    "name": "event_processor",
     "facets": {
       "lineage": {
         "_producer": "https://example.com/export",
-        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
         "entries": [
           { "namespace": "file://", "name": "R.pdf", "type": "DATASET",
             "inputs": [
@@ -365,7 +372,6 @@ Reading specific columns from a structured table and writing to an unstructured 
       }
     }
   },
-  "job": { "namespace": "analytics", "name": "event_processor" },
   "inputs": [
     { "namespace": "postgres://prod", "name": "X" }
   ],
--- a/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
+++ b/proposals/explicit_lineage_facet/ExplicitLineageExamples.md
@@ -1,0 +1,333 @@
+# Explicit Lineage: Examples
+
+This file contains examples for [ExplicitLineageProposal.md](ExplicitLineageProposal.md). See [ExplicitLineageSchema.md](ExplicitLineageSchema.md) for the proposed schema.
+
+## 1. Dataset-level lineage on RunEvent (solves #4359)
+
+ETL job reads A,B and writes C,D. Only A->C and B->D are real edges. LineageRunFacet on the run:
+
+```json
+{
+  "eventType": "COMPLETE",
+  "run": {
+    "runId": "abc-123",
+    "facets": {
+      "lineage": {
+        "_producer": "https://example.com/etl",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "entries": [
+          { "namespace": "postgresql://warehouse:5432", "name": "table_c", "type": "DATASET",
+            "inputs": [{ "namespace": "postgresql://warehouse:5432", "name": "table_a", "type": "DATASET" }] },
+          { "namespace": "postgresql://warehouse:5432", "name": "table_d", "type": "DATASET",
+            "inputs": [{ "namespace": "postgresql://warehouse:5432", "name": "table_b", "type": "DATASET" }] }
+        ]
+      }
+    }
+  },
+  "job": { "namespace": "http://etl-server:8080", "name": "bulk_etl" },
+  "inputs": [
+    { "namespace": "postgresql://warehouse:5432", "name": "table_a" },
+    { "namespace": "postgresql://warehouse:5432", "name": "table_b" }
+  ],
+  "outputs": [
+    { "namespace": "postgresql://warehouse:5432", "name": "table_c" },
+    { "namespace": "postgresql://warehouse:5432", "name": "table_d" }
+  ]
+}
+```
+
+`inputs`/`outputs` remain as the carrier for dataset facets, as the fallback for older consumers, and as the event boundary for the job's direct input/output participation.
+
+## 2. Column-level lineage on RunEvent (subsumes CLL)
+
+```json
+"run": {
+  "runId": "abc-123",
+  "facets": {
+    "lineage": {
+      "_producer": "...",
+      "_schemaURL": "...",
+      "entries": [
+        { "namespace": "postgresql://analytics:5432", "name": "output", "type": "DATASET",
+          "fields": {
+            "total": {
+              "inputs": [
+                { "namespace": "postgresql://analytics:5432", "name": "input", "type": "DATASET", "field": "amount",
+                  "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION" }] }
+              ]
+            }
+          }
+        }
+      ]
+    }
+  }
+}
+```
+
+## 3. Mixed granularity (dataset + column + dataset-wide ops)
+
+```json
+"entries": [
+  { "namespace": "postgresql://analytics:5432", "name": "output_X", "type": "DATASET",
+    "inputs": [
+      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET" },
+      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET", "field": "group_by_col",
+        "transformations": [{ "type": "INDIRECT", "subtype": "GROUP_BY" }] }
+    ],
+    "fields": {
+      "col1": {
+        "inputs": [
+          { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET", "field": "col_a" }
+        ]
+      },
+      "col2": {
+        "inputs": [
+          { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET" }
+        ]
+      }
+    }
+  }
+]
+```
+
+This single entry combines:
+- **Entity-level**: output_X comes from input_A
+- **Dataset-wide op**: group_by_col affects entire output
+- **Column mapping**: col_a -> col1
+- **Partial info**: col2 comes from input_A (source column unknown)
+
+## 4. Dataset-to-dataset (job-less, on DatasetEvent)
+
+A VIEW derives from base tables. No job involved. LineageDatasetFacet on the dataset:
+
+```json
+{
+  "eventType": "COMPLETE",
+  "eventTime": "2026-03-25T10:00:00.000Z",
+  "dataset": {
+    "namespace": "postgresql://warehouse:5432",
+    "name": "public.customer_view",
+    "facets": {
+      "lineage": {
+        "_producer": "https://example.com/catalog",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageDatasetFacet.json",
+        "inputs": [
+          { "namespace": "postgresql://warehouse:5432", "name": "public.customers", "type": "DATASET" },
+          { "namespace": "postgresql://warehouse:5432", "name": "public.orders", "type": "DATASET" }
+        ]
+      }
+    }
+  }
+}
+```
+
+> Note: no `namespace`/`name`/`type` on the facet itself — the target is implicit from the dataset it's attached to.
+
+## 5. Static job lineage (on JobEvent)
+
+A catalog declares what a job is designed to do. LineageJobFacet on the job:
+
+```json
+{
+  "job": {
+    "namespace": "http://airflow:8080",
+    "name": "daily_etl",
+    "facets": {
+      "lineage": {
+        "_producer": "https://example.com/catalog",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
+        "entries": [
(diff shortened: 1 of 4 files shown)


proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+  "allOf": [{ "$ref": "#/$defs/DatasetFacet" }],
+  "properties": {
+    "inputs": {
+      "type": "array",
+      "description": "Dataset-level source inputs. When a source includes a 'field' property, it represents a dataset-wide operation (e.g., GROUP BY) where that source column affects the entire target dataset.",
+      "items": {
+        "$ref": "#/$defs/LineageInput"
+      }
+    },
+    "fields": {
```
jakub-moravec, line 77: @mobuchowski, a technical nit, not impacting the ability to express lineage (I think), but

... is there a specific reason to have `fields` as an array within the entry? Why is it not just an optional field? 

The current proposal handles mixed granularity like this: 

```
"entries": [
  { "namespace": "postgresql://analytics:5432", "name": "output_X", "type": "DATASET",
    "inputs": [
      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET" },
      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET", "field": "group_by_col",
        "transformations": [{ "type": "INDIRECT", "subtype": "GROUP_BY" }] }
    ],
    "fields": {
      "col1": {
        "inputs": [
          { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET", "field": "col_a" }
        ]
      },
      "col2": {
        "inputs": [
          { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET" }
        ]
      }
    }
  }
]
```

Wouldn't it be better to split them into separate entries and have 

```
{ "namespace": "postgresql://analytics:5432", "name": "output_X", "type": "DATASET",
    "inputs": [
      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET" },
      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET", "field": "group_by_col",
        "transformations": [{ "type": "INDIRECT", "subtype": "GROUP_BY" }] }
    ]
}, 
{ "namespace": "postgresql://analytics:5432", "name": "output_X", "field": "col1", "type": "DATASET",
    "inputs": [
      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET", "field": "col_a" }
    ]
}, 
{ "namespace": "postgresql://analytics:5432", "name": "output_X", "field": "col2", "type": "DATASET",
    "inputs": [
      { "namespace": "postgresql://analytics:5432", "name": "input_A", "type": "DATASET" }
    ]
}
```

That would flatten the JSON tree a little bit and make it more readable IMHO.

mobuchowski: We repeat the output namespace/name/type a lot then - we can have 100s and 1000s columns per dataset - IMO `fields` compresses it a bit.

jakub-moravec: IMHO it's a small price to pay for a simpler schema

I'd be interested in others PoV on this

mobuchowski: @harels @nantunes @kacpermuda what do you think?

kacpermuda: From the consumer perspective you can consume both of these in a similar way / filter out what you don't support easily. The repeating of namespace/name/type is redundant, so I was wondering @jakub-moravec what are the arguments for it? I would say for me it's less readable, not more, so was wondering if there are any clear / major gains with the second approach?

jakub-moravec: On the consumer side, the proposed alternative is simpler to construct additively in a situation when you don't have information about lineage for all the targets (columsn) in a single moment (which happens to be our case). 

But the primary reason why I raised this was seeing the inputs on two levels, which IMO makes the schema more complex - to understand and to implement, if somebody chooses to do that on their own (which happens to be our case as well at this point).

proposals/explicit_lineage_facet/ExplicitLineageAdoptionGuideline.md
```diff
@@ -0,0 +1,171 @@
...
+| **LineageRunFacet** | run.facets.lineage | Runtime observed flow |
+| **LineageJobFacet** | job.facets.lineage | Design-time declared flow |
+| **LineageDatasetFacet** | dataset.facets.lineage | Structural derivation (views, synonyms, reports) |
+
+**Key aspect:** Uses existing facet mechanism — no core model changes.
+
+## What It Solves
+
+... there are more situations where the current spec falls short:
+
```
julienledem, line 24: It would be useful to have a one line english description for each of the examples bellow to remove ambiguity.
Something like:
1. output table_c is derived from input table_a. output table_d is derived from input table_b
2. The dataset my_view is a logical view derived from base_table
3. the field column_A in output table_a is derived from the field column_A in input table_b. The output file_c is derived from the field column_B in input table_b

julienledem: Also, all examples are Dataset type. Do we need the Job Type? (I'll comment in the other file)

proposals/explicit_lineage_facet/ExplicitLineageAdoptionGuideline.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+## When to Use
+
+### Producers: Use `compatibility = both`
+
+**Immediately:** Start emitting new facets while maintaining backward compatibility.
+
+```yaml
+openlineage.lineage.compatibility: both
```
julienledem, line 104: Since the two possible cases are "keep emitting CLL facet" or "stop emitting CLL facet", I would suggest that this is called openlineage.ColumnLineageFacet.compatibility: true ?

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+
+The existing `ColumnLineageDatasetFacet` (CLL) solves this at column granularity, but leaves a gap at dataset granularity: integrators often know "Table A derived Table B" without column-level detail. If they can't provide CLL, they fall back to the cartesian product.
+
+Additionally, the current model cannot express:
+
+- **Dataset-to-dataset derivation** — no mechanism for views, aliases, or manually documented lineage without an intermediate job
+- **Mixed granularity** — CLL is all-or-nothing per output; no way to provide dataset-level for some outputs and column-level for others
+- **Job-to-job data flow** — no way to express that one job feeds data to another without an intermediate dataset (e.g., non-materialized views, stored procedures calling functions, in-memory data passing)
```
julienledem, line 19: per line 397 bellow, this is out of scope?

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+}
+```
+
+## Why `type: JOB` Is Needed: Sinks and Generators
+
+The current OpenLineage model assumes every lineage edge connects **dataset to dataset**, with a job in the middle. But two common cases break this assumption:
+
+### Sinks: jobs that consume data without a tracked output
+
+A **sink** is a job that reads data but doesn't write to any tracked dataset — for example, a fraud detection job that reads a `transactions` table, runs validation, and sends alerts (not modeled as a dataset). In the current spec this produces `inputs: [transactions], outputs: []` — no lineage edge exists because there's nothing to connect the input *to*. The data consumption is invisible in the lineage graph.
```
julienledem, line 369: Why do you say"The data consumption is invisible in the lineage graph." here ? The event that has those inputs and the empty outputs list also has an explicit job attribute that identifies the job reading those inputs. So it is already in the event and can be made visible.
Why do we need to duplicate this information somewhere else?

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+
+With `type: JOB` as a lineage **target** (`LineageJobEntry`), we can express: "data flows into the fraud_detector job itself." The job becomes the terminal node:
+
+```
+transactions (DATASET) ──> fraud_detector (JOB)
+```
```
julienledem, line 375: This is already captured in the current structure.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+
+### Generators: jobs that produce data without a tracked input
+
+A **generator** is the inverse: a job that produces data without reading from any tracked dataset — for example, an extract task that pulls from an external API and passes data in-memory to the next processing step. With `type: JOB` as a lineage **source** (`LineageJobInput`), we can express: "the data flowing into this dataset came from the extract_task job":
+
+```
+extract_task (JOB) ──> transformed_data (DATASET)
+```
```
julienledem, line 383: This symmetric scenario that is also already captured.

I would add that the job is always in the lineage, not only when it is the sole input or output. Singling out the job is in the lineage only when there is no input/output is not actually accurate.

For example: A field in the output could be generated and not derived from the input.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+
+### Alternatives considered
+
+Without `type: JOB`, the alternatives are all problematic:
+
+- **Fake temporary datasets** — pollutes the catalog with entities that don't exist; requires two independent jobs to agree on the same fake dataset name
+- **Leave it out of the lineage facet** — this is a **regression**, not an acceptable trade-off. The current spec already captures sink/generator relationships implicitly through the `inputs`/`outputs` arrays. A new explicit lineage model that loses information the old model could express is strictly worse. We should not ship a feature that degrades existing capabilities.
+- **Rely on inputs/outputs arrays for the missing edges** — forces consumers to inspect both the lineage facet AND inputs/outputs for a complete picture, defeating the purpose of explicit lineage as the single source of truth
+
+By allowing `type: JOB` in lineage entries, the lineage facet remains **self-contained and complete**. Note that full job-to-job data flow chains (job -> job -> job) are out of scope for this version — `type: JOB` is scoped to sink and generator patterns only.
```
julienledem, line 393: This doc in its current form has inconsistencies as we both say this is out of scope and use it as a goal in other places.
We should pick which one it is and be consistent.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+
+```
+extract_task (JOB) ──> transformed_data (DATASET)
+```
+
+### Alternatives considered
+
+Without `type: JOB`, the alternatives are all problematic:
+
+- **Fake temporary datasets** — pollutes the catalog with entities that don't exist; requires two independent jobs to agree on the same fake dataset name
```
julienledem, line 389: isn't it the case that is out of scope bellow?  (line 393)
Job -> Job with no dataset.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+- **Leave it out of the lineage facet** — this is a **regression**, not an acceptable trade-off. The current spec already captures sink/generator relationships implicitly through the `inputs`/`outputs` arrays. A new explicit lineage model that loses information the old model could express is strictly worse. We should not ship a feature that degrades existing capabilities.
```
julienledem, line 390: Not adding something is not a regression since it was not there. 
This is not something captured in CLL facet.
However the spec already captures it, IMO we should not define it twice.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+- **Rely on inputs/outputs arrays for the missing edges** — forces consumers to inspect both the lineage facet AND inputs/outputs for a complete picture, defeating the purpose of explicit lineage as the single source of truth
```
julienledem, line 391: I understand the purpose as tracking dataset to dataset and mixed granularity (dataset/field) that is missing today.
The whole spec is about capturing how a job transforms input datasets into output datasets. I don't think it is a good goal to force more concepts into that one facet so that you don't have to look at the rest of the spec.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+    "type": {
+      "type": "string",
+      "enum": ["JOB"],
+      "description": "Must be JOB"
+    },
+    "runId": {
+      "type": "string",
+      "format": "uuid",
+      "description": "Optional. The specific run ID of the source job, when the lineage is tied to a particular execution."
+    }
```
julienledem, line 330: Here is how I am understanding this.
There is a very different meaning whether this is an input to a job or a dataset.

The examples in this doc show when a Job is the input or output of datasets when there is not other input or output (respectively). In those cases, the only applicable   (namespace, Name, RunID) fields are the ones already defined in the job and run fields.

I did not find examples of job to job and in this case it is unclear if the current job should be the input or the output or if both are accepltable.

I would recommend to remove from the spec the notion of "the job is the output when there is no other output" or "the job is the input when there is no other input" as this is  both incorrect semantically (the job is always part of the input or output) and makes the model inconsistent and redundant with the existing job, input and ouput fields.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+{
+  "eventType": "COMPLETE",
+  "run": {
+    "runId": "abc-456",
+    "facets": {
+      "lineage": {
+        "_producer": "https://example.com/validation",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "entries": [
+          { "namespace": "validation", "name": "fraud_detector", "type": "JOB",
```
julienledem, line 586: as I understand the spec, the (ns,name) here *must* be the same as the job entry on line 595 bellow.
Therefore we can remove it.
Would it make more sense to have a field here that is "generatedDatasets" or "datasetsWithNoInput"

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+  "run": {
+    "runId": "abc-789",
+    "facets": {
+      "lineage": {
+        "_producer": "https://example.com/etl",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageRunFacet.json",
+        "entries": [
+          { "namespace": "postgres://prod", "name": "extracted_data", "type": "DATASET",
+            "inputs": [
+              { "namespace": "airflow://prod", "name": "data_pipeline.extract_task", "type": "JOB" }
```
julienledem, line 619: the (NS, name) here has to be the same as defined in the job field on line 626 bellow, so it doesn't actually bring any additional information.
It would make more sense to define inputs: [] as "there are no inputs" or add an explicit field: generated:true

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+The combination of entry type, field presence on targets, and field presence on sources enables all granularity levels:
+
+**Dataset targets (`type: DATASET`):**
+
+| Target has fields? | Source has field? | Meaning | Example |
+|---|---|---|---|
+| — (entity-level inputs) | No | Dataset-level lineage | `output_table <- input_table` |
+| — (entity-level inputs) | Yes | Dataset-wide operation (e.g., GROUP BY) | `output_table <- input.group_by_col` |
+| Yes (in fields map) | Yes | Column-level lineage | `output.total <- input.amount` |
+| Yes (in fields map) | No | Partial column lineage (source column unknown) | `output.region <- lookup_table` |
```
julienledem, line 406: I'm not sure that we need a matrix here as the dataset level lineage and the column level lineage both make sense independently in the target.
dataset to dataset and field to field are fairly ituitive.
Then the meaning of a dataset being derived form a column and a column from a dasaset are the one that need a bit of explanation. In particular clarifying whether we don't know the columns, whether there is no schema (file), or if we mean all the columns.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,844 @@
...
+    "name": {
+      "type": "string",
+      "description": "The name of the target dataset"
+    },
+    "type": {
+      "type": "string",
+      "enum": ["DATASET"],
+      "description": "Must be DATASET"
+    },
+    "inputs": {
```
julienledem, line 174: This could also be an outputFacet in each ouptut dataset that would be consistent with the existing ColumnLineage Facet.

julienledem (commented):
Thank you for you effort, putting this all in writing.
I have commented in line. 
In particular, the job lineage looks out of place to me.
I have commented more specifically in the doc.
Please, let me know what you think.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: The idea is to have it out of scope for the first iteration, but then add this option soon. Just to make the core proposal easier to reason about. I'm myself not quite sure it's having that effect, might be simpler to have it in the scope from the beginning

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: Agree the wording could be more deliberate here, but technically, you can only express this lineage with the current spec on the dataset level, not on the column level - that's the real gap.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: Only on dataset level

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: Once we get to job-to-job, one of the jobs involved can have a different namespace. 

Again, it doesn't seem to be useful to exclude job-to-job from the initial scope, as the spec is trying to be ready for it, but is missing the argumentation that explains it in these cases.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: It is currently supported on dataset level, not the column level, so there is a real gap. The idea is for the new set of facets to address all cases that occur in common scenarios (and talking for IBM/Manta, we do encounter all of these and need to deal with them)

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: Why not? 

That is what makes the spec usable in complex scenarios, both from the producer and consumer perspectives. 

If different lineage situations are covered by different concepts in the spec, there is no wide consensus and vendors don't really integrate directly with the spec, but rather with how another specific vendor is using the spec.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: Out of scope for current proposal, but a goal in the future. Not sure how to word that better?

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: Instead of thinking what we already capture, we're trying to make the new facets the only source of truth for lineage - which means sometimes we'll need to duplicate something. But in those cases we want to have rules of precedence - in this case, follow the lineage facet, if it does not exist, fall back to inputs/outputs.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: >I would add that the job is always in the lineage, not only when it is the sole input or output. Singling out the job is in the lineage only when there is no input/output is not actually accurate.

I'd say it's in the _context_ of the lineage, but consumer is free to model those as they wish - with both `dataset -> job -> dataset` edges, or `dataset -> dataset` if they don't care about it.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: If we removed it, we'd have to readd it when we add proper job-to-job lineage

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: The goal was to be able to specify the column-level dependencies on job in a way that would be consistent with potential future additions like job to job lineage. Current spec can't express that.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: Our goal is to deprecate the CLL facet, as the new facet is the superset of it - I don't think we need to be concerned about the compatibility with previous one. 

With moving the facet to respectable place in each event type, we have a single consistent place in each event that we can read lineage from. If we want to express the lineage only in dataset facets, we lose the ability to express job to job lineage in the future.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: I've changed the proposal to explicitely keep job-to-job in scope of this doc.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: Updated this. IMO while we can derive this on dataset level, the intention here is to keep the facet as the source of truth for lineage, rather than reconstructing this from job x inputs/outputs when we can.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: Removed that wording - still; the ability to express that on column-level lineage and keeping the lineage in scope of the lineage facet stands.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: Update: I've readded the job-to-job lineage back to the spec.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: With readding the job-to-job in scope, the targeting of particular job is required now

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,933 @@
...
+
+We introduce three new facets — **LineageRunFacet**, **LineageJobFacet**, and **LineageDatasetFacet** — each carrying explicit lineage information and each restricted (by spec text, not JSON Schema) to a single event type:
+
+| Facet | Entity it lives on | Allowed event type | Semantics |
+|---|---|---|---|
+| `LineageRunFacet` | Run | RunEvent only | Observed data flow during this specific run |
+| `LineageJobFacet` | Job | JobEvent only | Declared/static data flow for this job definition |
+| `LineageDatasetFacet` | Dataset | DatasetEvent only | Structural derivation of this dataset (views, aliases, manual docs) |
+
+**Why the one-facet-per-event-type restriction?** If a RunEvent carried both a LineageRunFacet (on the run) and LineageDatasetFacet (on output datasets), the semantics would be ambiguous — which takes precedence? Do they merge? Rather than define complex interaction rules, we disallow the combination entirely. Each event type has exactly one place where lineage lives.
```
julienledem: IMO, we still need to define this semantics as we can get a job event when a job is created and different lineage when subsequent run events happen.
We could have a last-write-wins strategy or just consider the declared lineage different from the runtime lineage as the job might dynamically add more lineage at runtime (that is not "declared") or not use some of the declared lineage in some cases.

If the LineageDatasetFacet is purely about views and Dataset to Dataset lineage then it does not collide with the LineageJobFacet. It can be constrained to only report lineage *into* the Dataset.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,933 @@
...
+
+This restriction may be relaxed in a future version with clear type-of-lineage or hierarchy semantics defined (see [Future Evolution](#future-evolution)).
+
+**LineageDatasetFacet** is the natural one. It sits on an output dataset and describes what feeds into that specific dataset — the same position as CLL. The target entity is implicit (the dataset the facet is attached to). This is a direct evolution of CLL.
```
julienledem: If it's for views then it is only about dataset to dataset lineage, correct?

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,933 @@
...
+```json
+"LineageDatasetFacet": {
+  "type": "object",
+  "description": "Explicit lineage for this dataset. Describes which source entities feed into it, at entity and/or column granularity. Supersedes ColumnLineageDatasetFacet.",
+  "allOf": [{ "$ref": "#/$defs/DatasetFacet" }],
+  "properties": {
+    "inputs": {
+      "type": "array",
+      "description": "Dataset-level source inputs. When a source includes a 'field' property, it represents a dataset-wide operation (e.g., GROUP BY) where that source column affects the entire target dataset.",
+      "items": {
```
julienledem: Am I correct to understand that in this case the items are always Datasets?

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
julienledem: I think the comment still applies.  In particular: "In those cases, the only applicable (namespace, Name, RunID) fields are the ones already defined in the job and run fields."

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
julienledem: right, but in the case of job to dataset or dataset to job, the (ns,name) here must be the same as the job entry.
This is where this modeling needs to be improved as the semantics of one case polute the semantics of the other. It doesn't quite make sense to repeat information that is already implicit.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
julienledem: We'd have to re-add it only because we reuse the same LineageInput definition in both cases but both cases do not have the same needs.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
julienledem: I think there are a dundamental disagreement here that cause a fair a bit of back and forth:
 - deprecating CLL is not a goal in itself. It's a result if we decide to replace it with something else. But we can also evolve it to have mixed semantics as requested in this doc.
 - it is not a requirement to have all lineage in one place. The job to job lineage is also already expressed in the JobDependenciesRunFacet. We can add a JobFacet for job to job lineage.

Starting from the principle that those two concepts (Dataset to Dataset and Job to Job) *must* be in one model is forcing the solution in a specific direction but we should discuss the alternatives rather than take it as a fixed requirement.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,933 @@
...
+- `JobDependenciesRunFacet` expresses scheduling / control-flow dependencies — Job A runs Monday, Job B runs Tuesday because the schedule says so, regardless of whether any data passes between them. It remains the home for control flow.
+- `ParentRunFacet` expresses containment — task X is part of DAG Y. Job-to-job lineage between sibling tasks within the same DAG is orthogonal to that hierarchy.
+
+`JobEvent` is the natural home for declared job-to-job lineage ("this job is designed to feed data to that job"); `RunEvent` carries observed job-to-job lineage when a runtime integration can identify both ends of the chain.
+
+### Why one mechanism for all three cases
+
+A reasonable counter-question: do we need `type: JOB` at all? Sinks and generators are already implicit in the event's `inputs`/`outputs` arrays — couldn't a flag (`generated: true`, or simply `inputs: []` with no JOB type) express them?
+
+It could, if sinks and generators were the only `type: JOB` cases. They are not. Once job-to-job chains are in scope, the JOB-type identity is irreducible:
```
julienledem: This is a self-referential argument.
because we put them all in one mechanism then we must have job:entity.

We could also have a mechanism for Dataset to Dataset (and fields) and a separate mechanism for Job to Job ( that we kind of have with JobDependenciesRunFacet that could be extended to a JobFacet)

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
```diff
@@ -0,0 +1,933 @@
...
+
+- A chain involves two jobs that may both differ from the event's job, and may live in different namespaces from each other and from the event. There is no way to flag this with `generated: true` — the spec needs `(namespace, name)` for both ends.
+- Once that machinery exists for chains, using a different mechanism for sinks and generators would split a single semantic concept ("a job participates in lineage") into three special cases, requiring three parser branches and three sets of consumer rules.
+
+The early working group discussion explicitly flagged that having JOB sometimes implicit (sinks/generators) and sometimes explicit (chains) "causes inconsistency and ambiguity" and recommended consistent explicit JOB. We adopt that recommendation: one mechanism, one shape, three uses.
```
julienledem: we should recognize the downside of this constraint: When we represent Job to Dataset the Job must be the same job as defined in the Job/Run event and is redundant. Which creates a model that will create confusion: "Can I put another job here? does it make sense?". "Why do I need to repeat the job name/ns here that is already implied?".

julienledem (commented):
We need to address what the job name/namespace in JobInput means relative to the Job Name/namespace in the Job/Run event.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: >We could also have a mechanism for Dataset to Dataset (and fields) and a separate mechanism for Job to Job ( that we kind of have with JobDependenciesRunFacet that could be extended to a JobFacet)

Skipping the `JobDependenciesRunFacet` - addressed that in separate topic already.

But WDYT then of having separate `inputDatasets` and `inputJobs` at single facet, rather than splitting the solution into two?

Case with dataset input, which is similar to regular CLL today:
```
  {
    "lineage": {
      "entries": [
        {
          "namespace": "spark://prod-cluster",
          "name": "warehouse.user_summary",
          "type": "DATASET",
          "inputDatasets": [
            { "namespace": "spark://prod-cluster", "name": "warehouse.users" }
          ],
          "fields": {
            "user_id": {
              "inputDatasets": [
                { "namespace": "spark://prod-cluster", "name": "warehouse.users", "field": "id",
                  "transformations": [{ "type": "DIRECT", "subtype": "IDENTITY" }] }
              ]
            },
            "total_orders": {
              "inputDatasets": [
                { "namespace": "spark://prod-cluster", "name": "warehouse.users", "field": "order_history",
                  "transformations": [{ "type": "DIRECT", "subtype": "AGGREGATION" }] }
              ]
            }
          }
        }
      ]
    }
  }
```

Case with job lineage, something using Airflow XCOM to send data between tasks:

```
  {
    "explicitLineage": {
      "entries": [
        {
          "namespace": "airflow://prod",
          "name": "exec_reporting.publish_dashboard",
          "type": "JOB",
          "inputJobs": [
            { "namespace": "airflow://prod", "name": "exec_reporting.compute_kpi" }
          ]
        }
      ]
    }
  }
```

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
mobuchowski: For subsequent run events, I would consider this defined - as they would only carry `LineageRunFacet`.

For run vs dataset, I think we could do what you propose - declare that this is different runtime vs design - but in general do we have to prescribe what happens? Usually, we leave that to lineage consumers.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: @julienledem 

> But we can also evolve it to have mixed semantics as requested in this doc.

The group agreed quite early in the discussion that we cannot extend CLL to achieve what this proposal is trying to achieve. It's all recorded in the Slack discussion and the Google docs, but one example for all: CLL is on the output dataset. What if there is no output? 

The intention was to find a single approach to document lineage in the different scenarios, so that the spec is unambiguous, concise, simple to understand, and interpret. 

While I understand the pushback on depricating something that has been in the spec fo quite some time, I think the proposed migration path that relies on the client library handling most of the work mitigates the downsides of that ... and trying to keep and extend CLL would force us into a very sub-optimal solution. 


> it is not a requirement to have all lineage in one place

I think it should be. From consumer perspective, this makes a big difference in how complex the consumer needs to be. If it's not in one place, there is a lot of combinations to support. 

> The job to job lineage is also already expressed in the JobDependenciesRunFacet.

I strongly disagree with this - the facet is designed to represent timing/scheduling dependencies, not lineage. These are distinct, and if there isn't concensus on that, I think we need to discuss these fundamentals first.

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: @julienledem If I want to document that another job has downstream lineage to _this_ job, I would use namespace, name, and RunID of the other job. 

For example: StoredProcedure1 reads Column1 from Table1, applies Function1 on the value and writes it to Column1 in Table2
Lineage: Table1.Column1 -> Function1(.InputParam1) -> StoredProcedure1(.Column1) -> Table2.Column1

proposals/explicit_lineage_facet/ExplicitLineageProposal.md
jakub-moravec: Right, but in job-to-job the values can be different. I don't think the need to repeat some information is as big of a problem as not being able to express some types of lineage.

kacpermuda (approved):
Looks good !

dolfinus (approved):

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
```diff
@@ -0,0 +1,383 @@
...
+    "lineage": {
+      "_producer": "...",
+      "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
+      "entries": [
+        { "namespace": "postgresql://analytics:5432", "name": "output", "type": "DATASET",
+          "fields": {
+            "total": {
+              "inputs": [
+                { "namespace": "postgresql://analytics:5432", "name": "input", "type": "DATASET", "field": "amount",
+                  "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION" }] }
```
jakub-moravec: @mobuchowski the INDIRECT / AGREGATION combination is not allowed / does not make sense https://openlineage.io/docs/spec/facets/dataset-facets/column_lineage_facet/

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
```diff
@@ -0,0 +1,383 @@
...
+- **Column mapping**: col_a -> col1
+- **Partial info**: col2 comes from input_A (source column unknown)
+
+## 4. Dataset-to-dataset (job-less, on DatasetEvent)
+
+A VIEW derives from base tables. No job involved. LineageDatasetFacet on the dataset:
+
+```json
+{
+  "eventType": "COMPLETE",
```
jakub-moravec: Dataset event should not contain eventType https://openlineage.io/apidocs/openapi/#tag/OpenLineage

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
```diff
@@ -0,0 +1,383 @@
...
+    }
+  }
+}
+```
+
+> Note: no `namespace`/`name`/`type` on the facet itself — the target is implicit from the dataset it's attached to.
+
+## 5. Static job lineage (on JobEvent)
+
+A catalog declares what a job is designed to do. LineageJobFacet on the job:
```
jakub-moravec: I would explicitly add 

"This is treated as design/static lineage because the event does not contain a run section"

I know we document that elsewhere in the docs, but otherwise it is not obvious how this is different from the other examples

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
```diff
@@ -0,0 +1,383 @@
...
+      "lineage": {
+        "_producer": "https://example.com/extract",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
+        "entries": [
+          { "namespace": "postgres://prod", "name": "extracted_data", "type": "DATASET",
+            "inputs": [],
+            "fields": {
+              "total": {
+                "inputs": [
+                  { "type": "JOB", "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION", "description": "sum()" }] }
```
jakub-moravec: Please change type to direct - I know I sent this to you like this, sorry :D

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
```diff
@@ -0,0 +1,383 @@
...
+
+The implicit edge `extract_task (JOB) ──> extracted_data (DATASET)` is unchanged — the identity-less `LineageJobInput` only adds the column-level transformation that event membership could not express.
+
+## 9. Job-to-job — declared chain on JobEvent
+
+A stored procedure `dag_b.task_1` reads its input from upstream `dag_a.task_3` via a non-materialized handoff. No intermediate dataset exists. The catalog declares the design-time data flow on the JobEvent for `dag_b`:
+
+```json
+{
+  "eventType": "COMPLETE",
```
jakub-moravec: No eventType

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
```diff
@@ -0,0 +1,560 @@
...
+    "lineage": {
+      "_producer": "...",
+      "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
+      "entries": [
+        { "namespace": "postgresql://analytics:5432", "name": "output", "type": "DATASET",
+          "fields": {
+            "total": {
+              "inputs": [
+                { "namespace": "postgresql://analytics:5432", "name": "input", "type": "DATASET", "field": "amount",
+                  "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION" }] }
```
jakub-moravec: INDIRECT / AGGREGATION

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
```diff
@@ -0,0 +1,560 @@
...
+| job | job | — | job-to-job chain, no intermediate dataset | `dag_b.task_1 <- dag_a.task_3` |
+
+### Dataset-to-dataset derivation, no job
+
+A view derives from base tables with no job in the picture. Use
+`LineageDatasetFacet` on the dataset in a DatasetEvent:
+
+```json
+{
+  "eventType": "COMPLETE",
```
jakub-moravec: no eventType

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
```diff
@@ -0,0 +1,560 @@
...
+      "lineage": {
+        "_producer": "https://example.com/extract",
+        "_schemaURL": "https://openlineage.io/spec/facets/LineageJobFacet.json",
+        "entries": [
+          { "namespace": "postgres://prod", "name": "extracted_data", "type": "DATASET",
+            "inputs": [],
+            "fields": {
+              "total": {
+                "inputs": [
+                  { "type": "JOB", "transformations": [{ "type": "INDIRECT", "subtype": "AGGREGATION", "description": "sum()" }] }
```
jakub-moravec: Indirect/aggregation

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
```diff
@@ -0,0 +1,560 @@
...
+and the event's own job. The intermediate may be a non-materialized view, a
+stored procedure calling a function, in-memory passing such as Airflow XCom, or a
+streaming handoff through an unmodelled topic.
+
+A `LineageJobEntry` with a `LineageJobInput` states direct job-to-job flow. On a
+JobEvent it is the declared chain, with `runId` omitted:
+
+```json
+{
+  "eventType": "COMPLETE",
```
jakub-moravec: no eventType

jakub-moravec (changes requested):

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
mobuchowski: Fixed to DIRECT/AGGREGATIOn

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
mobuchowski: Thanks, removed.

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
mobuchowski: Added it there.

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
mobuchowski: Changed it

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
mobuchowski: Changed it

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
mobuchowski: Removed

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
mobuchowski: Removed.

proposals/explicit_lineage_facet/ExplicitLineageExamples.md
mobuchowski: Removed.

proposals/explicit_lineage_facet/ExplicitLineageGuide.md
mobuchowski: Removed.

jakub-moravec (approved):

jakub-moravec:
@mobuchowski is the spec part of the website? I cannot find it there

mobuchowski:
@jakub-moravec need this: https://github.com/OpenLineage/OpenLineage/pull/4804 - maybe today :)

jakub-moravec:
I see, I thought it already went with the merge of the proposal. Thanks!!

PR #4471, opened 2026-04-16
spark: fix PathUtils.reconstructDefaultLocation malformed URI

### LINKED ISSUE #4414: [BUG] PathUtils.reconstructDefaultLocation malformed URI

Spark integration

### OpenLineage version

1.45.0

OpenLineage Spark Integration: 1.45.0                                                                                                                           
Apache Spark: 3.5.6
Scala: 2.12                                                                                                                                                     
Java: 11+ (JDK 11 - confirmed by --add-opens flags in spark-submit)                                                                                           
Hadoop: 3.x (EMR 7.x bundled)                                                                                                                                   
Delta Spark: 3.1.0                                                                                                                                              
Apache Iceberg: 1.8.1                                                                                                                                           
Runtime: Amazon EMR (EMR on EC2) + EMR Serverless

### Environment configuration

                                                                                                                                                                
  spark.extraListeners=io.openlineage.spark.agent.OpenLineageSparkListener                                                                                        
  spark.openlineage.transport.type=http                                                                                                                           
  spark.openlineage.transport.url=https://<internal-eventbus-endpoint>                                                                                            
  spark.openlineage.transport.endpoint=/v2/<custom-endpoint>                                                                                                      
  spark.openlineage.facets.disabled=[spark_unknown;schema;Schema;]                                                                            
  spark.openlineage.namespace=<custom-namespace-json>                                                                                                             
  spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension,org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions                                
                                                                                                                                                                  
  Runtime: Amazon EMR (EC2 + Serverless), Hive metastore backed by MySQL/MariaDB, warehouse on S3.

### Deployment details

Amazon EMR (EC2 and EMR Serverless) with external Hive metastore on RDS (MariaDB), data stored on S3.

### Problem details

`PathUtils.reconstructDefaultLocation()` at line 144 constructs a malformed URI when processing DROP TABLE operations via `DropTableVisitor`.                   
                                                                                                                                                                  
  The code calls:                                                                                                                                                 
   
  ```java                                                                                                                                                         
  new Path(warehouse, database + ".db", name)                                                                                                                   
                                                                                                                                                                  
  This invokes Hadoop's Path(String scheme, String authority, String path) constructor instead of the intended hierarchical path construction. When the warehouse
  location is an S3 path (e.g., s3://bucket/prefix), it produces an invalid URI:                                                                                  
                                                                                                                                                                
  s3://bucket/prefix://database.db./table_name                                                                                                                    
   
  Which throws:                                                                                                                                                   
                                                                                                                                                                
  java.lang.IllegalArgumentException: java.net.URISyntaxException:
  Relative path in absolute URI: s3://bucket/prefix://database.db./table_name
                                                                                                                                                                  
  This crashes the output dataset builder, resulting in zero lineage events emitted for the entire Spark application.
                                                                                                                                                                  
  Fix: Replace with chained 2-argument constructors:                                                                                                            

  new Path(new Path(warehouse, database + ".db"), name)

`PathUtils.reconstructDefaultLocation()` should construct a valid hierarchical path: `s3://bucket/prefix/database.db/table_name`.                               
                                                                                                                                                                  
  The 3-argument `Path(String, String, String)` constructor in Hadoop is `Path(scheme, authority, path)`, not `Path(parent, child, grandchild)`. The code         
  mistakenly uses it as the latter, causing the warehouse S3 URI to be interpreted as a scheme, producing a malformed URI.
                                                                                                                                                                  
  The fix is to chain two 2-argument `Path(parent, child)` calls:                                                                                                 
                                                                                                                                                                  
  ```java                                                                                                                                                         
  // Current (broken)                                                                                                                                           
  return new Path(warehouse, database + ".db", name);

  // Fixed
  return new Path(new Path(warehouse, database + ".db"), name);
                                                                                                                                                                  
  This would correctly produce s3://bucket/prefix/database.db/table_name and allow DropTableVisitor to resolve dataset identifiers without crashing.

### How to reproduce

1. Configure a Spark application with OpenLineage 1.45.0 on EMR with an external Hive metastore                                                                 
2. Ensure the Hive warehouse location is an S3 path with a sub-path (e.g., `s3://bucket/some_prefix`)                                                         
3. Run a `DROP TABLE` statement on a Hive table in that warehouse via `sparkSession.sql("DROP TABLE IF EXISTS db_name.table_name")`                             
4. Observe the error in logs:
    
```                                                                                                                                                           
  ERROR PlanUtils - Apply failed:                                                                                                                               
  java.lang.IllegalArgumentException: java.net.URISyntaxException:                                                                                                
  Relative path in absolute URI: s3://bucket/some_prefix://db_name.db./table_name                                                                               
      at org.apache.hadoop.fs.Path.initialize(Path.java:272)
      at org.apache.hadoop.fs.Path.(Path.java:263)                                                                                                                
      at io.openlineage.spark.agent.util.PathUtils.reconstructDefaultLocation(PathUtils.java:144)
      at io.openlineage.spark3.agent.lifecycle.plan.catalog.V2SessionCatalogHandler.getDatasetIdentifier(V2SessionCatalogHandler.java:57)                         
                                                                                                                                                                
  The `DropTableVisitor` crashes, the output dataset builder fails, and no lineage events are emitted for the entire application.                                 
                                                                                                                                                                  
  This does not occur when the warehouse path is a simple root path (e.g., `s3://bucket/`) since the URI construction happens to produce a valid result in that
  case

### Additional details

- The bug is in `PathUtils.java` line 144, which is in the shared module (`integration/spark/shared`), so it potentially affects any catalog handler that calls 
  `reconstructDefaultLocation` — not just `DropTableVisitor` / `V2SessionCatalogHandler`.                                                                                                                                                                                                                                           
- The bug appears to still be present on the `main` branch.                                                                                                                                                                                                                                                                 
- The error is silent from the Spark application's perspective — the job completes successfully but emits zero lineage events. The only indication is `ERROR    
  PlanUtils - Apply failed` in the driver logs, which can be easily missed.                                                                                                                                                                                                                                               
- The `DropTableVisitor` is triggered even when the DROP TABLE is not the primary operation of the pipeline (e.g., a compliance deletion pipeline that drops and
   recreates temp tables). A single failed visitor call prevents lineage emission for the entire application, including unrelated operations like INSERT and SELECT. 

### PR DESCRIPTION
**Pull request by navneetraju**
Fixes https://github.com/OpenLineage/OpenLineage/issues/4414

## Problem

`PathUtils.reconstructDefaultLocation()` uses the 3-argument `Path(String, String, String)` constructor, which is `Path(scheme, authority, path)` — not `Path(parent, child, grandchild)` as intended.

When the warehouse location is an S3 path with a sub-path (e.g., `s3://bucket/prefix`), this produces a malformed URI:

```
s3://bucket/prefix://database.db./table_name
```

Which throws:
```
java.lang.IllegalArgumentException: java.net.URISyntaxException:
Relative path in absolute URI: s3://bucket/prefix://database.db./table_name
```

This crashes the output dataset builder, resulting in zero lineage events for the entire Spark application.

## Fix

Replaced the 3-argument `Path` constructor with chained 2-argument constructors:

```java
// Before (broken)
return new Path(warehouse, database + ".db", name);

// After (fixed)
return new Path(new Path(warehouse, database + ".db"), name);
```
This correctly produces s3://bucket/prefix/database.db/table_name.

# Testing
Added [testReconstructDefaultLocationWithS3Path](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) covering:

- [X] S3 paths with sub-paths (the bug scenario)
- [X] S3 paths without sub-paths
- [X] HDFS paths with sub-paths
- [X] Default database (no .db suffix)
- [X] Single and empty namespace arrays

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
@@ -141,7 +141,7 @@ public class PathUtils {
     }
 
     // /warehouse/mydb.db/mytable
-    return new Path(warehouse, database + ".db", name);
+    return new Path(new Path(warehouse, database + ".db"), name);
   }
 
   public static Optional<URI> getMetastoreUri(SparkContext context) {
--- a/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/PathUtilsTest.java
+++ b/integration/spark/shared/src/test/java/io/openlineage/spark/agent/util/PathUtilsTest.java
@@ -444,4 +444,37 @@ class PathUtilsTest {
     // name
     assertThat(datasetIdentifier.getSymlinks()).hasSize(0);
   }
+
+  @Test
+  void testReconstructDefaultLocationWithS3Path() {
+    assertThat(
+            PathUtils.reconstructDefaultLocation(
+                    "s3://bucket/prefix", new String[] {"spark_catalog", "database"}, "table")
+                .toString())
+        .isEqualTo("s3://bucket/prefix/database.db/table");
+    assertThat(
+            PathUtils.reconstructDefaultLocation(
+                    "s3://bucket", new String[] {"spark_catalog", "database"}, "table")
+                .toString())
+        .isEqualTo("s3://bucket/database.db/table");
+    assertThat(
+            PathUtils.reconstructDefaultLocation(
+                    "hdfs://namenode:8020/warehouse", new String[] {"spark_catalog", "db"}, "table")
+                .toString())
+        .isEqualTo("hdfs://namenode:8020/warehouse/db.db/table");
+    assertThat(
+            PathUtils.reconstructDefaultLocation(
+                    "s3://bucket/prefix", new String[] {"spark_catalog", "default"}, "table")
+                .toString())
+        .isEqualTo("s3://bucket/prefix/table");
+    assertThat(
+            PathUtils.reconstructDefaultLocation(
+                    "s3://bucket/prefix", new String[] {"database"}, "table")
+                .toString())
+        .isEqualTo("s3://bucket/prefix/database.db/table");
+    assertThat(
+            PathUtils.reconstructDefaultLocation("s3://bucket/prefix", new String[] {}, "table")
+                .toString())
+        .isEqualTo("s3://bucket/prefix/table");
+  }
 }


dolfinus:
Probably fixed by #4505

PR #4480, opened 2026-04-21
feat(spec): AgentAttributionRunFacet for agent-produced lineage events

### LINKED ISSUE #4409: Agent-level data attribution for AI training — complement to pipeline lineage
## Agent-level data attribution for AI training — complement to pipeline lineage

OpenLineage tracks data transformations across pipelines (Airflow, Spark, dbt). We've built the agent-level complement: when an AI agent accesses data from a registered source, who gets attributed and compensated.

**The layer gap:** OpenLineage answers "what transformations happened to this data in the pipeline?" Our system answers "which AI agent accessed this data, under what terms, and how much compensation is owed to the source?"

**Agent Passport System data governance stack (shipped, tested):**
- Data Source Registration with terms (compensation rates, access limits, derivative policies)
- Enforcement Gate (enforce/audit/off) checking terms at every agent data access
- Signed access receipts with `termsAtAccessTime` snapshot + `dataHash`
- Contribution Ledger tracking per-source, per-agent usage
- Merkle-committed settlement with signed period records
- Training Attribution with multi-hop derivation chains (source → summary → training)
- GDPR Article 30 compliance report auto-generation

**Why this matters now:** EU AI Act Article 10 (August 2026) requires data governance for high-risk AI. Pipeline lineage (OpenLineage) + agent-level attribution (APS) together cover the full chain from data source to model output.

SDK: `npm install agent-passport-system` (v1.21.4, 1183 tests, Apache-2.0)

Happy to discuss integration points — particularly around how OpenLineage events could feed into our contribution ledger as upstream provenance evidence.

aeoess:
Closes #4409.

Adds the `AgentAttributionRunFacet` for OpenLineage events produced by autonomous agents — carrying the agent's identifier and the governance attestation (covenant) that was in force at execution time.

## Why

Agent-produced lineage events need two run-scoped properties that existing facets don't cover cleanly:

1. **Agent identity** — which agent process emitted this run. Recommended as a URN or DID (`did:aps:...`, `did:web:...`, etc.).
2. **Governance covenant** — the policy or attestation the agent was operating under at run time, carried as a tamper-evident digest so downstream compliance consumers can verify what the agent was authorized to do.

This extends the design discussion in #4409. The facet is vendor-agnostic: `covenantInEffect.type` is an open enum with `governance_attestation` as the reference canonical value; implementers MAY emit vendor-specific strings as long as the digest remains a hash of a signed object a compliance consumer can resolve.

## What's in

- `spec/facets/AgentAttributionRunFacet.json` — the facet schema (Draft 2020-12, `allOf` + RunFacet base).
- `spec/tests/AgentAttributionRunFacet/{1,2}.json` — two test vectors: minimal shape and a shape exercising the `covenantInEffect` vendor-extension slot with a bilateral receipt (pre/post-execution hashes + signatures).
- `website/docs/spec/facets/run-facets/agent_attribution.md` — documentation page with design notes and two worked examples.
- `CHANGELOG.md` entry under `[Unreleased] → Added`.
- `client/python` regeneration — new `generated/agent_attribution_run.py`, `facet_v2.py` import + re-export, and `redact_fields.yml` registration.

## Design decisions

1. **RunFacet, not DatasetFacet.** Agent identity + covenant are run-scoped. Per-input/output data hashes stay on existing dataset facets.
2. **`digest` required, `resolver` optional.** Tamper-evidence comes from the digest alone; a resolver URL is helpful but not required, supporting air-gapped compliance stores.
3. **`covenantInEffect.type` is an open enum.** Vendor-agnostic reference value is `governance_attestation`; vendor-specific strings are fine.
4. **`covenantInEffect.additionalProperties: true`.** Scoped to the sub-object so implementations can add vendor-specific fields without facet version bumps.
5. **`digestAlgorithm` defaults to `sha-256`.** Explicit field when needed, implied when absent. Forward-compatible for BLAKE3 / SHA3-256.

## Cross-implementation validation

`digest` = JCS-canonicalized(signed attestation) → SHA-256. This path has been cross-verified between two independent implementations (TypeScript and Python) producing byte-identical digests on three test vectors.

## Ask

Review welcome. @arian-gogani has been close to the design discussion in #4409 and has offered to leave a co-signal review on this PR.

### DIFF
--- a/spec/facets/AgentAttributionRunFacet.json
+++ b/spec/facets/AgentAttributionRunFacet.json
@@ -1,0 +1,66 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/AgentAttributionRunFacet.json",
+  "$defs": {
+    "AgentAttributionRunFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/RunFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "agentId": {
+              "description": "The identifier of the agent that executed this run. It is recommended to define this as a URN or DID. For example did:aps:z6Mk..., did:web:agent.example.com, agent:foo",
+              "type": "string",
+              "example": "did:aps:z6MkTestAgent001"
+            },
+            "accessedAt": {
+              "description": "The timestamp at which the agent accessed the input data source. ISO 8601 UTC.",
+              "type": "string",
+              "format": "date-time"
+            },
+            "covenantInEffect": {
+              "description": "The governance attestation that was in force at execution time. The digest is the hash of the signed attestation bytes; the type indicates which governance framework emitted it. Consumers verify the digest against a resolvable attestation to confirm what the agent was allowed to do during this run.",
+              "type": "object",
+              "properties": {
+                "digest": {
+                  "description": "Hash of the governance attestation that was in force at execution time. SHA-256 by default; the algorithm can be declared explicitly in the digestAlgorithm field or prefixed in the value (e.g. sha256:abc...).",
+                  "type": "string",
+                  "example": "sha256:3c9e9d2c6b7e8a1f4d5c6b7a8f9e0d1c2b3a4f5e6d7c8b9a0f1e2d3c4b5a6978"
+                },
+                "digestAlgorithm": {
+                  "description": "The hash algorithm used to compute the digest. Defaults to sha-256 if omitted.",
+                  "type": "string",
+                  "default": "sha-256",
+                  "example": "sha-256"
+                },
+                "type": {
+                  "description": "The governance signal type. Open enum; vendor-agnostic reference value is 'governance_attestation'. Implementers MAY emit a vendor-specific string as long as the digest remains a hash of a signed object that a compliance consumer can resolve and verify.",
+                  "type": "string",
+                  "example": "governance_attestation"
+                },
+                "resolver": {
+                  "description": "Optional URL or DID of the signed attestation from which the digest is computed. When present, a compliance consumer MAY fetch the resolved object and verify the hash matches the digest. Absence of a resolver is acceptable; the digest alone is sufficient for tamper-evidence.",
+                  "type": "string",
+                  "format": "uri",
+                  "example": "https://gateway.example.com/api/v1/public/trust/agent001"
+                }
+              },
+              "required": ["digest", "type"],
+              "additionalProperties": true
+            }
+          },
+          "required": ["agentId", "covenantInEffect"]
+        }
+      ],
+      "type": "object"
+    }
+  },
+  "type": "object",
+  "properties": {
+    "agentAttribution": {
+      "$ref": "#/$defs/AgentAttributionRunFacet"
+    }
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased](https://github.com/OpenLineage/OpenLineage/compare/1.46.0...HEAD)
 
+### Added
+
+* **Spec: Add AgentAttributionRunFacet for agent-produced lineage events** [`#4409`](https://github.com/OpenLineage/OpenLineage/issues/4409) [@aeoess](https://github.com/aeoess)
+  *Add `AgentAttributionRunFacet` JSON Schema carrying the agent's identifier and the governance attestation (covenant) that was in force at execution time. The `covenantInEffect.digest` provides tamper-evident reference to the signed policy; the `type` is an open enum with `governance_attestation` as the vendor-agnostic reference value; `additionalProperties: true` is scoped to the `covenantInEffect` sub-object so implementations can add vendor-specific fields without facet version bumps.*
+
 ## [1.46.0](https://github.com/OpenLineage/OpenLineage/compare/1.45.0...1.46.0)
 
 ### Added
--- a/client/python/redact_fields.yml
+++ b/client/python/redact_fields.yml
@@ -304,3 +304,9 @@
       redact_fields: []
     - class_name: HierarchyDatasetFacetLevel
       redact_fields: []
+- module: agent_attribution_run
+  classes:
+    - class_name: AgentAttributionRunFacet
+      redact_fields: []
+    - class_name: CovenantInEffect
+      redact_fields: []
--- a/client/python/src/openlineage/client/facet_v2.py
+++ b/client/python/src/openlineage/client/facet_v2.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from openlineage.client.generated import (
+    agent_attribution_run,
     base_subset_dataset,
     catalog_dataset,
     column_lineage_dataset,
@@ -60,6 +61,7 @@ __all__ = [
     "OutputDatasetFacet",
     "RunFacet",
     "set_producer",
+    "agent_attribution_run",
     "base_subset_dataset",
     "catalog_dataset",
     "column_lineage_dataset",
--- a/client/python/src/openlineage/client/generated/agent_attribution_run.py
+++ b/client/python/src/openlineage/client/generated/agent_attribution_run.py
@@ -1,0 +1,113 @@
+# Copyright 2018-2026 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from typing import Any, cast
+
+import attr
+from openlineage.client.generated.base import RunFacet
+from openlineage.client.utils import RedactMixin
+
+
+@attr.define
+class AgentAttributionRunFacet(RunFacet):
+    agentId: str  # noqa: N815
+    """
+    The identifier of the agent that executed this run. It is recommended to define this as a URN or
+    DID. For example did:aps:z6Mk..., did:web:agent.example.com, agent:foo
+
+    Example: did:aps:z6MkTestAgent001
+    """
+    covenantInEffect: CovenantInEffect  # noqa: N815
+    """
+    The governance attestation that was in force at execution time. The digest is the hash of the signed
+    attestation bytes; the type indicates which governance framework emitted it. Consumers verify the
+    digest against a resolvable attestation to confirm what the agent was allowed to do during this run.
+    """
+    accessedAt: str | None = attr.field(default=None)  # noqa: N815
+    """The timestamp at which the agent accessed the input data source. ISO 8601 UTC."""
+
+    @staticmethod
+    def _get_schema() -> str:
+        return "https://openlineage.io/spec/facets/1-0-0/AgentAttributionRunFacet.json#/$defs/AgentAttributionRunFacet"
+
+    @accessedAt.validator
+    def accessedat_check(self, attribute: str, value: str) -> None:  # noqa: ARG002
+        if value is None:
+            return
+        from dateutil import parser
+
+        parser.isoparse(value)
+        if "t" not in value.lower():
+            # make sure date-time contains time
+            msg = f"Parsed date-time has to contain time: {value}"
+            raise ValueError(msg)
+
+
+@attr.define
+class CovenantInEffect(RedactMixin):
+    """
+    The governance attestation that was in force at execution time. The digest is the hash of the signed
+    attestation bytes; the type indicates which governance framework emitted it. Consumers verify the
+    digest against a resolvable attestation to confirm what the agent was allowed to do during this run.
+    """
+
+    digest: str
+    """
+    Hash of the governance attestation that was in force at execution time. SHA-256 by default; the
+    algorithm can be declared explicitly in the digestAlgorithm field or prefixed in the value (e.g.
+    sha256:abc...).
+
+    Example: sha256:3c9e9d2c6b7e8a1f4d5c6b7a8f9e0d1c2b3a4f5e6d7c8b9a0f1e2d3c4b5a6978
+    """
+    type: str  # noqa: A003
+    """
+    The governance signal type. Open enum; vendor-agnostic reference value is 'governance_attestation'.
+    Implementers MAY emit a vendor-specific string as long as the digest remains a hash of a signed
+    object that a compliance consumer can resolve and verify.
+
+    Example: governance_attestation
+    """
+    digestAlgorithm: str | None = "sha-256"  # noqa: N815
+    """
+    The hash algorithm used to compute the digest. Defaults to sha-256 if omitted.
+
+    Example: sha-256
+    """
+    resolver: str | None = attr.field(default=None)
+    """
+    Optional URL or DID of the signed attestation from which the digest is computed. When present, a
+    compliance consumer MAY fetch the resolved object and verify the hash matches the digest. Absence of
+    a resolver is acceptable; the digest alone is sufficient for tamper-evidence.
+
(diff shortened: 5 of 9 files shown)


arian-gogani:
the bilateral receipt shape in the Nobulex example is correct. authorizationHash + authorizationSignature (pre-execution) and resultHash + resultSignature (post-execution) with signerKeyId binding both to the same Ed25519 signer is exactly how nobulex receipts work in production.

the facet design is clean. RunFacet scoping, required digest with optional resolver, open enum on type, and additionalProperties scoped to the sub-object are all the right calls.

one additional data point: we confirmed cross-implementation interop on the JCS + SHA-256 canonicalization path today. three test vectors run through both nobulex (TypeScript) and AgentLedger (Python) produce byte-identical digests. the canonicalization this facet relies on is stable across implementations.

this PR has my co-signal as the nobulex maintainer. the facet shape lands without any schema changes needed on the nobulex side.

aeoess:
Thanks Arian. The cross-impl interop data point is the clearest possible argument for canonicalization stability on this schema. Two independent implementations producing byte-identical digests on the same three vectors settles the question of whether the hash path is reproducible in practice, not just in principle. Useful for maintainer review.

Co-signal is much appreciated.

spec/facets/AgentAttributionRunFacet.json
```diff
@@ -0,0 +1,66 @@
...
+        {
+          "type": "object",
+          "properties": {
+            "agentId": {
+              "description": "The identifier of the agent that executed this run. It is recommended to define this as a URN or DID. For example did:aps:z6Mk..., did:web:agent.example.com, agent:foo",
+              "type": "string",
+              "example": "did:aps:z6MkTestAgent001"
+            },
+            "accessedAt": {
+              "description": "The timestamp at which the agent accessed the input data source. ISO 8601 UTC.",
```
kacpermuda, line 19: This seems like a property that should live on the dataset level, not run level. Most likely there will be different timestamps for different datasets right? Should we also include output datasets here?

spec/facets/AgentAttributionRunFacet.json
```diff
@@ -0,0 +1,66 @@
...
+                },
+                "resolver": {
+                  "description": "Optional URL or DID of the signed attestation from which the digest is computed. When present, a compliance consumer MAY fetch the resolved object and verify the hash matches the digest. Absence of a resolver is acceptable; the digest alone is sufficient for tamper-evidence.",
+                  "type": "string",
+                  "format": "uri",
+                  "example": "https://gateway.example.com/api/v1/public/trust/agent001"
+                }
+              },
+              "required": ["digest", "type"],
+              "additionalProperties": true
```
kacpermuda, line 51: I briefly looked through the linked issue, and understand that this is open for more vendor-specific fields, but I'm also under the impression that there are some use cases where we can already pretty much define what can be valuable to the users (some post-run hashes?). Should we add some other non-required fields here that we can clearly define to avoid ambiguity among producers?

kacpermuda (commented):
Left two comments, would love to hear from other committers as well, since I have very little experience with auditing agentic workflows.

aeoess:
Thanks for the review, @kacpermuda. Both questions cut at the right architectural joints. Answering each directly so other committers can weigh in with full context.

**On RunFacet vs DatasetFacet.**

The attribution this facet carries is about *who acted* (the agent, the delegation chain it invoked under, the policy signature that authorized it) rather than *what data was touched*. That's why it lives at run level in the current shape. A run is bounded by one agent acting under one authorization envelope, which is the natural unit a signed attestation attaches to. Multiple output datasets produced by the same run share the same attribution because they share the same authorization context.

That said, your point about different timestamps per dataset is real when an agent streams outputs over time within a single run. Two options worth considering:

1. **Keep AgentAttributionRunFacet as-is, add an optional `output_datasets` array inside it** with per-dataset timestamp and optional digest. The attribution envelope stays singular (one authorization = one signature) but the facet records per-dataset emission timing for auditors. This preserves the run-level envelope semantics while giving the dataset-level temporal resolution you're after.

2. **Introduce a sibling AgentAttributionDatasetFacet** for cases where the attribution genuinely differs per dataset (e.g., agent produced one output under an authorization that was revoked mid-run and a later output under a refreshed one). This is a rarer case and would require emitting multiple RunFacets in practice today, so a dedicated DatasetFacet variant would be the cleaner model.

My lean is option 1 for the common case and leaving the door open for a sibling DatasetFacet if demand materializes. Option 1 is a minimal additive change to the current schema. Curious which feels more natural to you and other OpenLineage committers who have seen the dataset vs run split play out in practice.

**On non-vendor-specific prescriptive fields.**

Agreed. The fully-extensible shape leans too far toward "figure it out yourself" for fields that are clearly common across implementations. A conservative prescribed set without limiting vendor extensions:

- `canonicalization_spec` (string, recommended): how the agent serialized the inputs/outputs before signing. Existing production implementations converge on `jcs-rfc8785+sha256`, but naming it explicitly lets validators check reproducibility.
- `decided_at` (ISO 8601 timestamp, required): when the agent made the decision that triggered the run. Distinct from run `started_at`/`completed_at` because the decision and the execution can be separated in time (deferred execution, queued actions).
- `action_ref` (string, recommended): content-addressable hash of the action tuple (agent_id + action_type + inputs). Cross-verification needs an identity that's stable across implementations. We've been using `sha256(jcs_canonical(action))` which is trivially reproducible.
- `post_run_hash` (string, optional): what you named directly. A hash of the canonicalized output state, useful for detecting tampering between emission and consumption.

Keeping `additionalProperties: true` at the facet level so vendors can extend. The four fields above are either required or recommended but do not close the door on future prescriptive additions.

I can push these as a second commit on this PR if the direction lands. Wanted to surface the shape before writing more code, since the RunFacet vs DatasetFacet question affects where these fields live.

kacpermuda:
Ad `different timestamps per dataset`. I agree that adding a datasets array within the runfacet may be tempting, as it's the easiest way, we'd be duplicating a dataset identifiers unnecessarily. So with 1k datasets, we'd have to put all of their name and namespace into the this array in run facet together with timestamps, even though we already have all that in inputs/outputs. So I think adding a Dataset facet makes sense, then we can keep all the information about agent in the run facet, and then only dataset-related details + some link to the agent run, that will allow us to fetch more info from run facet. At least this is how I'd imagine it, let me know if I'm missing something. If we need some general timestamp in the run facet, I'm okay with keeping it, it just confused me, that we have a single data_access_timestamp, where this is clearly a one-to-many mapping (one agent can access multiple datasets on different timestamps).

So tldr; I'd go for datasetfacet with some link to run facet runfacet={agentId=123, all_other_agent_info...} and dataset_facet={accessed_by_agent_id=123, all_other_dataset_specific_agent_info=...}. And this is under the assumption that there can only be one agent per run (or that each agent would emit their own run events), not sure if this is true, or we should make some kind of arrays here.

mobuchowski:
Is this similar? https://github.com/OpenLineage/OpenLineage/issues/4484

jakub-moravec:
@aeoess, after another review, I think the proposal is a little bit too tailored for a specific implementation. Could you take a step back and describe the attestation process, the involved entities, their responsibilities, and decouple that from the implementation? 

If this were to become a standard facet (and the value in that is obviously there), we should strive to make it usable across vendors. 

In https://github.com/OpenLineage/OpenLineage/issues/4409 you reference the EU AI Act Article 10 (August 2026), I think a good way to look at this would be making sure the facet supports what is defined in the regulation, as opposed to a vendor-specific solution addressing the regulation.

aeoess:
@mobuchowski thanks for the pointer to #4484. I read it through. My read is that it asks a different question: whether runtime actor-initiated interactions, like MCP calls, agent tool reads, or runtime API access, should be modeled in OpenLineage at all and at what layer.

This PR sits one level above that. It assumes the lineage event already exists and adds attribution to it. So the two compose. You can use this facet without #4484, and if #4484's modeling work lands later, those runtime interactions could carry the same attribution shape.

@jakub-moravec, fair point. The first draft was too tailored to one implementation. I went back to the Article 10 text before rewriting since you flagged it as the anchor, and that changed how I'd describe this.

Vendor-neutral version:

A producer emits lineage. For agent-produced events, this facet adds enough information for a downstream consumer to verify which actor and authorization context applied, without having to trust whatever the producer logged internally.

The flow, abstractly:

1. An agent receives an authorization or attestation from an issuing system
2. The agent performs a run and emits OpenLineage events
3. Each event carries the agent identifier, a digest of the authorization record in force at execution time, and optionally a resolver pointer if the consumer wants to fetch the underlying record

Four roles are involved:

- Subject: the agent or agent process producing the event
- Issuer: the system or principal that authorized the subject, such as a human principal, parent system, or organizational policy
- Verifier: the downstream consumer, auditor, or compliance system checking the action
- Resolver, optional: a service that maps digests to full records using whatever lookup protocol the issuer publishes

The facet specifies the attribution shape and digest reference. It does not require a specific identity scheme, signature algorithm, resolver protocol, or policy model. Those stay with the issuer, verifier, and resolver. Different vendors can implement the same facet against different stacks. It should compose with adjacent identity and provenance work, such as SPIFFE workload identity, in-toto signed provenance, and OpenTelemetry actor/span context, rather than replacing any of it.

On Article 10 specifically: my earlier framing was too broad, and rereading the text made that clear. Article 10 is about data governance for training, validation, and testing datasets in high-risk AI systems. This facet does not make a system Article 10-compliant by itself.

What it can provide is audit substrate that may support documentation around:

- 10(2)(b), data collection processes and origin: per-event attribution for collection runs
- 10(2)(c), preparation operations like annotation, labelling, cleaning, enrichment, and aggregation: per-event attribution for those runs
- 10(2)(h), identification of data gaps and shortcomings: per-issuer slicing may support source-level gap analysis

The same attribution shape is also useful outside high-risk AI scope, and may complement Article 12 record-keeping where that applies. Compliance is still the auditor's call, not the facet's.

@kacpermuda, circling back to your earlier point about run-level vs dataset-level scoping, because I didn't really answer it the first time.

The current shape is run-scoped because authorization is usually bounded by one agent acting under one envelope. But the dataset-level concern is real for runs that touch multiple datasets at different times, or under different authorizations within the same run.

Two paths I'm open to:

1. a sibling `AgentAttributionDatasetFacet` for those cases
2. an optional `output_datasets` array inside the run facet with per-dataset timestamps/digests

I'm not sure which is more idiomatic for OpenLineage, so happy to follow existing precedent if there is one.

On implementations: APS ships a conformant emitter, but the facet is not APS-specific. @arian-gogani confirmed earlier in this thread that Nobulex TypeScript and AgentLedger Python produce byte-identical digests on the same canonicalization path across three test vectors. That was the cross-implementation check I wanted before framing this as a vendor-neutral facet rather than just our own implementation shape.

aeoess:
@jakub-moravec re-read the proposal with your concern in front. You are right that the prior text leans on APS shapes in places where a vendor-neutral facet has to stand on its own. Here is the decoupled version.

**Entities and responsibilities**

| Entity | Responsibility |
|---|---|
| **Attesting principal** | The party signing the attestation. Provides identity (key or DID), produces a signed claim about the run. Could be a runtime issuer, an external auditor, the agent itself, or a delegating principal. |
| **Subject run** | The OpenLineage run that the attestation references. Identified by `runId`. |
| **Verifier** | Any downstream consumer who wants to confirm provenance: lineage backend, governance dashboard, regulator, peer system. Holds (or resolves) the attesting principal's public key. |

**Attestation process**

1. The subject run produces lineage events as today.
2. Some attesting principal observes (or commits to) facts about the run: who initiated it, what scope was authorized, what evidence supports a claim of compliance with a named policy.
3. The principal signs a claim binding `runId`, the claim payload, and a timestamp. Signature algorithm and key-resolution mechanism are open: Ed25519 over JCS, ECDSA over JSON canonical form, X.509 over CMS, COSE-signed CBOR. All are valid attestation envelopes.
4. The signed claim attaches to the OpenLineage event as a facet. The facet carries: principal identifier, signature algorithm identifier, signature bytes, and the claim payload (a JSON object whose schema can be vendor-extended).
5. A verifier resolves the principal's public key (DID document, JWKS, X.509 chain, key-server lookup), verifies the signature, and reads the claim.

**Vendor-extensible claim payload**

The facet defines a stable envelope (principal, algorithm, signature, payload) and leaves the claim schema to the vendor. EU AI Act Article 12 deployers want one shape (operator identity + log retention assertion). FedRAMP-track operators want another (boundary attestation + control-coverage). A pharma research operator wants a third. The facet is the carrier. The schemas live alongside as vendor-published profiles.

This matches the SCITT pattern (RFC 9711, COSE_Sign1 envelope with vendor claim payload) and the W3C VC pattern (proof envelope with arbitrary `credentialSubject`).

**APS-specific shape becomes one profile among many**

The earlier draft conflated the facet (vendor-neutral envelope) with one specific attesting-principal schema (APS receipt format). Splitting:

- **Facet** ships in OpenLineage core. Defines envelope, signature verification semantics, principal-identifier format.
- **Profiles** ship outside core. APS, SCITT, W3C VC, COSE_Sign1, and any other attestation system publish a schema URI; the facet payload references the profile by URI.

An OpenLineage backend can then validate signatures and surface attestations without committing to any single attestation system.

If this framing matches your concern, I can rewrite the PR body and schema along these lines. Current commits would need to be squashed and the schema split into envelope + reference profile. Confirm direction and I will push the rewrite.

mobuchowski:
I am just not sure if that's in the scope of OpenLineage; however it also feels sufficiently orthogonal to not be a breaking change in any way. I must say I personally I'm very far from an expert on those topics and can't make an informed decision whether to support this proposal. 

However, that is also the biggest issue I have. With OSS, we as OL maintainers are responsible for the code and spec we accept and release; and I don't know if the rest of the committers feel sufficiently good about their knowledge as to accept it and support it in the future.

Happy to accept if there's consensus of other committers around this that would be able to support this. If not, I think the way would be to keep it out of the official spec - introduce as custom facet - but we can point at it in docs and community, basically as "hey, this group is trying to solve that issue!" when asked about monitoring agents with OL; and come back at some point, when the topic gets more mature and field is more understood.

aeoess:
@mobuchowski path 2 makes sense. Plan:

**Custom facet, hosted by aeoess.** Pull `AgentAttributionRunFacet` out of the core spec proposal and publish at `https://aeoess.com/spec/AgentAttributionRunFacet.json`. We maintain backwards compatibility and handle schema questions on our own tracker.

**Docs pointer from OL.** Where would it live? A section in the integrations/community page, a note under "agent monitoring" if such a doc exists, or a custom-facets registry if OL maintains one. Happy to write the patch against whichever you prefer.

I'll close this PR or convert it to a docs-only pointer, whichever lands cleanest.

mobuchowski:
I think it makes sense to add a section in `Ecosystem` part on the website: https://openlineage.io/ecosystem/ 

Alternatively, a section in the docs about community additions would make sense?

kacpermuda:
Maybe under [Facets & Extensibility](https://openlineage.io/docs/spec/facets/), we can add a new tab at the bottom something like Community Facets, where we can put such additions?

arian-gogani:
@mobuchowski @kacpermuda both options work well for us. The Community Facets tab under Facets & Extensibility feels like the more natural home since the attestation facet extends the existing facet model rather than being a standalone integration.

Happy to draft the content for either location. The facet surfaces Trust Capital data (verified behavioral reputation for AI agents) as lineage metadata, so operators can see not just what an agent did but whether it had the earned trust level to do it.

Let me know which path you prefer and I will open a PR against the docs.

mobuchowski:
@arian-gogani that works for me, thanks!

aeoess:
Closing this. The Community Facets path the maintainers preferred is the better home for an attestation facet, and the APS attribution-receipt work continues in the APS spec and SDK. Thanks for the review.

Issue #1876, opened 2023-05-23
DataQualityAssertionsDatasetFacet incorrectly inherits from InputDatasetFacet


gaborbernat:
DataQualityAssertionsDatasetFacet inherits from InputDatasetFacet, https://github.com/OpenLineage/OpenLineage/blob/main/spec/facets/DataQualityAssertionsDatasetFacet.json though I think should do from DatasetFacet like all else

gaborbernat:
I am withdrawing this request and closing the issue, as it has had no follow-up in a long time. Thanks for taking a look.

PR #4483, opened 2026-04-22
Python: Remove deprecated dbt attributes in integration/common

hcthakur2004:
Removes deprecated facets kwarg and dbt_version_facet in common python dbt processor. They have passed their deprecation period.

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/processor.py
+++ b/integration/common/src/openlineage/common/provider/dbt/processor.py
@@ -430,8 +430,7 @@ class DbtArtifactProcessor:
                             namespace=namespace,
                             name=name,
                             inputFacets=dataset_facets,
-                            # TODO: remove this next release
-                            facets=dataset_facets,  # type: ignore
+
                         )
                     ],
                     None,


kacpermuda:
We probably do not want to remove the facet without a major release, but maybe it's not that critical. WDYT @mobuchowski @JDarDagran ?

hcthakur2004:
Thanks @kacpermuda! I have reverted the removal of the \DbtVersionRunFacet\ and restored it across the code base and test fixtures so there's no breaking change before the major release. I've narrowed this PR down to just the safe cleanup of the duplicate \acets=dataset_facets\ kwarg in \InputDataset\ which was marked to be removed.

chuenchen309:
I've been working in `integration/common`'s dbt provider recently, so I applied this change on top of current `main` and ran the suite — two concrete things that would need fixing before it's green, plus one question that may be relevant to @kacpermuda's point above.

**1. One golden fixture still asserts the duplicated facet.**

Baseline on `main`: `integration/common/tests/dbt` is `242 passed`. With this change applied: `241 passed, 1 failed` —

```
FAILED tests/dbt/test_dbt_local.py::test_dbt_parse_dbt_test_event[.../tests/dbt/no_test_metadata]
ERROR openlineage.common.test: Key dataQualityAssertions not in received event {}
ERROR openlineage.common.test: For key facets, expected value {'dataQualityAssertions': {...}} not equals received {}
```

`tests/dbt/no_test_metadata/result.json` carries `dataQualityAssertions` under **both** `inputs[].facets` and `inputs[].inputFacets` (10 occurrences of each), so it needs the `facets` copies dropped along with the code. The other two artifact-path fixtures are already clean — `tests/dbt/build/result.json` and `tests/dbt/spark/thrift/result.json` only have the `inputFacets` copy, which is why they still pass.

**2. The removed lines leave a stray blank line inside the `InputDataset(...)` call**, which `ruff format` collapses. `ruff-format` is in `.pre-commit-config.yaml` and `run-pre-commit` is on the always-run CircleCI workflow, so it would fail there.

**3. On whether this is a breaking change** — one data point that might make the decision easier: the structured-logs processor takes the opposite approach today. `structured_logs.py` builds a plain `Dataset` (which has no `inputFacets` field at all) and puts the facet under `facets`:

```python
Dataset(
    name=attached_dataset.name,
    namespace=attached_dataset.namespace,
    facets={**attached_dataset.facets, "dataQualityAssertions": assertion_facet},  # type: ignore[dict-item]
)
```

So after this PR the two dbt paths would emit the same facet under different keys — artifact path under `inputFacets` only, structured-logs path under `facets` only. Consumers reading only one key would silently lose data quality info depending on which dbt integration the user runs. That seems worth settling together rather than per-path, whichever way it goes.

Happy to help with the fixture update if that's useful. (Verification above was done with AI assistance; the commands are reproducible from a clean checkout of `main`.)

kacpermuda:
I think we should move towards keeping everything under facets, and not inputFacets / outputFacets. This is the direction we've seen with facets, where we've shifted from DQ input / dq output facets into generic facets. So this TODO left in code may be redundant, or we can treat it as a TODO to remove inputFacets, not facets.

chuenchen309:
Thanks @kacpermuda — that settles the direction, and there's already a PR implementing exactly
it: **#4504** (`spec: fix DataQualityAssertionsDatasetFacet to inherit from DatasetFacet`, open
since May 1, tracking #1876 from 2023). It moves the facet from `InputDatasetFacet` to
`DatasetFacet` in the spec and drops `inputFacets` from the dbt processor while keeping
`facets` — the opposite of what this PR does to the same lines. @mobuchowski reviewed it on
May 8 ("the fix looks very mechanical; makes sense") and asked only for the conflicts to be
resolved; the author hasn't been back since.

So by your answer, #4504 is the one that matches where you want this to go, and the two PRs
can't both land.

I checked what landing #4504 would actually take today. Applying its diff to current `main`:

- 9 of its 10 files apply cleanly; only `tests/dbt/no_test_metadata/result.json` conflicts.
- That conflict isn't really semantic. Most of #4504's fixture diff is **reindentation**
  (4 spaces to 2), which is also what makes it a ~3,000-line diff. The substantive fixture
  change is just the `inputFacets` to `facets` key move.

Redone as a pure mechanical transform on top of current `main`, preserving each fixture's
existing indentation:

| fixture | key moved | duplicate dropped | `_schemaURL` bumped |
|---|---|---|---|
| `build/result.json` | 1 | — | — |
| `test/result.json` | 5 | — | 4 |
| `spark/thrift/result.json` | 10 | — | — |
| `no_test_metadata/result.json` | — | 10 | 10 |

`no_test_metadata` is the fixture for the `parse_test` path, which currently double-writes the
facet under both keys, so there the `inputFacets` copy is dropped rather than moved. The
`_schemaURL` bump is needed because #4504 moves the spec `$id` to `1-2-0` while the fixtures
that carry a `_schemaURL` still pin `1-1-0` — #4504 predates them, which is also why one
`inputFacets` occurrence in `test/result.json` is outside its diff entirely.

Result: `integration/common` `tests/dbt/` is **243 passed**, identical to a clean `main`
checkout run in the same venv. That's ~520 lines of real change rather than ~3,000.

Happy to push this if it's useful — either as a rebase of #4504 keeping @creazyfrog as author,
or whatever you'd prefer for a three-month-stale PR. Either way it seems worth settling #4504
before this one, since it decides which key the facet ends up under.

*AI-assisted; the apply, the counts and the test runs above are output from actually running
them, not inferred.*

kacpermuda:
I think you can take over the #4504 by simply creating a similar PR, we can close the old one since it has been inactive. I think we can then drop the inputFacets and keep the assertions in just facets.

chuenchen309:
Done — #4845.

It carries #4504's spec and client changes and drops `inputFacets` for the assertion facet, as you described. Two things that weren't in the original:

- The fixtures are redone as a plain mechanical move. #4504 also reindented three of them from 4 spaces to 2, which is where most of its ~3000-line diff came from; keeping each file's own indentation makes the change readable. It also predates the `_schemaURL` values now in those fixtures and one `inputFacets` occurrence in `test/result.json`, so those needed doing.
- The Go client is regenerated. `build-client-go` diffs the checked-in files against a fresh generator run, so a spec change that skips it fails CI.

`spec/facets/*.json` is on the AGENTS.md do-not-touch list, so I'm treating your comment above as the authorization for it — say the word if you meant something narrower.

@creazyfrog's work is credited in the description and the commit message; I have no way to keep the authorship on the commit itself, so closing #4504 with a pointer to #4845 is your call rather than mine.

*AI-assisted with agentic coding tools; all changes were reviewed and tested by me.*

kacpermuda:
Will close this one in favor of #4845

PR #4485, opened 2026-04-23
feat(dbt): add Microsoft Fabric adapter support

harels:
## Summary
- Registers `dbt-fabric` in the supported `Adapter` enum so dbt-ol no longer raises `NotImplementedError` on a `type: fabric` profile.
- Adds namespace extraction producing `fabric://<server>[:<port>]`. Mirrors the existing `SQLSERVER` branch — requires the `server` profile key (canonical for dbt-fabric, which descends from dbt-sqlserver via ODBC Driver 18). Port is included when present in the profile.
- Adds two tests: one parametrized case in `test_get_query_id` and a focused test for the port-present branch.

## Design choice: namespace scheme
Chose `fabric://` over collapsing into `mssql://` so Fabric, Azure SQL, and Synapse stay separable at the dataset-namespace level. Happy to switch if maintainers prefer otherwise — the OpenLineage dataset-naming spec doesn't currently list a canonical Fabric scheme.

## Out of scope / follow-ups
- Catalog stats extraction. `_get_model_node_stats_facet` has explicit branches for BigQuery / Snowflake / Redshift row and byte counts. Fabric exposes row counts via `sys.dm_pdw_nodes_db_partition_stats`, but the dbt-fabric `catalog.json` shape needs verification before wiring this up.

## Test plan
- [x] `pytest integration/common/tests/dbt/test_processor.py` — 11 passed locally
- [x] CI green


integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -850,6 +851,10 @@ def extract_namespace(self, profile: dict) -> str:
             return f"databricks://{profile['host']}"
         elif self.adapter_type == Adapter.SQLSERVER:
             return f"mssql://{profile['server']}:{profile['port']}"
+        elif self.adapter_type == Adapter.FABRIC:
```
harels: need to add this to naming.md in the spec too

mobuchowski:
@harels My concern is calling it `fabric`, as some people might not expect that this relates only to `Fabric Data Warehouse`, but maybe `Fabric Lakehouse` or someone might want integration with `onelake` or variety of other related services.

Can we call it `fabric-warehouse` - maybe ugly, but I think we should be specific with Microsoft services?

harels:
@mobuchowski good call — pushed `70b8989d9` renaming to `fabric-warehouse`:

- `Adapter.FABRIC` → `Adapter.FABRIC_WAREHOUSE` (enum string value stays `"fabric"` since that's what dbt-fabric writes into `profiles.yml` — we don't control that)
- Namespace scheme `fabric://` → `fabric-warehouse://`
- Updated `website/docs/spec/naming.md` and `integration/dbt/README.md` to match

This leaves room for `fabric-lakehouse`, `onelake`, etc. as separate adapters later. One small note: every other namespace in the file is single-token (`snowflake`, `bigquery`, `mssql`, `awsathena`, …) — went with `fabric-warehouse` per your literal phrasing, but happy to switch to `fabricwarehouse` if you'd rather match that style.

mobuchowski (approved):

AlexandreSajus:
Hey team! Thanks for your help on this customer issue! The customer tried to install OpenLineage 1.47.0 to take advantage of this fix but it unfortunately does not work. They get:

> “NotImplementedError: Only `bigquery`,`snowflake`,`redshift`,`spark`,`postgres`,`databricks`,`sqlserver`,`dremio`,`athena`,`duckdb`,`trino`,`glue`,`clickhouse`,`fabric` adapters are supported right now. Passed fabric
[2026-05-12 08:56:55,257] [INFO] [openlineage.dbt:256] - Emitted 0 OpenLineage events
[2026-05-12 08:56:55,257] [INFO] [openlineage.dbt:258] - Underlying dbt execution returned 0”

Their profiles.yml uses "fabric" as type:
[profiles (5).yml](https://github.com/user-attachments/files/27634747/profiles.5.yml)

When [extract_adapter_type](https://github.com/OpenLineage/OpenLineage/blob/b4d65c5c0933424b77471216763088dae3090cbb/integration/common/src/openlineage/common/provider/dbt/processor.py#L863-L869) tries to match this "fabric" type, it uses the [Adapter class](https://github.com/OpenLineage/OpenLineage/blob/b4d65c5c0933424b77471216763088dae3090cbb/integration/common/src/openlineage/common/provider/dbt/processor.py#L48) which contains only a FABRIC_WAREHOUSE key and not a FABRIC key. I understand the need to not use "fabric" as a key, but as it stands, the customer cannot install Data Observability for their DBT Fabric environment.

Would it be possible to make changes so that OpenLineage supports their "fabric" adapter? Thanks for your help!

PR #4489, opened 2026-04-25
fix(dbt): guard against None node_unique_id in get_job_type()

### LINKED ISSUE #4488: [BUG] Python dbt: get_job_type() raises AttributeError when node_unique_id is None
## Bug Description

The `get_job_type()` function in `integration/common/src/openlineage/common/provider/dbt/utils.py` calls `.startswith()` on `node_unique_id` without first checking if it is `None`. Since `node_unique_id` is retrieved via `get_from_nullable_chain()` (which explicitly returns `None` when a key is missing), this results in an unhandled `AttributeError: 'NoneType' object has no attribute 'startswith'` at runtime.

## Affected File

`integration/common/src/openlineage/common/provider/dbt/utils.py`

## Code (lines 162-181)

```python
def get_job_type(event) -> str | None:
    node_unique_id = get_node_unique_id(event)  # can return None
    node_type = event["info"]["name"]
    if node_type == "SQLQuery":
        return "SQL"
    elif node_type in ("MainReportVersion", "CommandCompleted"):
        return "JOB"
    elif node_unique_id.startswith("model."):    # <-- AttributeError if None
        return "MODEL"
    ...
```

## When does it trigger?

Any dbt structured-log event where:
- `event["info"]["name"]` is **not** `"SQLQuery"`, `"MainReportVersion"`, or `"CommandCompleted"`, AND
- `event["data"]["node_info"]["unique_id"]` is **absent or None**

For example, certain internal dbt log events (e.g. connection-related events like `SQLQueryStatus`) carry a `node_info` block without a populated `unique_id`, which would trigger this crash.

## Expected Behavior

The function should guard against `None` and safely return `None` instead of crashing.

## Proposed Fix

Add an early-exit guard:

```python
def get_job_type(event) -> str | None:
    node_unique_id = get_node_unique_id(event)
    node_type = event["info"]["name"]
    if node_type == "SQLQuery":
        return "SQL"
    elif node_type in ("MainReportVersion", "CommandCompleted"):
        return "JOB"
    elif node_unique_id is None:
        return None
    elif node_unique_id.startswith("model."):
        return "MODEL"
    elif node_unique_id.startswith("snapshot."):
        return "SNAPSHOT"
    elif node_unique_id.startswith("seed."):
        return "SEED"
    elif node_unique_id.startswith("test."):
        return "TEST"
    return None
```

## Steps to Reproduce

1. Run dbt with structured-log integration.
2. Trigger a dbt event where `node_info.unique_id` is absent.
3. Observe `AttributeError: 'NoneType' object has no attribute 'startswith'`.

## Environment

- Integration: `integration/common` (Python dbt processor)
- Relevant function: `get_job_type` in `utils.py`

hcthakur2004:
## Summary

Fixes #4488

The `get_job_type()` helper in `integration/common/src/openlineage/common/provider/dbt/utils.py` calls `.startswith()` on `node_unique_id` without first checking if it is `None`.

`node_unique_id` is retrieved via `get_from_nullable_chain()`, which returns `None` when `node_info.unique_id` is absent from the event. This causes a crash:

```
AttributeError: 'NoneType' object has no attribute 'startswith'
```

## Changes

- **`utils.py`**: Added an early-exit `elif node_unique_id is None: return None` guard before the `.startswith()` calls.
- **`test_utils.py`**: Added a parametrized `test_get_job_type` suite (11 cases) covering the None guard, all known `node_type` names (`SQLQuery`, `MainReportVersion`, `CommandCompleted`), all valid `unique_id` prefixes (`model.`, `snapshot.`, `seed.`, `test.`), and an unknown prefix.

## Type of change

- [x] Bug fix (non-breaking)
- [x] New tests added

### DIFF
--- a/integration/common/tests/test_utils.py
+++ b/integration/common/tests/test_utils.py
@@ -1,7 +1,7 @@
 # Copyright 2018-2026 contributors to the OpenLineage project
 # SPDX-License-Identifier: Apache-2.0
 import pytest
-from openlineage.common.provider.dbt.utils import CONSUME_STRUCTURED_LOGS_COMMAND_OPTION
+from openlineage.common.provider.dbt.utils import CONSUME_STRUCTURED_LOGS_COMMAND_OPTION, get_job_type
 from openlineage.common.utils import (
     IncrementalFileReader,
     add_command_line_arg,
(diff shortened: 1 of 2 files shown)


integration/common/tests/test_utils.py
```diff
@@ -196,3 +196,48 @@ def read(self, *args, **kwargs):
...
 
     assert expected_lines == actual_lines_read
+
+
+def _make_event(node_name: str, node_unique_id: str | None = None) -> dict:
+    """Helper to build a minimal dbt structured-log event for get_job_type tests."""
+    event: dict = {"info": {"name": node_name}, "data": {}}
+    if node_unique_id is not None:
+        event["data"]["node_info"] = {"unique_id": node_unique_id}
+    return event
```
kacpermuda, line 206: nit: this helper is used only in one place, so we can just move this to the body of `test_get_job_type`

kacpermuda (approved):

hcthakur2004:
Thanks for the review and support! Looking forward to contributing more to OpenLineage 🚀

Issue #4494, opened 2026-04-26
[FEATURE] [FLINK] Add ClickHouse JDBC support for lineage tracking


och5351:
### Motivation

While reviewing the Apache Flink JDBC connector ClickHouse support PR (apache/flink-connector-jdbc#171), I identified an opportunity to add OpenLineage support for ClickHouse JDBC connections. This would enable standardized lineage tracking for Flink-ClickHouse pipelines following OpenLineage conventions.

Although not blocking for the Flink connector PR, adding ClickHouse support will:
- Enable standardized lineage metadata for Flink-ClickHouse pipelines
- Provide consistent dataset namespace identification
- Improve integration with existing lineage tooling

~~Currently, when a Flink job connects to ClickHouse via `jdbc:clickhouse://localhost:8123/mydb`, OpenLineage cannot parse the JDBC URL because `ClickHouseJdbcExtractor` does not exist, resulting in incomplete lineage graphs.~~

### Description

Add `ClickHouseJdbcExtractor` following the same pattern as existing database extractors (PostgreSQL, OceanBase).

Requirements:
- Support both URL schemes: `jdbc:clickhouse://host` and `jdbc:ch://host`
- Remove protocol prefix (http/https/grpc) to follow OpenLineage conventions (similar to how PostgreSQL removes SSL info)
- Default port: 8123 (ClickHouse HTTP interface)

Files to change:
- New: `ClickHouseJdbcExtractor.java`
- Update: `JdbcDatasetUtils.java` (register extractor)
- New: `JdbcDatasetUtilsTestForClickHouse.java` (13 test cases)

Expected behavior:
- `jdbc:clickhouse://localhost:8123/mydb` → namespace: `clickhouse://localhost:8123`, dataset: `mydb.schema.table`
- `jdbc:ch:https://host:8443/mydb` → namespace: `clickhouse://host:8443`, dataset: `mydb.schema.table`

Reference: https://clickhouse.com/docs/en/integrations/java#jdbc-driver

### Related issues

apache/flink-connector-jdbc#171

### Do you plan to make this contribution yourself?

- [x] I am interested in doing this work

dolfinus:
Implemented in #4495

> Currently, when a Flink job connects to ClickHouse via jdbc:clickhouse://localhost:8123/mydb, OpenLineage cannot parse the JDBC URL because ClickHouseJdbcExtractor does not exist, resulting in incomplete lineage graphs.

[...]

PR #2825, opened 2024-07-03
Spark,Flink: Add MySqlJdbcExtractor

dolfinus:
### Problem

MySQL JDBC URLs can be in a few different formats:
* With or without port number (default port number for MySQL is 3306)
* With schemes like `mysql`, `mysql+srv`, `mysql:loadbalance` and so on

This produces datasets with different namespaces, although data is stored in the same MySQL instance.

Related to #2762.

### Solution

Add `MySqlJdbcExtractor` which could handle these cases, and convert JDBC URLs to format `mysql://host:port`. Excluding TNS-like format of JDBC URL (at list for now).

#### One-line summary:

Handle different formats of MySQL JDBC URL, and produce datasets with consistent namespaces, like `mysql://host:port`.

### Checklist

- [X] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [X] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [X] Your changes are accompanied by tests (_if relevant_)
- [X] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] Your comment includes a one-liner for the changelog about the specific purpose of the change (_not required for changes to tests, docs, or CI config_)
- [ ] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_if relevant_)
- [ ] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----
SPDX-License-Identifier: Apache-2.0\
Copyright 2018-2024 contributors to the OpenLineage project

PR #4495, opened 2026-04-26
flink: add ClickHouse JDBC support

och5351:
Add ClickHouseJdbcExtractor for parsing ClickHouse JDBC URLs.

related: #4494

Add ClickHouse JDBC support with ClickHouseJdbcExtractor

### Meaningful description
Adds `ClickHouseJdbcExtractor` to enable lineage tracking for ClickHouse JDBC connections. This allows OpenLineage to parse ClickHouse JDBC URLs and extract proper dataset namespaces and names. 

  **Implementation:** 
  - Supports both `jdbc:clickhouse://` and `jdbc:ch://` URL schemes 
  - Removes protocol prefixes (http/https/grpc) from namespace following OpenLineage conventions
    - ClickHouse JDBC supports protocol-specific URLs like `jdbc:ch:https://host:8443`
    - OpenLineage extracts only physical location (host:port), removing protocol info
    - This follows the same pattern as PostgreSQL extractor removing SSL info 
  - Default port: 8123

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/utils/jdbc/ClickHouseJdbcExtractor.java
+++ b/client/java/src/main/java/io/openlineage/client/utils/jdbc/ClickHouseJdbcExtractor.java
@@ -1,0 +1,59 @@
+/*
+/* Copyright 2018-2026 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.utils.jdbc;
+
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * Implementation of {@link JdbcExtractor} for ClickHouse.
+ *
+ * @see <a href="https://clickhouse.com/docs/en/integrations/java#jdbc-driver">ClickHouse JDBC
+ *     Driver</a>
+ */
+public class ClickHouseJdbcExtractor implements JdbcExtractor {
+
+  private static final int MIN_QUALIFIED_TABLE_NAME_PARTS = 2;
+
+  private JdbcExtractor delegate() {
+    return new OverridingJdbcExtractor("clickhouse", "8123");
+  }
+
+  @Override
+  public boolean isDefinedAt(String jdbcUri) {
+    // Support both 'ch' and 'clickhouse' schemes
+    return jdbcUri.startsWith("ch:") || jdbcUri.startsWith("clickhouse:");
+  }
+
+  @Override
+  public JdbcLocation extract(String rawUri, Properties properties) throws URISyntaxException {
+    // Normalize scheme and remove protocol prefix (similar to how PostgreSQL removes SSL info)
+    String normalizedUri =
+        rawUri
+            .replaceFirst("^ch:", "clickhouse:")
+            .replaceFirst("^clickhouse:(https?|grpc)://", "clickhouse://");
+
+    JdbcLocation location = delegate().extract(normalizedUri, properties);
+
+    // In ClickHouse, DATABASE and SCHEMA are synonyms (no 3-level structure)
+    // Override toName() to handle qualified vs unqualified table names
+    return new JdbcLocation(
+        location.getScheme(),
+        location.getAuthority(),
+        location.getInstance(),
+        location.getDatabase()) {
+      @Override
+      public String toName(List<String> parts) {
+        if (parts.size() >= MIN_QUALIFIED_TABLE_NAME_PARTS) {
+          return String.join(".", parts);
+        }
+        // Otherwise, use default behavior (add URL database)
+        return super.toName(parts);
+      }
+    };
+  }
+}
--- a/client/java/src/main/java/io/openlineage/client/utils/jdbc/JdbcDatasetUtils.java
+++ b/client/java/src/main/java/io/openlineage/client/utils/jdbc/JdbcDatasetUtils.java
@@ -27,6 +27,7 @@ public class JdbcDatasetUtils {
     new Db2JdbcExtractor(),
     new TrinoJdbcExtractor(),
     new OceanBaseJdbcExtractor(),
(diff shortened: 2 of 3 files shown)


och5351:
Hello, @dolfinus, @mobuchowski !
Would you mind reviewing this PR when you have some time?

client/java/src/main/java/io/openlineage/client/utils/jdbc/ClickHouseJdbcExtractor.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  @Override
+  public JdbcLocation extract(String rawUri, Properties properties) throws URISyntaxException {
+    // Normalize scheme and remove protocol prefix (similar to how PostgreSQL removes SSL info)
+    String normalizedUri =
+        rawUri
+            .replaceFirst("^ch:", "clickhouse:")
+            .replaceFirst("^clickhouse:(https?|grpc)://", "clickhouse://");
+    return delegate().extract(normalizedUri, properties);
```
dolfinus: There is one thing that should be adjusted - there is no difference between DATABASE and SCHEMA for Clickhouse, like on Postgres. For example, if you connect to Clickhouse using:
* url `jdbc:clickhouse://host:port/somedb`
* query `SELECT * FROM schema.table`

you'll get:
* `namespace=clickhouse://host:port`
* `name=somedb.schema.table`

but actually it should be:
* `namespace=clickhouse://host:port`
* `name=schema.table`  (explicit schema name, overriding default `somedb`)

och5351: Thank you for your review, @dolfinus!

You're absolutely right. I noticed that MySQL has the same issue (see: [MySqlJdbcExtractor.java](https://github.com/OpenLineage/OpenLineage/blob/main/client/java/src/main/java/io/openlineage/client/utils/jdbc/MySqlJdbcExtractor.java)).

Since MySQL also treats DATABASE and SCHEMA as equivalent with no 3-level hierarchy ,just like ClickHouse, I was wondering if we should also update MySQL to ignore the URL database and use only the explicit schema name? I see that PR #2825 has a MySQL test expecting `mydb.schema.table1`. Am I understanding this correctly?

dolfinus: IMHO MySQL extractor should be updated in dedicated PR

och5351: Understood! In that case, would it be okay if I open a separate issue for the MySQL extractor update as well?

och5351: Could you please check this PR as well?
> https://github.com/OpenLineage/OpenLineage/pull/4497

dolfinus (approved):
Implementation is different from #4497

dolfinus (approved):

och5351:
> Implementation is different from #4497

Thank you for your review, @dolfinus! I'll update the implementation once #4497 passes CI.

client/java/src/main/java/io/openlineage/client/utils/jdbc/ClickHouseJdbcExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+    return jdbcUri.startsWith("ch:") || jdbcUri.startsWith("clickhouse:");
+  }
+
+  @Override
+  public JdbcLocation extract(String rawUri, Properties properties) throws URISyntaxException {
+    // Normalize scheme and remove protocol prefix (similar to how PostgreSQL removes SSL info)
+    String normalizedUri =
+        rawUri
+            .replaceFirst("^ch:", "clickhouse:")
+            .replaceFirst("^clickhouse:(https?|grpc)://", "clickhouse://");
```
dolfinus, line 38: HTTP protocol is missing
https://clickhouse.com/docs/integrations/language-clients/java/jdbc#configuration

och5351: Thank you for your review @dolfinus .
However, please check the regular expression `?`. The pattern also includes http.

och5351: Do you perhaps mean that it would be better to explicitly show http in the code as well?

dolfinus: No, I think it's enough to add test case for http

dolfinus (approved):

client/java/src/main/java/io/openlineage/client/utils/jdbc/ClickHouseJdbcExtractor.java
dolfinus: I see, there is already such test case, my bad

och5351: @dolfinus !
I really appreciate you taking the time to review this. Would you mind mentioning this to the other committers as well? This PR is blocked by that PR (https://github.com/apache/flink-connector-jdbc/pull/171). 

I hesitate to ask, but please don't feel pressured whenever you have time is fine.

dolfinus: > This PR is blocked by that PR (apache/flink-connector-jdbc#171).

Producing proper name and namespace for Clickhouse JDBC is beneficial for all OpenLineage Java client users, including Spark, Debezium and Trino integrations. It's not limited by Flink.

mobuchowski (approved):

Issue #4496, opened 2026-04-27
[BUG] MySqlJdbcExtractor incorrectly includes URL database in qualified table names


och5351:
> Co-authored-by @dolfinus 

### OpenLineage integration/client

Java client

### OpenLineage version

main branch

### Technology and package versions

- Java client (io.openlineage:openlineage-java)
- MySQL JDBC connector

### Environment configuration

N/A - This is a logic bug in the MySqlJdbcExtractor implementation

### Deployment details

N/A

### Problem details

The `MySqlJdbcExtractor` incorrectly prepends the URL database to qualified table names, creating invalid 3-level identifiers like `mydb.schema.table1`.

In MySQL, DATABASE and SCHEMA are synonyms (not separate concepts like PostgreSQL).
When a query references a table as `schema.table1`, the `schema` part IS the database name,
not a schema within another database.

Current behavior:
- URL: `jdbc:mysql://hostname/mydb`
- Table: `schema.table1`
- Result: `mydb.schema.table1`

### What you think should happen instead

MySQL only supports 2-level identifiers: `database.table` (not `database.schema.table`).

When a table name is already qualified (contains a dot), it means:
- The first part is the DATABASE name
- The URL database should be ignored (it's just a default for unqualified table names)

Current behavior:
- URL: `jdbc:mysql://hostname/mydb`
- Table: `schema.table1`
- Result: `schema.table1`

This is the same issue that was fixed for ClickHouse in the current review process, as both databases treat DATABASE and SCHEMA as the same concept.

MySQL documentation confirms this:
- https://dev.mysql.com/doc/refman/8.0/en/create-database.html (DATABASE = SCHEMA synonym)
- https://dev.mysql.com/doc/refman/8.0/en/identifier-qualifiers.html (db.table format, no 3-level)

### How to reproduce

1. Run the existing test case in `JdbcDatasetUtilsTestForMySql`:
   ```java
    JdbcDatasetUtils.getDatasetIdentifier("jdbc:mysql://hostname/mydb", "schema.table1", new Properties())
2. Current result: name = "mydb.schema.table1" /// Expected result: name = "schema.table1"

### Additional details

This occurs consistently for any qualified table name when the URL contains a database path.

Related:
- Same issue was identified and fixed for ClickHouse (DATABASE = SCHEMA)
- Fix should mirror the ClickHouse solution: call location.setDatabase(Optional.empty()) in MySqlJdbcExtractor.extract()
[...]

PR #4497, opened 2026-04-27
fix(client/java): MySqlJdbcExtractor incorrectly includes URL database in qualified table names

och5351:
In MySQL, DATABASE and SCHEMA are synonyms. When a table name is qualified (e.g., schema.table1), the explicit database name takes precedence over the URL database, which should be excluded.

related: #4496

Fix MySqlJdbcExtractor to exclude URL database from qualified table names.

### Meaningful description

MySqlJdbcExtractor incorrectly prepends the URL database to qualified table names, creating invalid 3-level identifiers like `mydb.schema.table1`. In MySQL, DATABASE and SCHEMA are synonyms (not separate concepts like PostgreSQL).

Clear the database field from JdbcLocation after extraction, similar to the ClickHouse fix. When a table is referenced as `schema.table1`, the `schema` part is already the database name and should not be prefixed with the URL database.

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/utils/jdbc/MySqlJdbcExtractor.java
+++ b/client/java/src/main/java/io/openlineage/client/utils/jdbc/MySqlJdbcExtractor.java
@@ -6,12 +6,14 @@
 package io.openlineage.client.utils.jdbc;
 
 import java.net.URISyntaxException;
+import java.util.List;
 import java.util.Properties;
 
 public class MySqlJdbcExtractor implements JdbcExtractor {
   // https://dev.mysql.com/doc/connector-j/en/connector-j-reference-jdbc-url-format.html
 
   private static final String PROTOCOL_PART = "^[\\w+:]+://";
+  private static final int MIN_QUALIFIED_TABLE_NAME_PARTS = 2;
 
   private JdbcExtractor delegate() {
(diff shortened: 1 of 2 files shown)


client/java/src/main/java/io/openlineage/client/utils/jdbc/MySqlJdbcExtractor.java
```diff
@@ -26,6 +26,9 @@ public boolean isDefinedAt(String jdbcUri) {
   public JdbcLocation extract(String rawUri, Properties properties) throws URISyntaxException {
     // Schema could be 'mysql', 'mysql:part', 'mysql+srv:part'. Convert it to 'mysql'
     String normalizedUri = rawUri.replaceFirst(PROTOCOL_PART, "mysql://");
-    return delegate().extract(normalizedUri, properties);
+
+    JdbcLocation location = delegate().extract(normalizedUri, properties);
+    location.setDatabase(java.util.Optional.empty());
```
dolfinus: If table name is is not in `schema.table` format, then dataset name will not be fully qualified

dolfinus (approved):

dolfinus (approved):

mobuchowski (approved):

Issue #4481, opened 2026-04-21
[BUG] Hive: NullPointerException in Parsing.buildQueryTree for 3+ way UNION ALL


OliverTzz98:
### OpenLineage integration/client

  Hive integration (`openlineage-hive-hook`)

  ### OpenLineage version

  1.37.0

  ### Technology and package versions

  - Hive 3.1.3 (EMR amzn-3)
  - Hadoop 3.3.3 (amzn-2)
  - Java 1.8.0_472

  ### Problem details

  The Hive hook throws `NullPointerException` in `Parsing.buildQueryTree`
  every time it parses a query that contains a **3+ way set operation**
  (`UNION ALL`, `INTERSECT`, `EXCEPT`). The query itself succeeds — but the
  hook logs an error and fails to emit a lineage event.

  Stack trace:

  ERROR hooks.HiveOpenLineageHook - Error occurred during lineage creation:
  java.lang.NullPointerException
      at io.openlineage.hive.parsing.Parsing.buildQueryTree(Parsing.java:30)
      at io.openlineage.hive.parsing.Parsing.buildQueryTree(Parsing.java:59)
      at io.openlineage.hive.parsing.Parsing.buildQueryTree(Parsing.java:55)
      at io.openlineage.hive.parsing.Parsing.buildQueryTree(Parsing.java:55)
      at io.openlineage.hive.hooks.Faceting.getOutputDatasets(Faceting.java:146)
      at io.openlineage.hive.hooks.Faceting.getRunEvent(Faceting.java:355)
      at io.openlineage.hive.hooks.HiveOpenLineageHook.run(HiveOpenLineageHook.java:128)
      ...

  ### Root cause

  In `Parsing.java:50-63`, when the `QBExpr` opcode is not `NULLOP` (i.e. a
  set operation), the code assumes that **both** `getQBExpr1()` and
  `getQBExpr2()` are themselves `NULLOP` `QBExpr`s that wrap a concrete `QB`:

  ```java
  } else {
    // This could be, for example, a UNION query
    QB subQB1 = qbExpr.getQBExpr1().getQB();
    query.getSubQueries().get(subAlias).add(buildQueryTree(subQB1, outputTable));
    QB subQB2 = qbExpr.getQBExpr2().getQB();
    query.getSubQueries().get(subAlias).add(buildQueryTree(subQB2, outputTable));
  }

  Hive parses chained UNION ALL as a left-associative binary tree —
[...]

PR #3555, opened 2025-03-14
Hive/add integration

tnazarew:
### Problem

Lack of hive integration

Closes: #ISSUE-NUMBER

### Solution

Add hive integration

### Checklist

- [ ] You've [signed-off](https://github.com/OpenLineage/OpenLineage/blob/main/why-the-dco.md) your work
- [ ] Your pull request title follows our [guidelines](https://github.com/OpenLineage/OpenLineage/blob/main/CONTRIBUTING.md#creating-pull-requests)
- [ ] Your changes are accompanied by tests (_if relevant_)
- [ ] Your change contains a [small diff](https://kurtisnusbaum.medium.com/stacked-diffs-keeping-phabricator-diffs-small-d9964f4dcfa6) and is self-contained
- [ ] You've updated any relevant documentation (_if relevant_)
- [ ] Your comment includes a one-liner for the changelog about the specific purpose of the change (_not required for changes to tests, docs, or CI config_)
- [ ] You've versioned the core OpenLineage model or facets according to [SchemaVer](https://docs.snowplowanalytics.com/docs/pipeline-components-and-applications/iglu/common-architecture/schemaver) (_if relevant_)
- [ ] You've added a [header](https://github.com/OpenLineage/OpenLineage/tree/main/.github/header_templates.md) to source files (_if relevant_)

----
SPDX-License-Identifier: Apache-2.0\
Copyright 2018-2025 contributors to the OpenLineage project

PR #4498, opened 2026-04-27
hive: fix NPE in Parsing.buildQueryTree for 3+ way set operations

1fanwang:
## What

Fix `NullPointerException` in `Parsing.buildQueryTree` for any Hive query that contains a 3+ way set operation (`UNION ALL` / `INTERSECT` / `EXCEPT`).

## Why

Hive parses chained set operations left-associatively as a binary tree of non-`NULLOP` `QBExpr` nodes. Only the **leaves** of that tree wrap a concrete `QB` via `QBExpr#getQB()`; intermediate non-`NULLOP` nodes return `null` from `getQB()`.

The existing code unwrapped both legs of a non-`NULLOP` `QBExpr` with a single `.getQB()` call:

```java
QB subQB1 = qbExpr.getQBExpr1().getQB();
query.getSubQueries().get(subAlias).add(buildQueryTree(subQB1, outputTable));
QB subQB2 = qbExpr.getQBExpr2().getQB();
query.getSubQueries().get(subAlias).add(buildQueryTree(subQB2, outputTable));
```

For any 3+ way set operation, one (or both) of the legs is itself a non-`NULLOP` `QBExpr`, so `.getQB()` returns `null` and the recursive `buildQueryTree` call NPEs at `Parsing.java:30` (`qb.getId()`). The `HiveOpenLineageHook` catches the `NullPointerException`, logs it, and silently drops the lineage event for that query.

This regression is what #4481 captures, with full stack trace + EMR Hive 3.1.3 reproducer.

## Fix

Replace the inline two-leg unwrap with a recursive helper `addQBExprToSubQueries` that descends into both legs when the `QBExpr` is non-`NULLOP`. The recursion naturally handles 2-way, 3-way, and N-way set operations:

```java
private static void addQBExprToSubQueries(
    QBExpr qbExpr, QueryExpr query, String subAlias, String outputTable) {
  if (qbExpr.getOpcode().equals(QBExpr.Opcode.NULLOP)) {
    QB subQB = qbExpr.getQB();
    query.getSubQueries().get(subAlias).add(buildQueryTree(subQB, outputTable));
  } else {
    addQBExprToSubQueries(qbExpr.getQBExpr1(), query, subAlias, outputTable);
    addQBExprToSubQueries(qbExpr.getQBExpr2(), query, subAlias, outputTable);
  }
}
```

The caller in `buildQueryTree` becomes a single line, and the existing 2-way `union()` integration test still passes (the `NULLOP` leaf path is unchanged).

## Test

Added `unionThreeWay()` in `ColumnLevelLineageTests`. It runs a 3-way `UNION ALL` CTAS and asserts that the lineage hook emits **at least one** lineage event. On master the hook NPEs and records zero events; with the fix the recursion completes and events flow.

I avoided introducing a new JSON event fixture for this regression because the witness ("the hook emitted *something*") is sufficient to pin the bug — we don't need to assert on the specific column-lineage shape, only that the parser no longer drops the event. If the maintainers want a full event-shape assertion, happy to follow up with `cll3WayUnionStart.json` / `cll3WayUnionComplete.json` once I can run the container test locally.

## Refs

- Issue: #4481
- Affected since: the initial Hive integration (#3555); 3+ way unions had no test coverage


mobuchowski (approved):

mobuchowski:
Thanks for the fix @1fanwang

PR #4501, opened 2026-04-28
[Go] change generator

tnazarew:
Change the client classes generator

### Meaningful description
The generator used in the first version relied heavily on text generated by the `quicktype` tool.
This required manipulation on text rather than generating it from structured data. 

Now the code is generated by
- loading the spec files and parsing them with `go-jsonschema`
- extracts the intermediate resources that contain the facet information, resolves references
- renders the facet classes and helpers

This change also enables extension of generated classes and keeping them up to date with the spec and the client. (e.g. byool resources) 

Additionally, because it's a new client and there arent many users I want change
- `Run` interface for interacting with RunEvent -> `RunWrapper`
- `RunInfo` (representation of OL run section) -> `Run` (make consistent with `Job` and `Dataset`)
 

### Checklist


generator/go/main.go
```diff
@@ -0,0 +1,79 @@
...
+	var discoveredOL []discover.Facet
+
+	for _, path := range paths {
+		loader := schemas.NewDefaultCacheLoader([]string{"json"}, nil)
+		schema, err := loader.Load(path, "")
+		if err != nil {
+			log.Fatalf("load schema %s: %v", path, err)
+		}
+		for name, def := range schema.Definitions {
+			if _, ok := definitions[name]; !ok {
```
kacpermuda: **Silent first-wins on duplicate `$defs` names across schema files.** When the same `$defs` name appears in two facet schemas (perfectly possible — e.g. several facet files happen to define a local `Variant`/`Item`/`Field` definition with different shapes), the file that gets loaded first wins and all subsequent definitions with the same name are silently dropped. Whether the surviving definition is the right one depends entirely on the lexicographic order returned by `discover.ResolveGlobs`.

If this is intentional (you assume `$defs` names are globally unique across all OL facet files), please add an explicit warning when a duplicate-with-different-shape is encountered:

```go
if existing, ok := definitions[name]; ok {
    if !reflect.DeepEqual(existing, def) {
        log.Fatalf("conflicting $defs %q across files (loaded from %s)", name, path)
    }
    continue
}
```

The `log.Fatalf` makes a future schema collision loud at generation time instead of producing subtly-wrong Go types.

generator/go/discover/facets.go
```diff
@@ -0,0 +1,145 @@
...
+	return out
+}
+
+// FindAllFacets returns all facet types (job, dataset, run, input, output) without
+// any exclusion filter. Used for OL client code generation.
+func FindAllFacets(schema *schemas.Schema) []Facet {
+	keys := containerKeys(schema)
+	var out []Facet
+
+	for name, def := range schema.Definitions {
```
kacpermuda: **Non-deterministic facet ordering from map iteration.** `for name, def := range schema.Definitions` traverses a Go map, whose iteration order is intentionally randomized. The output `out []Facet` carries that order downstream into `BuildFacet`. While `RenderStructs`/`RenderHelpers` re-sort the top-level facet list, the **type names** chosen for anonymous nested objects depend on first-visit order in `buildObject`/`buildType` (because `ctx.objects[t] = obj` registers on first visit, and `inverseMap` doesn't have an entry for anonymous nested objects).

End result: re-running the generator on the same input can produce slightly different output (different field-naming for anonymous nested types) and trip the CI "Check generated files are up to date" gate at random.

Fix is one line — sort by name before iterating:
```go
names := make([]string, 0, len(schema.Definitions))
for n := range schema.Definitions { names = append(names, n) }
sort.Strings(names)
for _, name := range names {
    def := schema.Definitions[name]
    ...
}
```

Same goes for `containerKeys` (line 65) and `flatProperties` / property-iteration in `ir/build.go` lines 105 and 367.

client/go/pkg/openlineage/spec.gen.go
```diff
@@ -12,26 +12,53 @@ import (
...
+	// The JSON Pointer (https://tools.ietf.org/html/rfc6901) URL to the corresponding version of the schema definition for this RunEvent
+	SchemaURL string `json:"schemaURL"`
+}
+
+// RunWrapper
+type Run struct {
+	// The run facets.
+	Facets *facets.RunFacets `json:"facets,omitempty"`
+	// The globally unique ID of the run associated with the job.
+	RunId string `json:"runId"`
```
kacpermuda: **`RunId` is a regression** — the previous file had `RunInfo.RunID` (proper Go convention), and this PR renames it back to `RunId` because `genutil.FormatGoFieldName` deliberately keeps `Id` lowercase (`util.go:42` — "Id is kept as-is").

Two problems with that decision:
1. **It's a public API breaking change** for every external Go consumer that reads `event.Run.RunID`.
2. **It will trip `golangci-lint`** on the next run — `revive` and `stylecheck` both flag `Id` as a known-acronym mistake. The CI in this PR newly enforces `golangci-lint run ./...` (continue_config.yml:1283), so the file you commit here may not lint cleanly on the first run after a future spec change introduces another `Id` field.

Recommend:
```go
func FormatGoFieldName(name string) string {
    if strings.HasSuffix(name, "Uri") { ... }
    if strings.HasSuffix(name, "Url") { ... }
    if strings.HasSuffix(name, "Id")  { return strings.TrimSuffix(name, "Id") + "ID" }
    return name
}
```

If you really want to keep `RunId` as-is for some compatibility reason, please document **why** in the comment on `FormatGoFieldName` — "Id is kept as-is" doesn't explain the trade-off.

generator/go/render/spec/renderer.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	b.WriteString("type EventType string\n\n")
+	b.WriteString("const (\n")
+	for _, v := range enumProp.Enum {
+		s, ok := v.(string)
+		if !ok {
+			continue
+		}
+		// Title-case the suffix: "ABORT" → "Abort"
+		suffix := strings.ToUpper(s[:1]) + strings.ToLower(s[1:])
```
kacpermuda, line 117: Panics on an empty enum value: `s[:1]` indexes the first byte unconditionally. The guard above (`if !ok { continue }`) only handles non-string types, not empty strings. A future schema change that introduced a `""` value in a string enum would crash the generator. One-line fix:

```go
if s == "" {
    continue
}
```

Same caveat applies to `enumValueSuffix` in `render/client/facets.go:744` — `w[:1]` would panic on an empty word, though `if w == "" { continue }` is already there. The renderer.go path is the gap.

generator/go/render/client/facets.go
```diff
@@ -0,0 +1,884 @@
...
+			continue
+		}
+		// Only scalar/union/object types go into the constructor; collections use With* setters.
+		scalar, ok := scalarConstructorParam(field)
+		if !ok {
+			continue
+		}
+		paramName := genutil.LowerFirst(genutil.FormatGoFieldName(field.GoName))
+		// Avoid collision with reserved words
+		if paramName == "type" {
```
kacpermuda: **Keyword escaping is one-off and brittle.** Only `type` is escaped to `typ`. Any JSON property whose camelCase name happens to match a Go reserved word — `func`, `chan`, `select`, `case`, `default`, `range`, `package`, `import`, `return`, `struct`, `map`, `interface`, `go`, `var`, `const`, `for`, `if`, `switch`, `goto`, `break`, `continue`, `else`, `defer`, `fallthrough` — would generate non-compiling Go.

This is fine *today* because no OL facet has such a property, but the fix is one line and makes the generator robust to spec evolution:
```go
var goReserved = map[string]bool{
    "break": true, "case": true, "chan": true, "const": true, "continue": true,
    "default": true, "defer": true, "else": true, "fallthrough": true, "for": true,
    "func": true, "go": true, "goto": true, "if": true, "import": true,
    "interface": true, "map": true, "package": true, "range": true, "return": true,
    "select": true, "struct": true, "switch": true, "type": true, "var": true,
}

if goReserved[paramName] {
    paramName = paramName + "_"
}
```
Apply at both call sites (`emitConstructor` line 269, `emitWithSetter` line 327).

generator/go/render/client/facets.go
```diff
@@ -0,0 +1,884 @@
...
+		fmt.Sprintf("\t\tSchemaURL: %q,", f.SchemaURL),
+	}
+
+	for i := range f.Root.Fields {
+		field := &f.Root.Fields[i]
+		if !field.Required {
+			continue
+		}
+		// Only scalar/union/object types go into the constructor; collections use With* setters.
+		scalar, ok := scalarConstructorParam(field)
```
kacpermuda, line 290: **Required `List` and `Map` fields silently fall out of the constructor.** `scalarConstructorParam` returns `("", false)` for `ir.List` and `ir.Map` (see line 487-525), so `emitConstructor` skips them at line 264. A facet with `required: ["items"]` where `items` is an array will produce a `New<Facet>(producer string)` constructor that leaves `Items` as `nil`.

The `Required` flag in the JSON schema then becomes a lie at runtime — the field is required by the spec but unset by the generated constructor. Two consistent options:
1. Include required collections in the constructor too (caller passes a slice/map at construction time).
2. Drop the `Required` flag for collection-only fields in the IR, so consumers don't reasonably expect them to be set.

At minimum, please document the current behavior in the constructor doc comment ("required collections must be set with `WithFoo` after construction").

generator/go/render/client/facets.go
```diff
@@ -0,0 +1,884 @@
...
+	}
+	return "", false
+}
+
+// ── Nested type collection ────────────────────────────────────────────────────
+
+// collectNestedTypes walks all facet IR trees, collects unique ObjectDef instances
+// (by TypeName) that are NOT the root facet struct, and returns them sorted by name.
+func collectNestedTypes(facets []ir.Facet) []*ir.ObjectDef {
+	seen := map[string]*ir.ObjectDef{}     // TypeName → first seen ObjectDef
```
kacpermuda, line 612: **`seen` keyed by `TypeName` silently drops collisions with different shapes.** If two distinct `*ir.ObjectDef` instances (from different facets, or from different paths within one facet) end up with the same suggested type name but different field sets, only the first one ever ends up in `nested` and the second is silently lost. Then a struct field elsewhere that references the second's shape is now pointing at a struct missing those fields → won't compile, or worse, will compile and behave wrong if field sets overlap enough.

In practice this only bites on anonymous nested objects whose suggested name path collides — combined with the non-deterministic property-iteration order in `ir/build.go:105`, this is a latent bug just waiting for a schema reorg to surface.

Minimal defense: when inserting, assert structural equality with the existing entry; if they differ, panic with a clear message naming the colliding type. Better fix: name nested anonymous types with a unique counter or full path so collisions don't happen.

Same concern in `collectEnumTypes` (line 602) and `collectUnionTypes` (line 665).

generator/go/main.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		log.Fatalf("load OpenLineage.json: %v", err)
+	}
+	olDir := "../../client/go/pkg/openlineage"
+	_ = os.MkdirAll(olDir, 0o755)
+	write(olDir+"/spec.gen.go", spec.RenderSpec(specSchema))
+}
+
+func write(path, contents string) {
+	if err := os.WriteFile(path, []byte(contents), 0o644); err != nil {
```
kacpermuda: **No `go/format.Source` step on the output.** All three writes (`facets.gen.go`, `facet_helpers.gen.go`, `spec.gen.go`) go straight from `strings.Builder` to disk. Standard practice for Go code generators is:

```go
func write(path, contents string) {
    formatted, err := format.Source([]byte(contents))
    if err != nil {
        log.Fatalf("format %s: %v\n----\n%s", path, err, contents)
    }
    if err := os.WriteFile(path, formatted, 0o644); err != nil {
        log.Fatal(err)
    }
}
```

Two benefits:
1. The generated output is gofmt-stable. Today the renderer hand-rolls some alignment (`Producer:  producer,` with two spaces vs `Framework: framework,` with one — see `render/client/facets.go:253`) and skips alignment elsewhere; struct field tags in `JobFacets` (etc.) aren't column-aligned the way `gofmt` would do them. Anyone running `gofmt -w` on the repo will produce a diff against the generator's output and CI will fail.
2. **Parse errors become loud at generation time** instead of producing files that `go build` then chokes on with cryptic error messages.

generator/go/resolve/resolver.go
```diff
@@ -0,0 +1,76 @@
...
+	return &Resolver{defs: defs}
+}
+
+// Resolve follows $ref chains and returns the final pointed-to schema.
+func (r *Resolver) Resolve(t *schemas.Type) *schemas.Type {
+	if t == nil {
+		return nil
+	}
+
+	for t.Ref != "" {
```
kacpermuda: Unbounded `for t.Ref != ""` loop. A self-referential `$ref` (e.g. `{"$ref": "#/$defs/Foo"}` on a `$defs/Foo` that's also `{"$ref": "#/$defs/Foo"}`) would spin forever. The OL spec doesn't have such cycles, but this resolver also runs over user-imported registry/facet files (per the globs in `main.go:23-27`), so a malformed user schema would hang the generator with no diagnostic.

```go
for depth := 0; t.Ref != ""; depth++ {
    if depth > 32 {
        return nil  // or log.Fatalf with the ref path
    }
    ...
}
```

client/go/pkg/openlineage/context.go
```diff
@@ -21,46 +21,46 @@ const currentRunKey runContextKeyType = iota
...
+func ContextWithRun(parent context.Context, run RunWrapper) context.Context {
 	return context.WithValue(parent, currentRunKey, run)
 }
 
-// Run is an instrumentation utility that allows for more ergonomic usage of the SDK.
+// RunWrapper is an instrumentation utility that allows for more ergonomic usage of the SDK.
 // It is loosely modeled after the OpenTelemetry Span/Trace APIs.
-type Run interface {
-	Parent() Run
+type RunWrapper interface {
```
kacpermuda, line 42: **Breaking-rename heads up.** This PR renames the public `Run` interface to `RunWrapper` (because the generated `Run` struct now claims that name) and renames `RunInfo` → `Run`. Also `facets.ParentJob` → `facets.ParentRunFacetJob` and `facets.ParentRun` → `facets.ParentRunFacetRun`. All public, all binary-incompatible for downstream consumers.

Requests:
1. CHANGELOG entry that names every renamed exported symbol so external SDK users have something to grep their codebase against.
2. Major version bump on the Go module (this is technically required by Go's module compatibility rules — `RunInfo` is a public type that's been removed).
3. Optional but very nice: keep a one-release deprecated alias
   ```go
   // Deprecated: use RunWrapper.
   type Run = RunWrapper  // would fail because Run is now a struct
   ```
   …which actually doesn't work *because* the new `Run` struct now occupies the name. So at minimum, please call this out as a hard break in the release notes.

generator/go/discover/facets.go
```diff
@@ -0,0 +1,145 @@
...
+	const suffix = "/$defs/"
+	i := strings.LastIndex(ref, suffix)
+	if i < 0 {
+		return ""
+	}
+	return ref[i+len(suffix):]
+}
+
+// FindFacets returns only the Job and Dataset facets (used for Terraform model generation).
+func FindFacets(schema *schemas.Schema) []Facet {
```
kacpermuda: `FindFacets` and `isExcluded` (line 136) are dead code — `main.go` only calls `FindAllFacets`, and `isExcluded` is only referenced from `FindFacets`. The doc comment says "used for Terraform model generation" but no such caller exists in this PR.

If the Terraform generator is coming in a follow-up, a `// TODO(<owner>): used by ../../../terraform-gen (incoming)` would prevent this from being deleted in a future cleanup. If it's leftover from the old generator, please drop it — `golangci-lint`'s `unused` linter will flag it on the next CI run anyway.

.circleci/continue_config.yml
```diff
@@ -1261,7 +1261,23 @@ jobs:
...
+      - run:
+          name: Download generator dependencies
+          command: cd ~/openlineage/generator/go && go mod download
+      - run:
+          name: Run facet generator
+          command: cd ~/openlineage/generator/go && go run .
+      - run:
+          name: Check generated files are up to date
+          command: |
+            if ! git diff --quiet -- pkg/facets/; then
```
kacpermuda: The drift check only inspects `pkg/facets/`, but the generator also writes `pkg/openlineage/spec.gen.go` (per `generator/go/main.go:72`). If the OL core spec changes and `spec.gen.go` becomes stale, this gate passes silently and the regression lands.

```yaml
if ! git diff --quiet -- pkg/facets/ pkg/openlineage/spec.gen.go; then
    echo "ERROR: Generated files are out of sync with the spec."
    ...
    git diff -- pkg/facets/ pkg/openlineage/spec.gen.go
    exit 1
fi
```

generator/go/main.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/atombender/go-jsonschema/pkg/schemas"
+
+	"github.com/OpenLineage/openlineage/generator/go/discover"
+	"github.com/OpenLineage/openlineage/generator/go/ir"
+	"github.com/OpenLineage/openlineage/generator/go/render/client"
+	"github.com/OpenLineage/openlineage/generator/go/render/spec"
+	"github.com/OpenLineage/openlineage/generator/go/resolve"
+)
+
+func main() {
```
kacpermuda, line 24: **No tests for the generator itself.** This PR adds 1700+ lines of generation logic across 9 Go files and zero `*_test.go` files. The only verification is the CI "regenerated output matches committed output" check, which catches drift but not correctness for inputs the generator hasn't seen yet.

A minimal test suite would catch the bulk of the issues I've flagged in this review:
- Determinism: feed a fixture facet schema, run twice, assert outputs are byte-equal.
- `FormatGoFieldName`: table-test for `Id`/`Uri`/`Url` suffixes (would have caught the `RunId` regression).
- `_apply_…`-style logic in `ir/build.go`: feed a oneOf with a discriminator, assert the correct `DiscriminatorField` is detected.
- `enumValueSuffix`: table-test for `"ALTER"`, `"GREATER_THAN"`, `"some_value"`, `""` (for the panic concern).
- `Resolver.Resolve`: assert behavior on cyclic refs, missing $defs, cross-file refs.

On a code generator this exposed (it gates the entire OL Go SDK), "trust the regenerated output" is a thin safety net.

generator/go/render/client/facets.go
```diff
@@ -0,0 +1,884 @@
...
+	fmt.Fprintf(b, "type %s interface {\n\t%s\n}\n\n", u.TypeName, sealMethod)
+
+	// 2. Raw wrapper struct
+	rawName := "Raw" + u.TypeName
+	fmt.Fprintf(b, "// %s wraps %s for JSON marshaling/unmarshaling with type dispatch.\n", rawName, u.TypeName)
+	fmt.Fprintf(b, "type %s struct{ V %s }\n\n", rawName, u.TypeName)
+
+	if u.DiscriminatorField != "" {
+		// UnmarshalJSON
+		fmt.Fprintf(b, "func (r *%s) UnmarshalJSON(data []byte) error {\n", rawName)
```
kacpermuda, line 873: **Generated unions are missing variant constructors that pre-set the discriminator.** The `Raw<Union>.UnmarshalJSON` peeks at `peek.Type` and dispatches to a concrete struct, but the concrete struct is emitted as a plain struct (via `emitNestedStruct`) where the discriminator field — say `Type string \`json:"type"\`` — is just an ordinary Go field with zero-value `""`.

That means a user who constructs `BinarySubsetCondition{Left: ..., Right: ...}` and marshals it gets JSON with `"type": ""`, which would then fail to round-trip through this same `UnmarshalJSON`. There's no compile-time signal for this.

Suggest emitting per-variant constructors that set the discriminator constant:
```go
func NewBinarySubsetCondition(left, right RawBaseSubsetCondition, op string) BinarySubsetCondition {
    return BinarySubsetCondition{
        Type:  "binary",
        Left:  left,
        Right: right,
        Op:    op,
    }
}
```
Or embed the discriminator value as a method (`func (BinarySubsetCondition) Type() string { return "binary" }`) and exclude it from the struct entirely.

At minimum, the variant struct's discriminator field should be private+const or have a runtime check in `MarshalJSON` that errors if the discriminator value is empty.

generator/go/main.go
```diff
@@ -0,0 +1,79 @@
...
+	paths, err := discover.ResolveGlobs(globs)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	definitions := map[string]*schemas.Type{}
+	var discoveredOL []discover.Facet
+
+	for _, path := range paths {
+		loader := schemas.NewDefaultCacheLoader([]string{"json"}, nil)
```
kacpermuda: `schemas.NewDefaultCacheLoader` is constructed inside the loop, so the cache is reset on every iteration. If two facet schemas reference the same external file (e.g. both `$ref` `OpenLineage.json`), it'll be loaded and parsed once per facet schema instead of once total.

Move it out:
```go
loader := schemas.NewDefaultCacheLoader([]string{"json"}, nil)
for _, path := range paths {
    schema, err := loader.Load(path, "")
    ...
}
```

Minor performance only — the generator runs in CI, not hot-path — but it's also more honest about what "default cache loader" means.

generator/go/render/client/facets.go
```diff
@@ -0,0 +1,884 @@
...
+		if required {
+			return "int64"
+		}
+		return "*int64"
+	case ir.Float:
+		if required {
+			return "float64"
+		}
+		return "*float64"
+	case ir.DateTime:
```
kacpermuda, line 423: **Required `DateTime` fields are silently optional.** `olTypeExpr` for `ir.DateTime` returns `*time.Time` regardless of `required`, and `olJSONTag` (line 473-477) always tags DateTime fields with `omitempty`. Together that means any required date-time field in a facet schema can be left nil at construction time and silently emitted as a missing JSON field.

For the OL spec right now this affects fields like `nominalStartTime` / `nominalEndTime` (NominalTimeRunFacet, both required per spec) — the generator marks them required in the IR but the runtime contract is broken.

Fix:
```go
case ir.DateTime:
    if required {
        return "time.Time"
    }
    return "*time.Time"
```
and in `olJSONTag`:
```go
if _, ok := t.(ir.DateTime); ok && !required {
    return fmt.Sprintf(`json:"%s,omitempty"`, jsonName)
}
```

Also requires updating `emitConstructor` so required `DateTime` fields go in as `time.Time` (no `&` prefix on assignment, line 275-277).

generator/go/render/spec/renderer.go
```diff
@@ -0,0 +1,363 @@
...
+
+	switch ptype {
+	case "string":
+		if t.Format == "date-time" {
+			if required {
+				return "time.Time"
+			}
+			return "*time.Time"
+		}
+		if len(t.Enum) > 0 {
```
kacpermuda: **Any string enum in the merged Event is hardcoded to `*EventType`.** This works *today* because `eventType` is the only enumerated string in `BaseEvent`/`RunEvent`/`DatasetEvent`/`JobEvent`. The moment the spec adds another (e.g. an `eventSource` enum, or an enum on `BaseEvent.priority`), this branch silently types it as `*EventType` too — which is just wrong.

The correct behavior mirrors the facets renderer: emit a uniquely-named string type per enum, plus the corresponding `const` block. Either lift `buildEnum`/`emitEnumType` into a shared helper between `render/spec` and `render/client`, or special-case the lookup so this branch only kicks in when `propName == "eventType"`.

generator/go/discover/facets.go
```diff
@@ -0,0 +1,145 @@
...
+	ContainerKey string // JSON key used in the facets container struct (e.g. "columnLineage")
+}
+
+func detectFacetType(t *schemas.Type) (FacetType, bool) {
+	for _, sub := range t.AllOf {
+		if sub.Ref == "" {
+			continue
+		}
+		switch {
+		case strings.HasSuffix(sub.Ref, "/JobFacet"):
```
kacpermuda: **Facet-type detection by `HasSuffix` is too loose.** A `$ref` ending in `/JobFacet` matches the canonical OL base, but it would *also* match a hypothetical custom ref like `#/$defs/MyExtensionJobFacet` (where `MyExtensionJobFacet` is itself a derived facet, not the base). The intent is presumably "matches exactly the OL base facet types", which means the check should anchor on the full ref path:

```go
func isBaseFacetRef(ref, baseName string) bool {
    return strings.HasSuffix(ref, "OpenLineage.json#/$defs/"+baseName)
}
```

This matches the stricter check `isIgnorableBaseFacetRef` in `resolve/resolver.go:63-76` already does. Worth keeping the two consistent — currently the discover-side is permissive and the resolve-side is strict, which is exactly the kind of subtle inconsistency that produces hard-to-debug code-gen drift.

generator/go/render/spec/renderer.go
```diff
@@ -0,0 +1,363 @@
...
+		suffix := strings.ToUpper(s[:1]) + strings.ToLower(s[1:])
+		fmt.Fprintf(b, "\tEventType%s EventType = %q\n", suffix, s)
+	}
+	b.WriteString(")\n\n")
+}
+
+// emitMergedEventStruct emits the combined Event struct from BaseEvent +
+// RunEvent + DatasetEvent + JobEvent. Fields from BaseEvent are required;
+// all event-specific fields are optional (any given event may lack them).
+func emitMergedEventStruct(b *strings.Builder, defs map[string]*schemas.Type) {
```
kacpermuda: **The merged `Event` struct makes every event-specific field optional.** Comment at line 126-127 acknowledges this ("Fields from BaseEvent are required; all event-specific fields are optional"). The reasoning is clear (a single struct must accommodate any event type), but the consequence is that the Go type system can no longer enforce "a RunEvent must have run + job" or "a DatasetEvent must have dataset" — caller can construct `Event{EventTime: now()}` and emit it, getting a `400 Bad Request` from the receiver at runtime instead of a compile error.

Two paths forward:
1. Emit three distinct typed event structs (`RunEventPayload`, `DatasetEventPayload`, `JobEventPayload`) each with the right requireds, and a discriminated union for the wire format. More work, more correct.
2. Keep the merged struct but document the constraint loudly in the doc comment on `Event` and add a `Validate()` method generated alongside it that returns an error on missing required-by-subtype fields.

At minimum, the doc comment on `Event` ("// Event represents an OpenLineage event.") should mention that subtype validation is the caller's responsibility.

generator/go/render/spec/renderer.go
```diff
@@ -0,0 +1,363 @@
...
+		required[r] = true
+	}
+	for k, v := range t.Properties {
+		if _, exists := props[k]; !exists {
+			props[k] = v
+		}
+	}
+	for _, sub := range t.AllOf {
+		if sub.Ref != "" {
+			if def, ok := defs[refName(sub.Ref)]; ok {
```
kacpermuda, line 278: `gatherProps` reimplements ref resolution by hand (`defs[refName(sub.Ref)]`) instead of going through `resolve.Resolver`. Two problems with that:

1. **It silently disagrees with the resolver on cross-file refs.** `resolve.Resolver.Resolve` returns nil for any ref that isn't `#/$defs/...` (in-document). `gatherProps` strips `/$defs/` from anywhere in the ref string, so a cross-file ref like `OpenLineage.json#/$defs/BaseEvent` *would* be looked up in this `defs` map (the local OpenLineage.json defs, which happens to contain `BaseEvent`) — works by coincidence, not by design.
2. **It doesn't apply the `isIgnorableBaseFacetRef` filter** that `resolve.Resolver` does. So if a non-facet schema happens to allOf-embed a base facet ref, this would gather those props (`_producer`, `_schemaURL`, etc.) into the merged Event struct.

Reuse `resolve.Resolver` here (or extract the resolution logic into a shared utility). One source of truth for `$ref` semantics is much easier to keep correct than two.

generator/go/ir/build.go
```diff
@@ -0,0 +1,424 @@
...
+		return Int{}
+
+	case "number":
+		return Float{}
+
+	case "array":
+		// Array element inherits the same suggested name; the inverseMap will
+		// override it with the $defs name if the items type is a named ref.
+		return List{
+			Elem: buildType(ctx, r.Resolve(t.Items), r, suggestedTypeName),
```
kacpermuda: Array elements inherit the array field's name verbatim as their suggested type name. So a JSON property `runs: [{...}]` whose element is an anonymous object gets the suggested type name `Runs` (plural) — and if the element isn't in `$defs` (so `inverseMap` doesn't override), the generated Go type ends up named `…Runs` instead of the more idiomatic singular `…Run`.

For the existing OL spec this is rarely visible because most array elements are `$defs`-backed. But if a future facet adds an inline array of anonymous objects, you'll get awkward Go names like `OwnershipDatasetFacetOwners` (plural) instead of `OwnershipDatasetFacetOwner`.

Lightweight fix: depluralize the suggested name when entering an array element. Simple heuristic — strip a trailing `s`, `es`, or `ies` — covers most cases and lets `inverseMap` still override when a $defs entry is available.

kacpermuda (commented):
Adding some comments from Opus4.7 review.

client/go/pkg/openlineage/spec.gen.go
tnazarew: yeah, I didn't know that it was the "go-way", changed back to ID

generator/go/render/spec/renderer.go
tnazarew: I don't think we will add those

generator/go/render/spec/renderer.go
tnazarew: changed to `BaseEvent` and embedding structs

generator/go/render/spec/renderer.go
tnazarew: actually in the facet generation we want to skip the refs to their base facets as they are defined elsewhere but in spec generation we actually want to use the refs

generator/go/render/spec/renderer.go
tnazarew: I think it's good enough for now, I doubt we can expect adding any fields to main spec that way

generator/go/main.go
tnazarew: this module will probably only be used by OL devs when there are changes to the spec, in that case we can check the generated files correctness 

at the very least it works that way for the java client

client/go/pkg/openlineage/context.go
tnazarew: in the first version the `Run` structure was called `RunInfo` which didn't make sense when the other structures were just `Job` or `Dataset`. When it became `Run` this had to be renamed to something else

I figured it's the best time to fix it, since the module was "just" created and there are probably no active users yet

generator/go/discover/facets.go
tnazarew: I'll re-add it later

generator/go/discover/facets.go
tnazarew: added sorting

mobuchowski (approved):

PR #4506, opened 2026-05-01
flink: extract generic schema and config facet visitors

jsingh-yelp:
### Problem

Flink 2 lineage datasets outside the Table API, such as DataStream API connector datasets, can expose schema and config metadata through Flink lineage facets. The OpenLineage listener previously handled config extraction only inside `TableLineageFacetVisitor`, which meant this metadata was not handled for non-table dataset APIs.

### Changes

- Add `ConfigFacetVisitor` to extract `DatasetConfigFacet` from any Flink lineage dataset.
- Add `SchemaFacetVisitor` to extract generic `DatasetSchemaFacet` from non-Table datasets.
- Keep Table API schema extraction in `TableLineageFacetVisitor`, where it can use `CatalogBaseTable` and preserve table-specific schema behavior.
- Register the new facet visitors in the Flink 2 listener factory.
- Add unit tests for config facet extraction, schema facet extraction, and Table API schema/documentation handling.


jsingh-yelp:
cc: @pawel-big-lebowski, @mobuchowski Can I please have review and some thoughts on this change?

mobuchowski (approved):

mobuchowski:
Thanks @jsingh-yelp!

PR #4509, opened 2026-05-01
fix(python): read config files as utf-8

### LINKED ISSUE #4507: Python client config file should be read as UTF-8
﻿## Problem

The Python client reads the OpenLineage YAML config file with `open(config_path)` in `OpenLineageClient._get_config_file_content`. That uses the platform default text encoding instead of UTF-8.

On systems where the default encoding is not UTF-8, a valid UTF-8 `openlineage.yml` containing non-ASCII characters can fail to load with a `UnicodeDecodeError`, causing the client to ignore the config and fall back to console transport.

## Expected behavior

OpenLineage config files should be decoded consistently as UTF-8 across platforms.

## Proposed fix

Open the YAML config file with `encoding="utf-8"` and add a regression test that simulates a non-UTF-8 default encoding failing on UTF-8 content.

hcthakur2004:
﻿## Summary

- read Python client YAML config files with explicit UTF-8 decoding
- add a regression test that simulates a non-UTF-8 platform default encoding

## Issue

Fixes #4507

## Tests

- `uv run pytest tests/test_client.py -k "get_config_file_content"`
- `uv run ruff check src/openlineage/client/client.py tests/test_client.py`

### DIFF
--- a/client/python/src/openlineage/client/client.py
+++ b/client/python/src/openlineage/client/client.py
@@ -290,7 +290,7 @@ class OpenLineageClient:
     @staticmethod
     def _get_config_file_content(config_path: str) -> dict[str, Any]:
         try:
-            with open(config_path) as f:
+            with open(config_path, encoding="utf-8") as f:
                 config: dict[str, Any] | None = yaml.safe_load(f)
                 if not config:
                     log.error("Empty OpenLineage config file: `%s`", config_path)
--- a/client/python/tests/test_client.py
+++ b/client/python/tests/test_client.py
@@ -370,6 +370,26 @@ def test_get_config_file_content(root: Path) -> None:
     }
 
 
+def test_get_config_file_content_reads_utf8(
+    tmp_path: Path,
+    mocker: MockerFixture,
+) -> None:
+    config_path = tmp_path / "openlineage.yml"
+    config_path.write_text("transport:\n  type: console\nmetadata: café\n", encoding="utf-8")
+    real_open = open
+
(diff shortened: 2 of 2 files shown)


chuenchen309:
Looks correct to me — config files are user-authored YAML that can legitimately contain non-ASCII (a `metadata: café`, non-ASCII paths, etc.), so relying on the platform default encoding is a real portability bug on ASCII/cp1252 defaults. Nice repro with the `open`-defaults-to-ascii mock.

One scope check I went and verified so it doesn't come up later: the other text write in the client, `FileTransport.emit` (`transport/file.py`, opens in `"a"`/`"w"` mode without `encoding=`), does **not** need the same fix — `Serde.to_json` calls `json.dumps` with the default `ensure_ascii=True`, so event output is pure ASCII (any non-ASCII in the payload is `\u`-escaped) and is safe under any locale. The asymmetry is intentional: config is read as raw bytes-from-disk, events are emitted as ascii-escaped JSON. So this PR's scope (config read only) is right.

*(Disclosure: I'm an AI agent — Claude Code, on this account — reviewing autonomously; I verified the `ensure_ascii` behavior against `main` rather than asserting it.)*

kacpermuda (approved):

chuenchen309:
Flagging this for the next merge pass: approved since Jul 20, and still `MERGEABLE` / `CLEAN` with checks green — no conflicts to resolve and nothing outstanding from the review. This repo normally merges within hours of an approval, so it looks like this one just slipped rather than being held for anything.

PR #4510, opened 2026-05-02
integrations: add Prefect integration

merobi-hub:
Adds a Prefect producer to integration

### Meaningful description
This integration maps Prefect's `task` and `task run` to the spec's `job` and `job run`, respectively. It also leverages the Prefect Event Clients, which offer a `get_events_subscriber()` method and `EventNameFilter` for collecting task and flow events. The Prefect `Running` event type is used to approximate the OL `START` type.

Facets leveraged:
- PrefectDeploymentRunFacet: adds deployment information to tasks (a new custom facet)
- ParentRunFacet: uses flow events to add parents to tasks
- NominalTimeRunFacet: uses `expected_start_time` datetimes in tasks
- JobDependenciesRunFacet: uses Prefect `parents` to add upstream dependencies to tasks
- ProcessingEngineRunFacet: adds the Prefect version to runs. 

This has been tested with OSS Prefect running locally with Marquez as the consumer. See the comments for some payloads from Marquez. 

A note about Datasets: Assets would be ideal for creating Datasets if the Assets API were available, but it is gated in Prefect Cloud. As a workaround, this uses Artifacts. Users are asked to pass information about datasets using the Artifact library's `create_artifact_table` function within task definitions.

For example:
```
ol_table = [{"database_uri":"duckdb:///customers_db", "table":"customers"}]

create_table_artifact(
    key="upstream-insert",
    table=ol_table,
    description="ol-dataset_output"
)     
```
### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added OpenLineage integration for Prefect flows and tasks.
  * Emits lifecycle events for flow and task starts, completions, and failures.
  * Captures deployment metadata, parent runs, dependencies, timing, and input/output datasets.
  * Supports asynchronous collection and processing of Prefect runs.
  * Adds Prefect deployment metadata to emitted run information.

* **Documentation**
  * Added setup, configuration, asynchronous execution, and dataset usage guidance.

* **Tests**
  * Added comprehensive coverage for event emission, metadata, facets, listeners, and dataset handling.


integration/prefect/listener.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2018-2026 contributors to the OpenLineage project
+# SPDX-License-Identifier: Apache-2.0
+
+import asyncio
+
+from prefect.events.clients import get_events_subscriber
+from prefect.client.orchestration import get_client
+from prefect.events.filters import EventFilter, EventNameFilter
+from prefect.runtime import task_run
+from . import adapter, PrefectOpenLineageAdapter
```
mobuchowski: let's do absolute imports only

integration/prefect/listener.py
```diff
@@ -0,0 +1,53 @@
...
+			if event.event.split(".")[-1] in ['Running', 'Complete', 'Failed']:
+
+				if 'Running' in event.event:
+					event_type = 'Running'
+
+				elif 'Complete' in event.event:
+					event_type = 'Complete'
+
+				elif 'Failed' in event.event:
+					event_type = 'Failed'
```
mobuchowski: we can use switch here :)

mobuchowski (changes requested):
Can we mark this as experimental? I think it's fine to iterate in the main repo as long as we make it clear.

integration/prefect/adapter.py
```diff
@@ -0,0 +1,56 @@
...
+from openlineage.client.run import Job, Run, RunEvent, RunState
+
+CLIENT: OpenLineageClient = OpenLineageClient.from_environment()
+PRODUCER: str = 'https://github.com/OpenLineage/OpenLineage/tree/$VERSION/integration/prefect'
+JOB_NAMESPACE: str = os.environ.get('OPENLINEAGE_NAMESPACE')
+
+class PrefectOpenLineageAdapter:
+
+    def generate_job_name(self, flow_name: str, task_name: str):
+        return self.job_namespace + '.' + flow_name + '.' + task_name
```
dolfinus: Why job name contains namespace? Also there is no `job_namespace` attribute in this class, as well as constructor

integration/prefect/adapter.py
```diff
@@ -0,0 +1,56 @@
...
+            job=Job(
+                self.JOB_NAMESPACE, 
+                self.generate_job_name(flowName, taskName),
+            ),
+            producer=PRODUCER
+        )
+
+        try:
+            CLIENT.emit(run_event)
+            print('OpenLineage event sent successfully')
```
dolfinus: Why using print instead of logging?

integration/prefect/adapter.py
```diff
@@ -0,0 +1,56 @@
...
+    
+    def create_and_emit_events(
+        self, 
+        eventType: str = None,
+        eventTime: datetime = None,
+        flowName: str = None,
+        taskName: str = None,
+    ) -> RunEvent:
+
+        run_id: str = str(uuid.uuid4())
```
dolfinus: Please use `openlineage.client.uuid.generate_new_uuid`

integration/prefect/adapter.py
```diff
@@ -0,0 +1,56 @@
...
+        elif eventType == 'Failed':
+            eventType: RunState = RunState.FAILED
+
+        run_event = RunEvent(
+            eventType=eventType,
+            eventTime=eventTime.isoformat(),
+            run=Run(run_id),
+            job=Job(
+                self.JOB_NAMESPACE, 
+                self.generate_job_name(flowName, taskName),
```
dolfinus: Please add jobType facet to distinct Prefect jobs from anything else

integration/prefect/listener.py
```diff
@@ -0,0 +1,53 @@
...
+from prefect.events.clients import get_events_subscriber
+from prefect.client.orchestration import get_client
+from prefect.events.filters import EventFilter, EventNameFilter
+from prefect.runtime import task_run
+from . import adapter, PrefectOpenLineageAdapter
+
+async def stream_task_runs():
+	"""Requires that PREFECT_API_URL environment variable be set"""
+
+	async def get_flow_name_from_task_id(task_run_id: str):
```
dolfinus: Why using nested `def` statements? Inner function doesn't capture any variable from outer one, so it can be moved outside

integration/prefect/adapter.py
```diff
@@ -0,0 +1,56 @@
...
+
+        elif eventType == 'Failed':
+            eventType: RunState = RunState.FAILED
+
+        run_event = RunEvent(
+            eventType=eventType,
+            eventTime=eventTime.isoformat(),
+            run=Run(run_id),
+            job=Job(
+                self.JOB_NAMESPACE,
```
dolfinus: This constant is defined on module level, not class level. Also I recommend to explicitly pass namespace instead of relying on env variables, for testing purposes

kacpermuda:
An important question, can should we try adding the integration directly in the prefect repo so it's more sustainable in the future? Not sure if we should be adding this in the OL repo without trying that path first.

integration/prefect/adapter.py
```diff
@@ -0,0 +1,56 @@
...
+                self.JOB_NAMESPACE, 
+                self.generate_job_name(flowName, taskName),
+            ),
+            producer=PRODUCER
+        )
+
+        try:
+            CLIENT.emit(run_event)
+            print('OpenLineage event sent successfully')
+        except:
```
dolfinus: This pattern hides the reason which exception was raised and why

integration/prefect/adapter.py
```diff
@@ -0,0 +1,67 @@
...
+                job_facets
+            ),
+            producer=PRODUCER
+        )
+
+        try:
+            CLIENT.emit(run_event)
+            logger.info('Emitted OpenLineage event successfully.')
+        except Exception as e:
+            logger.warning(f'Could not emit OpenLineage event. {e}')
```
dolfinus:
```suggestion
            logger.exception('Could not emit OpenLineage event')
```

integration/prefect/adapter.py
```diff
@@ -0,0 +1,67 @@
...
+import logging
+import os
+from typing import List, Dict, Optional
+
+from openlineage.client import OpenLineageClient
+from openlineage.client.facet import JobTypeJobFacet
+from openlineage.client.run import Job, Run, RunEvent, RunState
+from openlineage.client.uuid import generate_new_uuid
+
+CLIENT: OpenLineageClient = OpenLineageClient.from_environment()
```
dolfinus: Why client is initialized during module import? It probably should be passed via consteuctor arguments

dolfinus: Nothing changed, actually

integration/prefect/adapter.py
```diff
@@ -0,0 +1,67 @@
...
+from openlineage.client.run import Job, Run, RunEvent, RunState
+from openlineage.client.uuid import generate_new_uuid
+
+PRODUCER: str = 'https://github.com/OpenLineage/OpenLineage/tree/$VERSION/integration/prefect'
+JOB_NAMESPACE: str = os.environ.get('OPENLINEAGE_NAMESPACE')
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+class PrefectOpenLineageAdapter:
+    client: OpenLineageClient = OpenLineageClient.from_environment()
```
dolfinus:
```suggestion
    def __init__(
        self,
        client: OpenLineageClient | None = None,
        job_namespace: str | None = None,
    ):
      self.client = client or OpenLineageClient.from_environment()
      self.job_namespace = job_namespace or os.getenv("JOB_NAMESPACE", "default")
```
Constructor is not generated from field definition. Current version actually defines a static class field which is initialized during module import.

integration/prefect/adapter.py
```diff
@@ -0,0 +1,67 @@
...
+            integration="Prefect", 
+            jobType="TASK"
+        )}
+
+        run_event = RunEvent(
+            eventType=eventType,
+            eventTime=eventTime.isoformat(),
+            run=Run(run_id),
+            job=Job(
+                'prefect_test', # to do: replace with constant
```
dolfinus:
```suggestion
                self.job_namespace,
```

integration/prefect/adapter.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+from typing import List, Dict, Optional
+
+from openlineage.client import OpenLineageClient
+from openlineage.client.facet import JobTypeJobFacet
+from openlineage.client.run import Job, Run, RunEvent, RunState
+from openlineage.client.uuid import generate_new_uuid
+
+PRODUCER: str = 'https://github.com/OpenLineage/OpenLineage/tree/$VERSION/integration/prefect'
+JOB_NAMESPACE: str = os.environ.get('OPENLINEAGE_NAMESPACE')
```
dolfinus:
```suggestion
```

dolfinus (changes requested):

integration/prefect/listener.py
```diff
@@ -0,0 +1,59 @@
...
+						event_type: str = 'RUNNING'
+					case 'Completed':
+						event_type: str = 'COMPLETE'
+					case 'Failed':
+						event_type: str = 'FAILED'
+
+				task_run_id = event.resource.id.split(".")[-1]
+				flow_name = await get_flow_name_from_task_id(task_run_id)
+
+				ol_adapter = PrefectOpenLineageAdapter()
```
dolfinus: Why OL adapter is created for every event?

integration/prefect/adapter.py
```diff
@@ -19,7 +19,13 @@
 logger: logging.Logger = logging.getLogger(__name__)
 
 class PrefectOpenLineageAdapter:
-    client: OpenLineageClient = OpenLineageClient.from_environment()
+    def __init__(
+        self,
+        client: OpenLineageClient | None = None,
+        job_namespace: str | None = None,
+    ):
+      self.client = client or OpenLineageClient.from_environment()
```
kacpermuda: `.from_environment()` is deprecated for a while now, just `OpenLineageClient()` is enough.

kacpermuda:
@merobi-hub 

Any thoughts on my comment above?

>An important question, can should we try adding the integration directly in the prefect repo so it's more sustainable in the future? Not sure if we should be adding this in the OL repo without trying that path first.

merobi-hub:
@kacpermuda I've been thinking about your comment. It's a good point. For context, I'm following the pattern we used with Airflow, where we bootstrapped an integration here and iterated first, but it's true we could have success starting in Prefect. That said, the development process and any discussions will be more efficient here where there are no barriers. Perhaps more OL contributors will be involved if we start here, also. We can always contribute it to Prefect down the road -- in fact, that's the idea. I'm open to other approaches, though.

harels:
Throwing out an idea:
Once we get this PR to a good state and we are happy with how the integration looks like (but before we merge it), maybe we engage the Prefect team and ask them if we can open this PR on their repo?

integration/prefect/adapter.py
dolfinus: Still not resolved @merobi-hub

integration/prefect/listener.py
```diff
@@ -0,0 +1,94 @@
...
+				flow_name: str = await get_flow_name_from_task_id(prefect_task_run_id)
+
+				ol_run_id: str = str(generate_new_uuid())
+				task_context[task_name] = ol_run_id
+
+				parent_runs: list = None
+				parents: bool = False
+				try:
+					parent_runs = []
+					task_parents: List = event.payload["task_run"]["task_inputs"]["__parents__"]
```
dolfinus: Parent job facet should represent flow -> task or flow -> nested flow relations. If there are task -> task relations within the same flow, this is probably job dependency facet

integration/prefect/listener.py
```diff
@@ -0,0 +1,94 @@
...
+					for parent in task_parents:
+						task_id: str = parent["id"] if parent["input_type"] == "task_run" else None
+						if task_id:
+							parent_run = await get_task_run_from_task_id(task_id)
+							parent_name = parent_run.name.split("-")[0]
+							parent_runs.append({"name": parent_name, "namespace": PARENT_RUN_NAMESPACE, "id": task_context[parent_name]})
```
dolfinus: This relies on assumption that all task runs are received by the same event listener in sequential order, and context already contains run_id for parent task run. This may be not the case for listener restart.

If task_run has field like `created_at`, I recommend using `generate_static_uuid(task_run.created_at, task_run.id)` instead of `generate_new_uuid()`, so the same task_run will get the same OpenLineage run_id, regardless of order of events.

integration/prefect/listener.py
```diff
@@ -0,0 +1,94 @@
...
+					case 'Failed':
+						event_type: str = "FAILED"
+
+				OL_ADAPTER.create_and_emit_event(
+					ol_run_id,
+					event_type, 
+					event_time, 
+					flow_name, 
+					task_name, 
+					parentRuns = parent_runs if parents else None
```
dolfinus: There is no need in this guard, in Python `if []:` is already `False`

integration/prefect/listener.py
merobi-hub: Makes sense. Repurposed parentRun for flows and added jobDependenciesFacet

integration/prefect/adapter.py
```diff
@@ -0,0 +1,96 @@
...
+
+        job_facets = {"jobType": JobTypeJobFacet(
+            processingType="BATCH", 
+            integration="Prefect", 
+            jobType="TASK"
+        )}
+        
+        def build_run_facets():
+            if jobDeps:
+                dep = jobDeps[0] # TODO: support multiple dependencies
```
dolfinus: Job dependencies is already a list, why limiting to just first dependency?

integration/prefect/listener.py
```diff
@@ -0,0 +1,114 @@
...
+		flow_run = await client.read_flow_run(flow_run_id) # TODO: type
+		flow_id: UUID = flow_run.flow_id
+		flow = await client.read_flow(flow_id) # TODO: type
+		return {"start_time": flow_run.start_time, "name": flow.name}
+
+async def collect_and_process_task_runs():
+	"""Requires PREFECT_API_URL environment variable be set"""
+
+	filter_criteria = EventFilter(
+    	event = EventNameFilter(prefix=["prefect.task-run."])
```
dolfinus: This filter should probably include _flow_ runs. Otherwise task run will declare parent run of flow, but will never emit respective event

integration/prefect/listener.py
```diff
@@ -0,0 +1,114 @@
...
+		async for event in subscriber:
+
+			event_time: datetime = event.occurred
+			task_name: str = event.resource.name.split("-")[0]
+			event_state: str = event.event.split(".")[-1] 
+
+			if event_state in ["Running", "Completed", "Failed"]:
+
+				prefect_task_run_id: str = event.resource.id.split(".")[-1]
+				ol_task_run_id: str = build_run_id(event_time, task_name, JOB_NAMESPACE)
```
dolfinus: Instead of event_time, a task run start/creation time. Otherwise START and COMPLETE events will have different run_id

integration/prefect/listener.py
```diff
@@ -0,0 +1,114 @@
...
+from adapter import PrefectOpenLineageAdapter
+from openlineage.client.uuid import generate_static_uuid
+from prefect.events.clients import get_events_subscriber
+from prefect.client.orchestration import get_client
+from prefect.events.filters import EventFilter, EventNameFilter
+from prefect.runtime import task_run
+
+FLOW_NAMESPACE: str = os.environ.get('FLOW_NAMESPACE', 'flow_test')
+JOB_NAMESPACE: str = os.environ.get('OPENLINEAGE_NAMESPACE', 'prefect_test')
+PARENT_RUN_NAMESPACE: str = os.environ.get('PARENT_RUN_NAMESPACE', 'parent_test')
```
dolfinus: What is the case when these can be different? All these flows and runs are emitted by the same Prefect instance

integration/prefect/listener.py
```diff
@@ -0,0 +1,114 @@
...
+				# Get flow run info
+				flow_namespace: str = FLOW_NAMESPACE
+				flow_data = await get_flow_run_from_task_id(prefect_task_run_id)
+				flow_name = flow_data["name"]
+				flow_run_id: str = build_run_id(flow_data["start_time"], flow_name, FLOW_NAMESPACE)
+
+				# Get job dependencies (Prefect "parents") info
+				parents: bool = False
+				try:
+					parent_runs: list = []
```
dolfinus:
```suggestion
				parent_runs = []
				try:
```

integration/prefect/listener.py
```diff
@@ -0,0 +1,114 @@
...
+					task_parents: List = event.payload["task_run"]["task_inputs"]["__parents__"]
+					for parent in task_parents:
+						task_id: str | None = parent["id"] if parent["input_type"] == "task_run" else None
+						if task_id:
+							parent_run = await get_task_run_from_task_id(task_id)
+							parent_name = parent_run.name.split("-")[0]
+							parent_run_id = build_run_id(parent_run.start_time, parent_name, PARENT_RUN_NAMESPACE)
+							parent_runs.append({"name": parent_name, "namespace": PARENT_RUN_NAMESPACE, "id": parent_run_id})
+					parents = True
```
dolfinus:
```suggestion
```

integration/prefect/listener.py
```diff
@@ -0,0 +1,114 @@
...
+				except:
+					logger.info(f"No task parents found for {prefect_task_run_id}.")
```
dolfinus:
```suggestion
					logger.info("No task parents found for %s", prefect_task_run_id)
```

integration/prefect/adapter.py
merobi-hub: The function does not yet support more than one dependency. It iss intended to be temporary

dolfinus: I mean, Prefect task can have multiple parent tasks, job dependency facet can have multiple upstream jobs. `jobDep = jobDeps[0]` can be replaced with `for jobDep in jobDeps` list comprehension

integration/prefect/listener.py
merobi-hub: Thank you for the suggestion -- implemented

integration/prefect/listener.py
merobi-hub: This makes sense. Instead of `event.occurred`, we can use `start_time` in the `task_run` dict

integration/prefect/listener.py
merobi-hub: Prefect allows tasks to be triggered from other flows, hence `PARENT_RUN_NAMESPACE`. This is what I initially had in mind. But, as there is no programmatic way to distinguish externally triggered tasks currently, I admit we do not need these vars at this time.

dolfinus: > Prefect allows tasks to be triggered from other flows

But these flows run on the same Prefect instance, right?

integration/prefect/adapter.py
```diff
@@ -0,0 +1,96 @@
...
+        
+        def build_run_facets():
+            if jobDeps:
+                for dep in deps:
+                    upstream_job = [job_dependencies_run.JobDependency(
+                        job=job_dependencies_run.JobIdentifier(
+                            namespace=dep["namespace"], 
+                            name=self.generate_job_name(flowName, dep["name"]) # assume same flow
+                        )
+                    )]
```
dolfinus:
```suggestion
                upstream_jobs = [
                    job_dependencies_run.JobDependency(
                        job=job_dependencies_run.JobIdentifier(
                            namespace=dep["namespace"], 
                            name=self.generate_job_name(flowName, dep["name"]) # assume same flow
                        )
                    )
                    for dep in deps
                ]
```

integration/prefect/adapter.py
merobi-hub: Implemented

integration/prefect/listener.py
merobi-hub: added flow runs to filter

integration/prefect/listener.py
```diff
@@ -0,0 +1,133 @@
...
+from prefect.runtime import task_run
+
+FLOW_NAMESPACE: str = os.environ.get('FLOW_NAMESPACE', 'flow_test')
+JOB_NAMESPACE: str = os.environ.get('OPENLINEAGE_NAMESPACE', 'prefect_test')
+PARENT_RUN_NAMESPACE: str = os.environ.get('PARENT_RUN_NAMESPACE', 'parent_test')
+OL_ADAPTER = PrefectOpenLineageAdapter()
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+@staticmethod
```
dolfinus: This is not a class method to mark it as static

integration/prefect/listener.py
```diff
@@ -0,0 +1,133 @@
...
+def build_run_id(
+	execution_time: datetime, 
+	run_name: str, 
+	namespace: str
+) -> str:
+	return str(generate_static_uuid(
+		instant=execution_time,
+        data=f"namespace.{run_name}".encode("utf-8"),
```
dolfinus: Why "namespace" is literal string?

integration/prefect/adapter.py
merobi-hub: thanks, will apply

integration/prefect/listener.py
merobi-hub: These do, yes. But it is possible to trigger a flow run and, hence, task run, in another instance using the API

dolfinus: But API can be called from any host, so PARENT_RUN_NAMESPACE env variable will not help. In this case, parent job info should probably be passed via [context](https://docs.prefect.io/v3/api-ref/rest-api/server/flow-runs/create-flow-run#response-context) field.

As for JOB_NAMESPACE vs FLOW_NAMESPACE, these should merged into one env variable. Job namespace describes "where" job exists/running, and this is the same Prefect instance for both flow and task runs.

integration/prefect/listener.py
```diff
@@ -0,0 +1,127 @@
...
+
+	filter_criteria = EventFilter(
+    	event = EventNameFilter(prefix=["prefect.task-run.", "prefect.flow-run."])
+	)
+
+	async with get_events_subscriber(filter=filter_criteria) as subscriber:
+        
+		async for event in subscriber:
+
+			if event.event.split(".")[1] == "task-run": # get task runs
```
dolfinus: Flow runs are allowed by filter, but excluded by this check

integration/prefect/listener.py
merobi-hub: WDYT about using an env var in the "parent" deployment to provide the namespace? This is very easy on the user because you can add env vars to prefect.yaml. I added a function to get the namespace from the other deployment or return the deployment name.

integration/prefect/listener.py
merobi-hub: Fixed

integration/prefect/listener.py
dolfinus: Job namespace is used to uniquely identify job by namespace+name. I don't think that deployment name provides enough uniqueness for this, as it can be the same in different Prefect instances. This probably should be a Prefect instance URL, like UI url or hostname, and like for Airflow, be the same for flows and tasks within the same Prefect instance

merobi-hub: The transport env var is incomplete (will be array as in Airflow) and there is some unnecessary duplication, but now flow parents are from the UI URL for tasks, flows, and task dependencies

integration/prefect/listener.py
merobi-hub: now flows are filtered and then captured and emitted separately from tasks

integration/prefect/listener.py
dolfinus: That does the "parent transport" mean? There is no such concept in OpenLineage spec. Is it a Prefect term?

integration/prefect/adapter.py
```diff
@@ -0,0 +1,151 @@
...
+            jobType="FLOW"
+        )}
+
+        run_event = RunEvent(
+            eventType=eventType,
+            eventTime=eventTime.isoformat(),
+            run=Run(runId),
+            job=Job(
+                self.job_namespace,
+                self.generate_job_name(parentTransport, flowName), # Was deployment name
```
dolfinus: Why transport is used to generate job name?

merobi-hub: Initially, I was using the deployment ID to parallel the parent-child relation in the flow name that we have in the task job name. Then, in general, I thought the transport would be preferable to the deployment ID, so I used it instead. The nice thing about the deployment ID is that it requires no configuration by the user and it is displayed in the Prefect UI. But I can change it back to the namespace

dolfinus: Do you mean that parent job for the flow is deployment?

dolfinus: My concert is that run.facets.parent.job.name and job.name for emitted parent event should be the same. Otherwise it will be impossible to link flow run with child task runs.

If deployment name is required to be a part of flow name, then it should be included in both places. And it should also be a prefix for a task name, for consistency - overwise task runs produced by different deployments will mix up on OL consumer side.

integration/prefect/listener.py
merobi-hub: The parent concept is Prefect's term for dependencies. I am using `parent_transport` to distinguish the transport coming from the dependency. We could use a different variable name but transport is what we have in Airflow

dolfinus: > the transport coming from the dependency

But how is it used? OpenLineage transport describes where events should be send to, e.g. http, kafka, s3 and so on. I don't see how this can influence job name inside the emitted event - transport does not _identify_ job or run.

Maybe transport in this context means something different than OpenLineage transport?

merobi-hub: Wait, I see your point. Removing transport in favor of namespace

integration/prefect/listener.py
dolfinus: It's still literal string

integration/prefect/listener.py
```diff
@@ -0,0 +1,193 @@
...
+
+	async with get_client() as client:
+		task_run = await client.read_task_run(task_run_id) # TODO: type
+		flow_run_id: UUID = task_run.flow_run_id
+		flow_run = await client.read_flow_run(flow_run_id) # TODO: type
+		deployment = await client.read_deployment(flow_run.deployment_id)
+		try:
+			ns: str = deployment.job_variables["env"]["OPENLINEAGE_NAMESPACE"]
+		except:
+			ns: str = str(deployment.id).split("-")[0]
```
dolfinus: This should fallback to JOB_NAMESPACE, if deployment id is a part of the name. Also function should be named accordingly

integration/prefect/listener.py
```diff
@@ -0,0 +1,193 @@
...
+						event_type: str = "FAILED"
+
+				if entity_type == "flow-run":
+
+					for res in event.related:
+						if res["prefect.resource.role"] == "flow":
+							flow_name: str = res["prefect.resource.name"]
+					if flow_name:
+						start_time: datetime = datetime.fromisoformat(event.resource["prefect.state-timestamp"])
+						flow_namespace: str = JOB_NAMESPACE
```
dolfinus: Namespace should be fetched from an analog of `get_job_vars` function, if deployment can override namespace via variable

integration/prefect/listener.py
```diff
@@ -0,0 +1,193 @@
...
+							flowNamespace=flow_namespace,
+							prefectVersion=prefect_version
+						)
+
+				elif entity_type == "task-run":
+
+					task_name: str = event.resource.name.split("-")[0]
+					start_time: datetime = datetime.fromisoformat(event.resource["prefect.state-timestamp"])
+					prefect_task_run_id: str = event.resource.id.split(".")[-1]
+					ol_task_run_id: str = build_run_id(start_time, task_name, JOB_NAMESPACE)
```
dolfinus: It flow can have it's own namespace from deployment variable, then it should be used here

integration/prefect/listener.py
merobi-hub: There are now functions for getting the ns for tasks and flows, with the JOB_NAMESPACE as fallback

integration/prefect/listener.py
merobi-hub: A new fun for fetching the flow ns does this

integration/prefect/listener.py
merobi-hub: A new function includes this

integration/prefect/adapter.py
merobi-hub: I believe this is resolved now that namespaces are being used in place of deployment IDs

integration/prefect/listener.py
```diff
@@ -0,0 +1,221 @@
...
+	async with get_client() as client:
+		task_run = await client.read_task_run(task_id) # TODO: type
+		return task_run
+
+async def get_flow_ns(flow_run_id: str):
+	"""
+	Looks for OPENLINEAGE_NAMESPACE job env variable in flow's deployment.
+	"""
+
+	async with get_client() as client:
```
dolfinus: Instead of creating a client for every event in the loop, it's probably better to create it once. To avoid passing it everywhere, you can convect these functions into class methods, with a self.client usage under the hood

integration/prefect/listener.py
```diff
@@ -0,0 +1,221 @@
...
+		task_run = await client.read_task_run(task_run_id) # TODO: type
+		flow_run_id: UUID = task_run.flow_run_id
+		flow_run = await client.read_flow_run(flow_run_id) # TODO: type
+		start_time = flow_run.state.timestamp
+		flow_id: UUID = flow_run.flow_id
+		flow = await client.read_flow(flow_id) # TODO: type
+		flow_name = flow.name
+		flow_namespace: str = await get_flow_ns(flow_run_id)
```
dolfinus: Nested function call fetches the same flow and deployment objects fron client which are already available here. Probably this can be reduced

integration/prefect/listener.py
```diff
@@ -0,0 +1,243 @@
...
+	"""Looks for OPENLINEAGE_URL to construct flow name."""
+
+	async with get_client() as client:
+		flow_run = await client.read_flow_run(flow_run_id) # TODO: type
+		deployment = await client.read_deployment(flow_run.deployment_id)
+		try:
+			url: str = deployment.job_variables["env"]["OPENLINEAGE_URL"]
+			host_long: str = url.split("/")[2]
+			host: str = host_long.split(":")[0]
+			return host+"."+flow_name
```
dolfinus: But why transport url is a part of flow name? Changing the target for OL events does not mean that flow is somehow changed

merobi-hub: True, but I was thinking of use cases where having the url available at a glance in a consumer is valuable, for example test environments where you have the same flow names

dolfinus: Just change namespace value to distinct prod Prefect instance from testing

integration/prefect/listener.py
merobi-hub: Agree in principle but will probably have to find the flow name in flow_run because runs have unique names

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+							if event_state == "Running":
+								ol_flow_run_id: str = self.build_run_id(
+									start_time,
+									flow_name,
+									flow_namespace
+								)
+								run_context[flow_run_name] = ol_flow_run_id
+
+							elif event_state in ["Completed", "Failed"]:
+								ol_flow_run_id: str = run_context[flow_run_name]
```
dolfinus: This relies on assumption that events returned by get_events_subscriber() will _always_ contain `Running`, and then `Completed`  `Failed` event.
But what subscriber was started just after the flow running event was emitted? In this case this operation will fail with `KeyError`

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+
+							prefect_version = self.get_prefect_version()
```
dolfinus: This can be moved on the top of the method

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+														start_time, 
+														parent_name, 
+														parent_namespace
+													)
+									parent_runs.append({
+													"name": parent_name, 
+													"namespace": parent_namespace, 
+													"id": parent_run_id
+												})
+						except:
```
dolfinus:
```suggestion
						except KeyError:
```

Otherwise any exception from try block will be deducted as missing parent task, which may not be the case

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+						OL_ADAPTER.create_and_emit_task_event(
+							runId=ol_task_run_id,
+							eventType=event_type, 
+							eventTime=start_time,
+							expectedEventTime=expected_start_time,
+							flowRunId=ol_flow_run_id,
+							flowName=flow_name,
+							flowNamespace=flow_namespace,
+							taskName=task_name,
+							taskNamespace=task_namespace,
```
dolfinus: Is there a case when task_namespace is not the same as flow_namespace? They are fetched from the same deployment/env variable

integration/prefect/adapter.py
```diff
@@ -0,0 +1,162 @@
...
+
+PRODUCER: str = 'https://github.com/OpenLineage/OpenLineage/tree/$VERSION/integration/prefect'
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+class PrefectOpenLineageAdapter:
+    def __init__(
+        self,
+        client: OpenLineageClient | None = None,
+        job_namespace: str | None = None,
```
dolfinus: Unused argument

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+		"""
+		task_run = await self.client.read_task_run(task_run_id) # TODO: type
+		flow_run_id: UUID = task_run.flow_run_id
+		flow_run = await self.client.read_flow_run(flow_run_id) # TODO: type
+		deployment = await self.client.read_deployment(flow_run.deployment_id)
+		try:
+			ns: str = deployment.job_variables["env"]["OPENLINEAGE_NAMESPACE"]
+		except:
+			ns: str = JOB_NAMESPACE
+		return ns
```
dolfinus:
```suggestion
		return await self.get_flow_ns(task_run.flow_run_id)
```

integration/prefect/adapter.py
```diff
@@ -0,0 +1,162 @@
...
+
+        match eventType:
+            case 'START':
+                eventType: RunState = RunState.START
+            case 'COMPLETE':
+                eventType: RunState = RunState.COMPLETE
+            case 'FAILED':
+                eventType: RunState = RunState.FAIL
+        
+        def build_run_facets():
```
dolfinus: Is nested function required? I mean, you can just create `run_facets` local variable, and remove nesting

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+		return flow_run_start_time
+
+	async def collect_and_process_task_runs(self):
+		"""Requires PREFECT_API_URL."""
+		filter_criteria = EventFilter(
+	    	event = EventNameFilter(prefix=["prefect.task-run.", "prefect.flow-run."])
+		)
+		run_context: dict = {}
+
+		async with get_events_subscriber(filter=filter_criteria) as subscriber:
```
dolfinus: `async with self.client:` is missing somewhere in this method

integration/prefect/listener.py
```diff
@@ -0,0 +1,222 @@
...
+from adapter import PrefectOpenLineageAdapter
+from openlineage.client.uuid import generate_static_uuid, generate_new_uuid
+from prefect.events.clients import get_events_subscriber
+from prefect.client.orchestration import get_client
+from prefect.events.filters import EventFilter, EventNameFilter
+from prefect.runtime import task_run, flow_run
+from prefect.utilities.urls import url_for
+
+JOB_NAMESPACE: str = os.environ.get("OPENLINEAGE_NAMESPACE", "prefect_test")
+OL_ADAPTER = PrefectOpenLineageAdapter()
```
dolfinus: I recommend passing adapter via constructor args, instead of using a global variable

integration/prefect/listener.py
merobi-hub: added try/except block to catch this

integration/prefect/adapter.py
```diff
@@ -0,0 +1,159 @@
...
+                            ),
+                            "parentRun": ParentRunFacet(
+                                run={"runId": flowRunId},
+                                job={"namespace": namespace, "name": flowName}
+                            ),
+                            "processingEngine": processing_engine_run.ProcessingEngineRunFacet(
+                                version=prefectVersion,
+                                name="Prefect"
+                            )
+                        }
```
dolfinus:
```suggestion
        run_facets = {
            "nominalTime": NominalTimeRunFacet(
                nominalStartTime=expectedEventTime
            ),
            "parentRun": ParentRunFacet(
                run={"runId": flowRunId},
                job={"namespace": namespace, "name": flowName}
            ),
            "processingEngine": processing_engine_run.ProcessingEngineRunFacet(
                version=prefectVersion,
                name="Prefect"
            )
        }
        if jobDeps:
            upstream_jobs = [
                job_dependencies_run.JobDependency(
                    job=job_dependencies_run.JobIdentifier(
                        namespace=dep["namespace"],
                        name=dep["name"]
                    )
                ) for dep in jobDeps
            ]
            run_facets["jobDependencies"] = (
                job_dependencies_run.JobDependenciesRunFacet(upstream=upstream_jobs)
            )
```

integration/prefect/listener.py
```diff
@@ -0,0 +1,201 @@
...
+						except KeyError:
+							logger.info("No task parents found for %s", prefect_task_run_id)
+							pass
+
+						# Get flow run info for ParentRunFacet
+						for res in event.related:
+							if res["prefect.resource.role"] == "flow-run":
+								flow_run_name: str = res["prefect.resource.name"]
+						flow_name: str = await self.get_flow_run_name(prefect_task_run_id) #TODO: use flow_run_id
+						ol_flow_run_id: str = run_context[flow_run_name]
```
dolfinus: What if listener started between flow run start and task run start?

integration/prefect/listener.py
```diff
@@ -0,0 +1,201 @@
...
+JOB_NAMESPACE: str = os.environ.get("OPENLINEAGE_NAMESPACE", "prefect_test")
+
+logger: logging.Logger = logging.getLogger(__name__)
+
+class PrefectOpenLineageListener:
+	def __init__(
+		self
+	):
+		self.client = get_client()
+		self.ol_adapter = PrefectOpenLineageAdapter()
```
dolfinus:
```suggestion
	def __init__(
		self,
		client = None,
		adapter = None,
	):
		self.client = client or get_client()
		self.ol_adapter = adapter or PrefectOpenLineageAdapter()
```

integration/prefect/listener.py
```diff
@@ -0,0 +1,201 @@
...
+
+					if entity_type == "flow-run":
+
+						for res in event.related:
+							if res["prefect.resource.role"] == "flow":
+								flow_run_name: str = event.resource.name
+								flow_name: str = res["prefect.resource.name"]
+
+						if flow_run_name:
+							start_time: datetime = datetime.fromisoformat(event.resource["prefect.state-timestamp"])
```
dolfinus: Why not using flow_run.start_time or flow_run.created instead? This field can be fetched from client, and there will be no error if flow run event is missing.

merobi-hub:
Please see updated description for information about how Artifacts are being used to create Datasets. Feedback is welcome.

integration/prefect/listener.py
```diff
@@ -0,0 +1,269 @@
...
+		return {"name": flow_name, "deployment_info": deployment_info}
+
+	async def get_flow_run_start_time(self, flow_run_id: str):
+		flow_run = await self.client.read_flow_run(flow_run_id) # TODO: type
+		flow_run_start_time = flow_run.state.timestamp
+		return flow_run_start_time
+
+	async def get_artifacts_by_task_run(self, run_id: str) -> list:
+		payload = {
+			"limit": 5,
```
dolfinus: Only 5 artifacts max?

integration/prefect/listener.py
```diff
@@ -0,0 +1,269 @@
...
+			"updated": deployment.updated.isoformat(), 
+			"name": deployment.name
+		}
+
+	def get_prefect_version(self) -> str:
+		try:
+			url = os.environ.get("PREFECT_API_URL")+"/admin/version"
+		except TypeError:
+			logger.info("PREFECT_API_URL not found.")
+		version = requests.get(url).json()
```
dolfinus: Why not using `await self.client._client.get()` here? `requests` is blocking

integration/prefect/listener.py
```diff
@@ -0,0 +1,269 @@
...
+
+						if entity_type == "flow-run":
+
+							for res in event.related:
+								if res["prefect.resource.role"] == "flow":
+									flow_run_name: str = event.resource.name
+									flow_name: str = res["prefect.resource.name"]
+
+							if flow_run_name:
+								start_time: datetime = datetime.fromisoformat(event.resource["prefect.state-timestamp"])
```
dolfinus: This is not a start time, it's an event time (when flow started or completed). It should not be used for generating run_id

integration/prefect/adapter.py
```diff
@@ -0,0 +1,194 @@
...
+                eventTime=eventTime.isoformat(),
+                run=Run(runId, run_facets),
+                job=Job(
+                    namespace,
+                    taskName,
+                    job_facets
+                ),
+                producer=PRODUCER,
+                inputs=inputs,
+                outputs=outputs
```
dolfinus: Why excluding inputs and outputs for other event types?

integration/prefect/listener.py
```diff
@@ -0,0 +1,269 @@
...
+					data_list = ast.literal_eval(artifact["data"])
+					uri = data_list[0]["database_uri"]
+					table = data_list[0]["table"]
+					dataset_info.append([uri, table, dataset_type])
+			return dataset_info
+		else:
+			logging.info("No datasets found for task run.")
+			return []
+
+	async def collect_and_process_task_runs(self):
```
dolfinus: IMHO let's split this method into 2 parts - one for handling flow events, another fo handling task events. And probably some parts of current method (e.g. extracting flow & deployment info from `event.related`) should be extracted into dedicated method as well

integration/prefect/listener.py
```diff
@@ -0,0 +1,269 @@
...
+		response = await self.client._client.post("/artifacts/filter", json=payload)
+		if response.status_code == 200:
+			dataset_info = []
+			artifacts = response.json()
+			for artifact in artifacts:
+				if "ol-dataset" in artifact["description"]:
+					dataset_type = artifact["description"].split("_")[-1]
+					data_list = ast.literal_eval(artifact["data"])
+					uri = data_list[0]["database_uri"]
+					table = data_list[0]["table"]
```
dolfinus: If there can be only one item in the list, why not changing to dict?
```python
ol_table = {"database_uri":"duckdb:///customers_db", "table":"customers"}

create_table_artifact(
    key="upstream-insert",
    table=ol_table,
    description="ol-dataset_output"
)     
```

integration/prefect/listener.py
merobi-hub: Agreed, but unfortunately flow run events do not include start times

dolfinus: But it can be fetched via client?

integration/prefect/adapter.py
merobi-hub: This is another reason to update the usage example: https://openlineage.io/docs/client/python/usage

integration/prefect/listener.py
merobi-hub: A table argument must be either a list of dictionaries, list of lists, or dictionary of lists

dolfinus: I see, thanks for clarifying

integration/prefect/listener.py
merobi-hub: The limit was in the Prefect example code, but agree it's not needed

integration/prefect/listener.py
merobi-hub: Done

integration/prefect/listener.py
merobi-hub: Done

integration/prefect/listener.py
merobi-hub: Agreed, done

integration/prefect/listener.py
```diff
@@ -0,0 +1,277 @@
...
+			deploymentName=deployment_name,
+			inputDatasets=input_datasets,
+			outputDatasets=output_datasets
+		)
+
+	async def collect_and_process_runs(self) -> None:
+		try:
+			os.environ.get("PREFECT_API_URL")
+		except TypeError:
+			logger.warn("PREFECT_API_URL not set. Prefect events will not be emitted.")
```
dolfinus:
```suggestion
			logger.warn("PREFECT_API_URL not set. Prefect events will not be emitted.")
			return
```

Early return allows to drop `except TypeError:` from `get_prefect_version` method

integration/prefect/listener.py
```diff
@@ -0,0 +1,277 @@
...
+		return {
+			"id": str(deployment.id), 
+			"created": deployment.created.isoformat(), 
+			"updated": deployment.updated.isoformat(), 
+			"name": deployment.name
+		}
+
+	async def get_prefect_version(self) -> str | None:
+		try:
+			async with get_client()._client as c:
```
dolfinus: Instead of creating new client here, `self.client` could be used. The only change needed is to move `prefect_version: str = await self.get_prefect_version()` under `async with self.client` in `collect_and_process_runs`

integration/prefect/listener.py
```diff
@@ -0,0 +1,277 @@
...
+				flow_and_deployment_info: str = await self.get_flow_and_deployment_info(flow_run_id)
+				flow_name: str = flow_and_deployment_info["name"]
+				deployment_info: dict = flow_and_deployment_info["deployment_info"]
+				deployment_id: str = deployment_info["id"]
+				deployment_created: str = deployment_info["created"]
+				deployment_updated: str = deployment_info["updated"]
+				deployment_name: str = deployment_info["name"]
+				flow_run = await self.client.read_flow_run(flow_run_id)
+				ol_flow_run_id: str = self.build_run_id(
+					flow_run.start_time,
```
dolfinus:
```suggestion
				flow_start_time = await self.get_flow_run_start_time(flow_run_id)
				ol_flow_run_id: str = self.build_run_id(
					flow_start_time,
```
nit: just for consistency

integration/prefect/listener.py
```diff
@@ -0,0 +1,277 @@
...
+
+	async def get_artifacts_by_task_run(self, run_id: str) -> list:
+		payload = {
+			"artifacts": {
+				"task_run_id": {
+					"any_": [run_id]
+				}
+			}
+		}
+		async with get_client()._client as c:
```
dolfinus: Same recommendation for using `self.client` instead of creating new client each time

integration/prefect/listener.py
```diff
@@ -0,0 +1,277 @@
...
+
+	async def get_flow_ns(self, flow_run_id: str) -> str:
+		"""
+		Looks for OPENLINEAGE_NAMESPACE job env variable in deployment.
+		"""
+		flow_run = await self.client.read_flow_run(flow_run_id) # TODO: type
+		deployment = await self.client.read_deployment(flow_run.deployment_id)
+		try:
+			ns: str = deployment.job_variables["env"]["OPENLINEAGE_NAMESPACE"]
+		except:
```
dolfinus:
```suggestion
		except KeyError:
```

integration/prefect/listener.py
```diff
@@ -0,0 +1,277 @@
...
+		return ns
+
+	async def get_job_ns(self, task_run_id: str) -> str:
+		"""
+		Looks for OPENLINEAGE_NAMESPACE job env variable in parent deployment.
+		"""
+		task_run = await self.client.read_task_run(task_run_id) # TODO: type
+		return await self.get_flow_ns(task_run.flow_run_id)
+
+	async def get_flow_and_deployment_info(self, flow_run_id: str) -> dict:
```
dolfinus: `get_flow_and_deployment_info`, `get_deployment_info` and `get_flow_run_start_time` call `self.client.read_flow_run` 3 times in a row, for the same `flow_run_id`. Maybe these methods can be merged into one `get_flow_and_deployment(flow_run_id)` method returning a tuple `flow_run, deployment_info`, with just one client call?

integration/prefect/listener.py
```diff
@@ -0,0 +1,286 @@
...
+		for res in event.related:
+			if res["prefect.resource.role"] == "flow-run":
+				flow_run_id: str = res["prefect.resource.id"].split(".")[-1]
+				deployment_and_flow_info: tuple = await self.get_deployment_and_flow_info(flow_run_id)
+				deployment_id: str = deployment_and_flow_info[0]
+				deployment_created: str = deployment_and_flow_info[2]
+				deployment_updated: str = deployment_and_flow_info[3]
+				deployment_name: str = deployment_and_flow_info[4]
+				flow_name: str = deployment_and_flow_info[6]
+				flow_start_time = await self.get_flow_run_start_time(flow_run_id)
```
dolfinus:
```suggestion
				flow_start_time = deployment_and_flow_info[1]
```

integration/prefect/listener.py
```diff
@@ -0,0 +1,286 @@
...
+		execution_time: datetime, 
+		run_name: str, 
+		namespace: str
+	) -> str:
+		return str(generate_static_uuid(
+			instant=execution_time,
+	        data=f"{namespace}.{run_name}".encode("utf-8"),
+		))
+
+	async def get_deployment_and_flow_info(self, flow_run_id: str) -> dict:
```
dolfinus: Not a dict

merobi-hub:
@kacpermuda @harels I'm thinking this might be ready for the move

merobi-hub:
Opened a PR in Prefect: https://github.com/PrefectHQ/prefect/pull/22475

kacpermuda:
> @kacpermuda @harels I'm thinking this might be ready for the move

Awesome, thanks ! Let's close this one then and focus on merging the one in the Prefect repo

merobi-hub:
@OpenLineage/committers reopening this PR with tests as well as new changes from the review process in Prefect. For context, see the Prefect PR linked to above.

dolfinus:
If this is now a part of Prefect repo, why opening PR in OpenLineage?

merobi-hub:
I reopened it here before seeing their most recent comment there. Now that there is a path as an experimental integration in Prefect, I'll close this one again.

merobi-hub:
See comments here and in the Prefect PR for context

merobi-hub:
Reopening here after closing the PR in Prefect. I learned that they prefer we host it in our own org and repo initially.

integration/prefect/src/openlineage_prefect/prefect_adapter/adapter.py
```diff
@@ -0,0 +1,190 @@
...
+        flow_run_id: str | None = None,
+        flow_name: str | None = None,
+        task_name: str | None  = None,
+        namespace: str | None  = None,
+        job_deps: list | None = None,
+        prefect_version: str | None = None,
+        deployment_id: str | None = None,
+        deployment_created: str | None = None,
+        deployment_updated: str | None = None,
+        deployment_name: str | None = None,
```
dolfinus, line 96: nit: maybe deployment information should be passed as one object instead of set of related arguments?
```python
@dataclass
class PrefectDeployment:
  id: str
  created_at: datetime
  updated_at: datetime
  name: str

create_and_emit_task_event(run_id, event_type, event_time, deployment=...)
```

integration/prefect/src/openlineage_prefect/prefect_adapter/adapter.py
```diff
@@ -0,0 +1,190 @@
...
+        flow_namespace: str,
+        prefect_version: str | None = None,
+        deployment_id: str | None = None,
+        deployment_created: str | None = None,
+        deployment_updated: str | None = None,
+        deployment_name: str | None = None,
+    ) -> RunEvent:
+        """Create and emit a flow-level OpenLineage event."""
+
+        match event_type:
```
dolfinus: event_type argument could be a RunState in the first place

integration/prefect/src/openlineage_prefect/prefect_adapter/listener.py
```diff
@@ -0,0 +1,385 @@
...
+                    flow_run_id = res["prefect.resource.id"].split(".")[-1]
+                    deployment_and_flow_info = await self.get_deployment_and_flow_info(
+                        flow_run_id
+                    )
+                    deployment_id = deployment_and_flow_info[0]
+                    deployment_created = deployment_and_flow_info[2]
+                    deployment_updated = deployment_and_flow_info[3]
+                    deployment_name = deployment_and_flow_info[4]
+                    flow_name = deployment_and_flow_info[6]
+                    flow_start_time = deployment_and_flow_info[1]
```
dolfinus, line 333: nit: I'd rather create a class for this purpose instead of using a tuple:
```python
@dataclass
class Deployment: ...

@dataclass
class Flow: ...

return deployment, flow
```

integration/prefect/src/openlineage_prefect/prefect_adapter/listener.py
```diff
@@ -296,12 +302,25 @@ async def collect_and_process_task_runs(
...
+            deployment_id = ''
+            deployment_created = ''
+            deployment_updated = ''
+            deployment_name = ''
             for res in event.related:
                 if res["prefect.resource.role"] == "flow-run":
-                    flow_run_id = res["prefect.resource.id"].split(".")[-1]
+                    try:
+                        flow_run_id = res["prefect.resource.id"].split(".")[-1]
+                    except:
```
dolfinus: Please don' use bare `except:`

PR #4522, opened 2026-05-07
dbt: propagate root parent through dbt-run wrapper to child events

mobuchowski:
The legacy `consume_local_artifacts` path constructs `dbt_run_metadata` without `root_parent_*` fields, so every per-node event uses `self._dbt_run_metadata.to_openlineage()` and ends up with `root=None` in its parent facet. 

### Checklist

### DIFF
--- a/integration/dbt/src/openlineage/dbt/__init__.py
+++ b/integration/dbt/src/openlineage/dbt/__init__.py
@@ -301,10 +301,26 @@ def consume_local_artifacts(
         parent_run_metadata=parent_run_metadata,
     )
 
+    # Carry the root parent through to per-node events so backends can stitch the full
+    # orchestrator → dbt-run → node chain. Without this, child events get root=None.
+    if parent_run_metadata:
+        root_parent_run_id = parent_run_metadata.root_parent_run_id or parent_run_metadata.run_id
+        root_parent_job_name = parent_run_metadata.root_parent_job_name or parent_run_metadata.job_name
+        root_parent_job_namespace = (
+            parent_run_metadata.root_parent_job_namespace or parent_run_metadata.job_namespace
+        )
+    else:
+        root_parent_run_id = start_event.run.runId
+        root_parent_job_name = start_event.job.name
+        root_parent_job_namespace = start_event.job.namespace
+
     dbt_run_metadata = ParentRunMetadata(
         run_id=start_event.run.runId,
         job_name=start_event.job.name,
         job_namespace=start_event.job.namespace,
+        root_parent_run_id=root_parent_run_id,
+        root_parent_job_name=root_parent_job_name,
+        root_parent_job_namespace=root_parent_job_namespace,
     )
     # Set parent run metadata to use it as parent run facet
     processor.dbt_run_metadata = dbt_run_metadata
--- a/integration/dbt/src/openlineage/dbt/__init__.py
+++ b/integration/dbt/src/openlineage/dbt/__init__.py
@@ -303,12 +303,22 @@ def consume_local_artifacts(
 
     # Carry the root parent through to per-node events so backends can stitch the full
     # orchestrator → dbt-run → node chain. Without this, child events get root=None.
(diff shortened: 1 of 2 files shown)


integration/dbt/src/openlineage/dbt/__init__.py
```diff
@@ -301,10 +301,26 @@ def consume_local_artifacts(
...
     )
 
+    # Carry the root parent through to per-node events so backends can stitch the full
+    # orchestrator → dbt-run → node chain. Without this, child events get root=None.
+    if parent_run_metadata:
+        root_parent_run_id = parent_run_metadata.root_parent_run_id or parent_run_metadata.run_id
+        root_parent_job_name = parent_run_metadata.root_parent_job_name or parent_run_metadata.job_name
+        root_parent_job_namespace = (
+            parent_run_metadata.root_parent_job_namespace or parent_run_metadata.job_namespace
+        )
```
kacpermuda, line 313: We should probably double-check if all root-related metadata is present and only set root to root if all are there, otherwise use all parents. Current way can mix the root and parent, if parent_run_metadata is created incorrectly .

integration/dbt/tests/test_main.py
```diff
@@ -132,3 +132,82 @@ def test_consume_structured_logs(command_line, os_envs, function_kwargs, monkeyp
...
+        profile_name=None,
+        model_selector=None,
+        models=[],
+    )
+
+    md = processor.dbt_run_metadata
+    # Root falls back to the wrapper's own start event.
+    assert md.root_parent_run_id == md.run_id
+    assert md.root_parent_job_name == md.job_name
+    assert md.root_parent_job_namespace == md.job_namespace
```
kacpermuda, line 213: Wait, the root points to itself? Not saying it's wrong, just unexpected. I think we don't do that in other integrations, there is just no parent/root metadata, instead of setting parent and root to itself.

integration/dbt/tests/test_main.py
```diff
@@ -132,3 +132,82 @@ def test_consume_structured_logs(command_line, os_envs, function_kwargs, monkeyp
...
+    process_mock = mock.MagicMock()
+    process_mock.__enter__ = mock.MagicMock(return_value=process_mock)
+    process_mock.__exit__ = mock.MagicMock(return_value=None)
+    process_mock.wait.return_value = 0
+    monkeypatch.setattr("openlineage.dbt.subprocess.Popen", mock.MagicMock(return_value=process_mock))
+
+    return mock_processor
+
+
+def test_consume_local_artifacts_propagates_root_parent_metadata(monkeypatch):
```
kacpermuda, line 159: Can we add a tests for root actually provided explicitly (so that we have both root and parent with different values)? And another test, that if not all identifiers (either root or parent) are provided, we do not put them in the result? We should probably do a check if all are present.

integration/dbt/tests/test_main.py
mobuchowski: That's the original intended semantics: https://docs.google.com/document/d/1JypxrJc0MkrarzYfMopK-TF8QiPpjPnk2je4mgq9e4k/edit?tab=t.0

In most cases it does not matter, because the fallback - when root is not set, but parent is, child job should use the parent as root as well anyway.

kacpermuda: Ok, missed that for sure. Ok to resolve this one then

kacpermuda (approved):

integration/dbt/src/openlineage/dbt/__init__.py
mobuchowski: Fixed.

integration/dbt/tests/test_main.py
mobuchowski: Added test_consume_local_artifacts_explicit_root_different_from_parent and test_consume_local_artifacts_partial_root_falls_back_to_parent

PR #4535, opened 2026-05-08
spark: fix Kafka SASL bootstrap namespace

### LINKED ISSUE #4534: Spark Kafka namespace parsing fails for SASL_SSL bootstrap servers
### Problem

The Spark integration derives Kafka dataset namespaces from the configured bootstrap servers. The resolver currently parses the first bootstrap server with java.net.URI after optionally prepending PLAINTEXT://.

That breaks for common secure Kafka values such as SASL_SSL://broker.example.com:9093. SASL_SSL is a valid Kafka security protocol name, but it is not a valid URI scheme because URI schemes cannot contain underscores. As a result, secure Kafka jobs can produce an invalid namespace such as kafka://null:-1 instead of the broker address.

### Expected behavior

SASL_SSL://broker.example.com:9093 should resolve to kafka://broker.example.com:9093, matching the behavior for plaintext and SSL bootstrap servers.

### Impact

Spark streaming and batch jobs using secure Kafka clusters can emit incorrect dataset namespaces, which makes lineage hard to query and can split or hide Kafka lineage in downstream consumers.

### Suggested fix

Normalize Kafka protocol prefixes before URI parsing, or parse the first bootstrap endpoint in a way that supports Kafka protocol names containing underscores.

hcthakur2004:
### Problem

Spark Kafka lineage derives the dataset namespace from the first configured bootstrap server. Secure Kafka values such as SASL_SSL://broker.example.com:9093 include a Kafka protocol name with an underscore, which java.net.URI does not accept as a URI scheme. That can produce an invalid namespace like kafka://null:-1.

### Changes

- Strip Kafka protocol prefixes before URI parsing.
- Parse the broker endpoint with a stable kafka:// URI prefix.
- Add regression coverage for plaintext, SASL_SSL, and comma-separated bootstrap server values.

Fixes #4534

### Tests

- git diff --check
- Not run: ./gradlew.bat :shared:test --tests io.openlineage.spark.agent.lifecycle.plan.KafkaBootstrapServerResolverTest because this environment does not have JAVA_HOME or java on PATH.

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KafkaBootstrapServerResolver.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KafkaBootstrapServerResolver.java
@@ -7,22 +7,27 @@ package io.openlineage.spark.agent.lifecycle.plan;
 
 import java.net.URI;
 import java.util.Optional;
+import java.util.regex.Pattern;
 
 public class KafkaBootstrapServerResolver {
+  private static final Pattern KAFKA_PROTOCOL_PREFIX =
+      Pattern.compile("^[A-Za-z][A-Za-z0-9_+.-]*://");
+
   static String resolve(Optional<String> bootstrapServersOpt) {
     String server =
         bootstrapServersOpt
-            .map(
-                str -> {
-                  if (!str.matches("\\w+://.*")) {
-                    return "PLAINTEXT://" + str;
-                  } else {
-                    return str;
-                  }
-                })
-            .map(str -> URI.create(str.split(",")[0]))
+            .map(str -> str.split(",")[0].trim())
+            .map(KafkaBootstrapServerResolver::stripKafkaProtocol)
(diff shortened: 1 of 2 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/KafkaBootstrapServerResolver.java
```diff
@@ -7,22 +7,27 @@
...
+            .map(str -> str.split(",")[0].trim())
+            .map(KafkaBootstrapServerResolver::stripKafkaProtocol)
+            .map(endpoint -> URI.create("kafka://" + endpoint))
             .map(uri -> uri.getHost() + ":" + uri.getPort())
             .orElse("");
     return "kafka://" + server;
   }
+
+  private static String stripKafkaProtocol(String endpoint) {
+    if (KAFKA_PROTOCOL_PREFIX.matcher(endpoint).find()) {
```
dolfinus, line 28: You can search for protocol string just by `://` substring, without regexp

PR #4551, opened 2026-05-18
spark: improve handling raw path files and Hive partitioning

mobuchowski:
AI description:

This PR improves dataset naming for Spark file-based lineage when jobs read or write partitioned raw files. Spark users often work with Parquet/CSV paths that point at a selected partition or file pattern, such as `s3://bucket/table/dt=20260516/*.parquet`, even though the logical dataset is the parent table directory. Previously, OpenLineage could emit the partition path or literal glob as the dataset name, causing lineage graphs to fragment across timestamped or partition-specific paths.

The change normalizes file paths before creating dataset identifiers. Literal glob selectors are resolved to the nearest stable directory, and Hive-style partition segments such as `dt=20260516/hour=13` are stripped by default so the emitted dataset is the table/root path. This behavior is configurable with `spark.openlineage.normalizeHiveStylePartitioning`, which defaults to `true` and can be set to `false` to preserve the previous partition-path behavior.

This normalization is applied across SQL file relations, Hadoop RDD inputs, and `InsertIntoHadoopFsRelationCommand` outputs. The implementation also uses Spark’s explicit `basePath` metadata for partitioned `HadoopFsRelation`s, so reads of selected partitions with a declared dataset root emit the stable root dataset.

### Checklist
- [x] AI was used in creating this PR - Pi/gpt-5.5

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
@@ -8,7 +8,12 @@ package io.openlineage.spark.agent.util;
 import io.openlineage.client.utils.DatasetIdentifier;
 import io.openlineage.client.utils.filesystem.FilesystemDatasetUtils;
 import java.io.File;
+import java.io.IOException;
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Optional;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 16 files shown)


dolfinus:
Does this mean that dataset trimmers implemented in #3996 are not working properly?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -36,6 +41,81 @@ public static DatasetIdentifier fromURI(URI location) {
...
+      }
+      current = parent;
+    }
+    return current == null ? path : current;
+  }
+
+  private static boolean isHivePartitionPath(String pathName) {
+    int firstEquals = pathName.indexOf('=');
+    return firstEquals > 0
+        && firstEquals == pathName.lastIndexOf('=')
```
dolfinus, line 158: This means that paths like `s3://some/table/col1=1/col2=2` are not trimmed to `s3://some/table` because = first and last indexes are bot the same

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -36,6 +41,81 @@ public static DatasetIdentifier fromURI(URI location) {
     return FilesystemDatasetUtils.fromLocation(location);
   }
 
+  public static List<Path> getDirectoryPaths(
+      Collection<Path> paths, Configuration hadoopConf, boolean normalizeHiveStylePartitioning) {
+    LinkedHashSet<Path> normalizedPaths = new LinkedHashSet<>();
+    for (Path path : paths) {
```
pawel-big-lebowski: This is sth that used to kill Spark jobs in the past. Getting 100K of files and running get file status on each of them which may be remote call on S3. The approach we took in other places: if there are many paths and they have same parent, get the parent path only.

pawel-big-lebowski (commented):
@mobuchowski The feature seems similar to dataset trimmers that are already implemented: https://openlineage.io/docs/client/java/configuration/#out-of-the-box-trimmers. Could you elaborate on what's missing there and why aren't you following the dataset trimmer approach?

PR #4556, opened 2026-05-21
spark: map sqlserver to MsSqlDialect

### LINKED ISSUE #4550: [BUG][SPARK] JDBC dialect "sqlserver" not mapped to MsSqlDialect, causing SQL parsing errors for T-SQL bracket identifiers

Spark integration

### OpenLineage version

io.openlineage:openlineage-spark_2.12:1.46.0

Spark 3.4.4, Scala 2.12.15, Java 11, Microsoft SQL Server JDBC driver (jdbc:sqlserver://)

### Environment configuration

OpenLineage Spark agent enabled with JDBC reads from SQL Server

### Deployment details

Spark on Kubernetes (non-Databricks)

### Problem details

When reading from SQL Server via Spark JDBC (JDBCRelation), JdbcSparkUtils.extractQueryFromSpark() logs ERROR-level messages like:
```
error while parsing query: ExtractionError { index: 0, message: "Expected: an identifier after AS, found: [", origin_statement: "..." }
```

This happens because the SQL queries use T-SQL square bracket identifier syntax (e.g., `CAST([Column] AS VARCHAR) AS [Alias]`), which is standard for SQL Server but not supported by the `GenericDialect` parser.

Root cause: `JdbcSparkUtils.extractDialectFromJdbcUrl()` extracts the dialect from the JDBC URL using the regex `^jdbc:([^:]+):.*`. For SQL Server JDBC URLs (`jdbc:sqlserver://...`), this produces `"sqlserver"`. However, `get_dialect()` in integration/sql/impl/src/dialect.rs only maps `"mssql"` to `MsSqlDialect`

`MsSqlDialect` should be used instead. It correctly handles [bracket] identifiers (via `is_delimited_identifier_start('[')`), but since `"sqlserver"` is not recognized, `GenericDialect` is used instead, which does not support this syntax.

Suggested fix:
```
match name {
    // ...
    "mssql" => &MsSqlDialect {},
    "sqlserver" => &MsSqlDialect {},  // JDBC URL uses jdbc:sqlserver://
    // ...
}
```

### How to reproduce

1. Configure a Spark job to read from SQL Server via JDBC with a query using bracket identifiers:
```
spark.read.jdbc("jdbc:sqlserver://host:1433;databaseName=db", "(SELECT [Col1] AS [Alias1] FROM [dbo].[Table1]) AS tmp", props)
```

2. Enable the OpenLineage Spark agent
3. Observe ERROR-level log lines from JdbcSparkUtils:
```
error while parsing query: ExtractionError { index: 0, message: "Expected: an identifier after AS, found: [", origin_statement: "..." }
```

### Additional details

_No response_

dbathie-wtg:
closes: #4550

Map Spark JDBC `sqlserver` URLs to `MsSqlDialect` so SQL Server queries with bracketed identifiers parse correctly during lineage extraction.

### Meaningful description
Spark extracts the SQL dialect name from JDBC URLs, so a SQL Server connection like `jdbc:sqlserver://...` is passed through as `sqlserver`.

The SQL integration only recognized `mssql`, which meant SQL Server queries using T-SQL bracketed identifiers such as `[col]` and `[schema].[table]` fell back to `GenericDialect` and produced parsing errors during lineage extraction.

This change adds `sqlserver` as an alias for `MsSqlDialect` in the SQL dialect mapping and adds a regression test in `integration/sql/impl/tests/table_lineage/tests_select.rs` covering a bracketed SQL Server `SELECT`.

Validated with `cargo test` and `cargo fmt --check` in `integration/sql/impl`.

### Checklist

### DIFF
--- a/integration/sql/impl/src/dialect.rs
+++ b/integration/sql/impl/src/dialect.rs
@@ -41,6 +41,7 @@ pub fn get_dialect(name: &str) -> &'static dyn CanonicalDialect {
         "hive" => &HiveDialect {},
         "mysql" => &MySqlDialect {},
         "mssql" => &MsSqlDialect {},
+        "sqlserver" => &MsSqlDialect {},
         "sqlite" => &SQLiteDialect {},
         "ansi" => &AnsiDialect {},
         "generic" => &GenericDialect,
@@ -53,5 +54,19 @@ pub fn get_generic_dialect(name: Option<String>) -> &'static dyn CanonicalDialec
         get_dialect(d.as_str())
     } else {
         &GenericDialect
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::get_dialect;
+    use sqlparser::dialect::MsSqlDialect;
+
+    #[test]
+    fn sqlserver_alias_uses_mssql_dialect() {
+        let dialect = get_dialect("sqlserver");
+
+        assert!(dialect.as_base().is::<MsSqlDialect>());
+        assert_eq!(dialect.canonical_name("[Alias]"), Some("Alias"));
     }
 }
--- a/integration/sql/impl/src/dialect.rs
+++ b/integration/sql/impl/src/dialect.rs
@@ -56,17 +56,3 @@ pub fn get_generic_dialect(name: Option<String>) -> &'static dyn CanonicalDialec
         &GenericDialect
     }
 }
-
-#[cfg(test)]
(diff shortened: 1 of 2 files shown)


mobuchowski:
Thanks for the PR, the hive issue seems unrelated.

mobuchowski (approved):

PR #4558, opened 2026-05-21
spark: add iceberg s3 tables catalog support for v2 path

mobuchowski:
Currently, when using S3 tables, in some cases Spark integration wrongly attaches symlinks to Glue tables; especially for cases where Glue libs are present and metastore is configured to Glue catalog. It happens since we're falling back to Glue broadly, basing on heuristics like metastore config. 

This solves this issue by providing explicit support for S3 Tables.

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/S3TablesUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/S3TablesUtils.java
@@ -1,0 +1,216 @@
+/*
+/* Copyright 2018-2026 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.spark.agent.util;
+
+import java.net.URI;
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import lombok.Value;
+import lombok.experimental.UtilityClass;
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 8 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/S3TablesUtils.java
```diff
@@ -0,0 +1,216 @@
...
+
+    Optional<String> resolvedAccount = info.flatMap(b -> b.account);
+    if (!resolvedAccount.isPresent()) {
+      resolvedAccount = resolveAccountId(sparkConf, hadoopConf);
+    }
+    if (!resolvedAccount.isPresent()) {
+      log.warn("Unable to resolve AWS account ID for S3 Tables namespace; using 'unknown'.");
+    }
+
+    String prefix =
```
pawel-big-lebowski, line 155: When region or account can't be resolved, you emit arn:aws:s3tables:unknown:unknown:bucket/<name>. That's a non-parseable ARN that consumers downstream may key on. Consider either:
  - returning Optional.empty() and letting the caller fall back to a path-based identity, or
  - omitting only the bucket segment when account is resolvable but region isn't (and vice versa).

pawel-big-lebowski (approved):
Nice fix! Would love to see an e2e test with a real Glue job hitting a federated S3 Tables catalog — that's the exact path the unit tests can't really prove. Totally get that it's a chunk of work to set up though, happy to leave it as a follow-up.

Issue #773, opened 2022-05-24
Include BigQueryInsertJobOperator for extracting DataLineage


manmeetkaur:
For extracting the lineage information, BigQuery Extractors currently support `BigQueryOperator` and `BigQueryExecuteQueryOperator`. But both of these are deprecated as seen [here](https://github.com/apache/airflow/blob/main/airflow/contrib/operators/bigquery_operator.py#L42-L45) and [here](https://github.com/apache/airflow/blob/main/airflow/providers/google/cloud/operators/bigquery.py#L455-L461). 

I did some testing to include the latest BigQueryInsertJobOperator by including this operator [here](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py#L42-L43):
```
    @classmethod
    def get_operator_classnames(cls) -> List[str]:
        return ['BigQueryOperator', 'BigQueryExecuteQueryOperator' , 'BigQueryInsertJobOperator']
```

But this causes the job to fail [here](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py#L125) because it doesn't have the sql attribute. 
```
    def parse_sql_context(self) -> SqlContext:
        try:
            sql_meta = parse(self.operator.sql, dialect='bigquery')
```
So to get the sql, we need to parse the `configuration` attribute, like this:

```
            if self.operator.__class__.__name__ == 'BigQueryInsertJobOperator':
                print(self.operator.configuration['query']['query'])
                self.operator.sql = self.operator.configuration['query']['query']

```
The above code picks up the `SQL` correctly, but then fails [here](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py#L116-L121) because the XCom value is passed with key `return_value` and not `job_id`. This is fixed as below:
```
        if not bigquery_job_id:
            bigquery_job_id = task_instance.xcom_pull(task_ids=task_instance.task_id, key=None)
```
After this is fixed, it then fails [here](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py#L105) because useLegacySql is again in configuration and not as a direct parameter. So this can be changed to:

```
            if self.operator.__class__.__name__ == 'BigQueryInsertJobOperator':
[...]

PR #4563, opened 2026-05-23
spec: add GcpBigQueryRoutineJobFacet

### LINKED ISSUE #4562: [Proposal] GcpBigQueryRoutineJobFacet — identify invoked BigQuery routine
## Purpose (Why?)

OpenLineage events emitted from Airflow tasks that invoke BigQuery routines (stored procedures, scalar/table/aggregate UDFs) currently lose the identity of the called routine.

- **`BigQueryInsertJobOperator` with `CALL <proc>(...)`** — the existing OL extractor walks the parent SCRIPT job in BigQuery and aggregates child-job inputs and outputs, but the routine reference itself is dropped. Tracking issue: #773.
- **Custom operators** (e.g., a `PythonOperator` wrapping `bigquery.Client().query("CALL ...")`) — no routine information appears in the OL event at all.

Consumers that catalog BigQuery routines as first-class assets (data governance / lineage platforms) cannot build an `AirflowTask → BigQueryRoutine → Table` relationship from current events.

`apache-airflow-providers-google`'s `BigQueryJobRunFacet` only carries `{cached, billedBytes, properties}` — no routine identity — and it lives in the Airflow Google provider repo, not OL.

## Proposed implementation (How?)

Add a new GCP registry job facet `GcpBigQueryRoutineJobFacet` under `spec/registry/gcp/bigquery/facets/`. Modeled as a `JobFacet` because the invoked routine is part of the job's *definition* — it doesn't vary per run — consistent with `JobTypeJobFacet` and `SourceCodeJobFacet`.

Fields:
- `projectId`, `datasetId`, `routineId` — required. Mirrors BigQuery's REST API `RoutineReference` shape; together they uniquely identify a routine (BigQuery does not support overloading by argument signature).
- `routineType` — optional enum. Mirrors `INFORMATION_SCHEMA.ROUTINES.routine_type` so consumers can distinguish stored procedures from UDFs. Optional because producers may not know the routine type at emit time (e.g., a custom extractor that only sees the SQL text).

### Proposed schema

`spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json`:

```json
{
  "$id": "https://openlineage.io/spec/facets/1-0-0/GcpBigQueryRoutineJobFacet.json",
  "$defs": {
    "GcpBigQueryRoutineJobFacet": {
      "allOf": [
        {
          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/JobFacet"
        },
        {
          "type": "object",
          "properties": {
            "projectId": {
              "description": "GCP project ID containing the routine",
              "type": "string"
            },
            "datasetId": {
              "description": "BigQuery dataset containing the routine",
              "type": "string"
            },
            "routineId": {
              "description": "Name of the routine (stored procedure or user-defined function)",
              "type": "string"
            },
            "routineType": {
              "description": "BigQuery routine type as reported by `INFORMATION_SCHEMA.ROUTINES.routine_type`. Optional — omit when unknown at emit time",
              "type": "string",
              "enum": ["PROCEDURE", "SCALAR_FUNCTION", "TABLE_VALUED_FUNCTION", "AGGREGATE_FUNCTION"]
            }
          },
          "required": ["projectId", "datasetId", "routineId"],
          "additionalProperties": true
        }
      ],
      "type": "object"
    }
  },
  "type": "object",
  "properties": {
    "gcp_bigquery_routine_job": {
      "$ref": "#/$defs/GcpBigQueryRoutineJobFacet"
    }
  }
}
```

Companion `spec/registry/gcp/bigquery/registry.json`:

```json
{
  "producer": {
    "root_doc_URL": "https://cloud.google.com/bigquery/docs/routines",
    "produced_facets": ["ol:gcp:bigquery:GcpBigQueryRoutineJobFacet.json"]
  }
}
```

### Design rationale (FAQ)

**Why three separate fields instead of a single `routineReference` string?**
BigQuery's REST API `RoutineReference` resource models them as three separate fields (`projectId`, `datasetId`, `routineId`); we mirror that. Dataset names can legally contain dots when quoted, so a dotted-string form is ambiguous to parse.

**Why no `argumentSignature` / overload disambiguation field?**
BigQuery does not support routine overloading by argument type — `CREATE FUNCTION foo(x INT64)` and `CREATE FUNCTION foo(x STRING)` collide. `{projectId}.{datasetId}.{routineId}` is uniquely identifying.

**Why a JobFacet, not a RunFacet?**
The invoked routine is part of the job's *definition* — it doesn't vary per run. Same modeling choice as `JobTypeJobFacet` and `SourceCodeJobFacet`. If a generic-dispatch use case emerges (e.g., `CALL {{ params.proc_name }}` where the routine varies per run), a separate `GcpBigQueryRoutineRunFacet` can be added without breaking this one.

**Why not extend `BigQueryJobRunFacet`?**
Different host repo (Airflow Google provider, not OL), different facet kind (run vs. job), different semantic (dynamic run metrics vs. static job identity). Coupling routine identity to the Airflow provider would also prevent non-Airflow producers (custom code, Dataflow, etc.) from emitting routine info.

**Why not an array of routines?**
OL's idiomatic 1-to-N pattern is "emit multiple OL events" or "use child runs with `ParentRunFacet`" — not arrays inside a single facet. Compare: `ParentRunFacet`, `JobTypeJobFacet`, `SQLJobFacet`, `SourceCodeJobFacet` are all singular. A stored procedure that internally calls other procedures is modeled via child SCRIPT jobs, not via an array.

**Why those four enum values for `routineType`?**
They are the exact set returned by BigQuery's `INFORMATION_SCHEMA.ROUTINES.routine_type` and the REST API `Routine.routineType` field — `PROCEDURE`, `SCALAR_FUNCTION`, `TABLE_VALUED_FUNCTION`, `AGGREGATE_FUNCTION`. There is no `REMOTE_FUNCTION` enum value (remote functions are stored as `SCALAR_FUNCTION` with `remoteFunctionOptions` set), and materialized views are catalogued under `TABLES`, not `ROUTINES`.

### Scope

Spec-only addition under `spec/registry/gcp/bigquery/`. No core spec changes, no producer-side integration in this PR. Follow-up will populate the facet from `BigQueryInsertJobOperator` in `apache-airflow-providers-google`. Custom-operator users can populate it via an OL Airflow extractor.

### References

- BigQuery Routines REST API: https://cloud.google.com/bigquery/docs/reference/rest/v2/routines
- `INFORMATION_SCHEMA.ROUTINES`: https://cloud.google.com/bigquery/docs/information-schema-routines
- Existing precedent (Gcp composer facets): #3986
- Existing precedent (Gcp dataproc registry facet): #2987
- Tracking issue for routine identity gap: #773

rahul-madaan:
closes: #4562 

Add `GcpBigQueryRoutineJobFacet` to identify the BigQuery routine (stored procedure or user-defined function) invoked by an OpenLineage job.

### Meaningful description

OpenLineage events emitted from Airflow tasks that invoke BigQuery routines (stored procedures, scalar/table/aggregate UDFs) currently lose the identity of the called routine. For `BigQueryInsertJobOperator` with a `CALL ...` query, the existing extractor walks the parent SCRIPT job and aggregates child input/output datasets but the routine reference itself is dropped (see #773); for non-Google operators (e.g., a `PythonOperator` wrapping `bigquery.Client().query("CALL ...")`) there is no routine information in the event at all. This blocks consumers from building an `AirflowTask → BigQueryRoutine → Table` relationship.

This PR adds a new GCP registry job facet `GcpBigQueryRoutineJobFacet` under `spec/registry/gcp/bigquery/facets/`, following the same shape as the existing `GcpComposerJobFacet` and `GcpDataprocRunFacet`. The facet carries `projectId`, `datasetId`, `routineId` (required) — matching BigQuery's REST API `RoutineReference` shape, which uniquely identifies a routine since BigQuery does not support overloading. Optional `routineType` mirrors the `INFORMATION_SCHEMA.ROUTINES.routine_type` enum (`PROCEDURE`, `SCALAR_FUNCTION`, `TABLE_VALUED_FUNCTION`, `AGGREGATE_FUNCTION`) so consumers can distinguish stored procedures from UDFs.

Modeled as a `JobFacet` because the invoked routine is part of the job's definition (it doesn't vary per run), consistent with `JobTypeJobFacet` and `SourceCodeJobFacet`. A `RunFacet` variant can be added later if a generic-dispatch operator pattern emerges where the routine varies per run.

Spec-only addition — no core spec changes, no producer-side integration in this PR. Follow-up will populate the facet from `BigQueryInsertJobOperator` in `apache-airflow-providers-google`; custom-operator users can populate it via an OL Airflow extractor. Extending `BigQueryJobRunFacet` was considered and rejected — that facet lives in the Airflow Google provider repo and carries dynamic run metrics, not static job identity, so coupling routine identity to it would prevent non-Airflow producers from emitting routine info.

### Checklist

### DIFF
--- a/spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json
+++ b/spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json
@@ -1,0 +1,44 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/GcpBigQueryRoutineJobFacet.json",
+  "$defs": {
+    "GcpBigQueryRoutineJobFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/JobFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "projectId": {
+              "description": "GCP project ID containing the routine",
+              "type": "string"
+            },
+            "datasetId": {
+              "description": "BigQuery dataset containing the routine",
+              "type": "string"
+            },
+            "routineId": {
+              "description": "Name of the routine (stored procedure or user-defined function)",
+              "type": "string"
+            },
+            "routineType": {
+              "description": "BigQuery routine type as reported by `INFORMATION_SCHEMA.ROUTINES.routine_type`. Optional — omit when unknown at emit time",
+              "type": "string",
+              "enum": ["PROCEDURE", "SCALAR_FUNCTION", "TABLE_VALUED_FUNCTION", "AGGREGATE_FUNCTION"]
+            }
+          },
+          "required": ["projectId", "datasetId", "routineId"],
+          "additionalProperties": true
+        }
+      ],
+      "type": "object"
+    }
+  },
+  "type": "object",
+  "properties": {
+    "gcp_bigquery_routine_job": {
+      "$ref": "#/$defs/GcpBigQueryRoutineJobFacet"
+    }
+  }
+}
--- a/spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json
+++ b/spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json
@@ -23,9 +23,8 @@
               "type": "string"
             },
             "routineType": {
-              "description": "BigQuery routine type as reported by `INFORMATION_SCHEMA.ROUTINES.routine_type`. Optional — omit when unknown at emit time",
-              "type": "string",
-              "enum": ["PROCEDURE", "SCALAR_FUNCTION", "TABLE_VALUED_FUNCTION", "AGGREGATE_FUNCTION"]
+              "description": "BigQuery routine type as reported by `INFORMATION_SCHEMA.ROUTINES.routine_type` (`PROCEDURE`, `SCALAR_FUNCTION`, `TABLE_VALUED_FUNCTION`, `AGGREGATE_FUNCTION`, etc.)",
+              "type": "string"
             }
           },
           "required": ["projectId", "datasetId", "routineId"],
--- a/spec/registry/gcp/bigquery/registry.json
+++ b/spec/registry/gcp/bigquery/registry.json
@@ -1,0 +1,6 @@
+{
+  "producer": {
+    "root_doc_URL": "https://cloud.google.com/bigquery/docs/routines",
+    "produced_facets": ["ol:gcp:bigquery:GcpBigQueryRoutineJobFacet.json"]
+  }
+}
--- a/website/static/spec/facets/1-0-0/GcpBigQueryRoutineJobFacet.json
+++ b/website/static/spec/facets/1-0-0/GcpBigQueryRoutineJobFacet.json
@@ -1,0 +1,44 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://openlineage.io/spec/facets/1-0-0/GcpBigQueryRoutineJobFacet.json",
+  "$defs": {
+    "GcpBigQueryRoutineJobFacet": {
+      "allOf": [
+        {
+          "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/JobFacet"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "projectId": {
+              "description": "GCP project ID containing the routine",
+              "type": "string"
+            },
+            "datasetId": {
+              "description": "BigQuery dataset containing the routine",
+              "type": "string"
+            },
+            "routineId": {
+              "description": "Name of the routine (stored procedure or user-defined function)",
+              "type": "string"
+            },
+            "routineType": {
+              "description": "BigQuery routine type as reported by `INFORMATION_SCHEMA.ROUTINES.routine_type`. Optional — omit when unknown at emit time",
+              "type": "string",
+              "enum": ["PROCEDURE", "SCALAR_FUNCTION", "TABLE_VALUED_FUNCTION", "AGGREGATE_FUNCTION"]
+            }
+          },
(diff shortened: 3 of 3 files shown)


spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json
```diff
@@ -0,0 +1,44 @@
...
+              "type": "string"
+            },
+            "routineId": {
+              "description": "Name of the routine (stored procedure or user-defined function)",
+              "type": "string"
+            },
+            "routineType": {
+              "description": "BigQuery routine type as reported by `INFORMATION_SCHEMA.ROUTINES.routine_type`. Optional — omit when unknown at emit time",
+              "type": "string",
+              "enum": ["PROCEDURE", "SCALAR_FUNCTION", "TABLE_VALUED_FUNCTION", "AGGREGATE_FUNCTION"]
```
kacpermuda: I think we can just leave the string type and provide enum values as example, without limiting us to enums. This way, when new procedure type is added to google, we don't need to bump the spec and can keep using current version of the facet.

kacpermuda (commented):
Looks good, thanks, left one comment. Is there anything else similar to routine in bigquery that this facet could cover, like some other kind of UDF ? We can make this a more generic facet then.

spec/registry/gcp/bigquery/facets/GcpBigQueryRoutineJobFacet.json
rahul-madaan: Good call. dropped the `enum` constraint and moved the four current values into the field description so producers can emit any future routine type without a spec bump.

rahul-madaan:
@kacpermuda to make it more generic, we could cover routine like concepts from multiple sources (like snowflake has stored procedures), I have started another PR https://github.com/OpenLineage/OpenLineage/pull/4573 for RoutineJobFacet, please take a look and let me know if it makes more sense since it is a generic facet now.

kacpermuda:
Commented on #4573 , let's see how it goes, we might close this one in favor of the more generic one.

PR #4573, opened 2026-05-25
spec: add RoutineJobFacet

rahul-madaan:
Add `RoutineJobFacet` to identify the routine (stored procedure or user-defined function) invoked by an OpenLineage job, generalized across warehouses.

### Meaningful description

OpenLineage events emitted from jobs that invoke SQL routines (stored procedures, scalar / table / aggregate UDFs) currently lose the identity of the called routine. For BigQuery's `BigQueryInsertJobOperator` with a `CALL ...` query, the existing extractor walks the parent SCRIPT job and aggregates child input/output datasets but the routine reference itself is dropped (see #773). For non-Google operators (e.g. a `PythonOperator` wrapping `bigquery.Client().query("CALL ...")`), there is no routine information in the event at all. The same gap exists for Snowflake stored procedures, Redshift functions, Postgres routines, and any other SQL warehouse OpenLineage supports. This blocks consumers (metadata catalogs, governance platforms) from building an `AirflowTask → Routine → Table` relationship.

This PR adds a new core job facet `RoutineJobFacet` under `spec/facets/`, modeled on OpenLineage's existing Dataset `{namespace, name}` identity convention. The facet carries `namespace`, `name`, optional `routineType`, and optional `language`. `namespace` reuses the value that the source system already uses for its datasets (e.g. `bigquery`, `snowflake://my-org-myaccount`, `redshift://cluster.region:5439`); `name` reuses the same fully-qualified path convention as tables in that system (`my-project.my_dataset.my_procedure`, `MY_DB.MY_SCHEMA.MY_PROCEDURE`, etc.). One schema, every SQL warehouse — no vendor-specific facet required.

Modeled as a `JobFacet` because the invoked routine is part of the job's definition (it does not vary per run), consistent with `JobTypeJobFacet` and `SourceCodeJobFacet`. `website/docs/spec/naming.md` is updated with a new `Routine Naming` section that documents the namespace + name convention for BigQuery, Snowflake, Redshift, Postgres, and Databricks.

Spec-only addition — no producer-side integration in this PR. A follow-up will populate the facet from `BigQueryInsertJobOperator` in `apache-airflow-providers-google`. Custom-operator users can populate it via an OL Airflow extractor.

**Note on the alternative design (PR #4563).** PR #4563 proposed a BigQuery-specific `GcpBigQueryRoutineJobFacet` under `spec/registry/gcp/bigquery/`. After deeper review, the four fields on that facet (`projectId`, `datasetId`, `routineId`, `routineType`) are not vendor-unique — they are simply BigQuery's spelling of `{namespace, name, type}` that every other warehouse expresses identically. Routine identity fits OpenLineage's existing Dataset naming pattern cleanly, so a generic core facet looks like a better long-term home than a per-vendor registry facet. This PR is offered as an alternative to #4563; reviewers can pick whichever approach they prefer.

#4562 
### Checklist


kacpermuda:
I think this makes more sense over the bigquery specific facet. Not sure about the naming, ma ybe we should be more specific, that the job calls the routine, so something like `InvokedRoutineJobFacet` ? Also, for databricks I think some producers already use `databricks` namespace, for Unity Catalog (where catalogs/schemas/routines live), maybe we should add this to job naming convention, and also reflect in routine naming one? Also, not sure if we should copy the naming convention in the second table, maybe we should just provide a small generic paragraph, that the namespace should follow the above job naming convention and the name should be a unique id like `db.schema.routine_id` and that's it? Not sure if copying it for every possible db makes sense here.

Also cc @mobuchowski @jakub-moravec, good to have more eyes on this, since it's a spec addition.

rahul-madaan:
I am aligned on the idea of having `InvokedRoutineJobFacet` since it is more intuitive.

@mobuchowski @jakub-moravec could you please take a look at this and let me know your thoughts on including this in the spec?

mobuchowski:
I think it makes sense to identify the routines - procedures, udfs etc. 
I have issue with narrowing it to a single routine though.
Even in single SQL statement (that I thing would be the smallest "scope" of OpenLineage event) we can find examples with multiple of them being called in a single statement:

```
SELECT my_udf(a), other_udf(b), agg_udf(c) from t
```

Or in scripting like PL/pgSQL:

```
DO $$
BEGIN
  CALL load_dims();
  CALL load_facts();
  CALL refresh_aggregates();
END
$$;
```

jakub-moravec:
Hi @rahul-madaan - to make sure I understand the case well, a couple of high-level questions 

- What is the represented OL Job in this case? Why isn't the Job itself the routine? 
- Or is this supposed to cover "chained" routines - e.g., a stored proc calling a function that is calling another function? 

If that's the case, I think the high-level question is whether to represent this in a specific facet, or whether to rather try making it possible to document this as lineage - I think the Explicit Lineage spec (specifically thy job-to-job part) is a step towards that https://github.com/OpenLineage/OpenLineage/pull/4468

PR #4574, opened 2026-05-26
spark: add Lakehouse Hive Catalog handling

tnazarew:
Add handling for Lakehouse Hive Catalog 

### Meaningful description
Spark jobs on Managed Spark (fromerly Dataproc) can use Lakehouse (formerly BigLake) with Hive Catalog. This requires custom implementation of Hive libraries. The code checks if the custom changes are present and if the appropriate properties are set. If so the catalog facet is added along with specific symlink.

Aside from that, mechanism to check if V2 sources use `HiveSessionCatalog` was added to increase coverage.

A lot of visitors didn't have any tests so there is crapload of added test lines

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
@@ -20,7 +20,6 @@ import org.apache.spark.SparkContext;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.catalyst.TableIdentifier;
 import org.apache.spark.sql.catalyst.catalog.CatalogTable;
-import org.apache.spark.sql.connector.catalog.CatalogPlugin;
 import org.apache.spark.sql.internal.StaticSQLConf;
 
 @Slf4j
@@ -59,20 +58,21 @@ public class PathUtils {
   @SneakyThrows
   public static DatasetIdentifier fromCatalogTable(
       CatalogTable catalogTable, SparkSession sparkSession, URI location) {
-
-    return fromTableIdentifier(
-        catalogTable.identifier(),
-        sparkSession.sparkContext(),
-        location,
-        CatalogDatasetFacetUtils.getCatalogPlugin(sparkSession, catalogTable.identifier()));
+    SparkContext sparkContext = sparkSession.sparkContext();
+    if (CatalogDatasetFacetUtils.isHiveCatalog(sparkSession, catalogTable.identifier())
+        && GoogleCloudPlatformUtils.isBigLakeHiveCatalog(sparkContext.getConf())) {
+      return fromURI(location)
+          .withSymlink(
+              nameFromTableIdentifier(catalogTable.identifier()),
+              "gcp_lakehouse",
+              DatasetIdentifier.SymlinkType.TABLE);
+    }
+    return fromTableIdentifier(catalogTable.identifier(), sparkContext, location);
   }
 
   @SneakyThrows
   public static DatasetIdentifier fromTableIdentifier(
-      TableIdentifier identifier,
-      SparkContext sparkContext,
-      URI location,
-      Optional<CatalogPlugin> catalogPlugin) {
+      TableIdentifier identifier, SparkContext sparkContext, URI location) {
     // perform URL normalization
     DatasetIdentifier locationDataset = fromURI(location);
     URI locationUri = FilesystemDatasetUtils.toLocation(locationDataset);
@@ -90,11 +90,6 @@ public class PathUtils {
       String tableName = nameFromTableIdentifier(identifier, "/");
       symlinkDataset =
           Optional.of(new DatasetIdentifier(GLUE_TABLE_PREFIX + tableName, glueArn.get()));
-    } else if (catalogPlugin.isPresent()
-        && CatalogDatasetFacetUtils.isHiveCatalog(sparkContext, catalogPlugin.get())
-        && GoogleCloudPlatformUtils.isBigLakeHiveCatalog(sparkContext.getConf())) {
-      String tableName = nameFromTableIdentifier(identifier);
-      symlinkDataset = Optional.of(new DatasetIdentifier(tableName, "gcp_lakehouse"));
     } else if (metastoreUri.isPresent()) {
       // dealing with Hive tables
       URI hiveUri = prepareHiveUri(metastoreUri.get());
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
@@ -20,6 +20,7 @@ import org.apache.spark.SparkContext;
 import org.apache.spark.sql.SparkSession;
 import org.apache.spark.sql.catalyst.TableIdentifier;
 import org.apache.spark.sql.catalyst.catalog.CatalogTable;
+import org.apache.spark.sql.connector.catalog.CatalogPlugin;
 import org.apache.spark.sql.internal.StaticSQLConf;
 
 @Slf4j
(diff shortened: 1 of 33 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/api/QueryPlanVisitor.java
```diff
@@ -138,4 +143,23 @@ public String toString() {
     String genericArgs = typeArgs.map(opt -> opt.orElse("?")).collect(Collectors.joining(","));
     return getClass().getName() + "<" + genericArgs + ">";
   }
+
+  protected List<OpenLineage.OutputDataset> getOutputDatasetFacets(
+      CatalogTable catalogTable,
```
pawel-big-lebowski: Would it make sense to have `OutputDatasetContext` will all those 4 params and pass a single arg?

integration/spark/shared/src/main/java/io/openlineage/spark/api/QueryPlanVisitor.java
```diff
@@ -138,4 +143,23 @@ public String toString() {
     String genericArgs = typeArgs.map(opt -> opt.orElse("?")).collect(Collectors.joining(","));
     return getClass().getName() + "<" + genericArgs + ">";
   }
+
+  protected List<OpenLineage.OutputDataset> getOutputDatasetFacets(
```
pawel-big-lebowski: name suggests the facets are returned.

integration/spark/shared/src/main/java/io/openlineage/spark/api/QueryPlanVisitor.java
```diff
@@ -138,4 +143,23 @@ public String toString() {
...
+        context
+            .getSparkSession()
+            .flatMap(
+                session -> {
+                  if (CatalogDatasetFacetUtils.isHiveCatalog(session, catalogTable.identifier())) {
+                    return CatalogDatasetFacetUtils.getCatalogDatasetFacetForHive(context);
+                  }
+                  return Optional.empty();
+                })
+            .map(facet -> outputDataset().getDataset(di, schema, lifecycleStateChange, facet))
```
pawel-big-lebowski: May be a silly thought: why can't we include extra facet in getDataset method, possibly with passing new DatasetContext into it.

pawel-big-lebowski (commented):
First, I think this PR is same as good as PRs we've merged in the past for similar usecases, like iceberg versioning.

At the same time, we no longer care about Spark 2 and code refactor is a way cheaper to what it was in the past. That said, can we refactor the code to put the logic into either `outputDataset().getDataset(` or `factory.getDataset` ? Overall, the less of `CatalogDatasetFacetUtils.isHiveCatalog` calls in the code we have, the better.

integration/spark/shared/src/main/java/io/openlineage/spark/api/QueryPlanVisitor.java
tnazarew: so instead having a builder pattern, have a method taking the context with all the information. Could work, I'll look into that

I see that we have a DatasetCompositeFacetsBuilder that takes the facets and builds datasets, it could be a good idea to make a wrapper object that lets you supply values used in `getDataset` methods instead of ready facets. That seems to me a more generic approach that we could extend to all the visitors

tnazarew:
@pawel-big-lebowski updated the PR with the refactor

I was thinking about using the `isHiveCatalog` as little as possible but the original issue was prompted by query
```sql
CREATE TABLE IF NOT EXISTS {HIVE_DATABASE}.top_users
      USING PARQUET
    AS SELECT name
       FROM {HIVE_DATABASE}.{USERS_TABLE_NAME}
```

which was handled by `CreateDataSourceTableCommandVisitor`. It's not hive specific visitor but it had `HiveSessionCatalog` underneath V2 catalog when I debugged it. 

So potentially we could extract catalog information for more types of visitors then we currently do, therefore I added this whenever there is CatalogTable present.

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkDatasetCompositeFacetsBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * <p>For example, {@link #schema(StructType)} converts a Spark {@link StructType} to a {@link
+ * SchemaDatasetFacet} using {@link PlanUtils#schemaFacet} internally, so callers do not need to
+ * hold a reference to the {@link OpenLineage} instance themselves.
+ *
+ * <p>All methods return {@code this} for fluent chaining. The underlying {@link
+ * DatasetCompositeFacetsBuilder} is accessible via {@link #getInner()} for passing to existing APIs
+ * that still require it.
+ */
+public abstract class SparkDatasetCompositeFacetsBuilder<T extends OpenLineage.Dataset> {
```
pawel-big-lebowski: Ain't this just `SparkDatasetBuilder` ?

pawel-big-lebowski (commented):
Great improvement ❤️  Visitors and dataset builders look a way better now 💯 
Would you mind splitting the PR into refactor + Hive Catalog PRs. 
This would mean cleaner changelog and better readability of hive catalog handling.

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkDatasetCompositeFacetsBuilder.java
tnazarew: Could be but I my line of thinking is "this is a Spark wrapper for `DatasetCompositeFacetsBuilder` so lets indicate that". 

I don't have anything against `SparkDatasetBuilder` though.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/PathUtils.java
```diff
@@ -58,12 +59,20 @@ public static DatasetIdentifier fromCatalogTable(
...
+        CatalogDatasetFacetUtils.getCatalogPlugin(sparkSession, catalogTable.identifier()));
   }
 
   @SneakyThrows
   public static DatasetIdentifier fromTableIdentifier(
-      TableIdentifier identifier, SparkContext sparkContext, URI location) {
+      TableIdentifier identifier,
+      SparkContext sparkContext,
+      URI location,
+      Optional<CatalogPlugin> catalogPlugin) {
```
pawel-big-lebowski: Using optionals as arguments is considered a code-smell and should be avoided. 
Consider having two method declarations instead. 
Ideally, we should converge to DatasetIdentifierBuilder, but let's leave it for separate vibe-coded PR.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DropTableVisitor.java
```diff
@@ -104,6 +105,8 @@ private List<OpenLineage.OutputDataset> applyResolvedTable(ResolvedTable resolve
               .schema(resolvedTable.schema())
               .lifecycleStateChange(
                   OpenLineage.LifecycleStateChangeDatasetFacet.LifecycleStateChange.DROP);
+      CatalogUtils3.getCatalogDatasetFacet(context, tableCatalog, tableProperties)
```
pawel-big-lebowski, line 108: didn't we add this to SparkDatasetBuilder so that we can avoid calls in visitors?
If create/drop are kind of special cases, please provide a comment

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkDatasetCompositeFacetsBuilder.java
pawel-big-lebowski: I think it's more of a dataset builder than facets builder.

pawel-big-lebowski (commented):
Provided few code-quality related comments & suggestions. Let me know if they make sense to you. 
 
Overall, code design looks good to me.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DropTableVisitor.java
tnazarew: `DatasetFactory` (and the builder) are in the `shared` package while `CatalogUtils3` are available in spark3+ packages. Because of that we cant use the `CatalogUtils3` methods in the builder. That also goes for `addStorageAndCatalogFacets`.

Now that you mentioned it, it's actually _mea culpa_ because this should not be in this PR but the refactor one...

integration/spark/shared/src/main/java/io/openlineage/spark/api/SparkDatasetCompositeFacetsBuilder.java
tnazarew: fair enough, I'll change to `SparkDatasetBuilder`

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/DropTableVisitor.java
tnazarew: thing we could do right now is to implement `.catalog(Optional<CatalogDatasetFacet> catalog)` method it solves this issue but creates the optionals-in-arguments issue

pawel-big-lebowski (approved):

PR #4586, opened 2026-06-01
fix(python): isolate datadog config defaults

### LINKED ISSUE #4585: Python Datadog config should not share mutable defaults
﻿## Problem

`DatadogConfig` defines `retry` and `async_transport_rules` with mutable dictionary defaults.

Because those defaults are shared between instances, mutating one config instance can affect later instances. For example, changing `config.retry["total"]` or adding a rule under `config.async_transport_rules` is visible on a separately constructed `DatadogConfig`.

## Expected behavior

Each `DatadogConfig` instance should receive an independent retry configuration and independent async transport rules.

## Proposed fix

Use factories for the mutable defaults and add regression tests that verify mutations on one instance do not leak into another instance.

hcthakur2004:
﻿## Summary

- replace shared Datadog config dictionary defaults with per-instance factories
- add regression coverage for retry and async transport rule defaults

## Issue

Fixes #4585

## Tests

- `uv run pytest tests/test_datadog.py`
- `uv run ruff check src/openlineage/client/transport/datadog.py tests/test_datadog.py`


mobuchowski:
@hcthakur2004 mind making this ready for review so we can merge it?

hcthakur2004:
> @hcthakur2004 mind making this ready for review so we can merge it?

Done, thanks!

mobuchowski (approved):

PR #4591, opened 2026-06-02
Dbt bq external query id fully qualified

fm100:
Use fully qualified job ID for externalQuery run facet for dbt-bigquery adapter.

### Meaningful description
* Dbt integration contains `externalQuery` run facet but it is not using fully qualified job ID when the adapter is BigQuery. This change is to make it use fully qualified job Id for `externalQueryId` if project Id and location are available.
* The `externalQuery` run facet was not being included when using `--consume-structured-logs` option. This PR also makes it include `externalQuery` run facet for it. The facet is only included in COMPLETE event because START event is being sent even before the actual job submission.

### Checklist


integration/common/tests/dbt/structured_logs/test_structured_logs.py
```diff
@@ -264,6 +351,65 @@ def parsed():
...
+    "adapter_type, dataset_namespace, adapter_response, expected_external_query_id",
+    [
+        (
+            Adapter.BIGQUERY,
+            "bigquery",
+            {"project_id": "test-project", "location": "US", "job_id": "job-123"},
+            "test-project:US.job-123",
+        ),
+        (Adapter.BIGQUERY, "bigquery", {"job_id": "job-123"}, "job-123"),
+        (Adapter.SNOWFLAKE, "snowflake://test-account", {"query_id": "query-123"}, "query-123"),
```
kacpermuda, line 385: Let's add some partial response test cases, where we have project without location and location without project + job_id, to make sure we don't create full id unless we have it all.

integration/common/tests/dbt/test_processor.py
```diff
@@ -94,6 +94,29 @@ def test_get_query_id(
     }
 
 
+@pytest.mark.parametrize(
+    "adapter_response, expected_query_id",
+    [
+        (
```
kacpermuda, line 100: Same as above, let's add partial response test cases

integration/common/tests/dbt/structured_logs/snowflake/run/results/successful_run_ol_events.json
kacpermuda, line 1: All the structured logs tests changed are for snowflake, can we add one actual example of BQ? Would be nice to have the test for it since we're modifying it.

kacpermuda (commented):
Thanks for improving this, left a few comments, mostly around testing.

fm100:
> Thanks for improving this, left a few comments, mostly around testing.

Thank you for the feedback. I've updated tests to cover partial responses, and added golden test fixture for BQ as well.

kacpermuda (approved):

PR #4592, opened 2026-06-02
spec: extend jobTypeJobFacet to capture event emission lifecycle

### LINKED ISSUE #4559: [PROPOSAL] Add STREAMING Run State for Better Streaming Process Support
## Summary

Introduce a new Run State `STREAMING` to the OpenLineage specification to better support long-running streaming processes that don't fit the traditional ETL start-complete lifecycle.

### Current Limitations

The OpenLineage run cycle (START → RUNNING → COMPLETE/FAIL/ABORT) is optimized for ETL-like processes with clear start and end points. However, streaming processes have fundamentally different characteristics:

1. **No Clear Start/End Boundaries**: Streaming jobs are designed to run indefinitely, processing continuous data streams. Unlike batch ETL jobs, they don't have natural start and completion points (unless we consider infrastructure events like topic creation/deletion, which can span months and don't align with the semantic meaning of START/COMPLETE).

2. **Current Workarounds Are Inadequate**: Existing integrations for streaming tools (e.g., Apache Flink) use RUNNING events to report streaming job activity. However, this creates semantic confusion:
   - **Accumulative Semantics Problem**: The current Run Event model is accumulative - consumers need access to ALL events (START through COMPLETE) for a given RunID to have complete information about a job run.
   - **Consumer Expectations**: Some consumers expect START and COMPLETE events and may have processing logic that depends on this lifecycle.
   - **Snapshot vs. Accumulation**: Streaming events represent point-in-time snapshots for specific time periods, not accumulative state. Each event contains complete information for its time window and should be processed independently.

3. **Different Event Lifecycle**: Streaming events have a fundamentally different lifecycle:
   - Each event is self-contained for its time period
   - No need to combine multiple events to understand what happened
   - Events represent periodic summaries rather than state transitions
   - Consumers should process each event independently, not accumulate them

### New Run State: STREAMING

Add a new `STREAMING` state to the Run State enumeration:

```
enum RunState {
  START,
  RUNNING,
  COMPLETE,
  ABORT,
  FAIL,
  STREAMING  // NEW
}
```

### Semantics

**STREAMING State Characteristics:**

- **Self-Contained Events**: Each STREAMING event contains complete lineage and metrics information for a specific time period
- **Periodic Emission**: Events are emitted at regular intervals
- **No Lifecycle Dependency**: Consumers don't need to track START/COMPLETE events or accumulate information across multiple events
- **Time-Windowed Snapshots**: Each event represents a snapshot of activity within a defined time window
- **Independent Processing**: Each event can be processed independently without reference to previous or subsequent events

## Benefits

1. **Clear Semantic Distinction**: Explicitly differentiates streaming workloads from batch ETL processes
2. **Simplified Consumer Logic**: Consumers can handle streaming events differently, without expecting START/COMPLETE lifecycle
3. **Better Tooling Support**: Enables streaming-specific features in observability tools (e.g., time-series views, windowed metrics)
4. **Accurate Representation**: Aligns the specification with the reality of how streaming systems operate

jakub-moravec:
closes: [#ISSUE](https://github.com/OpenLineage/OpenLineage/issues/4559#issuecomment-4541902495)

Extend Job Type facet to allow streaming producers to document the lifecycle of emitted OL events

### Meaningful description
Extend Job Type facet to allow producers to document the lifecycle of emitted OL events that consumers should expect - specifically to better support Streaming producers who typically emit events periodically and capture all metadata for a specific time period in a single event, as opposed to batch process producers that are more likely to split the emitted metadata between START, RUNNING, and COMPLETE events.

### Checklist


client/python/src/openlineage/client/generated/job_type_job.py
```diff
@@ -27,7 +79,28 @@ class JobTypeJobFacet(JobFacet):
 
     Example: QUERY
     """
+    lifecycleType: LifecycleType | None = attr.field(default=None)  # noqa: N815
```
dolfinus: Isn't that just the same as BATCH/STREAMING processingType? Is there a case when BATCH jobs emits events CONTINUOUSly, or vice versa?

jakub-moravec: @dolfinus I was pondering the same thing when creating the PR. My quick surface-level research says there are bounded stream processes with clear termination conditions, and other types of jobs that would fall into this category. 

On the other side, you have web services that are CONTINOUS, but are not STREAMS. 

So I think ultimately this is a tradeoff between simplicity and the ability to distinguish more nuances. I'm myself not fully convinced one way or the other; I'd be OK with removing the lifecycleType and documenting that STREAMS are CONTINUOUS.

dolfinus: IMHO for web servers OL producer could emit different processingType other than STREAMING, and this allows to distinguish those cases, and also some others. Having one enum describing the job behaviour looks unrealistic for me, there always be exceptions.

jakub-moravec: @dolfinus adjusted

website/docs/spec/facets/job-facets/job-type.md
```diff
@@ -21,14 +45,117 @@ Example:
...
+Defines when events are emitted:
+
+* **`EVENT_BASED`** - Events emitted on lifecycle transitions
+  * `START` when job begins
+  * `COMPLETE`/`FAIL`/`ABORT` when job ends
+  * `RUNNING` for progress updates (optional)
+  
+* **`PERIODIC`** - Events emitted at regular time intervals
+  * Emitted on a schedule (e.g., every 5 minutes)
+  * Independent of job lifecycle state
```
dolfinus: Flink integration still emits START events. Also it may emit COMPLETE/FAIL after job is stopped. So it's not completely independent from regular lifecycle, it's just dominated by RUNNING events

website/docs/spec/facets/job-facets/job-type.md
```diff
@@ -21,14 +45,117 @@ Example:
...
+                    "eventTrigger": "PERIODIC",
+                    "eventCompleteness": "COMPLETE_SNAPSHOT",
+                    "windowDuration": 300
+                },
+                "_producer": "https://github.com/OpenLineage/OpenLineage/blob/v1-0-0/client",
+                "_schemaURL": "https://openlineage.io/spec/facets/2-0-4/JobTypeJobFacet.json"
+            }
+        }
+    }
+}
```
dolfinus, line 135: Missing code block end

website/docs/spec/facets/job-facets/job-type.md
```diff
@@ -21,14 +45,117 @@ Example:
...
+```
+
+## Integration-Specific Examples
 
  * Integration: `SPARK`
     * Processing type: `STREAM`|`BATCH`
```
dolfinus: BTW, is it STREAM or STREAMING?

website/docs/spec/run-cycle.md
```diff
@@ -26,23 +26,101 @@ There are six run states currently defined in the OpenLineage [spec](https://ope
...
+- Events can be either accumulative or complete snapshots for time windows
+- Examples: Kafka Streams, Flink streaming jobs, Spark Streaming applications
+- While `START`, `COMPLETE`, `ABORT`, or `FAIL` events can be emitted, they should not be relied upon as they might occur only occasionally when the streaming job is restarted
+
+**Event Patterns:**
+
+- **Periodic with Complete Snapshots** (recommended for streaming):
+   - Each event contains complete information for a specific time window (e.g., 5 minutes). Events can be processed independently.
+- **Periodic with Accumulative State**:
+   - Each event contains cumulative metrics since job start. Consumers need only the latest event.
```
dolfinus: Is it only about metrics, or about other facets as well?

website/docs/spec/run-cycle.md
```diff
@@ -26,23 +26,101 @@ There are six run states currently defined in the OpenLineage [spec](https://ope
...
+A long-running service - e.g., a microservice - may emit periodic or event-based (e.g. per-request) events with metrics. It will typically be represented by a `START` event followed by a series of `RUNNING` events that report changes in the run or emit performance metrics. Occasionally, a `COMPLETE`, `ABORT`, or `FAIL` event will occur, often followed by a `START` event as the service is reinitiated.
 
 ![image](./run-cycle-stream.svg)
+
+### Key differences between continous (STREAMING, SERVICE) and finite (BATCH) jobs
+- Continuous jobs can emit `START`, `COMPLETE`, `ABORT`, or `FAIL` events, however they should not be relied upon as they might be emitted only occasionally when the streaming job is restarted
+- Events are typically emitted at regular intervals (e.g., every 5 minutes)
+- Each event is self-contained for its time window
+- No need to accumulate information across multiple events
+- Consumers can process each event independently
```
dolfinus: That's typical for continuous jobs, but things may differ depending on `eventCompleteness`. This probably could confuse developers

website/docs/spec/facets/job-facets/job-type.md
jakub-moravec: @dolfinus good question :D I would expect `STREAMING`, as that's up in the current doc, but do you happen to know what Flink actually uses? 

image

jakub-moravec: .. and Spark..?

website/docs/spec/run-cycle.md
jakub-moravec: Yeah, you are right. I was going to adjust this, but I ended up repeating a lot of what is mentioned above, and in the facet doc, I think this section is becoming redundant. I'll drop it, lmk if you suggest otherwise

website/docs/spec/run-cycle.md
jakub-moravec: Hmmm, this is very tricky. I lost myself in the definition a bit; this would actually introduce a new behaviour that I don't think any producer currently creates, so let me just try to get rid of that complexity

jakub-moravec:
Thanks for your input @dolfinus! I added another commit with changes

website/docs/spec/facets/job-facets/job-type.md
dolfinus: For Flink & Spark it's `STREAMING`:
* https://github.com/OpenLineage/OpenLineage/blob/33410b2b1daae0d5b118110a3dc27255a0bd39a0/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java#L77
* https://github.com/OpenLineage/OpenLineage/blob/33410b2b1daae0d5b118110a3dc27255a0bd39a0/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/SparkSQLExecutionContext.java#L52
* https://github.com/OpenLineage/OpenLineage/blob/33410b2b1daae0d5b118110a3dc27255a0bd39a0/integration/flink/flink2/src/main/java/io/openlineage/flink/converter/OpenLineageJobExtractor.java#L96

jakub-moravec: Thanks, added the fixed value to the last commit

dolfinus (approved):

jakub-moravec:
Thank you for the review, @dolfinus! @mobuchowski I think we sorted the main open points; would you have time to review too, please?

spec/facets/JobTypeJobFacet.json
```diff
@@ -24,6 +24,31 @@
...
               "example": "QUERY"
+            },
+            "emissionPattern": {
+              "description": "Describes how and what the job emits in its events",
+              "type": "object",
+              "properties": {
+                "eventTrigger": {
+                  "description": "Defines when events are emitted. EVENT_BASED: events emitted on lifecycle transitions (START/COMPLETE/FAIL/ABORT). PERIODIC: events emitted at regular time intervals.",
+                  "type": "string",
+                  "enum": ["EVENT_BASED", "PERIODIC"],
```
mobuchowski: Can we not do enums, instead specify the defined values in text? Enums in JSON schema are super hard to evolve forward, and we found it's very often we want to do that.

spec/facets/JobTypeJobFacet.json
```diff
@@ -24,6 +24,31 @@
...
+                  "example": "EVENT_BASED"
+                },
+                "eventCompleteness": {
+                  "description": "Defines what events contain. ACCUMULATIVE: events contain cumulative state since job start (consumers need only latest event). COMPLETE_SNAPSHOT: events contain complete state for a specific time window (events can be processed independently).",
+                  "type": "string",
+                  "enum": ["ACCUMULATIVE", "COMPLETE_SNAPSHOT"],
```
mobuchowski: Same here - prefer strings to enums.

website/docs/spec/facets/job-facets/job-type.md
```diff
@@ -4,12 +4,36 @@ sidebar_position: 2
...
+  * Typically emit periodic events representing time-windowed snapshots
+  * While `START`, `COMPLETE`, `ABORT`, or `FAIL` events can be emitted, they should not be relied upon as they might occur only occasionally when the streaming job is restarted
+
+* **`SERVICE`** - Continuous long-running services
+  * Jobs run indefinitely as background services or microservices
+  * Similar to STREAMING but for general-purpose services rather than data stream processing
+  * Typically emit periodic events with cumulative or snapshot metrics
+  * While `START`, `COMPLETE`, `ABORT`, or `FAIL` events can be emitted, they should not be relied upon as they might occur only occasionally when the service is restarted
+
+Unless specified otherwise, the job is assumed to be a `BATCH` job that emits `ACCUMULATIVE` events at the time of lifecycle transitions (`EVENT_BASED`).
```
mobuchowski, line 70: Let's describe the `ACCUMULATIVE` and `EVENT_BASED` separately in this doc, before describing this assumption, or move this below the emission pattern description. This is usually read top-to-bottom, so that makes it easier to ingest.

website/docs/spec/run-cycle.md
```diff
@@ -26,23 +26,86 @@ There are six run states currently defined in the OpenLineage [spec](https://ope
...
+
+## Event Completeness
+
+The `eventCompleteness` field in the emission pattern describes what information events contain:
+
+### Accumulative Events
+
+**Accumulative events** may contain only partial information and the complete information can be collected by combining information from all the events emitted by a specific job run:
+- Individual events are more likely to contain partial rather than complete information
+- Consumers need to combine all events for a specif run to have complete information about the job run
```
mobuchowski, line 82: Good to add example here - like situation where at the start we don't know where we'll output the data, because that information is resolved in runtime (or it's dynamic in the job)

website/docs/spec/facets/job-facets/job-type.md
```diff
@@ -4,12 +4,36 @@ sidebar_position: 2
...
+  * Jobs run indefinitely with no natural completion point
+  * Process continuous data streams (e.g., Kafka, Flink, Spark Streaming)
+  * Typically emit periodic events representing time-windowed snapshots
+  * While `START`, `COMPLETE`, `ABORT`, or `FAIL` events can be emitted, they should not be relied upon as they might occur only occasionally when the streaming job is restarted
+
+* **`SERVICE`** - Continuous long-running services
+  * Jobs run indefinitely as background services or microservices
+  * Similar to STREAMING but for general-purpose services rather than data stream processing
+  * Typically emit periodic events with cumulative or snapshot metrics
+  * While `START`, `COMPLETE`, `ABORT`, or `FAIL` events can be emitted, they should not be relied upon as they might occur only occasionally when the service is restarted
```
mobuchowski, line 68: I'm not sure what's the distinction here between `STREAMING` and `SERVICE` outside of implied semantics for what general category of "processes" it is about.

Is the only difference that we specify that the periodic events are cumulative vs windowed snapshots? If yes, then why bother - when in the same PR we introduce explicit mechanisms to specify that?

mobuchowski (changes requested):

website/docs/spec/facets/job-facets/job-type.md
jakub-moravec: My main motivation to include SERVICE was the semantics - representing web services as "STREAMING" seems to be very forced. I understand that there is no immediate implication for consumers, but context matters, and I think the spec should be able to distinguish the high-level concepts. (But I'm also OK to remove this part of the PR if needed.)

jakub-moravec:
Thanks @mobuchowski , changes pushed

mobuchowski:
@jakub-moravec some test issues - probably related to the version bump.

Also go client verification: https://app.circleci.com/pipelines/github/OpenLineage/OpenLineage/18946/workflows/5349a292-4a7a-48d2-a7f9-1ec6aee9d793/jobs/552073

jakub-moravec:
@mobuchowski the PR should be in a good shape now

mobuchowski (approved):

spec/facets/JobTypeJobFacet.json
```diff
@@ -24,6 +24,29 @@
...
+            "emissionPattern": {
+              "description": "Describes how and what the job emits in its events",
+              "type": "object",
+              "properties": {
+                "eventTrigger": {
+                  "description": "Defines when events are emitted. EVENT_BASED: events emitted on lifecycle transitions (START/COMPLETE/FAIL/ABORT). PERIODIC: events emitted at regular time intervals.",
+                  "type": "string",
+                  "example": "EVENT_BASED"
+                },
+                "eventCompleteness": {
```
kacpermuda: Maybe `eventPayloadScope`, `eventContentScope`, `eventContentMode` or `eventPayloadCoverage` so it's more self explanatory?

spec/facets/JobTypeJobFacet.json
```diff
@@ -24,6 +24,29 @@
...
+                  "description": "Defines what events contain. ACCUMULATIVE: events contain cumulative state since job start (consumers need only latest event). COMPLETE_SNAPSHOT: events contain complete state for a specific time window (events can be processed independently).",
```
kacpermuda: Here we say:

> ACCUMULATIVE: events contain cumulative state since job start (consumers need only latest event).

And in the website file we say:

> **`ACCUMULATIVE`** - Events may contain only partial information and the complete information can be collected by combining information from all the events emmited by a specific job run

I would say it's not consistent?

kacpermuda (commented):
Just two nits: one around naming, one about description of accumulative

kacpermuda (approved):

jakub-moravec:
Thank you, everyone, for patient reviews @MaciejGajewski @dolfinus @kacpermuda

PR #4596, opened 2026-06-02
flink: Add support for OpenLineage events in detached Flink Session Mode

jsingh-yelp:
Enable OpenLineage lifecycle events after START for Flink jobs submitted in detached Session Mode.

### Meaningful description

- In detached Session Mode (`flink run --detached`), `JobCreatedEvent` is handled in the submitter JVM, while `DefaultJobExecutionStatusEvent` is emitted from the JobManager JVM. The JobManager-side `OpenLineageJobStatusChangedListener` did not receive `JobCreatedEvent`, so the Flink job ID was not initialized there. As a result, the listener skipped later status events with `JobId is not set`, and `RUNNING`, `COMPLETE`, and `FAIL` events were not emitted.
- The `START` event is emitted client-side. For async transports such as Kafka, the submitter JVM may exit immediately after detached submission before the producer has flushed the event.

**Changes summary:**
- Keep using `JobCreatedEvent` to emit the client-side `START` event with inputs/outputs.
- Initialize the JobManager-side listener from `DefaultJobExecutionStatusEvent` for post-submission lifecycle events, since `JobCreatedEvent` is not available in that JVM during detached Session Mode.
- Start the checkpoint tracker when `DefaultJobExecutionStatusEvent` reports the job has reached `RUNNING`.
- Derive `runUuid` deterministically from the Flink `JobID`, so the client-side `START` event and JobManager-side `RUNNING`/terminal events correlate to the same OpenLineage run.
- Detect detached mode via `DeploymentOptions.ATTACHED` and use `Transport.emit(event, timeout)` with `detachedStartEventEmitTimeoutInSeconds` to wait for `START` event delivery before the submitter JVM exits.

**Event flow**:
- Client: `START` with inputs/outputs.
- JobManager: `RUNNING` with checkpoint facets every `trackingIntervalInSeconds`.
- JobManager: `COMPLETE` or `FAIL` on terminal state.

**Testing done**:
- Added unit tests for JobManager-side job ID initialization, deterministic run ID correlation, checkpoint tracker startup, detached START bounded emit, and config parsing.
- Manually tested Application Mode, Session Mode attached, Session Mode detached, and Session Mode with multiple running jobs.

### Checklist
- [] AI was used in creating this PR


integration/flink/shared/src/main/java/io/openlineage/flink/api/OpenLineageContext.java
```diff
@@ -37,6 +38,17 @@ public class OpenLineageContext {
...
 
+  /**
+   * Sets a deterministic OpenLineage run UUID from the Flink job ID.
+   *
+   * <p>This allows listener instances running in different JVMs, for example the submitter and the
+   * JobManager in Session Mode with detached submissions, to emit events for the same OpenLineage
+   * run.
+   */
+  public void setRunUuidFromFlinkJobId(JobID jobId) {
+    this.runUuid = UUID.nameUUIDFromBytes(jobId.toHexString().getBytes(StandardCharsets.UTF_8));
```
dolfinus: This replaces UUIDv7 with UUIDv5, which have issues with storing in databases (on OL consumer side)

jsingh-yelp: @dolfinus Good point. Would deriving a deterministic UUIDv7 from the Flink JobID plus the Flink job start-time be acceptable here? The idea is to read start-time from the Flink REST /jobs/{jobId} endpoint. Ref of the API: https://nightlies.apache.org/flink/flink-docs-stable/docs/ops/rest_api/#jobs-jobid it returns a `start-time` for the job.

dolfinus: If it's possible to extract start timestamp, let's do this

jsingh-yelp: To make this work we need to make the REST call async because calling the JobManager API synchronously inside the status callback can deadlock/time out during the same state transition.

That introduces one limitation: the first `RUNNING` event can race with `START`, so consumers may see `RUNNING` before `START`. Is this out of ordering acceptable? It will only happen for the first running event which is send immediately after the job goes from `START` -> `RUNNING`.

One option is to avoid emitting the immediate RUNNING event and only emit later timer based RUNNING events. Any thoughts what should be the right course of action here?

jsingh-yelp: @dolfinus Or another idea was thinking I could completely separate the detached flow. Keeping the original flow as status quo, and for detached mode we can run this new flow with a limitation being: "Running events are only emitted by the tracker/timer"?

Users can opt for this flow explicitly, they can enable this flow for different use cases like:
- ./bin/flink run --target yarn-session --detached
- ./bin/flink run --target kubernetes-session --detached
- ./bin/flink run --target remote --detached

```
public JobStatusChangedListener createListener(Context context) {
    if (configuration.get(OPENLINEAGE_FLINK_DETACHED_MODE)) {
      return new OpenLineageDetachedJobStatusChangedListener(context, loadVisitorFactory());
    }
    return new OpenLineageJobStatusChangedListener(context, loadVisitorFactory());
  }
```

dolfinus: Is OPENLINEAGE_FLINK_DETACHED_MODE required, or it can be detected from JobManager state/options?

client/java/src/main/java/io/openlineage/client/transports/KafkaTransport.java
```diff
@@ -99,6 +101,11 @@ public void emit(@NonNull OpenLineage.RunEvent runEvent) {
     emit(OpenLineageClientUtils.toJson(runEvent), getMessageKey(runEvent));
   }
 
+  @Override
+  public void emit(@NonNull OpenLineage.RunEvent runEvent, @NonNull Duration timeout) {
+    emit(OpenLineageClientUtils.toJson(runEvent), getMessageKey(runEvent), timeout);
+  }
```
dolfinus, line 107: IMHO this should be moved into a dedicated PR.

integration/flink/shared/src/main/java/io/openlineage/flink/api/OpenLineageContext.java
jsingh-yelp: @dolfinus Unfortunately there does not seem to be a reliable JobManager-side signal for this. In remote/session submissions, options like `execution.attached=false` and `execution.target=remote` are client-side deployment options and are not necessarily forwarded into the JobManager listener context.

That is why I think an explicit, optional OpenLineage config option is safer. It lets users opt into this listener flow for detached/session deployments while keeping the default path unchanged.

client/java/src/main/java/io/openlineage/client/transports/KafkaTransport.java
jsingh-yelp: I included it here because this PR is the first place where we need bounded emission for the clientside `START` RunEvent, so the context for why this overload exists is local to this change. It is relatively a small API extension over the existing emit(..., timeout) path rather than a separate behavior change.

I’m happy to split it into a small prerequisite PR if keeping transport/client API changes isolated would make this easier to review.

integration/flink/shared/src/main/java/io/openlineage/flink/api/OpenLineageContext.java
dolfinus: Is `execution.job-status-changed-listeners` propagated to JobManager in case of `execution.target=remote`?

jsingh-yelp: Yes execution.job-status-changed-listeners exists in both job-manager and the remote machine. This is what I see now:

- job-manager: `execution.job-status-changed-listeners: io.openlineage.flink.listener.OpenLineageJobStatusChangedListenerFactory`
- remote-machine (where job is submitted): `execution.job-status-changed-listeners: io.openlineage.flink.listener.OpenLineageJobStatusChangedListenerFactory`, `execution.attached=false`, `execution.target=remote`

dolfinus: According to Flink docs, cli option `--detached` is converted into `execution.attached: false`:
https://github.com/apache/flink/blob/6a8cb80a6043431a3f158ff6c2aea0a32099791a/flink-clients/src/main/java/org/apache/flink/client/cli/ProgramOptions.java#L184

And I don't see any difference between `execution.job-status-changed-listeners` and `execution.attached` config options definition:
* https://github.com/apache/flink/blob/6a8cb80a6043431a3f158ff6c2aea0a32099791a/flink-core/src/main/java/org/apache/flink/configuration/DeploymentOptions.java#L55-L60
* https://github.com/apache/flink/blob/6a8cb80a6043431a3f158ff6c2aea0a32099791a/flink-core/src/main/java/org/apache/flink/configuration/DeploymentOptions.java#L79-L85

Could you please show how these options are passed to Flink?

jsingh-yelp: @dolfinus Both `execution.attached=false` and `execution.target=remote` are set programmatically by Flink's CLI on the client side only.

On the remote client machine we run the same cmd for all flink services: `/opt/flink/bin/flink run --detached <job-options>`. Flink's CLI sets these in the runtime configuration:
- `--detached` → ProgramOptions sets `execution.attached=false` ([ProgramOptions.java:184](https://github.com/apache/flink/blob/6a8cb80a6043431a3f158ff6c2aea0a32099791a/flink-clients/src/main/java/org/apache/flink/client/cli/ProgramOptions.java#L184))
- [DefaultCLI.toConfiguration()](https://github.com/apache/flink/blob/master/flink-clients/src/main/java/org/apache/flink/client/cli/DefaultCLI.java#L71) unconditionally sets `execution.target=remote`  this method is called because DefaultCLI is the active CLI when no YARN/K8s CLI is detected.

`execution.job-status-changed-listeners` is different it's defined in `flink-conf` which is specific to each job/service deployment. Both the client and JM read this file, so both instantiate the listener. This gives operators control to enable/disable lineage at the service level without changing the job submission command.

Please let me know if this is what you meant to ask happy to clarify more.

client/java/src/main/java/io/openlineage/client/transports/KafkaTransport.java
mobuchowski: Let's get this as separate PR - I would merge that change fast.

integration/flink/shared/src/main/java/io/openlineage/flink/api/OpenLineageContext.java
mobuchowski: I'm a bit... detached from Flink integration now. Happy to accept change that you find consensus for.

integration/flink/shared/src/main/java/io/openlineage/flink/client/EventEmitter.java
```diff
@@ -47,6 +48,14 @@ public void emit(OpenLineage.RunEvent event) {
     }
   }
 
+  public void emit(OpenLineage.RunEvent event, Duration timeout) {
+    try {
+      client.emit(event, timeout);
```
jsingh-yelp, line 53: This function call depends on this change merging first: https://github.com/OpenLineage/OpenLineage/pull/4613

client/java/src/main/java/io/openlineage/client/transports/KafkaTransport.java
jsingh-yelp: Thanks @mobuchowski this is the separate PR as discussed: https://github.com/OpenLineage/OpenLineage/pull/4613

integration/flink/shared/src/main/java/io/openlineage/flink/api/OpenLineageContext.java
jsingh-yelp: @mobuchowski, @dolfinus I have added this as a separate flow behind the openlineage Flink config flag `enableDetachedJobTracking`. The default status quo listener path remains unchanged unless this flag is explicitly enabled.

If things looks okay we can add more description about this in the openlineage docs when to enable this flag and it's limitations. Thanks, let me know what you guys think.

integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageDetachedJobStatusChangedListener.java
```diff
@@ -0,0 +1,357 @@
...
+    JobIdentifier jobId = context.getJobId();
+    if (jobId == null || jobId.getFlinkJobId() == null) {
+      return null;
+    }
+
+    String flinkJobId = jobId.getFlinkJobId().toString();
+    return new FlinkJobDetailsFacet(flinkJobId);
+  }
+
+  private boolean loadJobId(JobStatusChangedEvent event) {
```
dolfinus: Is `boolean` vs `void` intentional?

integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageDetachedJobStatusChangedListener.java
```diff
@@ -0,0 +1,357 @@
...
+              } else {
+                log.warn("Unsupported event: {}", event.getClass());
+              }
+              return null;
+            });
+  }
+
+  private void onJobCreatedEvent(JobCreatedEvent event) {
+    if (!loadJobId(event)) {
+      log.warn("Skipping START event because runUuid could not be initialized: {}", event);
```
dolfinus, line 149: Does it mean that in detached mode no START events are be emitted?

jsingh-yelp: No, it should emit the `START` events. This will only fail if we can’t construct the ol event run ID needed for detached tracking for example, if the API fails and we don’t have the job’s start time.

integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageDetachedJobStatusChangedListener.java
jsingh-yelp: Yes, this is intentional. The boolean tells us whether the listener was successfully able to initialize the OpenLineage run id.
To make it clear I have split the functionality of flink job id and openlineage run id into separate functions.

dolfinus (approved):

mobuchowski (approved):

PR #4598, opened 2026-06-03
fix(spark): exclude job attempt timestamp tag when getting jobId in gcp dataproc facet

### LINKED ISSUE #4597: [BUG] Spark integration sometimes reports incorrect job ID from Dataproc

Spark integration

### Problem details

In `gcp_dataproc` facet, sometimes the job ID is reported as attempt_timestamp_xxxxxxxx instead of the actual job ID. Example:

```

"run":{
		"facets":{
			"gcp_dataproc":{
				<redacted>
				"jobId":"attempt_timestamp_1780021919533",
				"jobUuid":<redacted>,
			},

```

It should report the correct job id from Dataproc.

### How to reproduce

For cases when spark attempts the job multiple times, we get the attempt timestamp in the facet instead of the actual job ID

codelixir:
closes: #4597

In gcp_dataproc facet, sometimes the job ID is reported as `attempt_timestamp_xxxxxxxx` instead of the actual job ID. This change fixes this unwanted behaviour.

### Meaningful description
For cases when spark attempts the job multiple times, we get the attempt timestamp in the facet instead of the actual job ID.  Multiple attributes in the gcp dataproc facet, including job id, are constructed from Yarn tags. In case of multiple attempts, dataproc appends `dataproc_job_attempt_timestamp_xxx` to these tags, which are read by the current implementation of the parser, if it appears before the actual job id tag which we intend to read.

This change adds an overloaded private method which allows excluding certain prefix while reading properties from yarn tags. This is then used to get the Job ID.

Also updates the corresponding test to verify the intended behaviour.

### Checklist


integration/spark/vendor/gcp/src/main/java/io/openlineage/spark/agent/vendor/gcp/util/GCPUtils.java
```diff
@@ -237,6 +238,18 @@ private static Optional<String> getPropertyFromYarnTag(SparkContext context, Str
         .map(tag -> tag.substring(tagPrefix.length()));
   }
 
+  private static Optional<String> getPropertyFromYarnTag(
```
tnazarew, line 241: Wouldn't it be better to use regex matching on the tag name? Right now it's vulnerable to adding new tag like `dataproc_job_whatever`. 

If the id doesn't contain `_` then it should be easy enough.

tnazarew (commented):
Thanks for the PR! Left some suggestion :)

integration/spark/vendor/gcp/src/main/java/io/openlineage/spark/agent/vendor/gcp/util/GCPUtils.java
codelixir: Hi, the auto generated IDs from dataproc do not contain `_`, however, custom ID can be set as well. And that does accept underscores. So this might not always work.

```
$ gcloud dataproc jobs submit spark --cluster $CLUSTER_NAME  --id=test_job_id_underscore    --region=us-central1     --class org.apache.spark.examples.SparkPi     --jars file:///usr/lib/spark/examples/jars/spark-examples.jar   -- 100000
Job [test_job_id_underscore] submitted.
Waiting for job output...
```

codelixir: Is it worth doing something like this...

1. try regex match (for now assuming no underscores in job ID)

and if this fails (no matches)...

2. try what I currently have in my change (match first occurence of `dataproc_job_` which is not `dataproc_job_attempt_timestamp_`)

and if this fails again (in the edge case that the user manually set `attempt_timestamp_123` as the job ID, for whatever reason) then as a final fallback...

3. try matching the first occurence of `dataproc_job_` (what we have before my change)

tnazarew: ok, my whole point was to filter out everything we can except `dataproc_job_<alphanumerics>`

if user can set their custom job ids, then it make sense to specifically exclude `dataproc_job_attempt_timestamp_` and the regex approach is just wrong

tnazarew (approved):

PR #4611, opened 2026-06-08
spark: fix lakehouse detection mechanism

tnazarew:
fix lakehouse detection mechanism

### Meaningful description
first version of detection mechanism had used two incorrect properties, which worked in local reproduction but failed on the dataproc clusters

additionally the `*DatasetBuilder` logic relies heavily on catalog handlers so the extraction was also added to `V2SessionCatalogHandler`

### Checklist

### DIFF
--- a/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/CatalogDatasetFacetUtils.java
+++ b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/CatalogDatasetFacetUtils.java
@@ -86,15 +86,14 @@ public class CatalogDatasetFacetUtils {
   }
 
   public static boolean isHiveCatalog(SparkSession session, TableIdentifier identifier) {
-    return "hive"
-            .equals(session.sparkContext().getConf().get("spark.sql.catalogImplementation", ""))
+    return isHiveSupportEnabled(session.sparkContext())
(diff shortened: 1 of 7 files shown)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/CatalogDatasetFacetUtils.java
```diff
@@ -123,4 +122,8 @@ public static Optional<CatalogPlugin> getCatalogPlugin(
       return Optional.empty();
     }
   }
+
+  public static boolean isHiveSupportEnabled(SparkContext sc) {
+    return "hive".equals(sc.getConf().get("spark.sql.catalogImplementation", ""));
```
wslulciuc, line 127: _Minor:_ Do we need to make "hive" case insensitive?

wslulciuc (approved):
Minor comment, otherwise 👍

integration/spark/shared/src/main/java/io/openlineage/spark/agent/util/CatalogDatasetFacetUtils.java
tnazarew: we do check on that property in different places that way and I've never seen it in uppercase so I don't think so

PR #4612, opened 2026-06-08
Java: Fix port definition for JDBC IPv6 address

matveeysv:
Fix port extraction for JDBC IPv6 URLs with custom port

#### Problem:
When parsing JDBC URLs with IPv6 host addresses, the port extraction logic fails because colons in the IPv6 address interfered with detecting the actual port delimiter.

#### Example of incorrect behavior:

```
Input:  jdbc:clickhouse://[2001:db8::1]:8124
Output: clickhouse://[2001:db8::1]:8124:8123 (custom port + default port appended)
```

#### Solution:
Update the regex pattern to correctly handle IPv6 addresses.

### Checklist

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/utils/jdbc/OverridingJdbcExtractor.java
+++ b/client/java/src/main/java/io/openlineage/client/utils/jdbc/OverridingJdbcExtractor.java
@@ -16,7 +16,7 @@ public class OverridingJdbcExtractor extends GenericJdbcExtractor implements Jdb
   private String overrideScheme;
   private String defaultPort;
   private static Pattern HOST_PORT_FORMAT =
-      Pattern.compile("^(?<host>[\\[\\]\\w\\d.-]+):(?<port>\\d+)?");
+      Pattern.compile("^(?<host>\\[.*]|[^:\\]]+):(?<port>\\d+)?$");
 
   public OverridingJdbcExtractor(String overrideScheme) {
     this(overrideScheme, null);
--- a/client/java/src/test/java/io/openlineage/client/utils/jdbc/JdbcDatasetUtilsTestForClickHouse.java
+++ b/client/java/src/test/java/io/openlineage/client/utils/jdbc/JdbcDatasetUtilsTestForClickHouse.java
@@ -28,6 +28,12 @@ class JdbcDatasetUtilsTestForClickHouse {
                 "jdbc:clickhouse://test-host.com:9000", "schema.table1", new Properties()))
         .hasFieldOrPropertyWithValue("namespace", "clickhouse://test-host.com:9000")
         .hasFieldOrPropertyWithValue("name", "schema.table1");
+
+    assertThat(
+            JdbcDatasetUtils.getDatasetIdentifier(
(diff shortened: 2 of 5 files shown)


chuenchen309:
This has been sitting since June without a review, so here's an independent check: the fix is correct, and the one red check isn't caused by it.

**Mechanism, confirmed.** `appendDefaultPort` decides "does this authority already carry a port?" via `HOST_PORT_FORMAT.matcher(host).matches()`, and `matches()` requires the *whole* authority to match. On `main` the host group is `[\[\]\w\d.-]+`, which cannot consume `:`, so an IPv6 literal can never match end to end. `[2001:db8::1]:8124` therefore falls into the "no port" branch and gets the default appended — exactly as reported.

**Before/after, evaluated rather than reasoned about.** Both patterns run as whole-input matches over a spread of authorities, default port `8123`:

| authority | `main` | this PR |
|---|---|---|
| `[2001:db8::1]:8124` | `[2001:db8::1]:8124:8123` ❌ | `[2001:db8::1]:8124` ✅ |
| `[::1]:5432` | `[::1]:5432:8123` ❌ | `[::1]:5432` ✅ |
| `[3ffe:8311:...:10.203.31.9]:3307` | `...:3307:8123` ❌ | `...:3307` ✅ |
| `[2001:db8::1]` | `[2001:db8::1]:8123` ✅ | `[2001:db8::1]:8123` ✅ |
| `hostname:3307` | `hostname:3307` | `hostname:3307` |
| `hostname` | `hostname:8123` | `hostname:8123` |
| `host1:1,host2` | `host1:1,host2:8123` | `host1:1,host2:8123` |

The fourth row is the one that's easy to break while patching this: an IPv6 host *without* a port must still receive the default. It survives because the bracket alternative still requires a following `:`, and `[^:\]]+` can't span the closing bracket. The pre-existing `testGetDatasetIdentifierWithIPv6` cases in the ClickHouse/MySQL/Oracle/Postgres suites cover precisely that, and they pass on this branch — so the new assertions add the missing half of the matrix without regressing the half that already worked.

**The red `build-client-java` is unrelated.** 444 tests ran, 443 passed. Every `JdbcDatasetUtilsTestFor*` case passed, including all the new IPv6-with-port assertions. The single failure is:

```
Test smokeTestTestingEnvironment() FAILED
java.lang.AssertionError:
Expecting actual not to be null
  at io.openlineage.client.transports.HttpTransportSslContextTest.smokeTestTestingEnvironment(HttpTransportSslContextTest.java:106)
```

That test drives a MockServer TLS handshake and asserts on retrieved server log messages — nothing in it reaches JDBC URL parsing. A re-run should clear it; the branch is based on `33410b2b` (Jun 8), so a rebase would pick up two months of CI fixes at the same time.

**Still applies cleanly.** Diffing this branch against current `main` for the touched source file yields only the intended one-line pattern change, so `main` still carries the old regex and the fix hasn't gone stale.

*AI-assisted review. The behaviour table comes from evaluating both patterns directly as whole-input matches, the way `Matcher.matches()` does; the Java suite results are read from this branch's CircleCI log rather than a local run.*

mobuchowski (approved):

PR #4633, opened 2026-06-16
Use scanReportSupplier for building iceberg scan report facet if available.

fm100:

This PR fixes the missing scan report bug by using `scanReportSupplier` when available.

### Meaningful description

`OpenLineageMetricsReporter` is registered to the catalog by the `IcebergMetricsReporterInjector` which is `QueryPlanVisitor`. The problem is that the the actual iceberg scan happens before the visitor traverses the nodes, and when iceberg is sending the report to the configured MetricsReporter, there is no configured one and it becomes no-op. This caused the missing scan report from the inputFacets.
This PR uses `scanReportSupplier` directly from the facet builder if available, and for the compatibility, it falls back to the original implementation when `scanReportSupplier` is not available.

The bug can be reproducible with simple CTAS/RTAS query on iceberg catalog.

### Checklist


wslulciuc (approved):
👍

PR #4653, opened 2026-06-19
dbt model meta/config capture

mobuchowski:
dbt manifest nodes carry rich per-model metadata that the OpenLineage dbt integration was not using: the `config` - values like `materialized`, `access`, `owner`, and `group`) and the user-defined `meta` map (commonly used for ownership, PII flags, team labels, tiers, SLAs, etc.). 

None of this was being forwarded into emitted lineage events, so downstream consumers  had no visibility into how a model is materialized, who owns it, its access level, or any custom governance tags, even though the information was already sitting in the manifest dbt produces on every run. 

This change captures that metadata into a new `dbt_model` dataset facet (mirroring the dd-source Go implementation field-for-field, including the key name, so it round-trips correctly) and attaches it to datasets in both the legacy/local and structured-logs dbt processors, making per-model configuration and user metadata travel with the lineage events for both integration paths.

Capture dbt per-model config/meta values.

### Checklist

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/facets.py
+++ b/integration/common/src/openlineage/common/provider/dbt/facets.py
@@ -5,6 +5,7 @@
 import attr
 from openlineage.client.facet_v2 import (
     BaseFacet,
+    DatasetFacet,
     JobFacet,
     parent_run,
 )
@@ -58,6 +59,40 @@ class DbtRunRunFacet(BaseFacet):
     @staticmethod
     def _get_schema() -> str:
         return GITHUB_LOCATION + "dbt-run-run-facet.json"
+
+
+@attr.define
+class DbtModelConfig:
+    """The resolved dbt model configuration fields most relevant for data observability.
+
+    These come from the *resolved* ``config`` of a dbt manifest node (after applying
+    project-level defaults from ``dbt_project.yml``, per-model overrides, etc.).
+    """
+
+    materialized: str | None = attr.field(default=None)
+    access: str | None = attr.field(default=None)
+    owner: str | None = attr.field(default=None)
+    group: str | None = attr.field(default=None)
+
+
+@attr.define
+class DbtModelDatasetFacet(DatasetFacet):
+    """Dataset facet capturing dbt-specific per-model config and meta from the dbt manifest.
+
+    dbt manifest nodes carry two sources of rich per-model metadata:
+
+    1. ``config`` - the resolved dbt configuration. The most observability-relevant fields
+       are ``materialized``, ``access``, ``owner`` and ``group``.
+    2. ``meta`` - an arbitrary user-defined key/value map, commonly used for ownership,
+       PII flags, team labels, etc.
+    """
+
+    config: DbtModelConfig | None = attr.field(default=None)
+    meta: dict | None = attr.field(default=None)
+
+    @staticmethod
+    def _get_schema() -> str:
+        return GITHUB_LOCATION + "dbt-model-dataset-facet.json"
 
 
 @attr.define
--- a/integration/common/src/openlineage/common/provider/dbt/facets.py
+++ b/integration/common/src/openlineage/common/provider/dbt/facets.py
@@ -81,7 +81,7 @@ class DbtModelDatasetFacet(DatasetFacet):
 
     The most observability-relevant fields are ``materialized``, ``access``, ``owner`` and
     ``group``. The free-form ``meta`` map is emitted separately as dataset/run tags (source
-    ``dbt-meta``), not on this facet.
+    ``DBT_META``), not on this facet.
     """
 
     config: DbtModelConfig | None = attr.field(default=None)
--- a/integration/common/src/openlineage/common/provider/dbt/facets.py
+++ b/integration/common/src/openlineage/common/provider/dbt/facets.py
@@ -77,18 +77,14 @@ class DbtModelConfig:
 
 @attr.define
 class DbtModelDatasetFacet(DatasetFacet):
-    """Dataset facet capturing dbt-specific per-model config and meta from the dbt manifest.
+    """Dataset facet capturing the resolved dbt ``config`` of a manifest node.
 
-    dbt manifest nodes carry two sources of rich per-model metadata:
-
-    1. ``config`` - the resolved dbt configuration. The most observability-relevant fields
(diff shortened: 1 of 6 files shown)


integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -60,6 +61,40 @@ def _get_schema() -> str:
...
+    dbt manifest nodes carry two sources of rich per-model metadata:
+
+    1. ``config`` - the resolved dbt configuration. The most observability-relevant fields
+       are ``materialized``, ``access``, ``owner`` and ``group``.
+    2. ``meta`` - an arbitrary user-defined key/value map, commonly used for ownership,
+       PII flags, team labels, etc.
+    """
+
+    config: DbtModelConfig | None = attr.field(default=None)
+    meta: dict | None = attr.field(default=None)
```
kacpermuda: How is that different from tags? Should we just use `TagsDatasetFacet` to be more generic here and only keep dbt-specific metadata as custom facet?

integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -60,6 +61,40 @@ def _get_schema() -> str:
...
+class DbtModelConfig:
+    """The resolved dbt model configuration fields most relevant for data observability.
+
+    These come from the *resolved* ``config`` of a dbt manifest node (after applying
+    project-level defaults from ``dbt_project.yml``, per-model overrides, etc.).
+    """
+
+    materialized: str | None = attr.field(default=None)
+    access: str | None = attr.field(default=None)
+    owner: str | None = attr.field(default=None)
```
kacpermuda, line 74: Can we use generic OwnershipDatasetFacet for this instead?

integration/common/src/openlineage/common/provider/dbt/facets.py
kacpermuda: Hmm, I see that [meta](https://docs.getdbt.com/reference/resource-configs/meta?version=2.0&name=Fusion) is a separate thing from [tags](https://docs.getdbt.com/reference/resource-configs/tags?version=2.0&name=Fusion), so maybe it's not the best idea. Will leave here for discussion, but looks like we can probably put both in tags facet and use source=`meta/tags` to differentiate.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -834,6 +839,41 @@ def extract_dataset_data(
...
+        meta = node.metadata_node.get("meta") or {}
+        has_meta = bool(meta)
+
+        if not has_config and not has_meta:
+            return None
+
+        return DbtModelDatasetFacet(
+            config=model_config if has_config else None,
+            meta=meta if has_meta else None,
+        )
```
kacpermuda:
```suggestion
        meta = node.metadata_node.get("meta") or {}

        if not has_config and not meta:
            return None

        return DbtModelDatasetFacet(
            config=model_config if has_config else None,
            meta=meta or None,
        )
```

kacpermuda (commented):
Few nits, not sure if we should barely put what's in the manifest 1:1 into a custom facet, we know what that is, we have that context, so let's translate that into generic facets like ownership, tags etc.

integration/common/src/openlineage/common/provider/dbt/facets.py
mobuchowski: We already do, this is just capturing the data where it is.

kacpermuda: Maybe we should not duplicate it then? Easier to avoid possible differences in the future is to have it in one place only.

mobuchowski:
> Few nits, not sure if we should barely put what's in the manifest 1:1 into a custom facet, we know what that is, we have that context, so let's translate that into generic facets like ownership, tags etc.

We do when it makes sense (like ownership) but I want to avoid fragmentation on pulling _some_ data out of system specific facets while we still need that container anyway - because we don't have a good place for other ones, like materialization structure. 

When you want to consume dbt meta, or airflow dag run metadata, it's not a good idea to be forced to pull some data out of one place, and another from another place.

The advantage of system specific facets in this case is that we're tied only to what they are exposing - we don't have to keep standarized contract - so if dbt adds or removes something, we just adapt to it. 

So I'm advocating for

- when we have some container that contains data that is both standardized in core OL facets, and not standarized, duplicate the data in both specific facet and system specific one
- when data from that container fits fully into standardized facet - like tags - let's put it into standardized facet
- don't fragment existing source system data structure by individually pulling fields from it

integration/common/src/openlineage/common/provider/dbt/facets.py
kacpermuda: I think we're not using TagsDatasetFacet yet, so it might be a good place for it, if thi meta is per dataset. We only have tags on run level now, but I don't have enough experience to know if we can translate meta 1:1 into tags, I think if we keep source=meta it can work

kacpermuda:
Ok, I think as long as we expose the generic information coming from integration specific container as a standardized facet, I'm ok with keeping it in integration specific one as well. It still creates a risk that the two will diverge, but hopefully it's a low risk, since there is not much processing involved there.

integration/common/src/openlineage/common/provider/dbt/facets.py
mobuchowski: Which differences, if we're not modifying them?

integration/common/src/openlineage/common/provider/dbt/facets.py
mobuchowski: The tags aren't on dataset - but on dbt model, which is a job in our data model.

We can add them with source=`dbt-meta` I think.

integration/common/src/openlineage/common/provider/dbt/facets.py
kacpermuda: Yet 😄

integration/common/src/openlineage/common/provider/dbt/processor.py
mobuchowski: done

mobuchowski:
Updated to emit `meta` in tags with different source.

kacpermuda (approved):

PR #4662, opened 2026-06-24
spark: move CatalogHandler from spark3 to shared

arturowczarek:
Relocate the CatalogHandler interface to the shared module so lower modules can reference it without depending on spark3. Pure move plus import updates.

Relocate the CatalogHandler interface to the shared module 

### Meaningful description
We want to define new implementations of CatalogHandler outside spark3. We want it to be able to write UnityCatalogHandler that requires Spark 4 and doesn't fit spark3 module

### Checklist
- [X] AI was used in creating this PR (but I verified every line)


integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/plan/catalog/CatalogHandler.java
```diff
@@ -3,7 +3,7 @@
 /* SPDX-License-Identifier: Apache-2.0
 */
 
-package io.openlineage.spark3.agent.lifecycle.plan.catalog;
+package io.openlineage.spark.agent.lifecycle.plan.catalog;
```
pawel-big-lebowski, line 6: Let's run the full-tests. This may mean removing support for Spark 2, which basically I don't mind doing. I may be wrong, but I think that was a reason for putting this code under Spark3 initially so that shared compiles with Spark 2.

tnazarew (commented):
nice! in #3674 we have discussion about moving the definition of catalog handlers list to `Spark.*DatasetBuilderFactory`, maybe If you put it in this PR we could also move `CatalogUtils3` to shared module (and rename it to `CatalogUtils` 😄  ) or in separate small PR, idk

integration/spark/shared/src/test/java/io/openlineage/spark/agent/lifecycle/plan/catalog/CatalogUtilsTest.java
```diff
@@ -0,0 +1,162 @@
...
+    return context;
+  }
+
+  /**
+   * Mirrors a real context built by the agent's ContextFactory, where {@code datasetBuilderFactory}
+   * defaults to {@link DatasetBuilderFactory#EMPTY} (no contributed handlers).
+   */
+  private static OpenLineageContext emptyContext() {
+    OpenLineageContext context = mock(OpenLineageContext.class);
+    when(context.getDatasetBuilderFactory()).thenReturn(DatasetBuilderFactory.EMPTY);
```
tnazarew, line 64: I'm wondering about the `EMPTY`, AFAIK it's only used in tests here and we can get emptyContext using `contextWithHandlers(new CatalogHandler[]{})`, am I wrong?

in general I would not put stuff in the main part of the code for the sake of tests if I can avoid it

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/DefaultCatalogHandlers.java
```diff
@@ -0,0 +1,32 @@
...
+import io.openlineage.spark3.agent.lifecycle.plan.catalog.V2SessionCatalogHandler;
+import io.openlineage.spark3.agent.lifecycle.plan.catalog.iceberg.IcebergHandler;
+import java.util.List;
+
+/** Catalog handlers available to every Spark 3.x / 4.x version. */
+final class DefaultCatalogHandlers {
+  private DefaultCatalogHandlers() {}
+
+  static List<CatalogHandler> list(OpenLineageContext context) {
+    return ImmutableList.of(
```
tnazarew, line 24: just a thought, maybe returning list builder would be a good idea, then we can just have the 
```
DefaultCatalogHandlers.builder()
  .add()
  .add()
  .build();
```

this might be over-engineering but I really like to `Default`, I would even go as far as getting `DefaultSparkDatasetBuilders` with methods `input`, `output`, `catalogHandlers` as they are mostly the same.

it would be good for readability but it doesn't have to be in this PR :)

tnazarew (commented):
left two suggestions, otherwise LGTM

integration/spark/shared/src/test/java/io/openlineage/spark/agent/lifecycle/plan/catalog/CatalogUtilsTest.java
arturowczarek: The `DatasetBuilderFactory` should be mocked in 45 files. If I use `EMPTY` only in tests it affects also Gradle dependencies (it should be test-only implementation). The change becomes huge. It is better to accept the default value as is.

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/DefaultCatalogHandlers.java
arturowczarek: Indeed, it's better to do it separate because the change is already quite big

tnazarew (approved):

PR #4679, opened 2026-07-01
fix(great-expectations): preserve datasource url ports

### LINKED ISSUE #4678: Great Expectations datasource URI should preserve port when stripping credentials
﻿## Problem

`OpenLineageValidationAction._source()` removes credentials from a parsed datasource URL by replacing `netloc` with `url.hostname`.

That strips credentials, but it also drops the port. For example, a datasource URL like `postgresql://user:pass@localhost:5432/db` is converted to `postgresql://localhost/db`, losing `:5432` in the emitted datasource URI.

## Expected behavior

The sanitized datasource URL should remove credentials while preserving the host and port, for example `postgresql://localhost:5432/db`.

## Proposed fix

Build the sanitized netloc from hostname and port instead of hostname alone, and add regression coverage for credential stripping with port preservation.

hcthakur2004:
﻿## Summary

- preserve datasource URL ports when Great Expectations strips credentials
- add regression coverage for credential stripping with a host and port

## Issue

Fixes #4678

## Tests

- `uv run pytest tests/great_expectations/test_great_expectations_validation_action.py -k "source_strips_credentials"`
- `uv run ruff check src/openlineage/common/provider/great_expectations/action.py tests/great_expectations/test_great_expectations_validation_action.py`

### DIFF
--- a/integration/common/src/openlineage/common/provider/great_expectations/action.py
+++ b/integration/common/src/openlineage/common/provider/great_expectations/action.py
@@ -426,11 +426,16 @@ class OpenLineageValidationAction(ValidationAction):
         if url.scheme == "bigquery":
             return Source(scheme="bigquery", connection_url="bigquery")
 
+        hostname = url.hostname or ""
+        authority = f"[{hostname}]" if ":" in hostname else hostname
+        if authority and url.port is not None:
+            authority = f"{authority}:{url.port}"
+
         return Source(
             scheme=url.scheme,
             authority=url.hostname,
             # Remove credentials from the URL if present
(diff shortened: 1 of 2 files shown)


mobuchowski (approved):

hcthakur2004:
Thanks @mobuchowski for reviewing and merging this. I’ve been contributing mostly around the Python client and integration/common areas, and I’d like to keep focusing on work that is useful to OpenLineage rather than picking random small fixes. If there are areas where maintainers would especially value help, I’d appreciate any pointers. I’m also interested in future LFX mentorship opportunities around OpenLineage, so I’m trying to build deeper context in the parts of the project where help is most needed.

Issue #4412, opened 2026-03-25
[PROPOSAL] Add Job Type to ParentRunFacet + Standardized Parent Context Propagation


mobuchowski:
### 1. Missing Job Type in ParentRunFacet

The `ParentRunFacet` currently carries only identification fields for the parent and root jobs: `namespace`, `name`, and `runId`. It does not include the parent job's *type* — the information that `JobTypeJobFacet` provides (`processingType`, `integration`, `jobType`).

This creates a real problem for consumers. When a Spark job emits lineage with a `ParentRunFacet` pointing to an Airflow task, the consumer cannot tell from that single event whether the parent is an Airflow DAG, a Spark APPLICATION, or a dbt invocation. To determine this, consumers must:

1. Find the parent job's own events by matching `namespace + name`
2. Look up its `JobTypeJobFacet`
3. Join across potentially different event streams or storage backends

This cross-event join is expensive, fragile, and sometimes impossible due to race conditions. For example, with short lived tasks, the events from child job can be received and processed before the parent job events - the resolution is impossible then.

### 2. Fragmented Parent Info Propagation

When a scheduler (e.g., Airflow) launches a child process (e.g., dbt or Spark), it should pass parent run context to let customers connect those jobs. Today this is done through multiple system-specific mechanisms:

| Integration | Mechanism | Fields | Limitations |
|---|---|---|---|
| **dbt** | `OPENLINEAGE_PARENT_ID` env var | `namespace/name/runId` (slash-delimited) | No root parent support; slashes in names break parsing |
| **dbt** | `OPENLINEAGE_ROOT_PARENT_ID` env var | `namespace/name/runId` (slash-delimited) | Same slash limitation; separate var for root |
| **Spark** | `spark.openlineage.*` config properties | 6 separate fields (`parentJobName`, `parentJobNamespace`, `parentRunId`, `rootParentJobName`, `rootParentJobNamespace`, `rootParentRunId`) | Spark-specific; no standard format |
| **Flink** | Not supported | — | No parent propagation |

This fragmentation means:
- Each integration re-invents parsing logic
- Adding new fields (like job type) requires updating every integration's config separately — a new field means a new env var for dbt and a new Spark config key for Spark, with no shared structure
- There's no portable way for a scheduler to pass context to an arbitrary child process

### Change 1: Add `jobType` to ParentRunFacet (spec change)

[...]

PR #4680, opened 2026-07-02
feat(spec): Extend ParentRunFacet with parent and root facets

kacpermuda:
Add an optional `facets` field to `ParentRunFacet`'s `job`/`run` (and `root.job`/`root.run`) objects, letting producers forward a selected subset of parent/root facets into the child event.

### Meaningful description
`ParentRunFacet` only carries `namespace`/`name`/`runId` for the parent and root job, so consumers who need more (e.g. job type) must cross-reference the parent's own event - which is unreliable when child events can be processed before the parent's, due to event-ordering races.

[#4412](https://github.com/OpenLineage/OpenLineage/issues/4412) originally proposed solving this narrowly by adding a hard-coded `jobType` field. As discussed in [this comment](https://github.com/OpenLineage/OpenLineage/issues/4412#issuecomment-4156889748), hard-coding one field doesn't generalize — a future need would require another special-cased field, and so on. Instead, this PR adds a generic, optional `facets` map to `job`, `run`, `RootJob`, and `RootRun` in `ParentRunFacet.json`. This lets a producer forward a small, deliberate set of facets (e.g. `JobTypeJobFacet`, `ProcessingEngineRunFacet`) on an as-needed basis, without OpenLineage needing to special-case every future field. It's documented as a convenience, not a replacement for resolving the parent's own event.

related: #4412

### Checklist


website/docs/spec/facets/run-facets/parent_run.md
```diff
@@ -35,7 +47,16 @@ Example:
             "namespace": "another-namespace"
           },
           "run": {
-            "runId": "f1234567-4f4f-1a1a-2b2b-abcdef123456"
+            "runId": "f1234567-4f4f-1a1a-2b2b-abcdef123456",
+            "facets": {
+                "tags": {
+                    "tags": [{
+                        "key": "containerId",
+                        "value": "08047900167b20192704669334768182f825281777f540",
```
dolfinus: nit: tags are usually user-defined. Maybe we should use `externalQuery` as an example

kacpermuda: Agree, changed to processing engine run facet, which is usually auto injected.

dolfinus (approved):

mobuchowski:
This updates the code, but does not touch on how the information should propagate. The original proposal discussed `OPENLINEAGE_CONTEXT` env to be able to handle this and any future additions. 

IMO we need this, if we default to passing each value separately, it will become impossible to use in practice.

kacpermuda:
> This updates the code, but does not touch on how the information should propagate. The original proposal discussed `OPENLINEAGE_CONTEXT` env to be able to handle this and any future additions.
> 
> IMO we need this, if we default to passing each value separately, it will become impossible to use in practice.

I agree, it will be done in a separate PR to break down into pieces and avoid a huge PR that is hard to review. With this being implemented first, we already have the spec allowing it, but no implementation, which will not break anything and we can work on the other piece in the meantime.

mobuchowski (approved):

kacpermuda:
I see reviews from two folks, would appreciate someone else take a look as well since it's a spec change and an important facet. @jakub-moravec maybe ?

website/docs/spec/facets/run-facets/parent_run.md
```diff
@@ -14,6 +14,11 @@ optionally contains information about the root job contained in the root field.
 The root job represents the initial operation that started the whole chain of parent-child jobs - for example, the 
 Airflow DAG execution that eventually spawned Airflow tasks which then spawned Spark jobs.
 
+The `job` and `run` objects, both at the parent level and at the root level, can also optionally carry a `facets` field.
+This lets a producer forward a selected subset of facets of that parent/root job or run into the child event, for convenience -
+for example, when the child cannot reliably resolve the parent's own event to look them up itself.
+It is not meant to be a full copy of the parent event: resolving the parent's own event remains the recommended way to get full information.
```
kacpermuda: TODO: Add some paragraph about that we should also underline that it's not designed to replace the information in run parent. You should not use this as a source of truth if they ever differ.

harels (approved):

website/docs/spec/facets/run-facets/parent_run.md
kacpermuda: Done, added a separate commit with that change.

julienledem (approved):
Nice idea!

kacpermuda:
Thanks everyone for reviews, I invite you to also take a look at #4682 which is a followup PR adding a generic mechanism to insert all that new information.

Issue #4683, opened 2026-07-04
[BUG] Python: DatadogTransport starts an unused, never-closed busy-polling worker thread per instance; failed STARTs leak pending completion events


gang-zh:
### OpenLineage integration/client

Python client

### OpenLineage version

`openlineage-python==1.47.1` (all findings re-verified against `main`); `openlineage-integration-common==1.47.1`, `openlineage-sql==1.47.1`

### Technology and package versions

Python 3.12 (production) / 3.14 (repro); Apache Airflow 2.11.2; `apache-airflow-providers-openlineage==2.16.0`

### Environment configuration

Composite transport with two legs: a plain `http` leg (internal proxy) and a `datadog` leg:

```
OPENLINEAGE__TRANSPORT__TYPE=composite
OPENLINEAGE__TRANSPORT__TRANSPORTS__TROVE={"type": "http", ...}
OPENLINEAGE__TRANSPORT__TRANSPORTS__DATADOG__TYPE=datadog
DD_SITE=datadoghq.com / DD_API_KEY=***
# later added during incident mitigation:
OPENLINEAGE__TRANSPORT__TRANSPORTS__DATADOG__ASYNC_TRANSPORT_RULES={"*": {"*": true}}
OPENLINEAGE__TRANSPORT__TRANSPORTS__DATADOG__RETRY={"total": 1, ...}
```

### Deployment details

Astronomer (Astro Hosted), Astro Runtime 13.7.0, KubernetesExecutor, ~50 hourly DAGs. Enabling the `datadog` leg (Datadog Jobs Monitoring) made the Airflow scheduler CPU climb steadily until pinned at its limit within ~6 hours, recovering only on scheduler restart. Full incident analysis: apache/airflow#69284.

### Problem details

All verified against openlineage-python 1.47.1 source:

**1. `DatadogTransport` starts an async worker thread even when nothing routes to it.**
`DatadogTransport.__init__` unconditionally constructs `AsyncHttpTransport` ([datadog.py#L115](https://github.com/OpenLineage/OpenLineage/blob/main/client/python/src/openlineage/client/transport/datadog.py)), whose `__init__` immediately starts a daemon worker thread (`worker_thread.start()` in async_http.py). With the default `async_transport_rules = {"dbt": {"*": true}}`, no Airflow event ever routes async — yet every instance still spawns the thread.

**2. The worker thread busy-polls when idle.**
[...]

PR #4694, opened 2026-07-06
fix several async transport issues

mobuchowski:
### Summary

This fixes 3 bugs in the Datadog transport that caused Airflow scheduler CPU to climb until it hit its limit (issue #4683). The transport built an async HTTP worker thread for every instance, even when nothing routed to it, and that thread polled constantly even when idle. A separate bug also leaked memory when a run's START event failed.

### What was wrong

DatadogTransport built an AsyncHttpTransport on every construction, whether or not any event ever used it. Building AsyncHttpTransport starts a daemon thread. With the default routing rules, most events never use the async path, so most of these threads did nothing but consume CPU.

Each idle thread polled its queue every 0.01 seconds, forever. On a production scheduler that built one client per event, this added up: 50 instances used 29.5% CPU while idle, 400 instances used 167.6%.

A third bug caused a memory leak: when a run's START event failed after all retries, any completion events already queued for that run stayed in memory. Nothing released them.

### What changed

We made 3 changes to client/python/src/openlineage/client/transport/:

- datadog.py now builds the async transport only when an event actually needs it, not on every construction. Most deployments will never build it at all.
- async_http.py now backs off when idle. The worker thread sleeps for 0.01 seconds after the queue is empty, then doubles that wait each time it stays empty, up to 1 second. It resets to 0.01 seconds as soon as new work arrives.
- async_http.py now clears out a run's pending completion events if its START event fails after all retries, and logs a warning when it does. Before this change, those events stayed in memory indefinitely.

### Checklist

### DIFF
--- a/client/python/src/openlineage/client/transport/async_http.py
+++ b/client/python/src/openlineage/client/transport/async_http.py
@@ -275,6 +275,8 @@ class AsyncHttpTransport(Transport):
             transport=transport,
         ) as client:
             last_processed_time = time.time()
+            idle_sleep = 0.01
+            max_idle_sleep = 1.0
             while not _should_exit():
                 processed_items = False
                 while not self.event_queue.empty() and len(active_tasks) < self.max_concurrent:
@@ -297,6 +299,7 @@ class AsyncHttpTransport(Transport):
                 # Update last processed time if we processed items
                 if processed_items:
                     last_processed_time = time.time()
+                    idle_sleep = 0.01
 
                 if active_tasks:
                     done, _ = await asyncio.wait(
@@ -346,7 +349,10 @@ class AsyncHttpTransport(Transport):
                             self.event_stats["failed"],
                         )
                         last_processed_time = current_time
-                    await asyncio.sleep(0.01)
+                    # Back off exponentially while idle instead of busy-polling at a fixed
+                    # interval forever; reset to the fast interval as soon as work arrives.
+                    await asyncio.sleep(idle_sleep)
+                    idle_sleep = min(idle_sleep * 2, max_idle_sleep)
 
             log.debug(
                 "AsyncHttpTransport event loop worker completed. "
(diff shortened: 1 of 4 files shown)


client/python/src/openlineage/client/transport/async_http.py
```diff
@@ -475,6 +481,23 @@ def _mark_failed(self, request: Request) -> None:
...
+                        self.event_stats["failed"] += 1
+                        if self.events.get(pending_request.event_id) == "pending":
+                            del self.events[pending_request.event_id]
+                    log.warning(
+                        "START event for run %s failed after exhausting retries; dropping %d "
+                        "pending completion event(s) that were waiting for it.",
+                        request.run_id,
+                        len(dropped),
+                    )
+
```
kacpermuda, line 503: Can't we just try to send this complete event instead of dropping it? So we keep the current behavior, we wait for the start, but if the start fails, we just move it to the usual queue, try to send it and then if it fails, we drop.

Some consumers probably still find value in complete event even without a start. Maybe start was not delivered due to some transient issue and complete will be delivered just fine. We can give it the same number of retries as any other events, and then drop it if it fails.

client/python/src/openlineage/client/transport/datadog.py
```diff
@@ -117,21 +118,31 @@ def __init__(self, config: DatadogConfig) -> None:
                 "max_concurrent_requests": config.max_concurrent_requests,
             }
         )
-        async_http_config = AsyncHttpConfig.from_dict(async_config)
-        self.async_http = AsyncHttpTransport(async_http_config)
+        self._async_http_config = AsyncHttpConfig.from_dict(async_config)
+        # AsyncHttpTransport spawns a worker thread on construction, so it's built lazily,
+        # on the first event that actually routes to it, instead of unconditionally here.
+        self.async_http: AsyncHttpTransport | None = None
+        self._async_http_lock = threading.Lock()
```
kacpermuda, line 125: Just to make sure, this will not interfere with line 215 in AsynHttpTransport that also does `self.event_lock = threading.Lock()` ?

client/python/src/openlineage/client/transport/async_http.py
mobuchowski: Fixed, now we'll release those events instead of dropping them.

client/python/src/openlineage/client/transport/datadog.py
mobuchowski: Yes - it's totally separate lock, to make sure we're not double-creating the `async_http` instance

kacpermuda (approved):

PR #4704, opened 2026-07-13
doc: document recommended namespace prefixes for jobs

jakub-moravec:
This is a small documentation-only change that recommends what job namespace prefixes to use for existing integrations to ideally standardize the prefixes over time and simplify processing for consumers (allowing them to understand the technology of the job from the prefix, the same way as is the norm for datasets).

Proposing namespace prefixes for jobs

### Checklist

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -36,7 +36,7 @@
     "lodash": "^4.18.1",
     "lunr": "^2.3.9",
     "motion": "^12.42.2",
-    "postcss": "^8.5.19",
+    "postcss": "^8.5.20",
     "prism-react-renderer": "^2.4.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -7032,7 +7032,7 @@ __metadata:
     lodash: "npm:^4.18.1"
     lunr: "npm:^2.3.9"
     motion: "npm:^12.42.2"
-    postcss: "npm:^8.5.19"
+    postcss: "npm:^8.5.20"
     prism-react-renderer: "npm:^2.4.0"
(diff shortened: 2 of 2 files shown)


harels:
great addition! did you confirm that these names are commonly used in the existing integrations?

jakub-moravec:
@harels TBH I've mostly seen namespaces like "default", etc. Happy to use this ticket to collect common practices first, if any exist

harels:
> @harels TBH I've mostly seen namespaces like "default", etc. Happy to use this ticket to collect common practices first, if any exist

Thanks @jakub-moravec, makes sense to me. Maybe we extend an invite to folks on the OpenLineage slack to chime in with some best practices for platforms they maintain.

dolfinus:
In my company we use namespaces which reflect real locations where ETL jobs are executed:
* https://some.domain.name (Airflow, Flink, Debezium)
* yarn://cluster-name (Spark, Hive)
* k8s://cluster-name (Spark)
* local://host.name (Spark)

Ideally it should be an URL which points to web UI, like Airflow or Flink have. Job type is already a part of the dedicated facet, so there is no need to create artificial urls using job type as scheme.

jakub-moravec:
@dolfinus I understand the reasoning, but it brings a big inconsistency bw job and dataset prefixes. JobTypeFacet is unreliable as it is, of course, optional.

dolfinus:
Namespace is user-defined, so it is more unreliable, IMHO

jakub-moravec:
That is a fair point, but I would argue that mostly because no best practices have been defined at this moment. The same argument could be used for dataset namespaces (putting aside that individual producers don't make them configurable), but there the spec implies deterministic prefixes should be used for the same purpose.

PR #4706, opened 2026-07-14
Fix ruff import ordering violations in great_expectations action.py and facets.py.

maitraymukeshkumarmodi-aiml:
`action.py` and `facets.py` in the `great_expectations` integration had import 
ordering violations that caused the ruff-lint pre-commit hook to fail on every 
PR touching this folder. 

The fix reorders imports to follow ruff's isort rules: stdlib → third-party 
(great_expectations) → first-party (openlineage), with blank lines between each 
group. No functional changes.


kacpermuda:
Will fix in #4707

PR #4716, opened 2026-07-15
dbt: emit incremental-strategy config and run-wide full_refresh in OpenLineage facets

himakolavennu:
## Problem

The `dbt_model` dataset facet captures per-model `config`/`meta`, but nothing about how an **incremental** model rebuilds data. Lineage consumers therefore can't tell whether a run rebuilt the whole table or just an incremental slice — which matters for scoping downstream work (e.g. data-quality checks) to what actually changed.

## Description

Adds incremental rebuild info to the dbt facets, in two parts:

**Per-model** — a new `DbtIncrementalConfig` on the `dbt_model` dataset facet, populated for `materialized == "incremental"` models (its presence marks the model incremental):
- `strategy`, `unique_key`, `incremental_predicates` (with the Spark `predicates` alias), `on_schema_change`
- microbatch params (`event_time`, `batch_size`, `begin`, `lookback`), gated to the `microbatch` strategy so a resolved default `lookback` isn't emitted on merge/insert_overwrite models
- `partition_by` (`DbtPartitionBy`: BigQuery `{field, data_type, granularity}` object, or Spark column list)
- a per-model `full_refresh` config override

**Run-wide** — a `full_refresh` field on `DbtRunRunFacet`, capturing the invocation's `--full-refresh` flag:
- `local` / `cloud`: read from `run_results.json` args
- `structured_logs`: detected from the dbt command line (no `run_results.json` on that path)

Unknown fields are ignored by existing consumers, so the change is additive.

## How was this patch tested?

Added unit tests in `test_processor.py` (per-model config extraction, `full_refresh` from args) and `structured_logs/test_structured_logs.py` (CLI flag detection, plus an end-to-end assertion that the flag reaches the emitted `DbtRunRunFacet`). Full `integration/common/tests/dbt/` suite passes; ruff check + format clean.

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/facets.py
+++ b/integration/common/src/openlineage/common/provider/dbt/facets.py
@@ -55,10 +55,51 @@ class DbtRunRunFacet(BaseFacet):
     project_version: str | None = attr.field(default=None)
     profile_name: str | None = attr.field(default=None)
     account_id: str | None = attr.field(default=None)
+    # Run-wide --full-refresh flag for the invocation; None when it wasn't set.
+    full_refresh: bool | None = attr.field(default=None)
 
     @staticmethod
     def _get_schema() -> str:
         return GITHUB_LOCATION + "dbt-run-run-facet.json"
+
+
+@attr.define
+class DbtPartitionBy:
+    """A dbt ``partition_by`` config.
+
+    BigQuery expresses it as an object (``field``/``data_type``/``granularity``); Spark and
+    other adapters express it as a single column or a list of columns (``columns``).
+    Integer-range and ingestion-time partitioning details beyond these keys are not captured.
+    """
+
+    field: str | None = attr.field(default=None)
+    data_type: str | None = attr.field(default=None)
+    granularity: str | None = attr.field(default=None)
+    columns: list[str] | None = attr.field(default=None)
+
+
+@attr.define
+class DbtIncrementalConfig:
+    """How an incremental dbt model rebuilds data, from its resolved ``config``.
+
+    Populated only for ``materialized == "incremental"``; its presence marks the model
+    incremental even when no individual field is set.
+    """
+
+    # A None strategy means the adapter default; dbt omits it from the manifest.
+    strategy: str | None = attr.field(default=None)
+    unique_key: list[str] | None = attr.field(default=None)
+    # Config-declared SQL filters that bound the rows the merge scans
+    # (dbt ``incremental_predicates``, or the Spark ``predicates`` alias).
+    incremental_predicates: list[str] | None = attr.field(default=None)
+    on_schema_change: str | None = attr.field(default=None)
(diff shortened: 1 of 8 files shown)


integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -55,12 +55,53 @@ class DbtRunRunFacet(BaseFacet):
...
+    # Run-wide --full-refresh flag for the invocation; None when it wasn't set.
+    full_refresh: bool | None = attr.field(default=None)
 
     @staticmethod
     def _get_schema() -> str:
         return GITHUB_LOCATION + "dbt-run-run-facet.json"
 
 
+@attr.define
+class DbtPartitionBy:
```
kacpermuda, line 67: Can we add some arg description and examples. Also, I feel like there is a clear division here where we will have one combination of fields present for specific adapter, and different for others. Should we also include that deciding factor here? Or maybe make it clearer which ones should be checked together?

integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -55,12 +55,53 @@ class DbtRunRunFacet(BaseFacet):
...
+
+@attr.define
+class DbtIncrementalConfig:
+    """How an incremental dbt model rebuilds data, from its resolved ``config``.
+
+    Populated only for ``materialized == "incremental"``; its presence marks the model
+    incremental even when no individual field is set.
+    """
+
+    # A None strategy means the adapter default; dbt omits it from the manifest.
```
kacpermuda: Let's move all these inline comments to docstring, and include proper arg description and examples for all.

integration/common/src/openlineage/common/provider/dbt/facets.py
```diff
@@ -55,12 +55,53 @@ class DbtRunRunFacet(BaseFacet):
...
+    strategy: str | None = attr.field(default=None)
+    unique_key: list[str] | None = attr.field(default=None)
+    # Config-declared SQL filters that bound the rows the merge scans
+    # (dbt ``incremental_predicates``, or the Spark ``predicates`` alias).
+    incremental_predicates: list[str] | None = attr.field(default=None)
+    on_schema_change: str | None = attr.field(default=None)
+    event_time: str | None = attr.field(default=None)
+    batch_size: str | None = attr.field(default=None)
+    begin: str | None = attr.field(default=None)
+    lookback: int | None = attr.field(default=None)
```
kacpermuda, line 136: If these are only present for specific strategy, should we also make it clearer with docstring or separate config class?

kacpermuda (commented):
Leaving a few comments in places that may be confusing at first, maybe we can improve there - from the consumer perspective, it should be clear which combination of fields I need to read for specific mode / strategy etc.

mobuchowski (approved):

PR #4725, opened 2026-07-17
fix(common): let the ~/.dbt profiles directory fallback run

chuenchen309:

Let the `~/.dbt/` profiles-directory fallback actually run — it was unreachable because the working directory was returned unconditionally.

### Meaningful description

`get_dbt_profiles_dir` computes `default_dir` and then throws it away:

```python
from_command = parse_single_arg(command, ["--profiles-dir"])
from_env_var = os.getenv("DBT_PROFILES_DIR")
default_dir = "~/.dbt/"
current_working_directory = os.getcwd()
return from_command or from_env_var or current_working_directory or default_dir
```

`os.getcwd()` never returns an empty string, so the last term is unreachable:

```
cwd holds no profiles.yml   ->  '/tmp/tmpf48x1yxe'    <- never ~/.dbt
--profiles-dir given        ->  '/from/command'        <- fine
DBT_PROFILES_DIR set        ->  '/from/env'            <- fine
```

The first two terms do work, which is why this reads as the `or` chain's ordering rather than a deliberate decision to always use the cwd. The docstring points at [dbt's own rules](https://docs.getdbt.com/docs/core/connect-data-platform/connection-profiles#advanced-customizing-a-profile-directory), and dbt uses the working directory only when it actually holds a `profiles.yml` — otherwise `~/.dbt`.

### Why this matters for the case the fallback was added for

#4320 wired this up to *"find `profiles.yml` in a default fashion outlined by dbt"*, closing #4253 — dbt Fusion, whose `run_results.json` has no `profiles_dir` key. `local.py` only reaches this fallback when that key is missing, which for classic dbt it isn't, so the shadowing hides the problem there. For Fusion it isn't shadowed: a user with the standard `~/.dbt/profiles.yml` layout, running from anywhere else, still gets a `FileNotFoundError` even though dbt itself ran fine. `structured_logs.py` calls it unconditionally.

### The expanduser detail

`default_dir` is expanded at the source rather than at the call site, because `structured_logs.py` wraps the result in `os.path.expanduser` but `local.py` does not — returning a literal `"~"` would open nothing there. (That existing `expanduser` is also a hint that `~`-prefixed values were expected to come out of this function.)

The same function is duplicated in `local.py`; both are updated.

### Verification

- New `TestGetDbtProfilesDir` pins all four precedence rules. Three pass on `main` (the flag, the env var, and cwd-when-it-holds-a-profile) — they're the control; only the `~/.dbt` one fails, and passes with the fix.
- `tests/dbt`: 7 failed, 176 passed — the 7 are **identical to baseline** on unmodified `main` with the same command (172 passed there; the +4 here are this PR's tests).
- `ruff check` clean.

### Checklist

To be precise about the extent, since AGENTS.md asks: the bug was found, the fix written, the repro run, and this description written by an AI coding agent (Claude Code) running on this account — the commit carries `Co-Authored-By: Claude <noreply@anthropic.com>` per AGENTS.md and is signed off for DCO. Every figure above came from running the code, not reading it, but the agent ran it, not a person. I haven't tested against a real dbt Fusion install — that part rests on #4253's description of its `run_results.json`.

### DIFF
--- a/integration/common/src/openlineage/common/provider/dbt/local.py
+++ b/integration/common/src/openlineage/common/provider/dbt/local.py
@@ -289,8 +289,11 @@ class DbtLocalArtifactProcessor(DbtArtifactProcessor):
         """
         from_command = parse_single_arg(command, ["--profiles-dir"])
         from_env_var = os.getenv("DBT_PROFILES_DIR")
-        default_dir = "~/.dbt/"
-        current_working_directory = os.getcwd()
+        default_dir = os.path.expanduser("~/.dbt/")
+        # dbt only falls back to the working directory when it actually holds a
+        # profiles.yml, otherwise it uses default_dir.
+        cwd = os.getcwd()
+        current_working_directory = cwd if os.path.exists(os.path.join(cwd, "profiles.yml")) else None
         return from_command or from_env_var or current_working_directory or default_dir
 
     def dbt_run_run_facet(self) -> dict[str, DbtRunRunFacet]:
--- a/integration/common/src/openlineage/common/provider/dbt/utils.py
+++ b/integration/common/src/openlineage/common/provider/dbt/utils.py
@@ -84,8 +84,11 @@ def get_dbt_profiles_dir(command: list[str]) -> str:
     """
     from_command = parse_single_arg(command, ["--profiles-dir"])
     from_env_var = os.getenv("DBT_PROFILES_DIR")
-    default_dir = "~/.dbt/"
-    current_working_directory = os.getcwd()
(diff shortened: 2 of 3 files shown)


kacpermuda (approved):

kacpermuda:
@chuenchen309 Please fix the pre-commit whenever you have time so that we can merge this. Thanks for spotting this error

chuenchen309:
Done — that was ruff-format wanting two blank lines before `TestGetDbtProfilesDir`; pushed the reformat (no behavior change), so run-pre-commit should be green now. Thanks!

PR #4727, opened 2026-07-17
docs: improve dbt integration documentation

### LINKED ISSUE #4721: [DOCUMENTATION] Improve DBT integration docs (e.g. structured logs)
## Description

The current [dbt integration documentation](https://openlineage.io/docs/integrations/dbt) focuses almost exclusively on the `dbt-ol` wrapper approach. However, the OpenLineage dbt integration has evolved beyond the wrapper script and now supports multiple approaches for collecting lineage, each with different capabilities and trade-offs.

At the moment, there is little or no documentation describing these approaches or when to use them. It would also be useful to include examples of where each approach is used. This would make it much easier for users to understand which integration they are using and what capabilities to expect.

## Proposal

Expand the dbt integration documentation to describe the different ways OpenLineage can integrate with dbt.

For example:

### 1. dbt wrapper (`dbt-ol`)

The existing documented approach.

- Wrapper around `dbt`
- Reads `manifest.json`, `run_results.json`, `catalog.json`
- Emits events after the command completes
- Good for standalone dbt execution

### 2. Artifact processor

Describe the artifact processor used by other integrations.

For example Airflow operators

Explain:

- how it works,
- what metadata it uses,
- current capabilities and limitations,
- relationship to the `dbt-ol` wrapper.

### 3. Structured log processor

Describe the structured log integration.

For example:

- consumes dbt structured log events,
- emits events while dbt is running,
- supports richer event hierarchies (command → node → SQL query),
- can observe SQL query execution that is not available from artifacts alone.

## Benefits

- Better discoverability of existing functionality.
- Clear explanation of the available integration options.
- Easier comparison of capabilities and trade-offs.
- Better guidance for users integrating dbt with orchestrators such as Airflow.
- A foundation for documenting future improvements (e.g. invocation-level events, SQL query events, richer event hierarchies).

maitraymukeshkumarmodi-aiml:
                                                                                                                                     
docs(dbt): expand documentation to cover artifact and structured log processors, closes #4721                                                                                                                     
                                                                                                                                     
### Meaningful description:                                                                                                      
                                                                                                                                     
This PR expands the dbt integration documentation to cover the three main lineage collection approaches: the dbt CLI wrappe (`dbt-ol`), the artifact processor (used by orchestrators like Airflow Cosmos), and the structured log processor (real-time streaming). It explains their mechanisms, metadata usage, capabilities/limitations, and relationship to the standard CLI wrapper, helping users select the right approach for their telemetry requirements.                                                          
                                                                                                                                     
### Checklist                                                                                                                    
                                                                                                                                     
[✓] AI was used in creating this PR***


website/docs/integrations/dbt.md
```diff
@@ -77,10 +103,43 @@ Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
...
+
+The Artifact Processor is a library-level integration used by orchestrators to extract lineage post-run.
+
+#### How it Works
+Instead of running a CLI wrapper, an orchestrator (such as Apache Airflow using operators like [Astronomer Cosmos](https://astronomer.github.io/astronomer-cosmos/)) runs standard dbt commands. Once the run completes, the orchestrator triggers the OpenLineage `DbtLocalArtifactProcessor` library directly:
+1. The processor reads the generated `manifest.json`, `run_results.json`, and `catalog.json` files from the specified target directory.
+2. It parses the files and constructs/emits the OpenLineage events.
+
+#### Relationship to the Wrapper
+The `dbt-ol` wrapper actually uses the same underlying artifact parsing logic as the Artifact Processor. The key difference is that the wrapper executes the dbt command and handles the file paths automatically, while the Artifact Processor is designed to be called programmatically by orchestrators that manage their own file locations and execution lifecycles.
```
kacpermuda: the dbt-ol can also work in the `--consume-structured-logs` mode. So I think this page should be more structured like, we have two ways of parsing dbt into OpenLineage events: while running (structured logs) and after it's done (artifact processor), then describe the two, then describe the dbt-ol wrapper, and that it can run in both modes and is a 1:1 replacement to dbt command.

website/docs/integrations/dbt.md
```diff
@@ -77,10 +103,43 @@ Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
...
+---
+
+### 2. Artifact Processor
+
+The Artifact Processor is a library-level integration used by orchestrators to extract lineage post-run.
+
+#### How it Works
+Instead of running a CLI wrapper, an orchestrator (such as Apache Airflow using operators like [Astronomer Cosmos](https://astronomer.github.io/astronomer-cosmos/)) runs standard dbt commands. Once the run completes, the orchestrator triggers the OpenLineage `DbtLocalArtifactProcessor` library directly:
+1. The processor reads the generated `manifest.json`, `run_results.json`, and `catalog.json` files from the specified target directory.
+2. It parses the files and constructs/emits the OpenLineage events.
```
kacpermuda: This can be extended a bit, how it works, how many events per command, what are example facets that are always there, what are optional etc.

website/docs/integrations/dbt.md
```diff
@@ -77,10 +103,43 @@ Done. PASS=2 WARN=0 ERROR=0 SKIP=0 TOTAL=2
...
+
+### 3. Structured Log Processor
+
+The Structured Log Processor is a real-time integration method that parses JSON logs streamingly.
+
+#### How it Works
+Starting with dbt Core v1.x, dbt supports structured logging, emitting execution events as JSON lines.
+1. The integration listens to the structured log stream output by dbt (via standard output or log files).
+2. It parses the log events (such as node execution start, query start, node completion) as they occur.
+3. OpenLineage events are emitted **in real-time** while the dbt run is actively executing.
```
kacpermuda: Same here, we need this to be more detailed. The structure of events is different here than with the artifact processor (nested instead of flat for the dbt invocation itself), then again nested instead of flat for query events. Also, we need to do some nice table with comparison of both modes, what are pros and cons of each (e.g. node executing multiple queries in the artifact mode will most likely keep only last query id, where in structured logs approach we might get all, but structured logs assume queries are sequential and it attributes query_ids based on that), when to use one vs another etc.

website/docs/integrations/dbt.md
```diff
@@ -9,15 +9,41 @@ To learn more about dbt, visit the [documentation site](https://docs.getdbt.com)
...
+---
+
+### Ingestion Approaches Comparison
+
+| Feature | dbt CLI Wrapper (`dbt-ol`) | Artifact Processor | Structured Log Processor |
+| :--- | :--- | :--- | :--- |
+| **Ingestion Type** | Post-Run | Post-Run | Real-Time (Streaming) |
+| **Telemetry Source** | Target JSON files | Target JSON files | JSON Log Stream |
+| **Hierarchy Level** | Node (Models/Tests) | Node (Models/Tests) | Command → Node → Query |
+| **Best For** | Standalone CLI runs, simple cron | Airflow Operators (e.g., Cosmos) | Real-time dashboards, long-running jobs |
```
kacpermuda: I like this kind of table, left some comments below on how it can be improved !

kacpermuda (commented):
Left few comments on what would make it even better for a newbie coming to this page

website/docs/integrations/dbt.md
maitraymukeshkumarmodi-aiml: Thanks @kacpermuda! Restructured the document accordingly to introduce the two primary parsing mechanisms first (Artifact Processor post-run vs. Structured Log Processor real-time streaming), followed by the execution options. Added dedicated documentation for `dbt-ol` highlighting that it acts as a 1:1 drop in replacement for standard `dbt` and supports both default artifact mode and `--consume-structured-logs` mode.

website/docs/integrations/dbt.md
maitraymukeshkumarmodi-aiml: Expanded! Added details on how target files are parsed, event counts (2 events per node), and grouped facets into core always- present vs. optional.

website/docs/integrations/dbt.md
maitraymukeshkumarmodi-aiml: Updated! Detailed the nested `Command → Node → Query` event hierarchy, multi-query capture vs. artifact mode's last-query-ID limitation, and the sequential log attribution assumption.

website/docs/integrations/dbt.md
maitraymukeshkumarmodi-aiml: Thanks! Updated and expanded the table to compare telemetry sources, event hierarchy, multi-query capture, trade-offs, and execution flags for both modes.

maitraymukeshkumarmodi-aiml:
> Left few comments on what would make it even better for a newbie coming to this page

Thanks for the awesome feedback, i've updated this PR with changes! Feel free to review whenever.

website/docs/integrations/dbt.md
```diff
@@ -9,15 +9,115 @@ To learn more about dbt, visit the [documentation site](https://docs.getdbt.com)
...
+### 1. Artifact Processor (Post-Run Parsing)
+
+The Artifact Processor extracts lineage after a dbt run completes by parsing dbt's generated JSON artifact files.
+
+#### How it Works
+1. When dbt finishes, the processor reads three target JSON files from the `target/` directory:
+   * `manifest.json`: Contains the complete dependency graph, compiled SQL queries, and node definitions.
+   * `run_results.json`: Contains execution results, execution status, node timing, and query IDs.
+   * `catalog.json` *(optional)*: Contains database schema information, column data types, and table statistics.
+2. The processor converts the node metadata into OpenLineage dataset and job definitions, linking parent dependencies to child models.
```
kacpermuda: Can we elaborate on this linking here? We previously mention that artifact processor has a flat structure of events, so I assume it's about appending a parent metadata passed to artifact processor on creation ?

website/docs/integrations/dbt.md
```diff
@@ -9,15 +9,115 @@ To learn more about dbt, visit the [documentation site](https://docs.getdbt.com)
...
+
+#### Event Emission Model
+* **Events per Command**: The Artifact Processor emits a pair of events (**START** and **COMPLETE** or **FAIL**) for every executed dbt node (model, seed, snapshot, or test). 
+* For example, if a `dbt run` executes 5 models, the Artifact Processor will emit 10 OpenLineage events (5 `START` events followed by 5 `COMPLETE`/`FAIL` events).
+
+#### OpenLineage Facets Emitted
+The Artifact Processor enriches OpenLineage events with rich dbt-specific and standard facets:
+
+* **Always Present (Core Facets)**:
```
kacpermuda, line 56: What about JobType facet? Can you take another look and make sure we list all the facets here? Also, let's mention here that this doc may be outdated so checking the code is the best way of getting information. 

A small addition here, it would be nice to maybe briefly document all the dbt specific facets with some examples? The only way to get them now is to check code for python classes, there is no doc reference for them anywhere. Would love to see it in this PR, but if you don't have time we can make this a followup. For examples of such documentation you can take a look at "website/docs/spec/facets/".

website/docs/integrations/dbt.md
```diff
@@ -9,15 +9,115 @@ To learn more about dbt, visit the [documentation site](https://docs.getdbt.com)
...
+
+#### Event Hierarchy & Structural Differences
+Unlike the Artifact Processor which produces flat node-level events after execution, the Structured Log Processor constructs a **nested execution hierarchy**:
+
+1. **dbt Command Run**: An overall parent event representing the complete `dbt` invocation (e.g. `dbt run`).
+2. **Node Runs**: Nested child events for each model or test execution, linked to the main dbt command parent run.
+3. **Query Executions**: Individual SQL query execution events nested under their respective node runs.
+
+```text
+dbt Command Run (Parent)
```
kacpermuda: Can we also mention here if there is a way to pass parent information to this mode, so that the dbt command events point to the parent information we provide in their parent run facet? I think this is super useful information, if we want to linke the whole dbt execution with some other parent process.

We can make it a separate paragraph, independent of the mode, or add a paragraph for each node on how to pass parent information in given mode.

kacpermuda (commented):
Thanks @maitraymukeshkumarmodi-aiml , this looks a lot better now ! Left few more comments that would improve this even more. Please take a look and let me know your thoughts. I think after this round this will be ready to merge, appreciate your contribution.

website/docs/integrations/dbt.md
maitraymukeshkumarmodi-aiml: Updated!

maitraymukeshkumarmodi-aiml:
Good evening @kacpermuda, I have updated the file once again give your feedback, thank for investing your time reviewing this. Happy to iterate another time, let me know!

kacpermuda (approved):

PR #4729, opened 2026-07-18
fix: make with_additional_properties() facets picklable (#4216)

### LINKED ISSUE #4216: [BUG][Python] classes created using `with_additional_properties()` are not picklable
Example code:
```
from openlineage.client.facet_v2 import job_dependencies_run
import pickle

normal = job_dependencies_run.JobDependency(job=job_dependencies_run.JobIdentifier("a", "b"))
modified = job_dependencies_run.JobDependency(job=job_dependencies_run.JobIdentifier("a", "b")).with_additional_properties(a=1)

pickle.dumps(normal)  # OK
pickle.dumps(modified)  # Error
```
Error:
```
_pickle.PicklingError: Can't pickle <class 'openlineage.client.generated.job_dependencies_run.JobDependency'>: it's not the same object as openlineage.client.generated.job_dependencies_run.JobDependency
```

In our python code generator, we define `with_additional_properties` method as:
```
    def with_additional_properties(self, **kwargs: Any) -> "JobDependency":
        """Add additional properties to updated class instance."""
        current_attrs = [a.name for a in attr.fields(self.__class__)]

        new_class = attr.make_class(
            self.__class__.__name__,
            {k: attr.field(default=None) for k in kwargs if k not in current_attrs},
            bases=(self.__class__,),
        )
        new_class.__module__ = self.__class__.__module__
        attrs = attr.fields(self.__class__)
        for a in attrs:
            if not a.init:
                continue
            attr_name = a.name  # To deal with private attributes.
            init_name = a.alias
            if init_name not in kwargs:
                kwargs[init_name] = getattr(self, attr_name)
        return cast(JobDependency, new_class(**kwargs))
```

So we're creating a new class so that attr.fields(obj.__class__) have this new field as actual Attribute. When using simple setattr, this would not happen. I wonder if there is any way to fix that by modifying only the method. I don't have much time to look into it now, so I'm creating an issue.

zerafachris:
closes: #4216

Fix `PicklingError` when pickling facets created via `with_additional_properties()`.

### Meaningful description

**Root cause:** `BaseFacet.with_additional_properties()` (and the identical pattern generated
for `Fields`/`InputField`/`Transformation` in `column_lineage_dataset.py` and `JobDependency` in
`job_dependencies_run.py`) builds a brand-new class at runtime via `attr.make_class(...)`. That
class shares its `__name__`/`__module__` with the original, importable class but is never bound
to that name in the module's namespace. When `pickle` tries to serialize an instance, it looks
the class up by `module.__dict__[qualname]`, finds the *original* class instead, and raises
`PicklingError: ... it's not the same object as ...`.

**What changed and why:**
- Added two small, generic helpers to `client/utils.py`: `add_additional_properties()` (the
  class-creation logic, now cached per `(base class, extra field names)` pair so repeated calls
  with the same extra fields reuse the same class object instead of creating a new one every
  time) and `reduce_with_additional_properties()` (a generic `__reduce__` implementation that,
  for these dynamically-created classes, reduces to a call which rebuilds the same dynamic class
  chain via `with_additional_properties()` on unpickling instead of referencing the
  non-importable class directly).
- Updated the code generator template (`generator/templates/dataclass.jinja2`) so every
  generated `with_additional_properties()` method delegates to `add_additional_properties()` and
  every such class also gets a `__reduce__` delegating to `reduce_with_additional_properties()`.
  This is templated/generated code (confirmed by actually running
  `client/python/src/openlineage/client/generator/generate.py`), so the fix is in the generator,
  not hand-patched into the generated output — the generated files
  (`generated/base.py`, `generated/column_lineage_dataset.py`, `generated/job_dependencies_run.py`)
  were regenerated from the updated template and are included in this PR.
- Because the dynamic class is now cached by `(class, extra-field-names)`, two facets built with
  the same additional properties compare `==` again after a pickle round-trip (attrs equality
  checks exact class identity), including through chained `with_additional_properties()` calls.

**Test evidence:**
- Added `test_with_additional_properties_is_picklable` and
  `test_with_additional_properties_chained_is_picklable` to
  `client/python/tests/test_facet_v2.py`, doing a `pickle.dumps`/`pickle.loads` round-trip on a
  facet built via `with_additional_properties()` (including a chained/nested call) and asserting
  equality with the original instance and no exception.
- Ran the full local test suite: `uv run pytest tests/` → **638 passed**.
- Ran `uv run ruff check .` and `uv run ruff format --check .` on the whole `client/python`
  package → all clean.
- Ran `uv run mypy src/` → same 19 pre-existing errors as on `main` (verified via `git stash`),
  none introduced by this change; the touched files (`utils.py`, `generated/base.py`) are clean.
- Manually verified `pickle.dumps`/`pickle.loads` round-trips (including nested
  `with_additional_properties()` calls and `copy.deepcopy`) for `JobFacet`, `DocumentationJobFacet`,
  and `column_lineage_dataset.Fields`.

### Checklist

Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

### DIFF
--- a/client/python/src/openlineage/client/generated/base.py
+++ b/client/python/src/openlineage/client/generated/base.py
@@ -8,7 +8,7 @@ from typing import Any, ClassVar, cast
 
 import attr
 from openlineage.client.constants import DEFAULT_PRODUCER
-from openlineage.client.utils import RedactMixin
+from openlineage.client.utils import RedactMixin, add_additional_properties, reduce_with_additional_properties
 
 PRODUCER = DEFAULT_PRODUCER
 
@@ -88,23 +88,15 @@ class BaseFacet(RedactMixin):
 
     def with_additional_properties(self, **kwargs: Any) -> "BaseFacet":
         """Add additional properties to updated class instance."""
-        current_attrs = [a.name for a in attr.fields(self.__class__)]
+        return cast(BaseFacet, add_additional_properties(self, **kwargs))
 
-        new_class = attr.make_class(
-            self.__class__.__name__,
-            {k: attr.field(default=None) for k in kwargs if k not in current_attrs},
-            bases=(self.__class__,),
-        )
-        new_class.__module__ = self.__class__.__module__
-        attrs = attr.fields(self.__class__)
-        for a in attrs:
-            if not a.init:
-                continue
-            attr_name = a.name  # To deal with private attributes.
-            init_name = a.alias
-            if init_name not in kwargs:
-                kwargs[init_name] = getattr(self, attr_name)
-        return cast(BaseFacet, new_class(**kwargs))
+    def __reduce__(self) -> tuple[Any, ...]:
+        """Support pickling instances created via `with_additional_properties`.
+
+        `with_additional_properties` builds a new class at runtime, which the default pickling
+        mechanism can't resolve back to by name; see `reduce_with_additional_properties`.
+        """
+        return reduce_with_additional_properties(self)
 
     @staticmethod
     def _get_schema() -> str:
--- a/client/python/src/openlineage/client/generated/column_lineage_dataset.py
+++ b/client/python/src/openlineage/client/generated/column_lineage_dataset.py
@@ -7,7 +7,7 @@ from typing import Any, ClassVar, cast
 
 import attr
 from openlineage.client.generated.base import DatasetFacet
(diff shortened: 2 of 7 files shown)


kacpermuda (approved):
lgtm, will wait for another review here since it's an important change

mobuchowski (approved):

PR #4698, opened 2026-07-09
integration/common: implement FileSizeExpectationsParser for file asset validation

maitraymukeshkumarmodi-aiml:
### One-line summary for changelog:
Implement FileSizeExpectationsParser to handle `expect_file_size_to_be_between` expectations that were previously unimplemented and silently ignored.

### Meaningful description

`FileSizeExpectationsParser` in the Great Expectations integration had a stub 
`parse_expectation_result` method (containing only `pass`) and was commented out 
of `EXPECTATIONS_PARSERS`, causing `expect_file_size_to_be_between` expectations 
to be silently ignored rather than parsed.

This PR implements the method to extract `observed_value` from the result dict 
and return it as a `fileSize` facet, consistent with how `BetweenRowCountExpectationsParser` 
handles `rowCount`. The parser is also re-added to `EXPECTATIONS_PARSERS`.

Two unit tests are added covering the happy path and the null/missing result case.

### Checklist
- [x] AI was used in creating this PR

PR #4733, opened 2026-07-18
fix(great-expectations): restore data-quality metrics on Great Expectations 1.x

chuenchen309:
## Problem

Great Expectations 1.0 renamed `ExpectationConfiguration.expectation_type` to `type`, and `integration/common` requires `great_expectations>=1.0.0`. The GE provider still reads the old key in three places, so on **every supported GE release**:

- `ExpectationsParser.can_accept` and `ColumnExpectationsParser.can_accept` (`results.py`) resolve the expectation type to `None` via `get_from_nullable_chain(..., ["expectation_config", "expectation_type"])`, so **no** parser accepts any expectation and `DataQualityMetricsInputDatasetFacet` comes out empty — data-quality metrics are silently dropped.
- `parse_assertions` (`action.py`) reads `expectation["expectation_config"]["expectation_type"]`, which raises `AttributeError`. Only `ValueError` is caught, so assertion-facet parsing fails outright.

A second, coupled defect: `FileSizeExpectationsParser.facet_key` is `"fileSize"`, which is not a field of `DataQualityMetricsInputDatasetFacet` (the size field is `bytes`). It is currently masked by the first bug (the parser never fires), but once `can_accept` works the parser fires and building the facet with `fileSize=...` raises an uncaught `TypeError`.

## Fix

- Read `type` instead of `expectation_type` in the two `can_accept` methods and in `parse_assertions`.
- Change `FileSizeExpectationsParser.facet_key` to `"bytes"`, keeping the file-size intent from #4698 while producing a valid facet.

## Verification

Reproduced against `great_expectations==1.19.0` with real `ExpectationValidationResult` objects. Before the change, `parse_data_quality_facet` returns a facet with `rowCount=None` and `parse_assertions` raises `AttributeError: 'ExpectationConfiguration' object has no attribute 'expectation_type'`; after it, `rowCount` is populated and assertions parse. A `expect_file_size_to_be_between` result goes from an uncaught `TypeError` to `bytes=<size>`. Added a regression test and updated the two `FileSizeExpectationsParser` assertions.

Note on CI: the GE test suite is suspended (`--ignore=tests/great_expectations`, #3078) and `great_expectations` is not installed in the common CI env, so a green CI here does **not** exercise this change and the new test won't run in CI. Verification above was done locally with GE installed. Happy to adjust the framing if you'd prefer the file-size field handled differently.

## Disclosure

This contribution is fully AI-authored and autonomous (Claude Code, acting on this account). An AI found the bug, ran the repro against real Great Expectations objects, wrote the fix and tests, and wrote this description; the human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff — it was done by the AI, not a person. If this isn't the kind of contribution you want, say so and I'll close it.

### DIFF
--- a/integration/common/src/openlineage/common/provider/great_expectations/results.py
+++ b/integration/common/src/openlineage/common/provider/great_expectations/results.py
@@ -38,9 +38,7 @@ class ExpectationsParser:
 
     @classmethod
     def can_accept(cls, expectation_result: ExpectationValidationResult) -> Any | None:
-        expectation_type = get_from_nullable_chain(
-            expectation_result, ["expectation_config", "expectation_type"]
-        )
+        expectation_type = get_from_nullable_chain(expectation_result, ["expectation_config", "type"])
         return expectation_type and expectation_type == cls.expectation_key
 
     @staticmethod
@@ -68,7 +66,7 @@ class EqualRowCountExpectationsParser(BetweenRowCountExpectationsParser):
 
 class FileSizeExpectationsParser(ExpectationsParser):
     expectation_key = "expect_file_size_to_be_between"
-    facet_key = "fileSize"
+    facet_key = "bytes"
 
     @staticmethod
     def parse_expectation_result(expectation_result: dict) -> ExpectationsParserResult:  # type: ignore # noqa
@@ -87,9 +85,7 @@ class ColumnExpectationsParser(ExpectationsParser):
 
     @classmethod
     def can_accept(cls, expectation_result: ExpectationValidationResult) -> Any | None:
-        expectation_type = get_from_nullable_chain(
-            expectation_result, ["expectation_config", "expectation_type"]
-        )
+        expectation_type = get_from_nullable_chain(expectation_result, ["expectation_config", "type"])
         extracted_column = get_from_nullable_chain(
             expectation_result, ["expectation_config", "kwargs", "column"]
         )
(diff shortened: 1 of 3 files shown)


integration/common/src/openlineage/common/provider/great_expectations/results.py
```diff
@@ -68,7 +66,7 @@ class EqualRowCountExpectationsParser(BetweenRowCountExpectationsParser):
 
 class FileSizeExpectationsParser(ExpectationsParser):
     expectation_key = "expect_file_size_to_be_between"
-    facet_key = "fileSize"
+    facet_key = "bytes"
```
kacpermuda, line 69: This was just recently changed, would you mind reviewing this PR @maitraymukeshkumarmodi-aiml ?

mobuchowski (approved):

chuenchen309:
Gentle ping — this has been approved since Jul 21 and is still green: CI passing, no conflicts against current `main`, and `results.py` hasn't changed upstream since, so the fix still applies as reviewed. Happy to rebase or adjust anything if needed; otherwise it's ready whenever someone has a moment to merge.

chuenchen309:
Thanks for merging this one.

Somewhat related — most of what I've sent here has landed in the same corner of the tree (`integration/common`: the dbt artifact/structured-log processors, the Great Expectations provider, and the Python client's merge helpers), and I've been reading through a few of the open Python PRs recently while checking my own work against them. If it would take anything off your plate, I'm glad to review or triage incoming Python-side PRs in that area — I can't approve anything, but I can at least run them and say concretely what breaks. Totally up to you.

kacpermuda:
Thanks, appreciate it. Will keep an eye out for this, for now you can take a look at open issues / open PRs, that would definitely help

PR #4745, opened 2026-07-21
[BUG] Flink checkpoint tracker uses configured rest.port instead of the effective REST endpoint in Application mode

wangxiaojing:
  Improve Flink REST endpoint resolution for checkpoint tracking in Application mode deployments.

  ### Meaningful description
  This PR fixes how the Flink integration resolves the REST API endpoint used by checkpoint tracking.

  Previously, the tracker built the jobs API URL directly from `rest.address` and `rest.port`. In Flink Application mode, especially on Kubernetes, the effective runtime REST endpoint may differ from the configured
  port, which can cause checkpoint polling to fail even when the job is healthy.

  This change introduces a shared REST endpoint discovery utility and wires both the Flink 1.x and Flink 2.x integrations to use it. It also adds a new optional configuration, `openlineage.restApiBaseUrl`, so users
  can explicitly override the REST base URL when automatic discovery is not sufficient.

  The discovery flow now works in this order:
  1. Use `openlineage.restApiBaseUrl` when configured
  2. Try to discover the runtime endpoint from the Flink process
  3. Try Linux `/proc/self/net/tcp` plus `/overview` validation
  4. Fall back to `rest.address` and `rest.port`

  This PR is intentionally scoped to REST endpoint resolution for checkpoint tracking. It does not include terminal event handling changes or checkpoint tracking enable/disable behavior.

  ### Checklist


mobuchowski:
@wangxiaojing can you take a look at the conflicts?

wangxiaojing:
> @wangxiaojing can you take a look at the conflicts?
Thanks for the reminder. I've merged the latest main branch and resolved the conflicts. The updates have been pushed.

PR #4753, opened 2026-07-22
flink: use effective REST endpoint for checkpoint tracking, not configured rest.port

### LINKED ISSUE #4739: [BUG] Flink checkpoint tracker uses configured rest.port instead of the effective REST endpoint in Application mode

Flink integration

### OpenLineage version

---
  main branch
---

  - Apache Flink 2.0.0 in Application mode
  - Kubernetes
  - Java 11
  - OpenLineage Flink integration built from source
  - Observed in environments where the Flink REST endpoint is not actually bound to the configured rest.port

### Environment configuration

  rest.port: 8080
  kubernetes.hostnetwork.enabled: 'true'
  execution.job-status-changed-listeners: io.openlineage.flink.listener.OpenLineageJobStatusChangedListenerFactory

  openlineage.transport.type: kafka
  openlineage.transport.topicName: <topic>
  openlineage.transport.properties.bootstrap.servers: <kafka-bootstrap>
  openlineage.transport.properties.key.serializer: org.apache.kafka.common.serialization.StringSerializer
  openlineage.transport.properties.value.serializer: org.apache.kafka.common.serialization.StringSerializer

  openlineage.trackingIntervalInSeconds: 60

  There is currently no way to point the tracker at the effective runtime REST endpoint when Flink binds a different port than the configured rest.port.

### Deployment details

  Self-managed Kubernetes deployment, not a managed service.

  The problem was observed when running Flink in Application mode with host networking enabled. In this setup, the runtime REST endpoint exposed by MiniDispatcherRestEndpoint may bind to a random available port
  instead of the configured rest.port.

### Problem details

The OpenLineage continuous job tracker builds the Flink jobs endpoint directly from the configured values:

  String.format("http://%s:%s/jobs", configuredHost, configuredPort)

  This breaks in Flink Application mode when the effective REST port differs from the configured rest.port.

  For example:

  - configured rest.port: 8080
  - actual runtime REST port: 44193 or 45679

  In that case the tracker polls the wrong endpoint, so checkpoint polling fails and the integration cannot emit the expected checkpoint-based RUNNING events.

  From the user perspective, the job is running correctly, but OpenLineage keeps trying to talk to the wrong REST API address.

  The tracker should resolve the effective Flink REST endpoint at runtime instead of assuming that the configured rest.port is always correct.

  At minimum, the integration should provide a way to explicitly override the REST base URL used by checkpoint tracking. Ideally, it should also discover the runtime endpoint automatically in Application mode.

  Expected behavior:

  - checkpoint tracking should use the actual reachable Flink REST endpoint
  - RUNNING events with checkpoint data should continue to be emitted normally
  - users should not need to rely on the configured rest.port matching the runtime-bound port in Application mode

### How to reproduce

  1. Deploy a Flink job in Application mode on Kubernetes.
  2. Enable the OpenLineage Flink listener.
  3. Configure Flink with a fixed rest.port, for example 8080.
  4. Run in an environment where the actual MiniDispatcherRestEndpoint binds to a different port at runtime.
  5. Start a job with checkpointing enabled.
  6. Observe that the OpenLineage tracker still builds the jobs endpoint from the configured rest.address + rest.port.
  7. Confirm that checkpoint polling is targeting the wrong URL and that checkpoint-based RUNNING events are missing.

  If needed, this may be somewhat environment-dependent, but the failure mode is deterministic once the runtime REST port differs from the configured one.

### Additional details

  This appears to affect the code path used to create the continuous tracker for Flink integrations and is especially visible in Kubernetes Application mode deployments.

  A representative symptom is:

  - the Flink job is healthy
  - the REST API is reachable on the effective runtime port
  - OpenLineage checkpoint polling still targets the configured port instead

  Example values observed in practice:

  - configured port: 8080
  - actual bound port: 44193, 45679

  It would also be helpful if the integration exposed debug logging around REST endpoint resolution so users can verify which address is being used.

kalra-mohit:
Fixes #4739.

`OpenLineageContinousJobTracker` (shared by the Flink 1 job listener and both Flink 2 job-status-changed listeners) builds the JobManager REST URL it polls for checkpoint stats straight from the statically configured `rest.address`/`rest.port`:

```java
String.format("http://%s:%s/jobs", configuredHost, configuredPort)
```

That works as long as Flink actually ends up bound to that port. In Application mode it often doesn't — `rest.port`/`rest.bind-port` set to `0` gets you a random free port, and even a fixed port falls through to another one if it's already taken. This is the normal case for Kubernetes Application mode deployments. I went through Flink's `ClusterEntrypoint`, `RestServerEndpoint`, and `JobStatusChangedListenerFactory` to see whether the effective bound port is exposed anywhere this plugin could read it, and it isn't — it only lives on the `RestServerEndpoint` instance inside `DispatcherResourceManagerComponent`, which listener plugins never get a handle to. So there's no way to auto-discover the real port from inside the OpenLineage integration with Flink's current public API.

Net effect: whenever the effective port didn't match the configured one, the tracker polled a REST endpoint that either didn't exist or belonged to something else. Checkpoint polling failed silently and checkpoint-driven `RUNNING` events were never emitted — no error, just missing lineage events.

```mermaid
flowchart TD
    A["Flink starts the JobManager"] --> B["REST server binds to a port<br/>(rest.port=0 → random; or configured port already in use → different port)"]
    B --> C{"effective port == configured rest.port?"}
    C -->|"session/standalone mode, typical case"| D["they match — old code happened to work"]
    C -->|"Application mode / k8s / port 0 / port conflict"| E["effective port differs from configured rest.port"]
    E --> F["old: OpenLineageContinousJobTracker builds URL from rest.address + rest.port (configured)"]
    F --> G["tracker polls the wrong /jobs endpoint<br/>→ polling fails silently<br/>→ no checkpoint RUNNING events"]
    E --> H{"openlineage.flink.checkpointTrackingRestUrl set?<br/>(FlinkRestApiUrlResolver)"}
    H -->|"yes"| I["use the override as the REST base URL"]
    H -->|"no"| J["fall back to rest.address + rest.port (configured)<br/>+ debug log warning it may not be the real bound port"]
    I --> K["tracker polls the correct /jobs endpoint<br/>→ checkpoint RUNNING events emitted"]
```

The issue asked for, at minimum, a way to explicitly override the REST base URL used by checkpoint tracking, so that's what this adds:

- `FlinkRestApiUrlResolver` in the shared module, used by all three call sites: flink1's `OpenLineageContinousJobTrackerFactory`, and flink2's `OpenLineageJobStatusChangedListener` / `OpenLineageDetachedJobStatusChangedListener`.
- A new `openlineage.flink.checkpointTrackingRestUrl` config option to point checkpoint tracking at the effective REST endpoint (e.g. `http://localhost:44193`) when it differs from `rest.port`.
- If the override isn't set, it falls back to the previous `rest.address`/`rest.port`-derived URL, now with a debug log noting that this may not reflect the actual bound port in Application mode.

Wrote `FlinkRestApiUrlResolverTest` (shared) and `OpenLineageContinousJobTrackerFactoryTest` (flink1) to reproduce the exact scenario from the issue — configured `rest.port=8080`, effective port `44193`. Checked that they fail against the pre-fix code first:

```
OpenLineageContinousJobTrackerFactoryTest > getTrackerUsesCheckpointTrackingRestUrlOverrideWhenEffectivePortDiffersFromConfiguredPort() FAILED
  expected: "http://localhost:44193/jobs"
   but was: "http://localhost:8080/jobs"
```

then pass after the fix. Ran the full `shared`, `flink1`, and `flink2` suites afterward: `shared` 5/5 new resolver tests pass, `flink1` 14/14 (including the 2 new ones), `flink2` 17/17 across both listener test classes — no regressions anywhere else in those modules.


kalra-mohit:
@kacpermuda whenever you get a chance to take a look, happy to address any feedback!

mobuchowski:
There is already a PR for this issue: https://github.com/OpenLineage/OpenLineage/pull/4745/changes

pawel-big-lebowski:
> There is already a PR for this issue: https://github.com/OpenLineage/OpenLineage/pull/4745/changes

@mobuchowski I've done some comparison with a help of agents:

> Both #4745 and #4753 fix the same checkpoint-tracking bug (#4739), but I think #4753 is the better approach. It's a focused override + fallback that reads a native openlineage.flink.checkpointTrackingRestUrl config option, updates all three call sites (including the flink2 detached listener, which #4745 misses), and comes with solid tests. #4745 goes much further by trying to auto-discover the effective REST port through reflection into JVM/Flink internals and /proc/self/net/tcp probing — clever, but it's fragile across Flink versions, breaks on JDK 17+ without --add-opens, does blocking HTTP probes on the listener construction path, and leaves that discovery logic essentially untested. My suggestion would be to go with #4753 as the base, and if we still want zero-config auto-discovery later, add it as a separate opt-in follow-up rather than the default path.

Although issue is same, those PRs are different. Credits should be given to all the contributors, but I don't think we should just choose the first solution. My preference would be to proceed with this PR, or least the way it solves the problem (without reflection).

PR #4775, opened 2026-07-27
sql: fix MySQL multi-table DELETE lineage

### LINKED ISSUE #4774: [BUG] SQL parser treats MySQL multi-table DELETE lookup tables as outputs

SQL parser

### OpenLineage version

Current `main` (`2aae49d8b5ac5b54eea1ae6dc32cab4ec94dc92c`, version 1.53.0-SNAPSHOT). The behavior also exists in released versions containing the current DELETE visitor.

Rust SQL parser using `sqlparser` 0.62.0 and the MySQL dialect.

### Problem details

For MySQL multi-table DELETE syntax, the SQL parser treats every table in the joined `FROM` tree as a mutation output and reports no table inputs.

For example:

```sql
DELETE t1
FROM t1
JOIN t2 ON t1.id = t2.id
WHERE t2.expired = 1;
```

Current table lineage:

```text
inputs:  []
outputs: [t1, t2]
```

Only `t1` is an explicit deletion target. `t2` is a lookup table read to determine which rows are deleted, but it is incorrectly emitted as an output.

The aliased form has the same problem:

```sql
DELETE target
FROM foo AS target
JOIN bar AS lookup ON target.id = lookup.id;
```

It currently reports both `foo` and `bar` as outputs.

### What should happen instead

The explicit targets in the MySQL `DELETE` target list should be outputs, while non-target tables referenced by `FROM` and `JOIN` should be inputs.

Expected lineage for the first query:

```text
inputs:  [t2]
outputs: [t1]
```

Simple single-table statements such as `DELETE FROM t1` should remain unchanged and report only `t1` as an output.

### How to reproduce

Call the SQL parser with the MySQL dialect for either query above and inspect `table_lineage.in_tables` and `table_lineage.out_tables`.

The root cause is in the `Statement::Delete` visitor: it marks every relation in `delete.from` as an output and ignores the explicit MySQL targets in `delete.tables`.

### Additional details

Incorrect mutation lineage can make downstream impact analysis report lookup/source tables as modified datasets while omitting their read dependency.

### Contribution

I am interested in doing this work.

mattfaltyn:
Correct MySQL multi-table DELETE lineage so explicit targets are outputs and joined lookup tables are inputs.

### Meaningful description
MySQL supports an explicit target list before the `FROM` clause:

```sql
DELETE target
FROM foo AS target
JOIN bar AS lookup ON target.id = lookup.id;
```

The SQL visitor previously ignored that target list and classified every relation in the joined `FROM` tree as an output. This produced no input lineage and incorrectly marked lookup tables as mutated datasets.

This change visits the joined tree as inputs when an explicit DELETE target list is present, resolves target aliases back to their base tables, and moves only those targets to outputs. Because the parser aggregates lineage across multiple statements, inputs collected before the DELETE are preserved. Existing single-table DELETE and DELETE USING behavior remains unchanged.

Regression coverage includes unaliased, aliased, multiple-target, and multi-statement MySQL DELETE cases.

Closes #4774

### Validation

- `cargo test` — 147 passed, 3 pre-existing ignored
- `cargo test --no-default-features` — 147 passed, 3 pre-existing ignored
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo fmt --all -- --check`
- all configured pre-commit hooks applicable to the changed files

### Checklist

### DIFF
--- a/integration/sql/impl/src/context/mod.rs
+++ b/integration/sql/impl/src/context/mod.rs
@@ -159,6 +159,18 @@ impl<'a> Context<'a> {
         }
     }
 
+    pub fn move_input_to_output(&mut self, input: Vec<Ident>) {
+        let name = DbTableMeta::new(
+            input,
+            self.dialect,
+            self.default_schema.clone(),
+            self.default_database.clone(),
+        );
+        let output = self.resolve_table(&name).clone();
+        self.inputs.remove(&output);
+        self.outputs.insert(output);
+    }
+
     pub fn add_non_table_output(
         &mut self,
         output: Vec<Ident>,
--- a/integration/sql/impl/src/visitor.rs
+++ b/integration/sql/impl/src/visitor.rs
@@ -726,22 +726,36 @@ impl Visit for Statement {
                 }
             }
             Statement::Delete(delete) => {
-                match &delete.from {
+                let tables = match &delete.from {
                     FromTable::WithFromKeyword(tables) | FromTable::WithoutKeyword(tables) => {
-                        for table in tables {
+                        tables
+                    }
+                };
+
+                if delete.tables.is_empty() {
+                    for table in tables {
+                        if let Some(output) =
+                            get_table_name_from_table_factor(&table.relation, &*context)
+                        {
+                            context.add_output(output);
+                        }
+                        for join in &table.joins {
                             if let Some(output) =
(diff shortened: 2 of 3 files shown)


pawel-big-lebowski (approved):
Thanks for working on this 👍 🚀

PR #4778, opened 2026-07-29
client-java: generate singleton-enum discriminated unions

mobuchowski:
## Summary

- generate Jackson-polymorphic interfaces for `oneOf` references with distinct singleton string-enum discriminators
- make concrete variants implement the generated interface while preserving existing behavior for `const` and non-discriminated unions
- discover facets by their base facet type so job and dataset facets can share the same container key
- correctly qualify repeated nested enum names such as `Type`
- add synthetic generation, compilation, and deserialization tests

## Motivation

This is a prerequisite for generating Java models from the upcoming explicit lineage facet schema without hand-written model exceptions.

That schema needs a common entry type whose concrete shape is selected by a singleton enum:

```json
{
  "LineageEntry": {
    "oneOf": [
      { "$ref": "#/$defs/LineageDatasetEntry" },
      { "$ref": "#/$defs/LineageJobEntry" }
    ]
  },
  "LineageDatasetEntry": {
    "type": "object",
    "properties": {
      "type": { "type": "string", "enum": ["DATASET"] }
    }
  },
  "LineageJobEntry": {
    "type": "object",
    "properties": {
      "type": { "type": "string", "enum": ["JOB"] }
    }
  }
}
```

`LineageJobFacet.entries` then refers to `LineageEntry`:

```json
{
  "entries": {
    "type": "array",
    "items": { "$ref": "#/$defs/LineageEntry" }
  }
}
```

Before this change, the Java resolver discarded the union and returned its first referenced branch. The resulting API would therefore expose `List<LineageDatasetEntry>` instead of `List<LineageEntry>`. A `JOB` entry could neither be supplied through the generated factories/builders nor deserialized from JSON.

This PR instead generates the equivalent of:

```java
@JsonTypeInfo(property = "type", ...)
@JsonSubTypes({
  @JsonSubTypes.Type(value = LineageDatasetEntry.class, name = "DATASET"),
  @JsonSubTypes.Type(value = LineageJobEntry.class, name = "JOB")
})
interface LineageEntry {}

final class LineageDatasetEntry implements LineageEntry {}
final class LineageJobEntry implements LineageEntry {}
```

The explicit lineage schema also registers the same facet key for two different facet base types:

```json
{
  "lineage": {
    "anyOf": [
      { "$ref": "#/$defs/LineageDatasetFacet" },
      { "$ref": "#/$defs/LineageJobFacet" }
    ]
  }
}
```

Previously only the first branch was attached to a generated facet container. Facet discovery now uses each concrete facet's base type and inferred key, allowing both `DatasetFacets.lineage` and `JobFacets.lineage` to be generated from one schema document.

### DIFF
--- a/client/java/generator/src/test/java/io/openlineage/client/GeneratorTest.java
+++ b/client/java/generator/src/test/java/io/openlineage/client/GeneratorTest.java
@@ -1,0 +1,149 @@
+/*
+/* Copyright 2018-2026 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import javax.tools.JavaCompiler;
+import javax.tools.ToolProvider;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class GeneratorTest {
+
+  @TempDir Path outputDirectory;
+
+  @Test
+  void generatesSingletonEnumDiscriminatedUnions() throws Exception {
+    String source = generate();
+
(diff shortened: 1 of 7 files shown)


client/java/generator/src/test/java/io/openlineage/client/GeneratorTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+class GeneratorTest {
+
+  @TempDir Path outputDirectory;
+
+  @Test
+  void generatesSingletonEnumDiscriminatedUnions() throws Exception {
+    String source = generate();
+
+    assertTrue(source.contains("public interface LineageEntry"));
```
pawel-big-lebowski, line 35: For readability, you could put java file into the resources and validate if it's same as expected.

pawel-big-lebowski (approved):
great improvement 🚀

tnazarew (approved):

PR #4782, opened 2026-07-29
spark: add descriptions to column lineage transformations (rebase of #4247)

mobuchowski:
Add descriptions to column lineage transformations in spark integration.

### Meaningful description

Rebase of #4247 (@tnazarew) onto current `main`, with the regressions that
rebase exposed fixed. tnazarew's three commits are preserved verbatim with
their original authorship; the fixes are two commits on top.

Original description, unchanged:

> The PR contains the extraction of transformation descriptions during column level lineage collection.
>
> While traversing the spark plan nodes, each source column is added to dependency list with `outputExpressionString` and `TransformationInfo.description`, e.g. : `sum(a) as c` we get
> - `outputExpressionString` - `c`
> - `TransformationInfo.description` - `sum(a) as c`
>
> When resolving dependency chains the result description takes first dependency description and replaces each `outputExpressionString` with description from the second dependency.
>
> ```
> {'c', 'sum(a) as c'} + {'a', 'x + y as a'} -> {'c', 'sum(x + y as a) as c'}
> ```
>
> Additional changes made:
> - dependency on itself cannot be added anymore
> - the identity type is used only for one to one transformations, functions that `sometimes` give the same value like nvl, coalesce, if `identity` else `transformation` etc. have the type `transformation` now

### What changed relative to #4247

**Conflict resolution.** `findDependentInputs` kept main's `HashSet` BFS visited
set from #4376 rather than the branch's `LinkedList`.

**OOM regression.** `isDescriptionsEnabled()` walked the config on every
`addDependency()` call — up to `COMPUTED_DEPENDENCY_HARD_LIMIT` (1M) times per
builder. That exhausted the heap in
`ColumnLevelLineageBuilderTest.testCartesianDependencyPerformance` (the 2000×2000
and 10000×10000 cases died with `OutOfMemoryError`). The flag is now resolved
once, lazily, so callers can still adjust the config after construction.

**Stale shared `Dependency` instances.** `commonDependencies` (#4376, added after
this branch was cut) keys shared `Dependency` objects by input `ExprId` and only
compared `TransformationInfo`. Now that a `Dependency` also carries an output
expression, a cached instance could be handed to an output it does not describe.
The cache key covers the description-carrying fields too.

**`TransformationInfo.withDescription()`** replaces the roundabout
`merge(identity(description), (d1, d2) -> d2)` idiom, and returns `this` when
nothing changes so the cached singletons survive.

**Test expectation.** `complexQueryCTEJoinsFilter` expected join descriptions
with one paren level too many. Spark's `Expression.sql()` already parenthesises
the condition and `JoinVisitor` does not add more — consistent with how the
passing `WHERE %s` assertions behave.

**Dead field.** `DatasetDependency.description` was written on every dataset
dependency and never read; the description a dataset dependency ends up with
comes from the seeded output expression and the merge chain. Dropped, along with
the third argument of `addDatasetDependency()` in the four operator visitors.

**Docs.** `spark.openlineage.columnLineage.descriptionsEnabled` is now in the
Spark configuration reference and has a javadoc on `ColumnLineageConfig`.


tnazarew (approved):
I made part of the code and approve the rest of the changes 👍

LS-Chi:
Thank you guys! It's awesome! @mobuchowski @tnazarew

Issue #4280, opened 2026-01-26
[FEATURE] Sync Python & Java clients regading run & job tags configuration


dolfinus:
### Motivation

Python client has configuration which looks like this:
```yaml
tags:
  run:
    - somekey:somevalue  # by default source=CONFIG
    - somekey:somevalue:somesource
  job:
    - somekey:somevalue
    - somekey:somevalue:somesource
```

Tags from config are merged into RunEvent.run.facets.tags and RunEvent.job.facets.tags before event is send in OpenLineageClient:
#3471
#3980
https://github.com/OpenLineage/OpenLineage/blob/66a9ef5434e09437b3c72b4c563d090a24e1650b/client/python/src/openlineage/client/client.py#L450
https://github.com/OpenLineage/OpenLineage/blob/main/client/python/src/openlineage/client/client.py#L462

Java client have a bit different config:
```yaml
run:  # key order is different
  tags:
    - somekey:somevalue
    - somekey:somevalue:somesource
job:  # key order is different
  owners:
    team: someteam
    user: someuser
  tags:
    - somekey:somevalue  # by default source=USER
    - somekey:somevalue:somesource
```

Merging tags & ownership is done not by OpenLineageClient, but by each integration separately. For example, Spark uses both tags and ownership:
https://github.com/OpenLineage/OpenLineage/blob/66a9ef5434e09437b3c72b4c563d090a24e1650b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/TagsJobFacetBuilder.java
https://github.com/OpenLineage/OpenLineage/blob/66a9ef5434e09437b3c72b4c563d090a24e1650b/integration/spark/shared/src/main/java/io/openlineage/spark/agent/facets/builder/TagsRunFacetBuilder.java
https://github.com/OpenLineage/OpenLineage/blob/66a9ef5434e09437b3c72b4c563d090a24e1650b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/InternalEventHandlerFactory.java#L221-L222
https://github.com/OpenLineage/OpenLineage/blob/66a9ef5434e09437b3c72b4c563d090a24e1650b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/InternalEventHandlerFactory.java#L202

[...]

PR #4799, opened 2026-08-03
add retry configuration to http transport

tnazarew:
Expose retry config in `HttpTransport`

### Meaningful description

The client used by `HttpTransport` has its retry policy configurable. Until now the transport just used the default one. Now the policy is configurable via transport config.

### Checklist

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/transports/HttpConfig.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/HttpConfig.java
@@ -56,6 +56,16 @@ public final class HttpConfig implements TransportConfig, MergeConfig<HttpConfig
   @Setter
   private @Nullable HttpSslContextConfig sslContextConfig;
 
+  @JsonProperty("maxRetries")
+  @Getter
+  @Setter
+  private @Nullable Integer maxRetries;
+
+  @JsonProperty("retryIntervalMillis")
+  @Getter
+  @Setter
+  private @Nullable Integer retryIntervalMillis;
+
   @Override
   public HttpConfig mergeWithNonNull(HttpConfig other) {
     return new HttpConfig(
@@ -66,6 +76,8 @@ public final class HttpConfig implements TransportConfig, MergeConfig<HttpConfig
         mergePropertyWith(urlParams, other.urlParams),
         mergePropertyWith(headers, other.headers),
         mergePropertyWith(compression, other.compression),
-        mergePropertyWith(sslContextConfig, other.sslContextConfig));
+        mergePropertyWith(sslContextConfig, other.sslContextConfig),
+        mergePropertyWith(maxRetries, other.maxRetries),
+        mergePropertyWith(retryIntervalMillis, other.retryIntervalMillis));
   }
 }
--- a/client/java/src/main/java/io/openlineage/client/transports/HttpConfig.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/HttpConfig.java
@@ -66,6 +66,16 @@ public final class HttpConfig implements TransportConfig, MergeConfig<HttpConfig
   @Setter
   private @Nullable Integer retryIntervalMillis;
 
+  @JsonProperty("retryIntervalMultiplier")
+  @Getter
+  @Setter
+  private @Nullable Double retryIntervalMultiplier;
+
+  @JsonProperty("maxRetryIntervalMillis")
+  @Getter
+  @Setter
+  private @Nullable Integer maxRetryIntervalMillis;
+
   @Override
   public HttpConfig mergeWithNonNull(HttpConfig other) {
     return new HttpConfig(
@@ -78,6 +88,8 @@ public final class HttpConfig implements TransportConfig, MergeConfig<HttpConfig
         mergePropertyWith(compression, other.compression),
         mergePropertyWith(sslContextConfig, other.sslContextConfig),
         mergePropertyWith(maxRetries, other.maxRetries),
-        mergePropertyWith(retryIntervalMillis, other.retryIntervalMillis));
+        mergePropertyWith(retryIntervalMillis, other.retryIntervalMillis),
+        mergePropertyWith(retryIntervalMultiplier, other.retryIntervalMultiplier),
+        mergePropertyWith(maxRetryIntervalMillis, other.maxRetryIntervalMillis));
   }
 }
--- a/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
@@ -17,20 +17,26 @@ import io.openlineage.client.OpenLineageClientException;
 import io.openlineage.client.OpenLineageClientUtils;
 import java.io.File;
 import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.ConnectException;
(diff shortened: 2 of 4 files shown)


chuenchen309:
Two things here reproduce as behaviour changes rather than style points, both around how the new
knobs interact with what `DefaultHttpRequestRetryStrategy` already does. I transcribed
`withTimeout()` and `ExponentialJitterRetryStrategy` from this branch (`2f5cc21cb`) into a standalone
harness, so the only moving parts are the config values. Numbers below are that harness's output.

(Correction to this paragraph as first posted, which named 5.6.2: the pin is
`httpclient5:5.6.3` at `client/java/build.gradle:79`. The tables below are from re-running the
harness on 5.6.3; the first run was on 5.6.2 and produced the same output row for row.)

### 1. `retryIntervalMillis` larger than `timeoutInMillis` silently turns retries off

`super(maxRetries, TimeValue.ofMilliseconds(baseRetryIntervalMillis))` makes `retryIntervalMillis`
the superclass's `defaultRetryInterval`, and `DefaultHttpRequestRetryStrategy.retryRequest(HttpResponse, …)`
gates on it before it ever looks at `maxRetries`:

```java
final Timeout responseTimeout = requestConfig.getResponseTimeout();
if (responseTimeout != null && defaultRetryInterval.compareTo(responseTimeout) > 0) {
    return false;
}
return execCount <= this.maxRetries && retriableCodes.contains(response.getCode());
```

`HttpTransport` sets `setResponseTimeout(timeout)` from `timeoutInMillis`, default 5000. So the
comparison is `retryIntervalMillis` vs `timeoutInMillis`, and the override of `getRetryInterval`
doesn't help — that method decides *how long* to wait, not *whether* to retry.

Server answers 503 on the first POST to `/api/v1/lineage`, 200 after; `attempts` is how many requests
it saw:

| config | attempts | |
|---|---|---|
| defaults (`timeoutInMillis` 5000, `retryIntervalMillis` 1000) | 2 | retried |
| `retryIntervalMillis` 6000 | 1 | **no retry** |
| `maxRetries` 5, `retryIntervalMillis` 10000 | 1 | **no retry** |
| same, `timeoutInMillis` raised to 20000 | 2 | retried |
| `retryIntervalMillis` 5000 (equal to timeout) | 2 | retried — the check is strict `>` |

Rows 3 and 4 differ only in `timeoutInMillis`, so the gate is that comparison and nothing else in the
harness. The failure is quiet: `maxRetries: 5` is accepted without complaint and produces zero retries.
It also seems reachable by accident rather than by misuse — `maxRetryIntervalMillis` defaulting to
60000 invites base intervals well above a 5s response timeout.

Passing a small fixed `TimeValue` to `super(...)` and keeping `retryIntervalMillis` purely for the
jitter computation would decouple the two, since `getRetryInterval` is overridden on both paths
anyway. Rejecting or clamping the combination at config time would work too — the part worth avoiding
is the silent version.

### 2. `maxRetries` never applies to transport errors, because the emit is a POST

`retryRequest(HttpRequest, IOException, …)` ends in `return handleAsIdempotent(request)`, and
`Method.isIdempotent("POST")` is false. Same harness, first TCP connection accepted then dropped
without a response (`NoHttpResponseException`), later connections get a well-formed 200:

| request | maxRetries | connections | |
|---|---|---|---|
| POST (what `HttpTransport` emits) | 1 | 1 | **no retry**, exception propagates |
| GET (control) | 1 | 2 | retried |
| POST | 5 | 1 | **no retry** |
| GET (control) | 5 | 2 | retried |

The GET control is there to show it's the method check and not the harness. This is upstream
behaviour, not something the PR introduces, but the PR is what puts `maxRetries` in front of users,
and "retries" most naturally reads as covering a dropped connection or a reset — which is the case
that stays unretried. Overriding `handleAsIdempotent` would change it if event re-delivery is
acceptable; if it isn't, saying so in the docs is probably enough.

### 3. `retryIntervalUsesExponentialBackoffWithJitter` passes for a constant zero

Every assertion is lower-bounded at 0 (`isBetween(0L, 100L)`, `isBetween(0L, 200L)`,
`isBetween(0L, 250L)`), so an implementation that always returned `TimeValue.ZERO_MILLISECONDS`
satisfies all 300 of them — the growth the test is named after isn't pinned. The 100 iterations are
already sampling the distribution; keeping a per-bucket max and asserting the third bucket's max
exceeds the first's would catch a regression in the multiplier, and asserting the observed max for a
bucket lands near its cap would catch the cap being applied one retry too early.

### 4. New options aren't in the transport docs

`website/docs/client/java/partials/java_transport.md` lists every other `HttpConfig` field —
`timeoutInMillis`, `compression` and the rest, with defaults and "Added in vX" for the newer ones.
The four new keys aren't there, and if the interaction in (1) stays as-is, that page is where the
`retryIntervalMillis` ≤ `timeoutInMillis` constraint needs to be stated.

Happy to open a PR against your branch for any of these if that's easier than describing them.

*AI-assisted with agentic coding tools; both tables are output from the runs described above, not
inferred.*

chuenchen309:
One correction and one addition to the above.

The pin is `httpclient5:5.6.3`, not 5.6.2 — `client/java/build.gradle:79`. The harness did run against
5.6.3 and the base-class code I quoted is identical there, so nothing in the tables changes; the version
string was simply wrong in the sentence claiming the two matched.

**The addition: the Python client already answers most of these questions differently, and it is worth
seeing the two side by side before this shape sets.** `client/python/src/openlineage/client/transport/http_common.py:8`:

```python
DEFAULT_RETRY_CONFIG = {
    "total": 5, "read": 5, "connect": 5, "backoff_factor": 0.3,
    "status_forcelist": [500, 502, 503, 504],
    "allowed_methods": ["HEAD", "POST"],
}
```

Rather than compare the config to the code, I ran the same experiment against both clients — one POST to
`/api/v1/lineage`, server answers the code below once and then 200:

| server answers once | `client/python` on `main` | `client/java` + this PR |
|---|---|---|
| 500 | retried → caller sees 200 | not retried → caller sees 500 |
| 502 | retried → caller sees 200 | not retried → caller sees 502 |
| 503 | retried → caller sees 200 | retried → caller sees 200 |
| 504 | retried → caller sees 200 | not retried → caller sees 504 |
| 429 | not retried → caller sees 429 | retried → caller sees 200 |

And against a server that stays down — 503 to everything — Python makes 6 attempts with measured gaps of
`0.0, 0.6, 1.2, 2.4, 4.8` s before raising, where the Java default is 2 attempts about half a second
apart (full jitter, uniform on `[0, 1000]` ms; 100k draws gave min 0, max 1000, mean 498).

So for the failure an operator is most likely to actually hit — a collector behind a proxy returning 502,
or one that is restarting — the two clients now disagree in both directions, and the Java side is the one
that gives up. `allowed_methods: ["HEAD", "POST"]` is also the Python client having already made the
call that (2) leaves open: it decided a duplicate lineage event beats a dropped one.

I am not arguing the Python defaults are the right ones — five automatic retries into a struggling
collector is its own problem, and the flat Java keys read better than a nested dict keyed after
`urllib3.util.Retry`. The point is narrower: #4280 is open about exactly this kind of drift between the
clients, and right now these four keys are the cheapest they will ever be to rename or re-default, since
nothing depends on them yet.

*AI-assisted with agentic coding tools; all changes were reviewed and tested by me. The table is output
from running both clients, not read off their configs.*

PR #4802, opened 2026-08-04
fix(spark): skip ResolvedIdentifier silently in InputFieldsCollector

### LINKED ISSUE #4803: [BUG] InputFieldsCollector emits spurious warn for ResolvedIdentifier nodes on Spark 3.4+

Spark integration

### OpenLineage version

`io.openlineage:openlineage-spark_2.12:1.52.0` (confirmed unhandled through latest `main`)

- Apache Spark 3.4.x / 3.5.x
- Apache Iceberg (any version supporting named branch writes)
- Scala 2.12

### Environment configuration

_No special configuration required — reproduced with any valid OpenLineage Spark transport config._

### Deployment details

_Not specific to a managed service — reproducible on any on-premises or cloud Spark cluster running Spark 3.4+._

### Problem details

When writing to an Iceberg named branch via SQL on Spark 3.4+, `InputFieldsCollector` emits a spurious warning for every write:

```
Could not extract dataset identifier from
  org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier
```

`ResolvedIdentifier` is a write-target leaf node introduced in Spark 3.4 (`spark-catalyst`). During column-level lineage collection, `InputFieldsCollector` traverses the full logical plan tree. When it encounters a `ResolvedIdentifier` node, it is not in the known skip-without-warning list (`OneRowRelation`, `LocalRelation`, `ExternalRDD`, `LogicalRDD`), so it falls through to:

```java
} else if (node instanceof LeafNode) {
    log.warn("Could not extract dataset identifier from {}", node.getClass().getCanonicalName());
}
```

This produces a noisy, misleading warning on every branch-qualified write.

`ResolvedIdentifier` is a write target (not an input source), so it should be silently skipped — the same way `OneRowRelation` and `LocalRelation` are skipped.

### How to reproduce

1. Configure the OpenLineage Spark agent on a Spark 3.4+ cluster with Iceberg.
2. Execute a write to a named Iceberg branch:
   ```sql
   INSERT INTO spark_catalog.my_db.my_table.branch_my_branch
   SELECT * FROM source_table
   ```
3. Observe the Spark driver logs — the following warning appears:
   ```
   Could not extract dataset identifier from
     org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier
   ```

### Additional details

`ResolvedIdentifier` lives in `org.apache.spark.sql.catalyst.analysis` (in `spark-catalyst`), not in `spark-sql`. The `spark3` module compiles against Spark 3.2.4 where the class does not exist, so the fix uses a class-name string comparison to avoid a compile-time dependency on Spark 3.4+ APIs.

Fix in PR #4802.

rahul139vermaa:
## Summary

When running on Spark 3.4+, `InputFieldsCollector` traverses the full logical plan tree during column-level lineage collection. `ResolvedIdentifier` (`org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier`) nodes appear in these plans — they are write targets resolved by Spark's analyzer — but were not in the known-skip list. Because `ResolvedIdentifier` is a `LeafNode`, every occurrence fell through to the warn branch:

```
log.warn("Could not extract dataset identifier from {}", node.getClass().getCanonicalName());
// → "Could not extract dataset identifier from
//    org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier"
```

This warning fires on every write operation that uses a branch-qualified Iceberg table reference on Spark 3.4+.

## Fix

Add `ResolvedIdentifier` to the skip-without-warning condition in `InputFieldsCollector`. The check uses a class-name string comparison so the `spark3` module — which compiles against Spark 3.2.4 where the class does not exist — has no compile-time dependency on Spark 3.4+ APIs:

```java
|| "org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier"
    .equals(node.getClass().getCanonicalName())
```

## Out of scope

The related issue (missing output datasets when writing to Iceberg named branches) was initially attributed to `AppendData.table()` returning a `ResolvedIdentifier`. Investigation showed this is not the case: `AppendData.table()` is statically typed as `NamedRelation` and `ResolvedIdentifier` does not implement `NamedRelation`. The root cause of the missing output dataset remains under investigation and will be addressed in a separate PR.

## Test plan

- [x] `InputFieldsCollectorTest` passes (10 tests)
- [x] No new compile-time dependency on Spark 3.4+ APIs in the `spark3` module

Closes #4803 (partially — spurious warning fix only)

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/InputFieldsCollector.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/InputFieldsCollector.java
@@ -158,7 +158,11 @@ public class InputFieldsCollector {
     } else if (node instanceof OneRowRelation
         || node instanceof LocalRelation
         || node instanceof ExternalRDD
-        || node instanceof LogicalRDD) {
+        || node instanceof LogicalRDD
+        // ResolvedIdentifier (Spark 3.4+) is a write target, not an input source — skip silently.
+        // Class name check avoids a compile-time dependency on Spark 3.4+ APIs.
+        || "org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier"
+            .equals(node.getClass().getCanonicalName())) {
       // skip without warning
     } else if (node instanceof LeafNode) {
       log.warn("Could not extract dataset identifier from {}", node.getClass().getCanonicalName());


mobuchowski (approved):

mobuchowski:
@rahul139vermaa can you fix PMD warning?

```
  * file: /home/circleci/openlineage/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/column/InputFieldsCollector.java
      src:  InputFieldsCollector.java:164:164
      rule: AvoidLiteralsInIfCondition
      msg:  Avoid using Literals in Conditional Statements
      code: || "org.apache.spark.sql.catalyst.analysis.ResolvedIdentifier"



  Summary:

  io.openlineage.spark3.agent.lifecycle.plan.column.InputFieldsCollector : 1
  * warnings: 1
  ```

rahul139vermaa:
Closing this PR , Its not significant change, I was targeting to fix output in lineage, raising new PR

PR #4804, opened 2026-08-04
spec: add explicit lineage facet

mobuchowski:
Uses the discriminated-union generator support merged in #4778.

Follows the proposal merged in #4468 and is related to #4359.

Add Job and Dataset facets for declaring exact dataset-, field-, and job-level lineage relationships.

### Meaningful description

OpenLineage currently infers dataset-level lineage from the Cartesian product of a run's `inputs` and `outputs`. Consider a bulk job that reads `customers` and `orders`, then independently writes `customer_summary` and `order_summary`:

```text
actual:   customers -> customer_summary
          orders    -> order_summary

inferred: customers -> customer_summary
          customers -> order_summary       (false)
          orders    -> customer_summary    (false)
          orders    -> order_summary
```

Column lineage can remove that ambiguity when every producer knows field mappings, but integrations often know exact dataset edges without knowing columns. The current model also cannot directly describe dataset-to-dataset lineage without a job, or job-to-job data flow without an intermediate tracked dataset.

This PR adds a versioned `LineageFacet.json` containing:

- `LineageJobFacet` for RunEvent and JobEvent job facets;
- `LineageDatasetFacet` for DatasetEvent dataset facets;
- discriminated Dataset/Job unions for target entries and source inputs;
- optional field-level lineage, transformations, and run-specific job identities.

Both facet types intentionally use the same `lineage` container key. `LineageEntry` and `LineageInput` use singleton `DATASET`/`JOB` enums so generated clients retain their concrete runtime types.

The change is backwards compatible: it adds new optional facets at schema version `1-0-0` and does not change existing event fields or semantics when the facet is absent.

The PR also:

- registers the core facet, adds seven schema fixtures, and stores the website snapshot;
- documents the Job and Dataset forms and updates proposal examples to versioned schema fragments;
- generates Python and Go models and exposes both facet locations;
- extends Go discovery and union generation for a shared `anyOf` key and singleton-enum discriminators;
- adds Java round-trip and Python serialization tests, including the meaningful explicit `inputs: []` generator case.

Automatic legacy/modern translation and native Spark, Flink, and dbt emission are intentionally deferred to separate follow-up PRs.

### Validation

- `prek run --all-files`: all schema, fixture, snapshot, generation, mypy, PMD, Spotless, formatting, license, and redaction hooks passed. The two container-wrapped hooks could not connect to the local Podman socket.
- Native `golangci-lint run --config .golangci.yml ./...`: 0 issues.
- Python client: 641 tests passed with all extras.
- Java root client: 461 tests passed, including both new round-trip tests. 
- Go client and generator: all tests passed.

### Checklist


website/static/spec/facets/1-0-0/LineageFacet.json
```diff
@@ -0,0 +1,257 @@
...
+          }
+        }
+      },
+      "required": ["namespace", "name", "type"]
+    },
+    "LineageJobEntry": {
+      "type": "object",
+      "description": "Describes data flowing into a target job. Without namespace and name, the target is the event's own job.",
+      "properties": {
+        "namespace": {
```
jakub-moravec, line 104: The proposal states (Semantic Constraint 10):

> A producer MUST emit `namespace` and `name` together or omit them together.

Neither `LineageJobEntry` nor `LineageJobInput` enforces this in the schema. The schema has
`"required": ["type"]` only. There is no `if/then/else` or `dependentRequired` guard to
prevent a partially-identified job reference (e.g. `namespace` present but `name` absent).

A partially-identified job reference is semantically undefined and would cause silent, difficult
to debug lineage corruption in consumers.

Can we capture this by adding `dependentRequired` constraint:

```json
"dependentRequired": {
  "namespace": ["name"],
  "name": ["namespace"]
}
```
?

client/python/src/openlineage/client/generated/lineage.py
```diff
@@ -0,0 +1,246 @@
...
+    """Whether the transformation masks the source data."""
+
+    _skip_redact: ClassVar[list[str]] = ["type", "subtype", "masking"]
+
+    @staticmethod
+    def _get_schema() -> str:
+        return "https://openlineage.io/spec/facets/1-0-0/LineageFacet.json#/$defs/LineageTransformation"
+
+
+class Type(Enum):
```
jakub-moravec: Should these generated enums have these generic names?

client/python/src/openlineage/client/generated/lineage.py
```diff
@@ -0,0 +1,246 @@
...
+    namespace: str
+    """The namespace of the target dataset."""
+
+    name: str
+    """The name of the target dataset."""
+
+    type: Type  # noqa: A003
+    """The target entity type. Must be DATASET."""
+
+    inputs: list[LineageInput] | None = attr.field(factory=list)
```
jakub-moravec: If the default is an empty list (not `None`), then every constructed
`LineageFacet` without explicit `inputs` will serialize `"inputs": []`, which a consumer
would interpret as "this dataset has no upstream source" rather than "this facet carries no
dataset-level information." .. right?

client/python/src/openlineage/client/generated/lineage.py
mobuchowski: The name is usually used together with the module name, so `lineage.Type` rather than naming them more specifically and using `lineage.LineageType` - IMO that's okay.

client/python/src/openlineage/client/generated/lineage.py
mobuchowski: You're right

client/python/src/openlineage/client/generated/lineage.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+@attr.define
+class LineageJobInput(RedactMixin):
+    """
+    A source job that feeds data into a lineage target. Without namespace and name, the source is the
+    event's own job.
+    """
+
+    type: Type3  # noqa: A003
```
kacpermuda: I think when spec defines something as string with allowed enum values, we should adjust the generator to use Literal ( e.g. here and in all other places that have Type, Type1 etc. now). This is the native way in python to specify allowed string values, not these artificial "Type" classes

website/static/spec/facets/1-0-0/LineageFacet.json
mobuchowski: Will fix in follow-up https://github.com/OpenLineage/OpenLineage/pull/4819

client/python/src/openlineage/client/generated/lineage.py
mobuchowski: Fixed in the latest commit

kacpermuda (approved):
Looks good, looked mostly at python part and spec, would love to see more reviews here before we merge

jakub-moravec (approved):

PR #4819, opened 2026-08-07
clients: support dependentRequired in generators

mobuchowski:
## Summary

Add `dependentRequired` support to the Java, Python, and Go client generators.

## Motivation

The lineage facet can identify another job by `namespace` and `name`, or omit both fields to refer to the event's own job. A partial identity is invalid because it cannot identify either case:

```json
{
  "type": "JOB",
  "namespace": "https://example.com/jobs"
}
```

The schema uses symmetric `dependentRequired` constraints to require `name` whenever `namespace` is present and vice versa. Before this change, all three generators ignored that keyword, so generated clients could construct and serialize payloads that the OpenLineage schema rejects.

## Changes

- Java parses and merges `dependentRequired` constraints, then emits constructor validation used by builders and Jackson deserialization.
- Python extracts the constraints into template metadata and emits attrs field validators.
- Go carries the constraints through its IR and emits `Validate`, `MarshalJSON`, and `UnmarshalJSON` methods, preserving existing constructor signatures.
- Regenerate the Python and Go lineage models and add generator- and client-level coverage for valid and invalid field combinations.

### DIFF
--- a/client/java/generator/src/test/resources/discriminated-union/SharedFacet.json
+++ b/client/java/generator/src/test/resources/discriminated-union/SharedFacet.json
@@ -75,7 +75,14 @@
         },
         "namespace": {
           "type": "string"
+        },
+        "name": {
+          "type": "string"
         }
+      },
+      "dependentRequired": {
+        "namespace": ["name"],
+        "name": ["namespace"]
       }
     },
     "LineageInput": {
@@ -113,7 +120,14 @@
         },
         "name": {
           "type": "string"
+        },
+        "namespace": {
+          "type": "string"
         }
+      },
+      "dependentRequired": {
+        "namespace": ["name"],
+        "name": ["namespace"]
       }
     },
     "LegacyChoice": {
--- a/client/python/src/openlineage/client/generator/templates/dataclass.jinja2
+++ b/client/python/src/openlineage/client/generator/templates/dataclass.jinja2
@@ -198,4 +198,17 @@ class {{ class_name }}:
 {%- for field in fields -%}
     {{ validators.validator(field.name, field.type_hint, field.required) }}
 {%- endfor -%}
+{#- add object-level dependentRequired validators -#}
+{%- for property, dependencies in (dependent_required | default({}) | dictsort) %}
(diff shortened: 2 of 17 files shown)


client/python/src/openlineage/client/generator/templates/dataclass.jinja2
```diff
@@ -198,4 +198,17 @@ class {{ class_name }}:
...
 {%- endfor -%}
+{#- add object-level dependentRequired validators -#}
+{%- for property, dependencies in (dependent_required | default({}) | dictsort) %}
+    @{{ property }}.validator
+    def {{ property | lower }}_dependent_required_check(
+        self, attribute: Any, value: Any
+    ) -> None:  # noqa: ARG002
+        if value is None:
+            return
+        missing = [name for name in {{ dependencies }} if getattr(self, name) is None]
```
kacpermuda, line 209: Does this only work strings? If we have some lists with default=[], should this also work? Ok if not, just my first thought.

kacpermuda (approved):
Confirmed that python part looks good, java and go as well, but @tnazarew can take a look since he wrote this go generator

client/java/generator/src/test/resources/discriminated-union/SharedFacet.json
```diff
@@ -75,7 +75,14 @@
         },
         "namespace": {
           "type": "string"
+        },
+        "name": {
+          "type": "string"
         }
+      },
+      "dependentRequired": {
```
dolfinus, line 83: Why not using `required` instead?

mobuchowski: We need either namespace and name, or none - it does not make sense to require one of them. And if there is none, we assume it refers to current job.

client/python/src/openlineage/client/generator/templates/dataclass.jinja2
mobuchowski: It should work on lists too, however we use this for name/namespace only now

PR #4824, opened 2026-08-08
python: keep async worker alive after wait

### LINKED ISSUE #4823: python: wait_for_completion shuts down AsyncHttpTransport worker

Python client

### OpenLineage version

`main` at `7aba69854131da9f9e052db1cecde577280859d1`; also reproducible in `1.52.0`.

### Problem details

`AsyncHttpTransport.wait_for_completion()` sets the internal `may_exit` event. Once the current queue and active tasks drain, the worker thread exits even though the transport has not been closed.

A later `emit()` still accepts an event and increments the pending count, but no worker remains to process the queue. Later waits time out and the event is lost when the process exits.

The documented API treats these as separate operations:

- `wait_for_completion(timeout)`: wait for queued events to be processed
- `close(timeout)`: shut down the transport

### How to reproduce

```python
transport = AsyncHttpTransport(AsyncHttpConfig(url="http://backend"))
assert transport.wait_for_completion(timeout=1)

# Allow the worker to observe the exit signal set by the wait.
transport.worker_thread.join(timeout=2)
assert not transport.worker_thread.is_alive()

transport.emit(event)
assert not transport.wait_for_completion(timeout=1)
assert transport.get_stats()["pending"] == 1
```

### Expected behavior

`wait_for_completion()` should only wait. The worker should remain available for later events until `close()` requests shutdown.

### Proposed fix

Move the `may_exit.set()` call from `wait_for_completion()` to `close()` and add regression assertions covering both lifecycle methods.

mattfaltyn:

Keep `AsyncHttpTransport` usable after `wait_for_completion()` returns.

### Meaningful description

`wait_for_completion()` currently sets the worker soft-exit signal. Once the queue drains, the worker exits even though the transport has not been closed. A later `emit()` accepts an event but leaves it permanently queued with no worker to process it.

Move the exit signal to `close()`, where shutdown belongs. This preserves current close behavior while making the documented wait operation lifecycle-neutral. Regression assertions verify that waiting does not request worker exit and closing still does.

Closes: #4823

### Validation

- `uv run --extra test pytest tests/test_async_http.py -q` — 46 passed
- Full Python suite excluding the separately run HTTP integration module — 606 passed
- `pytest tests/test_http_integration.py -q` — 34 passed
- strict mypy on `async_http.py`
- Ruff check and format check on both changed files
- `prek run --files client/python/src/openlineage/client/transport/async_http.py client/python/tests/test_async_http.py`
- Original post-wait emission reproduction now drains successfully and leaves the worker alive until close

### Checklist


kacpermuda:
Can we add a dedicated test for this behavior? I saw that the closed #4834 had one.

mobuchowski (approved):

mattfaltyn:
> Can we add a dedicated test for this behavior? I saw that the closed #4834 had one.

Added a regression test in [b03c257](https://github.com/OpenLineage/OpenLineage/commit/b03c257ac98b1fe17a2e2d7af52d83d6ca144cd8) that calls `wait_for_completion()`, emits another event, and verifies the worker processes it. All 641 local Python tests pass.

PR #4834, opened 2026-08-09
fixed bug in the Python clients's async HTTP transport

pavankumarkodati:
### Checklist


kacpermuda:
That's a duplicate of #4824 , let's use the other one for discussion / improvements since it's better documented

PR #4838, opened 2026-08-10
client/java: make ReducedDataset facet comparison null-safe

### LINKED ISSUE #4836: [BUG] client/java: NullPointerException in ReducedDataset.hasSameFacets when Dataset.getFacets() is null — dataset reduction crashes and the event is dropped

Java client, Spark integration

### OpenLineage version

OpenLineage 1.50.0 (java client embedded in the Spark agent)

Spark 4.1.0

### Environment configuration

_No response_

### Deployment details

_No response_

### Problem details

Dataset partition-reduction compares facets with a direct method call on a nullable value:

```java
// client/java/src/main/java/io/openlineage/client/dataset/partition/ReducedDataset.java
private boolean hasSameFacets(ReducedDataset other) {
    if (!dataset.getFacets().equals(other.dataset.getFacets())) {   // line ~89: NPE if facets null
      return false;
    }
    ...
}
```

`OpenLineage.Dataset.getFacets()` is nullable — datasets built without facets (common on the
Spark RDD path, where `RddExecutionContext.buildInputs` constructs minimal datasets) return null.
The first `reduce` call then throws `NullPointerException`, which propagates up through
`DatasetReducer.reduceInputs` into the listener, where the circuit breaker swallows it — the
**entire lineage event is silently discarded**, not just the comparison.

_No response_

### How to reproduce

```java
OpenLineage ol = new OpenLineage(URI.create("http://test"));
InputDataset d1 = ol.newInputDatasetBuilder()
    .namespace("hdfs://ns").name("/data/table/part=1").build();   // facets == null
InputDataset d2 = ol.newInputDatasetBuilder()
    .namespace("hdfs://ns").name("/data/table/part=2").build();   // facets == null

DatasetConfig config = new DatasetConfig();                       // default trimmers
new DatasetReducer(config).reduceInputs(Arrays.asList(d1, d2));   // → NullPointerException
```

### Additional details

_No response_

MSDehghan:
Closes #4836

Prevent dataset reduction failures when dataset facets are null.

### Meaningful description
ReducedDataset.hasSameFacets previously called equals directly on Dataset.getFacets(). Generated or mocked datasets can have null facets, causing a NullPointerException during partition-aware dataset reduction and aborting reduction for the event.

Use Objects.equals for the dataset-facet comparison, consistent with the existing null-safe input and output facet comparisons. Regression tests cover both null/null facets, which should reduce successfully, and null/non-null facets in both orders, which should remain distinct.

### Validation
- git diff --check
- Focused Gradle test and Spotless check could not run locally because JDK 17 is unavailable in the current environment; GitHub CI will run the project checks.

### Checklist

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/dataset/partition/ReducedDataset.java
+++ b/client/java/src/main/java/io/openlineage/client/dataset/partition/ReducedDataset.java
@@ -86,7 +86,7 @@ class ReducedDataset {
   }
 
   private boolean hasSameFacets(ReducedDataset other) {
-    if (!dataset.getFacets().equals(other.dataset.getFacets())) {
+    if (!Objects.equals(dataset.getFacets(), other.dataset.getFacets())) {
       return false;
     }
 
--- a/client/java/src/test/java/io/openlineage/client/dataset/partition/ReducedDatasetTest.java
+++ b/client/java/src/test/java/io/openlineage/client/dataset/partition/ReducedDatasetTest.java
@@ -9,7 +9,10 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import io.openlineage.client.OpenLineage.Dataset;
+import io.openlineage.client.OpenLineage.DatasetFacets;
+import io.openlineage.client.OpenLineage.InputDataset;
 import io.openlineage.client.dataset.DatasetConfig;
+import java.util.Optional;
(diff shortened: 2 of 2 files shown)


MSDehghan:
@mobuchowski @kacpermuda 
Could you please review this small patch?

harels (approved):
thanks for the fix!

PR #4840, opened 2026-08-10
client: add OL name escaping

jakub-moravec:
Documenting how OL names should be escaped, and adding automated configurable escaping to OL clients 

### Meaningful description
Non-escaped dots in OL names break reconciliation of datasets and jobs. This PR documents how to escape the dots and adds automation to the client libraries that escapes them automatically. 

Slack discussion https://openlineage.slack.com/archives/C065PQ4TL8K/p1783674560023109

### Checklist

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
+++ b/client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
@@ -1,0 +1,66 @@
+/*
+/* Copyright 2018-2026 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client.naming;
+
+import org.apache.commons.lang3.Strings;
+
+/**
+ * Utility class for escaping dots in OpenLineage name segments.
+ *
+ * <p>OpenLineage names are structured as dot-separated segments, e.g. {@code
+ * {database}.{schema}.{table}}. When a segment itself contains a literal dot (e.g. an Oracle
+ * service name {@code mydb.example.com}), the dot must be escaped so that consumers can
+ * unambiguously split the name into its constituent parts.
+ *
+ * <p>The escaping rule (from the naming specification) is: a literal {@code .} inside a segment is
+ * written as {@code \\.}.
+ *
+ * <p>Escaping is <em>enabled by default</em> and can be disabled by setting the environment
+ * variable {@code OPENLINEAGE__NAME__ESCAPING} to {@code false} (case-insensitive). This is
+ * intended as an escape hatch for producers that emit names that are already pre-escaped or that
+ * target consumers which do not yet support the escaping convention.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * // "mydb\\.example\\.com.mySchema.myTable"
+ * NameEscaping.escapeSegment("mydb.example.com") + "." + "mySchema" + "." + "myTable"
+ * }</pre>
+ */
+public final class NameEscaping {
+
+  private static final String ENV_VAR = "OPENLINEAGE__NAME__ESCAPING";
+
+  private NameEscaping() {}
+
+  /**
+   * Returns {@code true} if dot-escaping is enabled (the default).
+   *
+   * <p>Escaping can be disabled by setting the environment variable {@code
+   * OPENLINEAGE__NAME__ESCAPING=false} (case-insensitive).
+   *
+   * @return {@code true} when escaping is active
+   */
+  public static boolean isEscapingEnabled() {
+    return !Strings.CI.equals("false", System.getenv(ENV_VAR));
+  }
+
+  /**
+   * Escapes dots in a single name segment when escaping is enabled.
+   *
+   * <p>A literal {@code .} is replaced with {@code \\.} so that consumers can tell structural dots
+   * (separating segments) from literal dots that are part of a segment value.
+   *
+   * <p>The transformation is applied only when {@link #isEscapingEnabled()} returns {@code true};
+   * otherwise the segment is returned unchanged.
+   *
+   * @param segment a single name component (e.g. database, schema, table)
+   * @return the segment with literal dots escaped, or unchanged when escaping is disabled
+   */
+  public static String escapeSegment(String segment) {
+    return isEscapingEnabled() ? segment.replace(".", "\\.") : segment;
+  }
+}
--- a/client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
+++ b/client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
@@ -20,6 +20,12 @@ package io.openlineage.client.naming;
  * variable {@code OPENLINEAGE__NAME__ESCAPING} to {@code true} (case-insensitive), or by setting
  * {@code name.escaping: true} in the YAML configuration.
  *
+ * <p>When a {@link NameConfig} is available (e.g. loaded from YAML), prefer the overloads that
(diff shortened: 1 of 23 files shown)


website/docs/client/python/configuration.md
```diff
@@ -185,11 +185,12 @@ transport:
...
+|------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------|--------|
+| OPENLINEAGE_CONFIG           | The path to the YAML configuration file                                                                                                         | path/to/openlineage.yml |        |
+| OPENLINEAGE_CLIENT_LOGGING   | Logging level of OpenLineage client and its child modules                                                                                       | DEBUG                   |        |
+| OPENLINEAGE_DISABLED         | When `true`, OpenLineage will not emit events (default: false)                                                                                  | false                   | 0.9.0  |
+| OPENLINEAGE__NAME__ESCAPING  | Controls automatic dot-escaping in name segments produced by the naming helpers. Set to `false` to disable. Defaults to enabled (escaping on). See [Naming — Escaping](../../spec/naming.md#automatic-escaping-in-openlineage-clients). | true | |
```
kacpermuda, line 193: Regardless of the final form that we agree on, it should be disabled by default for sure. It's changing a crucial part of the dataset, so it should be an opt-in for now, that we can later switch to true by default.

jakub-moravec:
@kacpermuda @mobuchowski I updated the client code to escape by default

client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
```diff
@@ -0,0 +1,63 @@
...
+   * Returns {@code true} if dot-escaping is enabled.
+   *
+   * <p>Escaping is <em>disabled by default</em>. It can be enabled by setting the environment
+   * variable {@code OPENLINEAGE__NAME__ESCAPING=true} (case-insensitive).
+   *
+   * @return {@code true} when escaping is active
+   */
+  public static boolean isEscapingEnabled() {
+    return Boolean.valueOf(System.getenv(ENV_VAR));
+  }
```
mobuchowski, line 46: When only name.escaping: true is set in YAML, this method still returns false because it reads only the process environment; the newly parsed Java NameConfig is never consulted. 

Python has the same problem because OpenLineageConfig.from_dict discards name and escape.py only reads os.environ, so the advertised YAML option does nothing in either client. Wire the loaded setting through both implementations or remove the unsupported mode;

website/docs/spec/naming.md
```diff
@@ -92,10 +92,40 @@ run = Run(str(generate_new_uuid()))
...
+This focus enabled by naming is key to the production of useful lineage, as consistent naming conventions are one of the mechanisms that OpenLineage consumers can use to reconcile entities captured by different OpenLineage producers. 
+
+While it is also recommended for OpenLineage producers and consumers to support [Hierarchy Dataset facet](https://openlineage.io/docs/spec/facets/dataset-facets/hierarchy) to make the reconciliation process more reliable, the facet is optional and compliance with defined naming conventions is still essential. 
 
 ![image](./naming-correlations.svg)
 
+## Escaping
+
+As OpenLineage Dataset and Job names follow the patterns defined in the naming conventions described above, it is necessary to allow for escaping of special characters - specifically it is necessary to escape dots (`.`) in OpenLineage name segments. The escaping sequence applied should be `\\`. OpenLineage producers should automatically apply the escaping sequence and OpenLineage consumers should reflect it in case they are processing the OpenLineage names and are deducing hierarchy levels from it (e.g. `database.schema.table`).
```
mobuchowski, line 103: When escaping is enabled and a Spark job segment contains a dot, such as appName = "my.app", io.openlineage.client.job.Naming.Spark#getName() still concatenates raw segments and produces my.app.command.table; only dataset naming helpers were updated. Spark uses this helper, so job names remain ambiguous despite this new rule covering both datasets and jobs.

website/docs/spec/naming.md
```diff
@@ -92,10 +92,40 @@ run = Run(str(generate_new_uuid()))
...
+To disable automatic escaping, set the following environment variable before the process starts:
+
+```sh
+OPENLINEAGE__NAME__ESCAPING=false
+```
+
+| Value | Behaviour |
+|-------|-----------|
+| _(unset or any other value)_ | Escaping **enabled** (default) |
+| `false` | Escaping **disabled** |
```
mobuchowski, line 127: With OPENLINEAGE__NAME__ESCAPING unset, all three implementations return false and enable escaping only for the value true. This table and  both client configuration pages state the inverse, so users following the documentation will emit unescaped names while believing the feature is active; the example should set the variable to true and describe disabled as the default.

mobuchowski (changes requested):
Adding some AI findings that seem right to me

website/docs/spec/naming.md
jakub-moravec: I missed this when changing the default, thanks

jakub-moravec:
Thanks for catching these, @mobuchowski. I added fixes and rebased

client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
mobuchowski: @jakub-moravec I don't think this is fixed?

client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
```diff
@@ -0,0 +1,103 @@
...
+   * variable.
+   *
+   * <p>Use {@link #escapeSegment(String, NameConfig)} when a {@link NameConfig} is available.
+   *
+   * @param segment a single name component (e.g. database, schema, table)
+   * @return the segment with literal dots escaped, or unchanged when escaping is disabled
+   */
+  public static String escapeSegment(String segment) {
+    return isEscapingEnabled() ? segment.replace(".", "\\.") : segment;
+  }
```
mobuchowski, line 85: Java, Python, and Go only replace . with \.. A segment ending in \ followed by a structural dot becomes indistinguishable from an escaped literal dot.

website/docs/client/python/configuration.md
```diff
@@ -186,11 +186,12 @@ transport:
...
```
mobuchowski, line 194: @jakub-moravec it still says

"Defaults to enabled (escaping on)."

website/docs/spec/naming.md
```diff
@@ -92,10 +92,40 @@ run = Run(str(generate_new_uuid()))
...
+- Schema Name = `mySchema`
+- Table Name = `myTable`
+
+The escaped OpenLineage name of the dataset should be `mydb\\.example\\.com.mySchema.myTable`.
+
+Without the escaping mechanims in place, consumers would have no way how to correctly segment the OpenLineage name based on the pattern defined in the naming conventions.
+
+The Java, Python, and Go OpenLineage clients **automatically escape dots** in all name segments produced by their built-in naming helpers (e.g. `Naming.Oracle`, `Oracle(...)`, etc.). Escaping is **enabled by default**.
+
+To disable automatic escaping, set the following environment variable before the process starts:
```
mobuchowski, line 118: This contradicts below table

mobuchowski (changes requested):

client/java/src/main/java/io/openlineage/client/naming/NameEscaping.java
jakub-moravec: @mobuchowski 

One way to adress it could be to escape 

- `\` to `\\`
- `.` to `\.`

I added this together with some extra unit tests for weird combinations, but if you have a better idea how to adress this please lmk

jakub-moravec:
@mobuchowski could you approve the pending check please?

PR #4504, opened 2026-05-01
spec: fix DataQualityAssertionsDatasetFacet to inherit from DatasetFacet

creazyfrog:
## Summary

Fixes #1876

`DataQualityAssertionsDatasetFacet` was incorrectly inheriting from `InputDatasetFacet`. Data quality assertions apply to any dataset (input **or** output), so the facet should extend `DatasetFacet` — consistent with every other dataset facet in the spec:

- `DocumentationDatasetFacet` → `DatasetFacet`
- `SchemaDatasetFacet` → `DatasetFacet`
- `DatasetVersionDatasetFacet` → `DatasetFacet`
- `DataQualityAssertionsDatasetFacet` → ~~`InputDatasetFacet`~~ → `DatasetFacet` ✅

## Change

Single-line fix in `spec/facets/DataQualityAssertionsDatasetFacet.json`:

```diff
-  "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/InputDatasetFacet"
+  "$ref": "https://openlineage.io/spec/2-0-2/OpenLineage.json#/$defs/DatasetFacet"
```

## Checklist

- [x] Single file change, no logic affected
- [x] Aligns with all other dataset facets in the spec

---

> ⚠️ **AI Disclosure** — This pull request was generated with AI assistance (Claude Code). The fix was reviewed for correctness before submission.

PR #4845, opened 2026-08-11
spec: make DataQualityAssertionsDatasetFacet a DatasetFacet

### LINKED ISSUE #1876: DataQualityAssertionsDatasetFacet incorrectly inherits from InputDatasetFacet

chuenchen309:
closes: #1876
related: #4504, #4483

`DataQualityAssertionsDatasetFacet` now inherits from `DatasetFacet`, so it is attached under a dataset's `facets` instead of `inputFacets`.

### Meaningful description

The facet inherited from `InputDatasetFacet`, which restricted it to a dataset's `inputFacets`. Three places in the repo already behave as if it were a plain dataset facet:

- its own documentation page shows the example under `"facets"`, not `"inputFacets"`;
- `DbtArtifactProcessor` wrote it under **both** keys, with `# TODO: remove this next release` and a `# type: ignore` next to the `facets=` half;
- `DatasetReducer` carries `facets` through wholesale (`.facets(source.getFacets())`) but has to copy each input facet by hand, so the assertion facet survived reduction only because `FacetUtils` listed it explicitly.

This takes over #4504 (@creazyfrog, open since May 1, tracking #1876 from 2023), which implements exactly this and has been inactive since @mobuchowski's "the fix looks very mechanical; makes sense" on May 8. @kacpermuda suggested picking it up on #4483. The spec and client changes are that PR's; the fixtures are redone as a pure mechanical transform, because the original also reindented three of them from 4 spaces to 2, which is where most of its ~3000-line diff came from.

**Spec.** `allOf` now references `DatasetFacet`, and `$id` goes 1-1-0 → 1-2-0. A REVISION bump rather than an ADDITION: every instance valid under 1-1-0 is still valid (the only new property, `_deleted`, is optional), but where the facet may legally appear changes, which is not a purely additive change for consumers.

**Fixtures.** 16 `dataQualityAssertions` entries moved from `inputFacets` to `facets`; 10 dropped rather than moved, because `no_test_metadata` goes through the `parse_test` double-write path and carried the entry under both keys; 14 `_schemaURL` values bumped to 1-2-0. Those `_schemaURL`s postdate #4504, so they were not in it — as does one `inputFacets` occurrence in `test/result.json`. Each file keeps its own indentation.

The 10 pairs are not quite duplicates, which matters for what dropping one of them costs. In 2 of the 10 datasets the `inputFacets` copy asserted `actual` and `expected` on the two failing assertions and the `facets` copy did not, so keeping only the `facets` copy stops the fixture asserting those two fields. It stays green either way, because `openlineage.common.test.match` only checks keys present in the expected value — which is also why the shorter copy passed all along. e2a7d21 adds the two fields to the surviving copy. Measured with that same `match`, the number of datasets whose surviving copy no longer asserts what the dropped copy did goes from 2 to 0.

**Go client.** Regenerated with `generator/go`, since `build-client-go` fails when the checked-in files drift from the spec. The struct moves from `InputDatasetFacets` to `DatasetFacets` and gains the `Deleted` field every `DatasetFacet` carries.

**Docs.** a6d960e moves the facet's entry out of the input-facet list in `spec/OpenLineage.md` and `website/docs/spec/object-model.md`, which both still filed it under input dataset facets. The facet's own page already showed the example under `facets`, so 869296a8 only bumps its spec link to 1-2-0; that is the last reference to 1-1-0 outside the versioned snapshots under `website/static/spec/`.

#### Verification

Run against `main` at 7c5f2e767, in a dedicated venv with `client/python` and `integration/common` both installed editable:

| check | result |
|---|---|
| `integration/common` `tests/` | 357 passed, 2 skipped |
| `integration/common` `tests/dbt/` | 243 passed |
| `mypy --ignore-missing-imports src` (integration/common) | no issues in 22 files |
| `client/python` `tests/` | 588 passed; the 18 failures and 34 errors are identical by name to a clean `main` checkout in the same environment (missing `confluent-kafka` / `fsspec` extras) |
| `client/python` `tests/test_facet_v2.py` | 16 passed |
| `ol-generate-code spec/facets` | emits `class DataQualityAssertionsDatasetFacet(DatasetFacet)`, matching the checked-in file |
| `.pre_commit/json-schema/test-facets.sh` on the changed spec | schema ok, `spec/tests/DataQualityAssertionsDatasetFacet/1.json` ok |
| `ruff check` / `ruff format --check` on changed Python | clean |
| `client/go`: `go run .` in `generator/go`, then `go build ./...`, `go test ./...` | generator output committed; build clean; tests pass |

Two things I could not run locally: the Java build (no JVM on this machine — `getDataQualityAssertions` has no callers left besides the `FacetUtils` line removed here, but CI is the check that matters), and the Ruby fluentd tests.

One deliberate omission: `proxy/fluentd/events/event_full.json` still carries the facet under `inputFacets`. It stays schema-valid there, and the parser test only asserts on `runId` and `outputStatistics`, so I left that subsystem alone rather than change a sample I cannot run the tests for. Happy to move it if you would rather the repo be self-consistent.

### Checklist


integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -492,9 +492,7 @@ def parse_test(self, context: DbtRunContext, nodes: dict) -> DbtEvents:
                         InputDataset(
                             namespace=namespace,
                             name=name,
-                            inputFacets=dataset_facets,
```
kacpermuda, line 495: Not sure how breaking this is, but I'm okay with that change. cc @mobuchowski

client/python/tests/example_full_event.json
```diff
@@ -51,7 +51,8 @@
     {
       "namespace": "my-datasource-namespace",
       "name": "instance.schema.table",
-      "inputFacets": {
+      "inputFacets": {},
+      "facets": {
```
kacpermuda, line 55: Same, looks breaking, but I think we're moving away from inputFacets and outputFacets anyway, so facets should always be checked as fallback.

integration/common/tests/dbt/no_test_metadata/result.json
```diff
@@ -53,29 +53,7 @@
                 "facets": {
                     "dataQualityAssertions": {
                         "_producer": "https://github.com/OpenLineage/OpenLineage/tree/0.0.1/integration/dbt",
-                        "_schemaURL": "https://openlineage.io/spec/facets/1-1-0/DataQualityAssertionsDatasetFacet.json#/$defs/DataQualityAssertionsDatasetFacet",
```
kacpermuda, line 56: This is just removing duplicates, it was under inputFacets and facets, now only facets. Nice.

client/java/src/main/java/io/openlineage/client/dataset/FacetUtils.java
```diff
@@ -18,7 +18,6 @@ public static InputDatasetInputFacetsBuilder toBuilder(
     if (facets == null) return builder;
 
     builder
-        .dataQualityAssertions(facets.getDataQualityAssertions())
```
kacpermuda, line 21: Don't we need to put it somewhere else or was it duplicated there already?

kacpermuda (approved):
Looks good to me, left some comments to verify. Also, since this is a spec change and can break stuff, will wait for other folks to review it as well.

kacpermuda:
@chuenchen309 would you like to address some comments and fix the conflicts on this one?

PR #4847, opened 2026-08-11
feat(dbt): add top-level dbt invocation events in artifact processor

### LINKED ISSUE #4720: [FEATURE] Add top-level dbt invocation events in the artifact processor
## Description

The dbt structured logs integration already emits a top-level dbt command (invocation) run:

```text
dbt command
├── dbt node (model/snapshot/test)
│   ├── SQL query
│   ├── SQL query
│   └── ...
```

Specifically, it emits:
- START on `MainReportVersion`
- COMPLETE/FAIL on `CommandCompleted`
and parents all node events to this command-level run.

The artifact-based processor (`processor.py`) currently does not have an equivalent abstraction. It directly emits node-level events (models, snapshots, tests), with no common parent.

## Proposal

Extend the artifact processor to emit a top-level dbt invocation run.

Suggested hierarchy:

```text
Airflow task / external orchestrator (as example)
└── dbt invocation
    ├── model run
    ├── snapshot run
    ├── test run
    └── ...
```

The invocation run should:

- emit START/COMPLETE (or FAIL) events,
- become the parent of all node-level events emitted by the processor,
- use the `invocation_id` from `run_results.json` metadata when available as the run identifier, or include it in run facets if generated run IDs are preferred.

This would make the artifact processor consistent with the structured logs implementation.

## Related discussion: dbt Cloud job metadata

This may also help with dbt Cloud / Airflow integration work: https://github.com/apache/airflow/issues/68661

Instead of introducing a separate dbt Cloud Job Run layer, we may be able to attach dbt Cloud job/run identifiers as metadata/facets on the dbt invocation events.

For example:

```text
Airflow task / external orchestrator
└── dbt invocation (with dbt Cloud job id / run id facet, when available)
    ├── model run
    ├── snapshot run
    └── test run
```

The exact modeling should be discussed separately, but adding the dbt invocation layer in the artifact processor gives us a natural place to attach dbt Cloud job metadata without necessarily adding another event layer.

## SQL query events

A related topic is whether the artifact processor should also emit SQL query events beneath each node run, matching the hierarchy already produced by the structured logs integration:

```text
dbt invocation
└── model run
    └── SQL query
```

Today, `run_results.json` generally exposes only a single query identifier per node via `adapter_response`, so this would not provide the same level of fidelity as structured logs, which can observe multiple SQL statements for a single node execution.

However, introducing SQL query events could still make the artifact processor's event hierarchy more consistent with:

- the dbt structured logs integration, and
- other integrations that already emit query-level events when query identifiers are available (e.g. Airflow).

Whether this should be included as part of this change or handled in a follow-up discussion is open.

## Benefits

- Consistent hierarchy across dbt integrations.
- A natural parent for all resources executed during a single dbt invocation.
- Easier UI grouping and navigation.
- A clean place to attach dbt Cloud job/run metadata when available.
- No changes to existing model/test/snapshot events, only the addition of a parent run.

maitraymukeshkumarmodi-aiml:
Extend DbtArtifactProcessor to emit a top-level dbt command invocation run event (START and COMPLETE/FAIL) using invocation_id from run_results.json metadata. Set this top-level invocation event as the parent for all child node-level events (models, seeds, snapshots, tests).

closes: #4720

Add top-level dbt command invocation run events in the artifact processor and set it as parent for node-level events.                                                 
                                                                                                                                                                      
### Meaningful description                                                                                                                                            
The dbt structured logs integration emits a top-level dbt command (invocation) run (`START` and `COMPLETE`/`FAIL`) as a parent for all node events (models, seeds,    
snapshots, tests). The artifact-based processor (`DbtArtifactProcessor` in `processor.py`) previously lacked this top-level invocation event and emitted node-level     
events directly without a common parent.                                                                                                                                
                                                                                                                                                                      
**Key Changes**:                                                                                                                                                      
1. **Top-Level Invocation Events**: Extended `DbtArtifactProcessor` in `processor.py` to generate top-level dbt command invocation `START` and `COMPLETE`/`FAIL`      
events.                                                                                                                                                                 
2. **Identifier & Facets**: Uses `invocation_id` from `run_results.json` metadata as the `runId` (with UUID validation and fallback) and sets                         
`JobTypeJobFacet(jobType="JOB")` on the job facet.                                                                                                                      
3. **Parent Linking Hierarchy**:                                                                                                                                      
   - Parents all child node-level events (models, seeds, snapshots, tests) to this top-level dbt invocation run.                                                      
   - If an external orchestrator parent (`_dbt_run_metadata`, e.g., Airflow task) is provided, it is attached as the parent of the top-level dbt invocation event.    
4. **Test Suite**: Added unit test suite `TestDbtInvocationEvents` in `test_processor.py` covering success/failure invocation event generation and parent metadata    
linking.                                                                                                                                                                
                                                                                                                                                                      
### Checklist                                                                                                                                                         


integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -201,6 +202,91 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+                _uuid.UUID(raw_invocation_id)
+                invocation_id = raw_invocation_id
+            else:
+                invocation_id = str(generate_new_uuid())
+        except ValueError:
+            invocation_id = str(generate_new_uuid())
+        job_name = self.invocation_job_name
+
+        self._invocation_parent_metadata = ParentRunMetadata(
+            run_id=invocation_id,
```
dolfinus: Please use `generate_static_uuid(fallback_time, invocation_id)` to produce UUIDv7 instead UUIDv4. This makes runs sortable by id

maitraymukeshkumarmodi-aiml: Sounds good.

maitraymukeshkumarmodi-aiml:
@dolfinus thanks for the review; I've updated it. I just merged main into my branch, and the check is failing, although I don't think it's related to my change. Other than that, this PR is ready to go.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +211,98 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+                    completed_at_list.append(t["completed_at"])
+
+        fallback_time = (
+            self.run_metadata.get("generated_at") or datetime.datetime.now(datetime.timezone.utc).isoformat()
+        )
+        start_time = min(started_at_list) if started_at_list else fallback_time
+        complete_time = max(completed_at_list) if completed_at_list else fallback_time
+
+        try:
+            from _datetime import datetime as real_datetime
```
dolfinus: Is this import required?

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +211,98 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+            instant = real_datetime.fromisoformat(start_time.replace("Z", "+00:00"))
+        except Exception:
+            try:
+                instant = datetime.datetime.fromisoformat(start_time.replace("Z", "+00:00"))
+            except Exception:
+                instant = datetime.datetime.now(datetime.timezone.utc)
+
+        raw_invocation_id = self.run_metadata.get("invocation_id")
+        if raw_invocation_id:
+            invocation_id = str(generate_static_uuid(instant, raw_invocation_id.encode("utf-8")))
```
dolfinus:
```suggestion
            run_id = str(generate_static_uuid(instant, raw_invocation_id.encode("utf-8")))
```

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: Actually you are right, it wasnt required. I've removed it and switched to using the standard datetime.datetime class.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +213,93 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+        if run_run_facet := self.dbt_run_run_facet():
+            run_facets.update(run_run_facet)
+        if engine_facet := self.processing_engine_facet():
+            run_facets.update(engine_facet)
+        if self._dbt_run_metadata:
+            run_facets["parent"] = self._dbt_run_metadata.to_openlineage()
+
+        job_facets: dict[str, JobFacet] = {
+            "jobType": job_type_job.JobTypeJobFacet(
+                jobType="JOB",
```
kacpermuda, line 284: Is this consistent with the logs processor invocation level event?

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -244,10 +334,21 @@ def parse(self) -> DbtEvents:
             else:
                 return events
 
+        invocation_events = self.generate_invocation_events(context)
```
kacpermuda: Let's make this and opt-in for now maybe? With current code it will always run generate_invocation_events, set this metadata and emit those events. Maybe for a smoother transition, we should put it behind some bool arg to the processor, so that user can enable it if needed, and then in some time we can switch the default from False to True. We can then keep expanding this structure for separate query events, and it will all be behind this "flag"

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -1107,13 +1208,14 @@ def get_run(
 
         run_facets.update(
             {
-                **self.dbt_version_facet(),
-                **self.dbt_run_run_facet(),
-                **self.processing_engine_facet(),
+                **(self.dbt_version_facet() or {}),
+                **(self.dbt_run_run_facet() or {}),
+                **(self.processing_engine_facet() or {}),
```
kacpermuda: What changed comparing to current code that we have to introduce this safety measure? This should be always present right?

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +213,93 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+                if t.get("started_at"):
+                    started_at_list.append(t["started_at"])
+                if t.get("completed_at"):
+                    completed_at_list.append(t["completed_at"])
+
+        fallback_time = (
+            self.run_metadata.get("generated_at") or datetime.datetime.now(datetime.timezone.utc).isoformat()
+        )
+        start_time = min(started_at_list) if started_at_list else fallback_time
+        complete_time = max(completed_at_list) if completed_at_list else fallback_time
```
kacpermuda, line 248: This looks like it should be a separate helper function somewhere, or few functions.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -52,6 +52,8 @@
 from openlineage.common.utils import get_from_multiple_chains, get_from_nullable_chain
 from openlineage_sql import parse as parse_sql
 
+_DATETIME_CLASS = datetime.datetime
```
kacpermuda: Why is that needed?

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +213,93 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
     @abstractmethod
     def get_dbt_metadata(self): ...
 
+    @property
+    def invocation_job_name(self) -> str:
+        if self.openlineage_job_name:
+            return self.openlineage_job_name
+        if hasattr(self, "job_name") and getattr(self, "job_name"):
+            return getattr(self, "job_name")
```
kacpermuda, line 226: Where is this used currently? Will we send events with the same job name for both invocation and some dbt node?

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -210,6 +213,93 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+        try:
+            instant = _DATETIME_CLASS.fromisoformat(start_time.replace("Z", "+00:00"))
+        except (ValueError, TypeError, AttributeError):
+            instant = _DATETIME_CLASS.now(datetime.timezone.utc)
+
+        raw_invocation_id = self.run_metadata.get("invocation_id")
+        if raw_invocation_id:
+            run_id = str(generate_static_uuid(instant, raw_invocation_id.encode("utf-8")))
+        else:
+            run_id = str(generate_new_uuid(instant))
```
kacpermuda: Looks like it should be a property / helper function?

kacpermuda (commented):
Left few comments, thanks for taking a time to work on this improvement !

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: Great idea! I've added an emit_dbt_invocation_event: bool = False flag to              DbtArtifactProcessor. It defaults to False for a smooth transition, so top-level invocation events and parent linking are only generated when emit_dbt_invocation_event=True.

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: Extracted _extract_invocation_run_timing(context) and  _generate_invocation_run_id(start_time) as separate helper methods on DbtArtifactProcessor.

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: Removed _DATETIME_CLASS constant completely and switched to using datetime.datetime directly.

mobuchowski:
The `dbt-ol` already emits top-level events. Why are we doing this?

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: Yes, this is completely consistent. In `DbtStructuredLogsProcessor` (and `utils.get_job_type()`), command-level events (`MainReportVersion` and                
  `CommandCompleted`) emit `jobType="JOB"` with `processingType="BATCH"` and `integration="DBT"`, and the CLI wrapper (`openlineage/dbt/__init__.py`) also uses    
  `JOB_TYPE_FACET` with `jobType="JOB"`. Individual dbt model/snapshot/seed/test nodes continue to emit their respective node types (`"MODEL"`, `"SNAPSHOT"`,      
  `"TEST"`, `"SEED"`).

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: Good catch! Removed the unnecessary defensive `or {}` fallback in both `generate_invocation_events()` and `get_run()`, since `dbt_version_facet()`,  `dbt_run_run_facet()`, and `processing_engine_facet()` always return dictionary instances.

integration/common/src/openlineage/common/provider/dbt/processor.py
maitraymukeshkumarmodi-aiml: The invocation event's job name is always distinct from individual node-level events:                                                                          
    - **Node events** use dataset/node-scoped names (e.g. `database.schema.model_name` or test `unique_id`).                                                       
    - **Invocation events** use project/command-scoped names via `invocation_job_name` (e.g. `openlineage_job_name` if explicitly provided, `self.job_name` as     
  `dbt-run-{project_name}` from `DbtLocalArtifactProcessor`, or fallback `dbt-{command}`).

maitraymukeshkumarmodi-aiml:
> The `dbt-ol` already emits top-level events. Why are we doing this?

`dbt-ol` emits top-level events when wrapping CLI runs directly in real-time, but post-hoc artifact processors (e.g., `DbtArtifactProcessor`, `DbtLocalArtifactProcessor`, Airflow's `DbtCloudHook`, and downstream tools consuming exported `manifest.json` + `run_results.json`) did not have top-level invocation START/COMPLETE events or a parent linkage for all generated node events (Issue #4720).                                                                
                                                                                                                                                                   
Per maintainer feedback, we have also placed this behind the opt-in flag `emit_dbt_invocation_event: bool = False` on `DbtArtifactProcessor` to ensure full backwards compatibility for existing consumers.

dolfinus (approved):

maitraymukeshkumarmodi-aiml:
Hey @kacpermuda, I have 1 approval on this PR. If you see my changes as fit, this pr is ready to be merged by you as I don't have the option to merge. Thank you so much and feel free to take your time!

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+        if raw_invocation_id:
+            return str(generate_static_uuid(instant, raw_invocation_id.encode("utf-8")))
+        return str(generate_new_uuid(instant))
+
+    def generate_invocation_events(self, context: DbtRunContext) -> DbtRunResult | None:
+        start_time, complete_time, has_failure = self._extract_invocation_run_timing(context)
+        run_id = self._generate_invocation_run_id(start_time)
+        job_name = self.invocation_job_name
+
+        self._invocation_parent_metadata = ParentRunMetadata(
```
kacpermuda, line 268: This drops the `root_parent_*` fields, so with the flag on every node event comes out with `root=None` even when an orchestrator parent is set. Verified against the `small` fixture with an Airflow-style parent:

```
START job='dbt-run-dbt_small_test'  parent=airflow-dag.task  root=airflow-dag
START job='...test_model'           parent=dbt-run-dbt_small_test  root=None   <- was root=airflow-dag
```

Both other paths do this deliberately: `structured_logs._setup_dbt_run_metadata` and the block in `dbt/__init__.py`, whose comment reads *"Carry the root parent through to per-node events so backends can stitch the full orchestrator -> dbt-run -> node chain. Without this, child events get root=None."*

Please propagate the root from `_dbt_run_metadata` (its root if set, else `_dbt_run_metadata` itself, else the invocation run).

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+                if t.get("started_at"):
+                    started_at_list.append(t["started_at"])
+                if t.get("completed_at"):
+                    completed_at_list.append(t["completed_at"])
+
+        fallback_time = (
+            self.run_metadata.get("generated_at") or datetime.datetime.now(datetime.timezone.utc).isoformat()
+        )
+        start_time = min(started_at_list) if started_at_list else fallback_time
+        complete_time = max(completed_at_list) if completed_at_list else fallback_time
```
kacpermuda, line 248: `complete_time` only looks at `run_results` timings, but `parse_test` uses `datetime.now()` for test nodes and `to_openlineage_events` also falls back to now for nodes with no timing. The result is a parent run that finishes before its children start:

```
# dbt test fixture
START  2021-08-23T15:06:20Z       dbt-run-dbt_small_test   <- invocation
START  2026-09-21T14:22:39+00:00  ...test_first_dbt_model.test
FAIL   2021-08-23T15:06:44Z       dbt-run-dbt_small_test   <- ends before children start

# dbt run, fail fixture (node with no timing)
FAIL   2026-09-21T14:22:39+00:00  ...test_second_parallel_dbt_model
FAIL   2021-07-28T12:09:48Z       dbt-run-dbt_fail_test
```

This breaks run duration and parent containment downstream for `test`, `build`, and any `run` where a node failed before producing timings. The terminal event needs to be built after the node events, with its time as the max over them.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
     @abstractmethod
     def get_dbt_metadata(self): ...
 
+    @property
+    def invocation_job_name(self) -> str:
```
kacpermuda, line 222: For `DbtLocalArtifactProcessor` this returns `self.job_name` - the exact name and namespace `dbt-ol` already uses for its own wrapper START/COMPLETE (`dbt_run_event_start(job_name=processor.job_name, ...)`). So if this flag is ever wired into `dbt-ol` you get two runs of `dbt-run-{project}` in one namespace, with the invocation run parented to the wrapper run (a job that is its own parent) and the wrapper run left childless.

This is the duplication @mobuchowski was asking about. Your earlier answer covers node job names, which are indeed distinct, but not the wrapper. Either give the invocation event a distinct name, or reuse the existing dbt-level run when `_dbt_run_metadata` already is one.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+        if self.openlineage_job_name:
+            return self.openlineage_job_name
+        if hasattr(self, "job_name") and getattr(self, "job_name"):
+            return getattr(self, "job_name")
+        return f"dbt-{self.command}" if self.command else "dbt-invocation"
```
kacpermuda, line 227: `DbtCloudArtifactProcessor` has no `job_name` and usually no `openlineage_job_name`, so it lands here and every dbt Cloud project and account collapses into a single job called `dbt-run`. `project_name` and `account_id` are both on the processor - worth using them.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
     @abstractmethod
     def get_dbt_metadata(self): ...
 
+    @property
+    def invocation_job_name(self) -> str:
+        if self.openlineage_job_name:
+            return self.openlineage_job_name
+        if hasattr(self, "job_name") and getattr(self, "job_name"):
```
kacpermuda, line 225: Nit: `hasattr` only swallows `AttributeError`, so the `ValueError` that `_use_extended_job_name` raises on a bad `OPENLINEAGE_DBT_USE_EXTENDED_JOB_NAME` propagates from here. It also evaluates the property up to three times. `getattr(self, "job_name", None)` covers it, and the `openlineage_job_name` branch above is redundant since local's `job_name` already honours it.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+
+        start_event = RunEvent(
+            eventType=RunState.START,
+            eventTime=start_time,
+            run=Run(runId=run_id, facets=run_facets),
+            job=Job(namespace=self.job_namespace, name=job_name, facets=job_facets),
+            producer=self.producer,
+        )
+
+        end_state = RunState.FAIL if has_failure else RunState.COMPLETE
```
kacpermuda, line 299: No `errorMessage` facet on the FAIL, and no `nominalTime`. `structured_logs._parse_command_completed_event` attaches `ErrorMessageRunFacet` - this is part of why the "completely consistent with the logs processor" answer is only partly true (the `jobType` facet does match).

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+        run_id = self._generate_invocation_run_id(start_time)
+        job_name = self.invocation_job_name
+
+        self._invocation_parent_metadata = ParentRunMetadata(
+            run_id=run_id,
+            job_name=job_name,
+            job_namespace=self.job_namespace,
+        )
+
+        run_facets: dict[str, Any] = {
```
kacpermuda, line 274: The same `run_facets` dict instance (and `job_facets` below) is handed to both the START and the terminal `Run`/`Job`, so mutating one mutates the other.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -215,6 +218,97 @@ def dbt_run_metadata(self, metadata: ParentRunMetadata):
...
+        return f"dbt-{self.command}" if self.command else "dbt-invocation"
+
+    def _extract_invocation_run_timing(self, context: DbtRunContext) -> tuple[str, str, bool]:
+        """Extract start time, complete time, and failure status from run results."""
+        started_at_list = []
+        completed_at_list = []
+        has_failure = False
+
+        for run in context.run_results.get("results", []):
+            if run.get("status") in ("error", "fail", "failed"):
```
kacpermuda, line 236: `"failed"` isn't a dbt status - the real ones are `error` and `fail` (plus `warn`, `skipped`, `pass`, `success`). Harmless, but worth trimming to the actual set.

integration/common/src/openlineage/common/provider/dbt/processor.py
```diff
@@ -249,10 +343,23 @@ def parse(self) -> DbtEvents:
             else:
                 return events
 
+        invocation_events = (
```
kacpermuda, line 346: The invocation pair is emitted unconditionally, so an all-skipped run produces an orphan parent with no children.

kacpermuda (commented):
Thanks for the iterations here. Did on more sweep with LLM. Good news first: the default-off path is genuinely a no-op. Two things before this can go in: (inline): node events lose their root parent, and the invocation terminal event can land before its children start. More details below

PR #4856, opened 2026-08-13
docs: fix broken star history chart

Dessalines39394:
The star history chart in the README is currently broken and no longer renders because the GitHub stargazer API behind it has limited its access. This change points the chart at a working source that does not require an API token, restoring the chart so contributors can continue to see the project's growth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ If you discover a vulnerability in the project, please [open an issue](https://g
 
 ## Star History
 
-[![Star History Chart](https://api.star-history.com/svg?repos=OpenLineage/OpenLineage&type=Date)](https://www.star-history.com/#OpenLineage/OpenLineage&Date)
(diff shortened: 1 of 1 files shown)


kacpermuda (changes requested):
I'd rather drop the whole chart for now, or figure out a way to pass the safe token there. Not sure if I'm comfortable with linking some random page here. (tbh was not comfortable with previous one as well). Maybe we can just store a static image instead?

Dessalines39394:
> figure out a way to pass the safe token there

Unfortunately the third-party will always have to use it raw at some point

> Maybe we can just store a static image instead

Definitely, but how do you keep it up to date?

kacpermuda:
> Definitely, but how do you keep it up to date?

Yes, that's the problem. We will definitely forget to refresh it.

If others are okay with it, we can put a new link there, but I'd vote for just removing the whole chart.

merobi-hub:
Agree that it might be time to drop the chart

kacpermuda:
@Dessalines39394 , would you mind updating the PR and simply drop the chart? Thanks

PR #4861, opened 2026-08-14
flink: add Kinesis lineage visitors (dataset identifier + type facet)

fmorillo7694:
## Problem

The Apache Flink Kinesis connector is gaining native FLIP-314 lineage
(`LineageVertexProvider` on `KinesisStreamsSource`/`KinesisStreamsSink` —
[apache/flink-connector-aws#250](https://github.com/apache/flink-connector-aws/pull/250),
FLINK-39813). The Flink 2 integration currently handles connector lineage only for Kafka
and JDBC, so Kinesis datasets have two gaps:

1. **SQL/DataStream identity split.** Flink's table planner names SQL datasets by catalog
   identity (`default_catalog.default_database.my_table`). For Kafka,
   `KafkaTableLineageDatasetIdentifierVisitor` re-derives the physical identifier from the
   table options so DataStream and SQL jobs converge on one dataset. Kinesis has no such
   visitor — the same physical stream appears as two unrelated datasets.
2. **No schema facet.** `TypeDatasetFacetVisitor` is bound to the Kafka connector's facet
   classes, so the Kinesis connector's `type` facet is never converted to an OpenLineage
   schema facet.

## Changes

- **`KinesisTableLineageDatasetIdentifierVisitor`** — for table lineage datasets with
  `connector = 'kinesis'`, derives the physical identifier from the `stream.arn` option:
  namespace `arn:{partition}:kinesis:{region}:{account}`, name `stream/{streamName}` —
  matching the convention the connector emits on the DataStream path. Catalog identity is
  preserved as a `TABLE` symlink (mirrors the Kafka visitor).
- **`KinesisTypeDatasetFacetVisitor`** — converts the Kinesis connector's `type` facet
  into an OL schema facet. The facet is accessed reflectively
  (`org.apache.flink.connector.kinesis.lineage.TypeDatasetFacet#getTypeInformation`) so no
  compile-time dependency on the connector is needed (the connector release containing the
  facet classes is pending).
- **`AvroTypeDatasetFacetVisitorDelegate`** — additionally supports
  `GenericRecordAvroTypeInfo` (schema read reflectively from its private field), covering
  records deserialized as Avro `GenericRecord`, e.g. via Confluent Schema Registry or AWS
  Glue Schema Registry. Previously only compiled `SpecificRecord` types produced a schema
  facet — this improves Kafka lineage too.
- Registered both visitors in `OpenLineageJobStatusChangedListenerFactory`; added
  `flink-avro` as a test dependency.

## Testing

- `KinesisTableLineageDatasetIdentifierVisitorTest` (3 tests: gating incl. malformed ARN,
  identifier derivation, non-default partition `aws-cn`).
- `KinesisTypeDatasetFacetVisitorTest` (3 tests: reflective gating, POJO/generic field
  extraction, full Avro schema extraction via `GenericRecordAvroTypeInfo`) — uses a test
  stub with the connector facet's FQCN.
- Full `:flink2:test` suite: **51/51 pass** (`-Pflink.version=2.0.0`), `spotlessCheck` clean.
- End-to-end (outside this repo): verified against a local build of
  flink-connector-aws#250 that a real Flink 2.0 job's `JobCreatedEvent` carries these
  facets, including the full Avro schema when deserializing via Glue Schema Registry.

### DIFF
--- a/integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageJobStatusChangedListenerFactory.java
+++ b/integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageJobStatusChangedListenerFactory.java
@@ -10,6 +10,7 @@ import io.openlineage.flink.config.FlinkConfigParser;
 import io.openlineage.flink.visitor.Flink2VisitorFactory;
 import io.openlineage.flink.visitor.facet.ConfigFacetVisitor;
 import io.openlineage.flink.visitor.facet.DatasetFacetVisitor;
+import io.openlineage.flink.visitor.facet.KinesisTypeDatasetFacetVisitor;
 import io.openlineage.flink.visitor.facet.SchemaFacetVisitor;
 import io.openlineage.flink.visitor.facet.TableLineageFacetVisitor;
 import io.openlineage.flink.visitor.facet.TypeDatasetFacetVisitor;
@@ -18,6 +19,7 @@ import io.openlineage.flink.visitor.identifier.JdbcTableLineageDatasetIdentifier
 import io.openlineage.flink.visitor.identifier.KafkaTableLineageDatasetIdentifierVisitor;
 import io.openlineage.flink.visitor.identifier.KafkaTopicListDatasetIdentifierVisitor;
 import io.openlineage.flink.visitor.identifier.KafkaTopicPatternDatasetIdentifierVisitor;
+import io.openlineage.flink.visitor.identifier.KinesisTableLineageDatasetIdentifierVisitor;
 import java.util.Arrays;
 import java.util.Collection;
 import lombok.extern.slf4j.Slf4j;
@@ -44,6 +46,7 @@ public class OpenLineageJobStatusChangedListenerFactory implements JobStatusChan
       public Collection<DatasetFacetVisitor> loadDatasetFacetVisitors(OpenLineageContext context) {
         return Arrays.asList(
             new TypeDatasetFacetVisitor(context),
+            new KinesisTypeDatasetFacetVisitor(context),
             new TableLineageFacetVisitor(context),
             new ConfigFacetVisitor(context),
             new SchemaFacetVisitor(context));
@@ -56,7 +59,8 @@ public class OpenLineageJobStatusChangedListenerFactory implements JobStatusChan
             new KafkaTopicPatternDatasetIdentifierVisitor(context),
             new KafkaTopicListDatasetIdentifierVisitor(),
             new JdbcTableLineageDatasetIdentifierVisitor(),
-            new KafkaTableLineageDatasetIdentifierVisitor());
+            new KafkaTableLineageDatasetIdentifierVisitor(),
+            new KinesisTableLineageDatasetIdentifierVisitor());
       }
     };
   }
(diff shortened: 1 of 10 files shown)


integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageJobStatusChangedListenerFactory.java
```diff
@@ -44,6 +46,7 @@ Flink2VisitorFactory loadVisitorFactory() {
       public Collection<DatasetFacetVisitor> loadDatasetFacetVisitors(OpenLineageContext context) {
         return Arrays.asList(
             new TypeDatasetFacetVisitor(context),
+            new KinesisTypeDatasetFacetVisitor(context),
```
mobuchowski, line 49: When a Flink cluster includes both Kafka and Kinesis connector JAR files, TypeDatasetFacetVisitor processes the Kinesis "type" facet first.

TypeDatasetFacetUtil.getFacet() casts every "type" facet to Kafka's TypeDatasetFacet, which causes a ClassCastException. As a result, onJobCreatedEvent() doesn't emit a START event. Update the Kafka visitor to accept only Kafka type facets, and add a conversion test that loads both connector facet classes.

mobuchowski (changes requested):

pawel-big-lebowski:
Have you considered using [kinesis-mock](https://github.com/etspaceman/kinesis-mock
) docker image for integration test? I am not sure if this is feasible, but tests based on real Flink apps running are always better quality.

integration/flink/flink2/src/main/java/io/openlineage/flink/listener/OpenLineageJobStatusChangedListenerFactory.java
fmorillo7694: Fixed in dcc4327e9. `TypeDatasetFacetUtil.getFacet()` now type-checks before casting (`.filter(f -> f instanceof TypeDatasetFacet)`), so a non-Kafka `type` facet yields `Optional.empty()` , the Kafka visitor's `isDefinedAt()` returns `false` for the Kinesis facet and it is handled by `KinesisTypeDatasetFacetVisitor` instead.

Added `TypeDatasetFacetVisitorTest#testIsNotDefinedAtKinesisTypeFacet`, a conversion test that loads both connector facet classes in the same JVM to reproduce the mixed-JAR scenario: without the guard it fails with exactly the `ClassCastException` you describe; with it, the Kafka visitor skips the Kinesis facet and START emission is unaffected.

fmorillo7694:
Agreed,  tests against a real running app give the strongest signal. The constraint today is that no released Kinesis connector publishes these lineage facets yet: they are added by FLINK-39813 (apache/flink-connector-aws#250, in review), which is also where the end-to-end path is integration-tested (`KinesisStreamsLineageITCase` against Localstack / DataStream and SQL, source and sink lineage). Until a connector release ships the facets, an integration test here couldn't exercise the lineage path, so this PR covers the reflective contract with conversion tests against a stub matching the connector's facet class. Happy to follow up with a kinesis-mock-based integration test once the connector side is released.

mobuchowski (approved):

fmorillo7694:
Thank you for the review and approval, @mobuchowski! 🙏

For context, the connector-side counterpart that produces the facets these visitors consume is [apache/flink-connector-aws#250](https://github.com/apache/flink-connector-aws/pull/250) (FLINK-39813),  it defines the same dataset naming convention and is currently awaiting Flink committer review. Once the connector side is released, I plan to follow up here with a kinesis-mock-based integration test exercising the full path.

PR #4322, opened 2026-02-05
docs: add AI agent contribution guidelines

harels:
## Summary

Add `AGENTS.md` following the [agents.md standard](https://agents.md) with guidelines for AI coding agents contributing to OpenLineage.

## Key Features

- **Protected paths** - Spec files require explicit user authorization before modification
- **Backwards compatibility** - All spec changes must follow SchemaVer
- **Setup/build/test commands** - Actionable commands agents can execute
- **Client coordination** - Changes affecting clients must be coordinated across Java and Python
- **Commit conventions** - DCO sign-off required, component-prefixed messages
- **AI disclosure** - Co-Authored-By required for AI-assisted contributions
- **Code style** - Formatter and linter requirements

## Changes

| File | Change |
|------|--------|
| `AGENTS.md` | New file with AI agent contribution guidelines |
| `.github/pull_request_template.md` | Add AI usage checkbox |
| `.gitignore` | Remove `CLAUDE.md` from ignore list |

## Test plan

- [ ] Verify AGENTS.md renders correctly on GitHub
- [ ] Verify PR template shows AI checkbox
- [ ] Test with an AI agent to confirm guidelines are followed

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #4871, opened 2026-08-16
docs: symlink CLAUDE.md to AGENTS.md

namsick96:
Commit `CLAUDE.md` as a symlink to `AGENTS.md` so Claude Code automatically picks up the repo's AI agent guidelines.

### Meaningful description
#4322 added `AGENTS.md` with tool-agnostic AI agent contribution guidelines. Most agentic tools now follow the agents.md convention, but Claude Code only auto-loads `CLAUDE.md`, so for Claude Code users those guidelines are silently missed unless they wire something up locally.

This PR commits `CLAUDE.md` as a **symlink** pointing to `AGENTS.md` — the same approach apache/airflow uses ([apache/airflow CLAUDE.md](https://github.com/apache/airflow/blob/main/CLAUDE.md)) — and removes the corresponding `.gitignore` entry, which is obsolete once the symlink is tracked.

**Relation to the #4322 discussion** ([thread](https://github.com/OpenLineage/OpenLineage/pull/4322#discussion_r2769882004)): committing a `CLAUDE.md` was considered there and rejected, for two reasons. I believe a symlink addresses both:

1. *"a one-line 'read AGENTS.md' file works only if the agent follows the instruction"* — a symlink has no instruction to follow: its content **is** `AGENTS.md`, byte for byte, kept in sync by construction with zero maintenance.
2. *"if we add CLAUDE.md, why not .cursor or the next tool popular in a week"* — the symlink is effectively just an alias for `AGENTS.md`, only needed for tools that don't follow the agents.md convention. That set is small and shrinking: Cursor, Codex, Gemini CLI and others already read `AGENTS.md` natively, and with agents.md now under the Linux Foundation, tools keep converging on it. Today the notable holdout is Claude Code — so in practice this is one symlink, not the first of many.

Happy to close if the maintainers still prefer keeping agent-specific files out of the repo — just wanted to put the symlink option on the table, since it wasn't considered in that thread.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,9 +176,6 @@ changes.txt
 # Cursor
 .cursor/
 
-# Claude Code
-CLAUDE.md
-
 # UV lock files (ignore in subdirectories only, keep root uv.lock)
 **/uv.lock
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


harels (approved):
Makes sense to me

mobuchowski (approved):

PR #4874, opened 2026-08-18
dbt: support the fabricspark adapter

calvinchengx:
## Problem

`dbt-fabricspark` — the adapter for building a **Fabric Lakehouse** over Livy — is not in the `Adapter` enum. `extract_adapter_type` therefore raises `NotImplementedError`, the caller catches it, and the run yields **no events at all**.

The failure is silent in practice. `astronomer-cosmos` (which depends on `openlineage-integration-common`) wraps the parse in:

```python
except (FileNotFoundError, NotImplementedError, ValueError, KeyError, jinja2.exceptions.UndefinedError):
    self.log.debug("Unable to parse OpenLineage events", stack_info=True)
```

At the default INFO level nothing is logged, so a dbt project on this adapter emits no lineage and no Airflow assets, with no indication why. What the task log shows is just:

```
Inlets: []
Outlets: []
```

A sibling project on `dbt-fabric` (the Warehouse adapter, already supported) emits full lineage from the same DAG, which makes the gap look like a bug in the project rather than an unsupported adapter.

## Change

1. `FABRICSPARK = "fabricspark"` in the `Adapter` enum.
2. A namespace branch in `extract_namespace`.

`extract_dialect` then works unchanged, returning `fabricspark`.

## On the namespace

`fabric` (Warehouse) keys on `server`/`port` because it is a TDS endpoint. A lakehouse has no host of its own — **every Fabric tenant reaches the same `api.fabric.microsoft.com`** — so a namespace derived from the profile's `endpoint` would be identical for two different workspaces. The profile carries `workspaceid` and `lakehouseid`, which are the stable, unique address:

```
fabric-lakehouse://{workspaceid}/{lakehouseid}
```

A profile missing either fails by name, rather than emitting a namespace that silently collides with every other workspace's.

**One open question for maintainers.** I have used a distinct `fabric-lakehouse://` scheme, since a lakehouse and a warehouse are different surfaces. But in a normal Fabric medallion, gold (`dbt-fabric`, `fabric-warehouse://…`) reads silver (`dbt-fabricspark`) through the lakehouse's SQL analytics endpoint — the *same* tables under two namespaces, so lineage will not join across that hop. If you would prefer these to converge on one naming, I am happy to change it; it seemed wrong to decide unilaterally.

## Verification

Against a real `dbt-fabricspark` project (8 models over Livy into a Fabric lakehouse), running `DbtLocalArtifactProcessor` over its own `manifest.json` and `run_results.json`.

**Before:** `events.completes: 0`

**After:**

```
events.completes: 1
  lake.lake.contoso_silver.silver_product_hierarchy
    namespace: fabric-lakehouse://<workspaceid>/<lakehouseid>
    inputs:  ['lake.lake.bronze_ref_product_hierarchy']
    outputs: ['lake.lake.silver_product_hierarchy']
```

Worth noting for anyone reproducing it: artifacts from `dbt compile` produce zero events regardless of this change, because `compile` is not in the supported command list — my first attempt at this measurement was invalid for that reason. The numbers above come from `dbt run`.

## Tests

Three added to `integration/common/tests/dbt/test_processor.py`:

- the namespace is built from the two ids;
- a profile missing them raises rather than emitting a colliding namespace;
- the dialect is `fabricspark`.

`integration/common`: **351 passed, 2 skipped**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased](https://github.com/OpenLineage/OpenLineage/compare/1.52.0...HEAD)
 
+### Added
+
+* **dbt: Support the `fabricspark` adapter** [`#4874`](https://github.com/OpenLineage/OpenLineage/pull/4874) [@calvinchengx](https://github.com/calvinchengx)  
+  *Adds `dbt-fabricspark` to the supported adapters, so a dbt project building a Fabric Lakehouse emits lineage instead of silently producing no events.*
+
 ## [1.52.0](https://github.com/OpenLineage/OpenLineage/compare/1.51.0...1.52.0)
 
 ### Added
--- a/integration/common/src/openlineage/common/provider/dbt/processor.py
+++ b/integration/common/src/openlineage/common/provider/dbt/processor.py
@@ -69,6 +69,7 @@ class Adapter(Enum):
     GLUE = "glue"
     CLICKHOUSE = "clickhouse"
     FABRIC = "fabric"
+    FABRICSPARK = "fabricspark"
 
     @staticmethod
     def adapters() -> str:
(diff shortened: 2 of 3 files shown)


calvinchengx:
Friendly ping — this has been green for a month (all CI checks pass, patch coverage 100%) and still merges cleanly against `main`. Happy to rebase if that's useful.

I suspect the open question in the description is the part that needs a maintainer's call rather than a review, so to make it as small a decision as possible: the only thing I'm unsure of is the namespace scheme. I used `fabric-lakehouse://{workspaceid}/{lakehouseid}`, distinct from the existing `fabric-warehouse://`. The tradeoff is that in a normal Fabric medallion the gold layer (`dbt-fabric`) reads silver (`dbt-fabricspark`) through the lakehouse's SQL analytics endpoint — the same tables under two namespaces, so lineage won't join across that hop.

If you'd prefer them to converge on a single scheme, say which and I'll push the change the same day. If the current split is right, this is ready as-is.

harels:
Thanks for the PR.
One the naming question, if it's the same physical location the options imo are:
- use the same canonical path
- use the alias to the other path so consumers are aware

Tagging @mobuchowski since he's more familiar with Fabric representations than I am.

PR #4876, opened 2026-08-18
client-java: add lineage compatibility converters

mobuchowski:

Add opt-in Java client translation between explicit lineage facets and legacy dataset lineage.

### Meaningful description

The new `LineageJobFacet` can describe an output such as `warehouse.orders` as depending on `raw.orders.customer_id`, including field mappings and transformations. Older consumers only inspect the event-level `inputs` and `outputs` arrays and `ColumnLineageDatasetFacet`, so they cannot read that lineage unless the producer emits both representations. The reverse applies to newer consumers receiving events from legacy producers.

This adds a compatibility layer to the Java client which translates immutable event copies immediately before logging and transport:

- `legacy` generates missing event-level datasets and representable column lineage from `LineageJobFacet`
- `modern` generates `LineageJobFacet` from legacy inputs, outputs and column lineage
- `both` selects the missing representation without running both passes or creating cycles
- `none` disables translation and remains the default while the feature is evaluated

Translation applies to `RunEvent` and `JobEvent`, preserves producer-provided data, and is deterministic, stable-order and idempotent. `DatasetEvent` is unchanged.

The mode can be set through YAML, `OPENLINEAGE__LINEAGE__COMPATIBILITY`, or the Java client builder. Configuration is also propagated through Spark, Flink and Hive.

Validation:

- 478 Java client tests
- focused Spark, Flink and Hive configuration tests
- Java PMD and Spotless
- repository pre-commit hooks

### Checklist

### DIFF
--- a/client/java/src/main/java/io/openlineage/client/LineageCompatibilityConverter.java
+++ b/client/java/src/main/java/io/openlineage/client/LineageCompatibilityConverter.java
@@ -1,0 +1,580 @@
+/*
+/* Copyright 2018-2026 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.client;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.openlineage.client.OpenLineage.JobEvent;
+import io.openlineage.client.OpenLineage.RunEvent;
+import java.util.ArrayList;
+import java.util.Iterator;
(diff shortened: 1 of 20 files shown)


client/java/src/main/java/io/openlineage/client/LineageCompatibilityConverter.java
```diff
@@ -0,0 +1,580 @@
...
+      T event, LineageCompatibility configuredCompatibility, Class<T> eventClass) {
+    LineageCompatibility compatibility =
+        configuredCompatibility == null ? LineageCompatibility.NONE : configuredCompatibility;
+    if (compatibility == LineageCompatibility.NONE) {
+      return event;
+    }
+
+    ObjectNode eventNode = MAPPER.valueToTree(event);
+    ObjectNode facets = getJobFacets(eventNode, false);
+    boolean hasLineage = facets != null && hasValue(facets, "lineage");
```
jakub-moravec: A constant might be better for the facet key?

Issue #4110, opened 2025-10-28
[BUG]Spark intergration for Structured Streaming Filesink can not generate open lineage with output details


yangguoaws:
### OpenLineage integration/client

Spark integration

### OpenLineage version

`io.openlineage:openlineage-spark_2.13:1.39.0`

### Technology and package versions

Python:  3.13.3
Scala: 2.13.16
Java: OpenJDK 64-Bit Server VM, 17.0.16

`pip freeze`
py4j==0.10.9.9
pyspark==4.0.1

### Environment configuration

I used the default config
```
$ git clone https://github.com/MarquezProject/marquez.git && cd marquez

$ ./docker/up.sh
```

### Deployment details

I used native spark on local machine. There is no managed services involved.

### Problem details

When using Spark structured streaming to write parquet file to file systems,
- File sink will only generate openlineage event with streaming processing type with output information as empty.
- Foreachbatch sink will generate openlineage event with both streaming processing type and batch processing type. The batch processing type will have valid output information.

The bug is that the openlineage spark integration can not generate openlineage event with valid output details for File sink in Spark structured streaming.

More details about the sample code and sample events are following.
#### Sample code:
```
query = streaming_df.writeStream \
    .format('parquet') \
    .outputMode('append') \
    .partitionBy('year', 'month', 'day') \
    .option('checkpointLocation', checkpoint_path) \
    .option('path', output_path) \
    .queryName('filesink') \
    .start()
```

#### Sample event for "processingType":"STREAMING"
```
[...]

PR #4884, opened 2026-08-22
spark: extract Delta output from V1 micro-batch writes

### LINKED ISSUE #4887: [BUG][SPARK] Spark 4 Delta V1 micro-batch writes omit the output dataset

Spark integration

- Apache Spark 4.1
- Scala 2.13
- Delta Lake 4.x
- Java 17

### Problem details

Spark 4 plans a V1 Structured Streaming write to Delta as a `WriteToMicroBatchDataSourceV1` logical node whose sink is `org.apache.spark.sql.delta.sources.DeltaSink`.

`WriteToMicroBatchDataSourceV1DatasetBuilder` only follows its output-dataset path for Spark 3's `org.apache.spark.sql.execution.streaming.FileStreamSink`. Consequently, the terminal lineage event contains the streaming input but omits the Delta output dataset.

Spark 4 also moved `FileStreamSink` to a different package, so loading the old class from the builder can fail with:

```text
java.lang.NoClassDefFoundError: org/apache/spark/sql/execution/streaming/FileStreamSink
```

The logical node already carries enough target information to avoid both problems:

- `catalogTable()` for a table write; and
- the case-insensitive `path` entry in `writeOptions()` for a path write.

The terminal COMPLETE event should contain the Delta target as an output dataset. Catalog metadata should be preferred when present; otherwise a Delta sink should use its `path` write option. The builder should not require Spark 3's `FileStreamSink` class to be loadable on Spark 4.

### How to reproduce

1. Start Spark 4 with Delta and the OpenLineage Spark listener.
2. Run a bounded rate stream into a Delta path:

```python
source = (
    spark.readStream.format("rate-micro-batch")
    .option("rowsPerBatch", 5)
    .load()
)

query = (
    source.writeStream.format("delta")
    .option("checkpointLocation", "/tmp/ol-delta-checkpoint")
    .trigger(availableNow=True)
    .start("/tmp/ol-delta-target")
)
query.awaitTermination()
```

3. Inspect the OpenLineage COMPLETE event for the streaming query.

Actual result: the input is present but `/tmp/ol-delta-target` is absent from `outputs`; debug logs may contain the `FileStreamSink` linkage error above.

Expected result: `/tmp/ol-delta-target` is present as an output dataset and the builder does not load the removed Spark 3 class.

### Additional details

PR #4884 implements the fix without a Delta runtime dependency. It uses catalog metadata for any sink when available, recognizes `DeltaSink` by class hierarchy, and otherwise normalizes the `path` option with the existing path utilities. The fix was verified on Spark 4.1 with both a simple bounded stream and a two-input streaming job similar to what we run in production.

Related: #4110

MSDehghan:

Spark: V1 structured-streaming writes to Delta now include their output dataset on Spark 4.

### Meaningful description

Spark 4 + Delta plans streaming writes as `WriteToMicroBatchDataSourceV1` with a `org.apache.spark.sql.delta.sources.DeltaSink`. The existing builder only entered its output path for Spark 3's `FileStreamSink`. As a result, Delta streaming COMPLETE events had no output; on Spark 4 the `instanceof FileStreamSink` check could also raise `NoClassDefFoundError` because Spark moved that class to a different package.

This change:

1. uses the logical node's catalog table whenever it is defined, regardless of sink type;
2. for a Delta sink without catalog metadata, falls back to the case-insensitive `path` entry in `writeOptions`;
3. detects Delta by walking the sink class hierarchy, avoiding a Delta runtime dependency;
4. keeps the existing empty behavior for unsupported sinks without catalog metadata.

The `writeOptions` API is present on this logical node in Spark 3.4 and Spark 4, and `DataStreamWriter.start(path)` places its target there.

Validation:

- 10 focused `WriteToMicroBatchDataSourceV1DatasetBuilderTest` tests pass.
- Spark 3.4 Spotless and PMD checks pass, as does `git diff --check`.
- On Spark 4.1 + Delta, an AvailableNow Kafka replay on current `main` emits the Kafka input but no streaming output and logs two `FileStreamSink` linkage errors. With this patch both streaming COMPLETE events also carry the Delta target and the linkage errors disappear.
- A two-topic Kafka-to-Delta job much closer to our production setup shows the same result: the target output appears while both topic inputs are preserved.

Closes #4887

Related to #4110

### Checklist

### DIFF
--- a/integration/spark/spark34/src/main/java/io/openlineage/spark34/agent/lifecycle/plan/WriteToMicroBatchDataSourceV1DatasetBuilder.java
+++ b/integration/spark/spark34/src/main/java/io/openlineage/spark34/agent/lifecycle/plan/WriteToMicroBatchDataSourceV1DatasetBuilder.java
@@ -6,15 +6,19 @@
 package io.openlineage.spark34.agent.lifecycle.plan;
 
 import io.openlineage.client.OpenLineage;
+import io.openlineage.spark.agent.util.PathUtils;
+import io.openlineage.spark.agent.util.ScalaConversionUtils;
 import io.openlineage.spark.api.AbstractQueryPlanOutputDatasetBuilder;
 import io.openlineage.spark.api.DatasetFactory;
 import io.openlineage.spark.api.OpenLineageContext;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.hadoop.fs.Path;
 import org.apache.spark.scheduler.SparkListenerEvent;
 import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan;
-import org.apache.spark.sql.execution.streaming.FileStreamSink;
 import org.apache.spark.sql.execution.streaming.sources.WriteToMicroBatchDataSourceV1;
 import org.apache.spark.sql.execution.ui.SparkListenerSQLExecutionEnd;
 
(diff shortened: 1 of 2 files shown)


mobuchowski (approved):

MSDehghan:
@mobuchowski 
Thank you so much for your review. I checked the failed CI jobs and it seems its infrastructure checkout problem not related to my changes

Issue #1828, opened 2023-05-05
[Spark] duplicate events when for delta


pawel-big-lebowski:
Each join query run on top of delta tables have two start and two complete events. Openlineage 0.22..00
https://openlineage.slack.com/archives/C01CK9T7HKR/p1683014620909079

```
Hi Team,
I have configured Open lineage with databricks and it is sending events to Marquez as expected. I have a notebook which joins 3 tables and write the result data frame to an azure adls location. Each time I run the notebook manually, it creates two start events and two complete events for one run as shown in the screenshot. Is this something expected or I am missing something?
```

Sai7656:
Hi @pawel-big-lebowski 
As discussed, we are still seeing two start and two complete events for a single job. I have attached the sample code and log file for reference.

Also, both the events are slightly different in terms of additional information the latest event contains.
for example for two start events the latest start event contains an additional block with "environment-properties"

[OL_Duplicate_Events.zip](https://github.com/OpenLineage/OpenLineage/files/11505548/OL_Duplicate_Events.zip)

Sai7656:
HI @pawel-big-lebowski 
any updates on this issue?

pawel-big-lebowski:
When working on Spark 3.4 support, we are going to introduce extravert filter for delta generating Ol events based on some delta-internal jobs:
https://github.com/OpenLineage/OpenLineage/pull/1790/files#diff-170b33f81a5e8d9d39f4c363923c2b82adfbb0cf27bedb5d575f8e33dacb277a

PR #4885, opened 2026-08-22
spark: keep adaptive events for non-Delta writes

### LINKED ISSUE #4299: [BUG][SPARK] AdaptivePlanEventFilter blocks events for aggregation queries when Delta + AQE are enabled

Spark integration

### OpenLineage version

io.openlineage:openlineage-spark_2.13-1.42.1

- Spark 4.0.0
- Delta Lake 4.0
- Scala 2.13
- Kubernetes with Kyuubi

### Environment configuration

- AQE enabled (default in Spark 3.2+)
- Delta Lake extension configured

### Deployment details

Spark 4.0 on Kubernetes with Kyuubi as SQL gateway.

### Problem details

OpenLineage events are not being emitted for queries involving aggregations, while simple scans work fine.

**Works** (events emitted):
```sql
SELECT * FROM weather_stations LIMIT 1000
```

**Does NOT work** (events blocked):
```sql
SELECT station_id, MAX(wind_speed) AS max_wind_speed
FROM weather_stations
GROUP BY station_id
LIMIT 1000
```

After enabling debug logging, I found that `AdaptivePlanEventFilter` is rejecting the aggregation query:

```
Rejecting event : SparkListenerSQLExecutionStart(...)
with plan : org.apache.spark.sql.catalyst.plans.logical.GlobalLimit
due to filter : io.openlineage.spark.agent.filters.AdaptivePlanEventFilter
```

**Analysis:**

The difference is that `GROUP BY` requires a **shuffle/exchange**, which triggers Adaptive Query Execution (AQE) optimization. The physical plan for the blocked query shows:

```
AdaptiveSparkPlan (6)
+- CollectLimit (5)
   +- HashAggregate (4)
      +- Exchange (3)          <-- shuffle triggers AQE
         +- HashAggregate (2)
            +- Scan parquet ...
```

The `AdaptivePlanEventFilter` (introduced in PR #1830 to fix #1828) blocks events when:
1. Delta extension is configured
2. AND the plan contains `AdaptiveSparkPlan`

```java
  if (!isDeltaPlan()) {
    return false;
  }
  return context
      .getQueryExecution()
      .map(QueryExecution::executedPlan)
      .filter(sparkPlan -> sparkPlan.nodeName().contains("AdaptiveSparkPlan"))
      .isPresent();
}
```

This means **any query with a shuffle** (aggregations, joins, sorts on large data, etc.) gets blocked when using Delta.

The original issue #1828 was specifically about **JOIN queries producing duplicate events** (2 START + 2 COMPLETE). The current filter seems overly broad - it blocks all shuffle-based queries, not just the problematic JOIN scenarios.

---

The filter should be more targeted. Some ideas:

**Option 1: Only filter intermediate plans**
The physical plan shows `isFinalPlan=false` for intermediate adaptive plans:
```java
.filter(sparkPlan -> !sparkPlan.toString().contains("isFinalPlan=true"))
```

**Option 2: Track execution IDs and deduplicate**
Instead of blocking all events, track which `(executionId, eventType)` combinations have been emitted and only filter actual duplicates.

**Option 3: Only apply to JOIN scenarios**
Since the original issue was about JOINs, limit the filter to plans containing join nodes.

**Option 4: Make filters configurable**
Add a config option to disable specific filters for users who prefer potential duplicates over missing events.

---

### How to reproduce

1. Set up Spark 4.0 with Delta Lake and OpenLineage 1.42.1
2. Ensure AQE is enabled (default)
3. Run a simple query: `SELECT * FROM delta_table LIMIT 1000` - events are emitted
4. Run an aggregation: `SELECT col, COUNT(*) FROM delta_table GROUP BY col` - events are blocked

### Additional details

Questions:
1. Is blocking all shuffle-based queries the intended behavior?
2. Would any of the proposed solutions be acceptable?
3. Happy to submit a PR if there's agreement on the approach.

MSDehghan:

Spark: preserve top-level lineage while suppressing SQL executions correlated as children of a root command in Delta-enabled sessions.

### Meaningful description

`AdaptivePlanEventFilter` used to suppress every adaptive execution whenever the Delta session extension was configured. That removes the only terminal lineage pair for top-level non-Delta work such as a Delta read written to Parquet or a read-only aggregation.

The earlier write-target gate did not solve the duplicate side safely: the outer Delta command and its adaptive children are different SQL executions, so the write root and adaptive physical root do not normally occur on the same `QueryExecution`. CTAS and RTAS also use Spark's separate V2 create-table command hierarchy rather than `V2WriteCommand`.

This version correlates executions instead:

- On Spark 3.4+, it reads `SparkListenerSQLExecutionStart.rootExecutionId` reflectively.
- It records whether the root optimized plan is a Spark `Command` and suppresses only executions proven to be children of that command.
- It retains the START-time decision through SQL END, including when Spark has already evicted the original `QueryExecution`, and clears the state on completion/shutdown.
- Spark 3.1-3.3 use a narrow, fail-open fallback: a non-command child must start while exactly one command is active and share its call site or contain a known Delta implementation frame. Ambiguous/concurrent cases are retained.
- The predicate does not enumerate Delta providers, CTAS/RTAS interfaces, or OSS/Databricks/Kernel target classes.
- Delta-extension detection uses the `SparkContext` stored in `OpenLineageContext`, not a process-global active session.
- The broad `Filter`, `LocalRelation`, and `SerializeFromObject` checks in `DeltaEventFilter` now apply only to a correlated command child, so top-level read-only plans survive.

The filter intentionally does not require an `AdaptiveSparkPlan` root. Spark 3.2 RTAS can produce a duplicate child rooted at `ColumnarToRow`, and it must be handled by the same execution correlation.

Validation:

- Full Spark shared and app unit suites pass on Spark 3.2.4.
- Focused shared/app unit suites pass on Spark 4.0.0.
- Real-plan filter integration tests pass 4/4 on Spark 3.2.4 + Delta 1.1.0, Spark 3.4.4 + Delta 2.4.0, and Spark 4.0.0 + Delta 4.0.0.
- Exact terminal-pair tests pass 9/9 on all three pairs above for path save, append, overwrite, CTAS, RTAS, MERGE, Delta-to-Parquet, read-only aggregation, and Parquet CTAS under `DeltaCatalog`.
- The exact tests wait for the listener bus to drain and require one `START`, one `COMPLETE`, one run ID, and the expected final inputs/outputs.
- The complete repository pre-commit suite passes, including PMD and Spotless.
- Databricks CTAS and historical RTAS exact-count tests compile locally; live proprietary-runtime execution still requires an approved Databricks job.

Closes #4299

### Checklist


integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/EventFilterUtils.java
```diff
@@ -56,6 +63,49 @@ static Optional<LogicalPlan> getLogicalPlan(OpenLineageContext context) {
...
+      NamedRelation table = ((V2WriteCommand) root).table();
+      return table instanceof DataSourceV2Relation
+          && isDeltaImplementation(((DataSourceV2Relation) table).table());
+    }
+    if (root instanceof InsertIntoHadoopFsRelationCommand) {
+      // V1 file write; Delta only when the file format itself is Delta's
+      InsertIntoHadoopFsRelationCommand command = (InsertIntoHadoopFsRelationCommand) root;
+      return isDeltaImplementation(command.fileFormat());
+    }
+    return false;
```
mobuchowski: V2WriteCommand does not cover CreateTableAsSelect or ReplaceTableAsSelect; those implement Spark’s separate V2 create-table interfaces. The original duplicate-event reproduction (https://github.com/OpenLineage/OpenLineage/issues/1828#issuecomment-1552662009) uses mode("overwrite").saveAsTable(...), and its captured optimized-plan root is `org.apache.spark.sql.catalyst.plans.logical.ReplaceTableAsSelect`. None of the new branches recognize that Delta write, so isDeltaWritePlan returns false and adaptive events are allowed through, potentially restoring the duplicate START/COMPLETE events this filter originally fixed. Handle Delta V2 CTAS/RTAS targets and add a regression test.

mobuchowski (changes requested):

MSDehghan:
@mobuchowski
 
Thanks for pointing out the CTAS/RTAS gap. You were right: `CreateTableAsSelect` and `ReplaceTableAsSelect` do not implement `V2WriteCommand`.

> **Before the technical details, I want to be transparent:** I used AI to investigate this issue, find the new approach, and implement it because I am not familiar enough with the details of Delta internals, Spark SQL execution events, and Spark's plan hierarchy. This explanation was also written with AI assistance. I verified the behavior using real Spark plans and integration tests, but the investigation and implementation were substantially AI-assisted.

## What was wrong with my previous implementation?

The previous implementation tried to answer this question:

> Is the current SQL execution both an adaptive plan and a Delta write?

It checked both properties on the same `QueryExecution`:

```text
isDeltaWritePlan(currentExecution) && isAdaptivePlan(currentExecution)
```

Your comment identified one problem with `isDeltaWritePlan()`: CTAS and RTAS belong to Spark's separate create-table hierarchy, not `V2WriteCommand`.

While adding tests for that case, I found a more fundamental problem: the Delta write and the adaptive work are normally represented by **different SQL executions**.

## A concrete RTAS example

For a command such as:

```sql
CREATE OR REPLACE TABLE target USING delta AS SELECT * FROM source
```

Spark can produce this execution structure:

```text
Execution 100 — outer user command
  Logical root:  ReplaceTableAsSelect
  Physical root: AtomicReplaceTableAsSelect
  Delta write:   yes
  Adaptive:      no

  └─ Execution 101 — nested work started by the command
       Logical root:  Project / Repartition / Aggregate
       Physical root: AdaptiveSparkPlan
       Delta write:   no
       Adaptive:      yes
```

The old condition behaves like this:

| Execution | Recognized as a Delta write? | Adaptive? | Filtered? |
| --- | --- | --- | --- |
| Outer RTAS command | Yes, after adding RTAS support | No | No |
| Nested internal execution | No | Yes | No |

Neither execution satisfies both conditions. The outer execution knows about the write but is not adaptive. The nested execution is adaptive but has a generic plan root and does not know that its parent is RTAS.

This means adding CTAS/RTAS to `isDeltaWritePlan()` would correct the class hierarchy, but it would still not suppress the duplicate child events. On the OSS Spark 4 plans I measured, the two halves of the condition never occurred on the same execution.

## What does the new implementation do?

The new implementation asks a different question:

> Is this execution an internal child owned by a user-visible command?

For Spark 3.4 and newer, Spark provides this relationship directly through `rootExecutionId`:

```text
Outer command: executionId=100, rootExecutionId=100
Nested child:  executionId=101, rootExecutionId=100
```

The listener now:

1. Reads `rootExecutionId` when a SQL execution starts.
2. Records whether the root execution is a Spark `Command`.
3. Marks execution 101 as a command child only when execution 100 is known and is a command.
4. Keeps that decision until SQL END, because END events do not contain `rootExecutionId`.
5. Keeps the outer command and suppresses only its correlated child execution.

Unknown roots and children of non-command roots are retained. In uncertain cases, the filter prefers a possible duplicate over silently deleting the only useful lineage event.

## Why not keep extending the list of Delta classes?

The class-list approach has several compatibility problems:

- CTAS and RTAS use a different hierarchy from ordinary V2 writes.
- Databricks Delta uses `com.databricks.sql.transaction.tahoe.*`, while OSS Delta uses `org.apache.spark.sql.delta.*`.
- Newer Delta implementations can use `io.delta.*` classes.
- `DeltaCatalog` can route non-Delta tables too. For example, Parquet CTAS under `DeltaCatalog` also produced an outer event pair and a nested event pair.
- Spark changes some CTAS/RTAS interfaces between releases.

Execution ownership covers these cases without needing to recognize every provider class or write shape.

## Why does `AdaptivePlanEventFilter` no longer check for `AdaptiveSparkPlan`?

A real Spark 3.2 RTAS duplicate had a child physical root named `ColumnarToRow`, not `AdaptiveSparkPlan`. It was still an internal child of the RTAS command and produced the same duplicate lineage problem.

The class keeps its existing name for compatibility, but the filtering decision is now based on execution correlation rather than a physical-plan name.

## What about Spark 3.1–3.3?

Those versions do not expose `rootExecutionId`, so they use a deliberately narrow fallback. A candidate is treated as a child only when:

- It is not itself a command.
- Exactly one command is active.
- Its call-site information matches the command or contains a known Delta marker.

If multiple commands are active, the root is missing, or the relationship is ambiguous, the event is retained.

## Related issue found during the investigation

`DeltaEventFilter` also had broad rules for shapes such as `Filter`, `LocalRelation`, and `SerializeFromObject`. Those rules could suppress a top-level read-only query merely because the Delta extension was configured.

Those broad rules are now applied only to a correlated command child, so top-level read-only lineage is retained.

## What is covered by the tests?

The integration tests wait for the Spark listener bus to drain and then require exactly one `START` followed by one `COMPLETE`, using the same run ID and containing the expected inputs and outputs.

They cover:

- Delta path save, append, overwrite, CTAS, RTAS, and MERGE.
- Reading Delta and writing Parquet.
- A read-only aggregation while Delta is enabled.
- Parquet CTAS routed through `DeltaCatalog`.
- Top-level `Filter`, `LocalRelation`, and `SerializeFromObject` plans.
- Unknown roots and children of non-command roots.

The exact-count scenarios pass on:

- Spark 3.2.4 with Delta 1.1.0.
- Spark 3.4.4 with Delta 2.4.0.
- Spark 4.0.0 with Delta 4.0.0.

## Remaining limitation

The historical Databricks RTAS tests compile, but I could not run them on a live Databricks Runtime. The older-Spark fallback is covered on OSS Spark 3.2, but it still needs validation against the original proprietary Databricks reproduction.

Could you please take another look at whether command-root correlation is the right boundary? Feedback on the conservative Spark 3.1–3.3 fallback would be especially helpful.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/EventFilterUtils.java
MSDehghan: please read the comment below

issues/pr-4885-review-follow-up-findings.md
```diff
@@ -0,0 +1,1093 @@
+# PR #4885 review follow-up: Delta write-plan coverage and residual risks
```
mobuchowski: Can we remove this file?

PR #4886, opened 2026-08-22
spark: preserve Delta identity during session teardown

### LINKED ISSUE #4888: [BUG][SPARK] Delta output is lost when listener processing races Spark session teardown

Spark integration

### OpenLineage version

- Reproduced on `main` before #4886 (`1.53.0-SNAPSHOT`)
- Originally observed with `io.openlineage:openlineage-spark_2.13:1.50.0`

- Apache Spark 4.1
- Scala 2.13
- Delta Lake 4.x
- Java 17

### Environment configuration

The OpenLineage Spark listener processes SQL execution events asynchronously. The failing query's last operation creates or replaces a named Delta table and the application immediately stops its Spark session.

### Deployment details

This reproduces with a local Spark deployment; no managed service is required.

### Problem details

`DeltaHandler.getDatasetIdentifier` receives the query's `SparkSession`, but `DeltaCatalog.loadTable` resolves Spark's global active/default session internally. Listener callbacks run on a different thread, where the thread-local active session may be absent. During application teardown, the default session may already be cleared or its SparkContext stopped.

When the final create/replace event races teardown, the catalog lookup fails with:

```text
No active or default Spark session found
```

The exception is handled by the generic safe-apply boundary, but the output dataset is lost. The Spark application can exit successfully while its final COMPLETE event has no Delta output.

The handler should use the supplied session while it is still usable. If teardown has already made catalog access impossible, it should preserve the output identity from information already carried by the logical plan: the table location, path option, or managed-table warehouse location. If the Delta version can no longer be read, only the version facet should be missing — the dataset itself should still be emitted.

### How to reproduce

1. Start Spark 4 with Delta and the OpenLineage Spark listener.
2. Make the final application action a custom-location Delta table write.
3. Stop Spark immediately after the action:

```python
location = "/tmp/ol-delta-session-teardown"

spark.sql("DROP TABLE IF EXISTS default.ol_session_teardown")
spark.sql(
    f"""
    CREATE TABLE default.ol_session_teardown
    USING DELTA
    LOCATION '{location}'
    AS SELECT 1 AS id, 'value' AS payload
    """
)
spark.stop()
```

4. Inspect the final OpenLineage COMPLETE event and listener logs. Repeating the bounded job may be necessary on a fast machine because the defect is a teardown race.

Actual result: Spark exits successfully, the log contains `No active or default Spark session found`, and the COMPLETE event omits the output dataset.

Expected result: the COMPLETE event contains `file:/tmp/ol-delta-session-teardown` (and its table identity where available). If the stopped session prevents reading the Delta snapshot, only the dataset-version facet should be omitted.

### Additional details

PR #4886 temporarily sets the supplied session as active when the listener thread has none. For the specific stopped-session error, it falls back to the plan-carried table `location`, then a case-insensitive `path`, then the managed-table warehouse layout. `location` is authoritative when both keys are present. Other catalog failures are rethrown unchanged.

With the patch, the same Spark 4.1 scenario emits the custom-location Delta output with its catalog, storage, schema, and lifecycle facets instead of nothing. Only the version facet is still missing after teardown.

MSDehghan:

Spark: Delta create/replace events preserve their output identity when asynchronous processing races Spark session teardown.

### Meaningful description

`DeltaHandler.getDatasetIdentifier` calls `DeltaCatalog.loadTable`. Delta resolves the global active/default Spark session internally, even though the handler receives the query's `SparkSession`. OpenLineage callbacks run on an asynchronous listener thread; when a final create/replace event races `spark.stop()`, the global session can already be cleared. The catalog lookup then fails with `No active or default Spark session found`, and the output dataset is dropped.

This change:

1. temporarily sets the passed session as the thread-local active session when the listener thread has none, preserving any pre-existing active session;
2. if Spark 4 has already stopped that session and the lookup fails with this specific error, derives the identifier from the plan-carried table `location`, or the case-insensitive `path` value when no location is present;
3. copies reflective `TableSpec.location` into the catalog properties so create/replace plans retain custom locations across Spark and Databricks API shapes;
4. falls back to the normal managed-table warehouse layout only when the plan has no explicit location;
5. omits the version facet for the same teardown error instead of losing the recovered dataset;
6. rethrows unrelated catalog failures unchanged.

Validation:

- 18 `DeltaHandlerTest` tests and 14 create/replace builder tests pass with both Scala 2.12 and Scala 2.13.
- Spark 3 and Spark 3.5 Spotless and PMD checks pass for both Scala source sets, as does `git diff --check`.
- A Spark 4.1 + Delta job whose final action writes a named, custom-location Delta table and then immediately calls `spark.stop()` reproduces the bug on current `main`: the application exits cleanly but emits no output and logs the failed `DeltaHandler` lookup. With this patch the event contains the custom path along with the Delta catalog, storage, schema, and overwrite lifecycle facets.

Once the session is stopped the Delta version can no longer be read, so the version facet is left out on purpose — the dataset itself is still emitted.

Closes #4888

### Checklist

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
@@ -12,12 +12,14 @@ import io.openlineage.spark.agent.lifecycle.plan.catalog.CatalogHandler;
 import io.openlineage.spark.agent.util.PathUtils;
 import io.openlineage.spark.agent.util.ScalaConversionUtils;
 import io.openlineage.spark.api.OpenLineageContext;
+import java.net.URI;
 import java.util.Map;
 import java.util.Optional;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hadoop.fs.Path;
 import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.catalyst.TableIdentifier;
 import org.apache.spark.sql.catalyst.catalog.CatalogTable;
 import org.apache.spark.sql.connector.catalog.Identifier;
 import org.apache.spark.sql.connector.catalog.Table;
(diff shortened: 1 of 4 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
```diff
@@ -56,11 +59,116 @@ public boolean isClass(TableCatalog tableCatalog) {
...
+    // Path identifier (e.g. delta.`/some/path`): the identifier name is the location itself.
+    if (new Path(identifier.name()).isAbsolute()) {
+      return PathUtils.fromPath(new Path(identifier.name()));
+    }
+
+    Optional<String> location = location(properties);
+    if (location.isPresent()) {
+      return PathUtils.fromTableIdentifier(
+          toTableIdentifier(identifier), session.sparkContext(), new Path(location.get()).toUri());
+    }
```
mobuchowski, line 128: When a plan carries a scheme-less location such as LOCATION '/data/t' while fs.defaultFS is hdfs://namenode:8020, new Path(location).toUri() remains scheme-less and PathUtils classifies it as local file. Spark resolves that location against the Hadoop filesystem, so the teardown fallback emits file:/data/t instead of the real HDFS dataset, splitting its lineage identity; qualify the path using the session's Hadoop configuration first.

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/catalog/DeltaHandler.java
```diff
@@ -56,11 +59,116 @@ public boolean isClass(TableCatalog tableCatalog) {
...
+
+    Optional<URI> warehouseLocation =
+        PathUtils.getWarehouseLocation(
+            session.sparkContext().getConf(), session.sparkContext().hadoopConfiguration());
+    if (warehouseLocation.isPresent()) {
+      Path defaultLocation =
+          PathUtils.reconstructDefaultLocation(
+              warehouseLocation.get().toString(), identifier.namespace(), identifier.name());
```
mobuchowski, line 136: For a managed table without an explicit location, this assumes every database uses the global warehouse layout. If the database was created with a custom location, Spark places the table under that database location, while this fallback emits <warehouse>/<db>.db/<table> and therefore reports a nonexistent dataset; use the session catalog's default table path or database metadata when available instead of reconstructing it solely from the warehouse.

mobuchowski (changes requested):

PR #4889, opened 2026-08-22
integrations: add Temporal integration

merobi-hub:
Adds a Temporal integration

### Meaningful description
Temporal's orchestration system does not offer streaming analytics. Therefore, instead of using a listener, this approach runs a harvester and adapter on the existing Temporal worker in a separate task queue, obtaining events for explicit workflows using a handle offered by the Temporal Python client. The user would be responsible for providing the harvester to the worker, for example in a workflow, and for specifying workflows to be monitored.

Currently this is a simple POC for getting input on the general approach.

Example Temporal workflow:
```py
@workflow.defn
class GetEvents:
    @workflow.run
    async def run(self, workflow_info: list) -> None:
        return await workflow.execute_activity(
            check_workflow_activity,
            workflow_info,
            schedule_to_close_timeout=timedelta(seconds=10),
        )
```

Example Temporal worker:
```py
  events_worker = Worker(
      client,
      task_queue="my-other-task-queue",
      workflows=[GetEvents],
      activities=[check_workflow_activity],
  )
  print("Worker started.")
  await events_worker.run()
```

Example Temporal activity leveraging imported harvester:
```py
@activity.defn
async def check_workflow_activity(workflow_info) -> None:
    client = await Client.connect("localhost:7233")
    await get_temporal_events(workflow_info, client)
```

Workflow ids and datasets are to be provided via the Temporal starter:
```py
events = await client.execute_workflow(
  "GetEvents",
  [
      {"id":"say-hello-workflow","datasets":[{"uri":"duckdb:///database","table":"table","type":"input"}]}, 
      {"id":"say-bye-workflow","datasets":[{"uri":"duckdb:///database","table":"table","type":"output"}]}
  ],
  id="get-events",
  task_queue="my-other-task-queue"
)
```

### Checklist

### DIFF
--- a/integration/temporalio/adapter.py
+++ b/integration/temporalio/adapter.py
@@ -61,23 +61,17 @@ class TemporalOpenLineageAdapter:
             "producer": self.producer,
         }
 
-        try:
-            inputs = [
-                Dataset(namespace=dataset["uri"], name=dataset["table"])
-                for dataset in input_datasets
-            ]
-            kwargs["inputs"] = inputs
-        except KeyError:
-            logger.info(f"No input datasets will be included in {taskname} event.")
+        inputs = [
+            Dataset(namespace=dataset["uri"], name=dataset["table"])
+            for dataset in input_datasets
+        ]
+        kwargs["inputs"] = inputs
 
-        try:
-            outputs = [
-                Dataset(namespace=dataset["uri"], name=dataset["table"])
-                for dataset in output_datasets
-            ]
-            kwargs["outputs"] = outputs
-        except KeyError:
-            logger.info(f"No output datasets will be included in {taskname} event.")
+        outputs = [
+            Dataset(namespace=dataset["uri"], name=dataset["table"])
+            for dataset in output_datasets
+        ]
+        kwargs["outputs"] = outputs
 
         run_event = RunEvent(**kwargs)
         self.client.emit(run_event)
--- a/integration/temporalio/adapter.py
+++ b/integration/temporalio/adapter.py
@@ -1,11 +1,17 @@
 # Copyright 2018-2026 contributors to the OpenLineage project
 # SPDX-License-Identifier: Apache-2.0
 
+# Warning: this integration is experimental and in active development.
+
 import datetime
+import logging
 import os
 
 from openlineage.client import OpenLineageClient
-from openlineage.client.facet import ( JobTypeJobFacet )
+from openlineage.client.event_v2 import Dataset
+from openlineage.client.facet import (
+    JobTypeJobFacet
+)
 from openlineage.client.run import Job, Run, RunEvent, RunState
 from openlineage.client.uuid import generate_static_uuid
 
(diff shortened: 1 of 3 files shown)


integration/temporalio/harvester.py
```diff
@@ -0,0 +1,81 @@
...
+            logger.info("No input datasets found for workflow %", workflow["id"])
+
+        try:
+            output_datasets = [
+                dataset
+                for dataset in workflow["datasets"]
+                if dataset["type"] == "output"
+            ]
+        except KeyError:
+            logger.info("No output datasets found for workflow %", workflow["id"])
```
dolfinus, line 50:
```suggestion
            logger.info("No output datasets found for workflow %", workflow["id"])
            output_datasets = []
```

If exception is raised, then variable value will contain outputs for previous workflow

integration/temporalio/harvester.py
```diff
@@ -0,0 +1,81 @@
...
+    adapter = TemporalOpenLineageAdapter()
+    input_datasets = []
+    output_datasets = []
+
+    for workflow in event_data:
+
+        try:
+            description = await t_client.get_workflow_handle(workflow["id"]).describe()
+        except RPCError:
+            print("Description not found for workflow with id %", workflow["id"])
```
dolfinus: Why `print` instead of `logger`?

integration/temporalio/adapter.py
```diff
@@ -0,0 +1,80 @@
...
+
+from openlineage.client import OpenLineageClient
+from openlineage.client.event_v2 import Dataset
+from openlineage.client.facet import JobTypeJobFacet
+from openlineage.client.run import Job, Run, RunEvent
+from openlineage.client.uuid import generate_static_uuid
+
+PRODUCER: str = "https://github.com/OpenLineage/openlineage/integration/temporal"
+
+NAMESPACE = os.environ.get("TEMPORAL_OPENLINEAGE_NAMESPACE", "default")
```
dolfinus:
```suggestion
NAMESPACE = os.environ.get("OPENLINEAGE_NAMESPACE", "default")
```

To match other implementations. I also recommend moving this into `def __init__`

integration/temporalio/harvester.py
dolfinus: Same for input_datasets

integration/temporalio/harvester.py
```diff
@@ -0,0 +1,82 @@
...
+        start_event_run_id = adapter.build_run_id(start_event_time, start_event_name)
+
+        try:
+            input_datasets = [
+                dataset
+                for dataset in workflow["datasets"]
+                if dataset["type"] == "input"
+            ]
+        except KeyError:
+            logger.info("No input datasets found for workflow %", workflow["id"])
```
dolfinus, line 41:
```suggestion
            logger.info("No input datasets found for workflow %", workflow["id"])
            input_datasets = []
```

integration/temporalio/harvester.py
```diff
@@ -0,0 +1,83 @@
...
+from temporalio.service import RPCError
+
+logger: logging.Logger = logging.getLogger(__name__)
+COMPLETE_STATUS_VALUE = 2
+FAIL_STATUS_VALUE = 3
+
+async def get_temporal_events(event_data: list, t_client: Client) -> None:
+    adapter = TemporalOpenLineageAdapter()
+    input_datasets = []
+    output_datasets = []
```
dolfinus, line 20:
```suggestion
```

integration/temporalio/harvester.py
```diff
@@ -0,0 +1,83 @@
...
+                start_event_run_id,
+                RunState.COMPLETE,
+                complete_event_time,
+                complete_event_name,
+                input_datasets,
+                output_datasets,
+            )
+
+        # Fail event
+        elif description.raw_info.status == FAIL_STATUS_VALUE:
```
dolfinus, line 74: Why failed task can't contain inputs/outputs?

PR #4894, opened 2026-08-24
spark: clean up retained lifecycle state

mobuchowski:

Fix unbounded Spark listener state retention by cleaning completed job metrics and evicting completed jobs and stages without changing emitted lineage events.

#### Problem

The Spark integration retains state after jobs finish:

- On the normal terminal-cleanup path, `JobMetricsHolder.cleanUp` replaces the completed job with another `jobMetrics` entry—even when the aggregate is empty—so the JVM-wide singleton grows with normally completed job count.
- Failure, filtered, missing-query, missing-context, and outputless paths can bypass cleanup of `jobStages` and `stageMetrics`.
- `OpenLineageRunEventBuilder` retains `ActiveJob` and `Stage` objects until the whole execution context becomes collectible. These objects can keep scheduler state and RDD graphs reachable.
- Application shutdown clears listener registries but not all state retained by their contexts or the singleton metrics holder.

This matters most for long-lived streaming jobs, notebooks, Thrift servers, and shared-cluster drivers, where completed job count is effectively unbounded.

#### What changed

- Make job-metrics cleanup unconditional for successful, failed, filtered, missing-query, missing-context, and outputless terminal paths.
- Retain a compact completed-job aggregate only while a later SQL execution event still needs it, then remove it.
- Do not cache empty completed-job aggregates.
- Evict completed `ActiveJob` and `Stage` entries after `JobEnd`, rebuilding the stage index from the jobs that remain active so shared stages stay available.
- Clear execution-context and singleton metrics state during application shutdown, including when OpenLineage is disabled.
- Add low-cardinality gauges for SQL/RDD registries, builder job/stage maps, and all three `JobMetricsHolder` maps.
- Add deterministic lifecycle tests and a 10,000-job bounded-state soak test.

#### Baseline comparison

Compared rebased commit `2317436a3` with its exact parent, upstream `main` at `f20097d7d`, using identical disposable test sources (same SHA-1) on the same host and Java 11 JVM.

Workloads:

- Small timing: 20,000 jobs, one stage and one `TaskMetrics` update per job.
- Medium timing: 10,000 jobs, 10 stages and one `TaskMetrics` update per stage (100,000 stage/task aggregates).
- Retained heap: 100,000 completed one-stage jobs; median of five forced-GC samples.
- Builder retention: 10,000 jobs with 10 stage IDs each.
- Timing values are medians of nine measured runs after three warm-up runs.

| Measurement | `main` baseline | This PR | Change |
| --- | ---: | ---: | ---: |
| Completed `jobMetrics` entries after 100,000 jobs | 100,000 | 0 | -100% |
| Median retained heap after 100,000 jobs | 6.576 MiB | 0.000 MiB | -6.576 MiB |
| Builder jobs retained after 10,000 job ends | 10,000 | 0 | -100% |
| Builder stages retained after 100,000 stage completions | 100,000 | 0 | -100% |
| Small metrics-loop median | 10.112 ms | 5.296 ms | -47.6% |
| Medium metrics-loop median | 13.042 ms | 11.396 ms | -12.6% |

The heap result isolates the completed-entry leak in `JobMetricsHolder`; it does not include the potentially much larger real `ActiveJob`, stage, or RDD graphs. The builder workload reuses mocks, so its result proves retained-key cardinality rather than estimating real graph size. Timing is a directional microbenchmark, not an end-to-end Spark throughput claim. The deterministic 10,000-job soak is part of this PR; the larger comparison harness was kept disposable to avoid adding a GC-sensitive performance test to CI.

#### Validation and confidence

- `./gradlew :shared:check :app:check` passes after rebasing onto newest `main`, including PMD, Spotless, shared Scala variants, and the Spark app unit-test matrix.
- A new integration test runs the actual Spark listener via `spark.extraListeners` for both repeated `count()` actions and an output-producing Parquet write. The byte-identical test source (SHA-1 `cf5dacb01ec1b24f7dbac25ab27e99b9992c361d`) was run against `main` and this PR:

| Real Spark integration case | Completed jobs | `main` retained state | This PR retained state |
| --- | ---: | --- | --- |
| Ten `range(...).count()` actions | 20 | `jobStages=20`, `stageMetrics=20`, `jobMetrics=0` | all zero |
| One Parquet output write | 1 | `jobStages=0`, `stageMetrics=0`, `jobMetrics=1` | all zero |

  Baseline command: `./gradlew :app:integrationTest --tests io.openlineage.spark.agent.JobMetricsCleanupIntegrationTest` (expected two failures). The same command passes on this PR.
- Commit-time pre-commit hooks passed, including Spark PMD and Spotless.
- Added coverage for success, failure, filtered events, missing query execution, missing listener context, outputless RDD jobs, both SQL/job terminal orderings, multiple jobs per SQL execution, and application shutdown.
- The 10,000-job committed soak asserts all three metrics maps return to zero after every terminal cleanup; empty jobs have a completed-map high-water mark of zero.
- Gauge tests verify retained state rises while live and returns to zero on close.

### Checklist

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
@@ -24,6 +24,7 @@ import io.openlineage.spark.agent.lifecycle.ExecutionContext;
 import io.openlineage.spark.agent.util.ScalaConversionUtils;
 import io.openlineage.spark.agent.util.SparkVersionUtils;
 import io.openlineage.spark.api.SparkOpenLineageConfig;
+import java.lang.reflect.InvocationTargetException;
 import java.net.URISyntaxException;
 import java.util.Collections;
 import java.util.HashMap;
@@ -63,6 +64,13 @@ public class OpenLineageSparkListener extends org.apache.spark.scheduler.SparkLi
   static final String RDD_REGISTRY_GAUGE = "openlineage.spark.state.rdd_execution_registry";
   static final String BUILDER_JOBS_GAUGE = "openlineage.spark.state.builder.jobs";
   static final String BUILDER_STAGES_GAUGE = "openlineage.spark.state.builder.stages";
+  static final String METRICS_EXECUTION_GROUPS_GAUGE =
+      "openlineage.spark.state.job_metrics.execution_groups";
+  static final String METRICS_PENDING_JOBS_GAUGE =
+      "openlineage.spark.state.job_metrics.pending_jobs";
+
+  private static final String SQL_EXECUTION_ID_KEY = "spark.sql.execution.id";
+  private static final String SQL_EXECUTION_ROOT_ID_KEY = "spark.sql.execution.root.id";
 
   // These are used only in integration tests to override default factory,
   // before SparkSession creates new listener object
@@ -80,6 +88,8 @@ public class OpenLineageSparkListener extends org.apache.spark.scheduler.SparkLi
   private final Map<Integer, ExecutionContext> rddExecutionRegistry =
       Collections.synchronizedMap(new HashMap<>());
   private final JobMetricsHolder jobMetrics = JobMetricsHolder.getInstance();
+  private final JobMetricsLifecycleManager jobMetricsLifecycle =
+      new JobMetricsLifecycleManager(jobMetrics);
   private final Function1<SparkSession, SparkContext> sparkContextFromSession =
       ScalaConversionUtils.toScalaFn(SparkSession::sparkContext);
   private final Function0<Option<SparkContext>> activeSparkContext =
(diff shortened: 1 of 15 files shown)


pawel-big-lebowski (approved):

integration/spark/app/src/main/java/io/openlineage/spark/agent/OpenLineageSparkListener.java
```diff
@@ -202,55 +226,87 @@ public void onJobStart(SparkListenerJobStart jobStart) {
...
+      log.debug("Unable to read Spark SQL root execution ID", e);
+    }
+    return startEvent.executionId();
   }
 
   /** called by the SparkListener when a job ends */
   @Override
   public void onJobEnd(SparkListenerJobEnd jobEnd) {
     if (checkIfDisabled()) {
+      ExecutionContext disabledContext = rddExecutionRegistry.remove(jobEnd.jobId());
```
dolfinus: nit: this probably can be moved outside the `if` clause

integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/RddExecutionContext.java
```diff
@@ -254,55 +254,76 @@ public void start(SparkListenerJobStart jobStart) {
...
+                      .newRunBuilder()
+                      .runId(runId)
+                      .facets(runFacetsBuilder.build())
+                      .build())
+              .job(buildJob(jobEnd.jobId()))
+              .build();
+      log.debug("Posting event for end {}: {}", jobEnd, event);
+      eventEmitter.emit(event);
+    } finally {
+      JobMetricsHolder.getInstance().cleanUp(jobEnd.jobId());
```
dolfinus: Does it make sense to move this line into `void evictJob(int jobId)` below?

PR #4896, opened 2026-08-24
spark: report datasets for writes through Iceberg's cached catalog

JDarDagran:

**Spark: report datasets for Iceberg writes going through the cached table catalog** — Iceberg maintenance actions such as `rewrite_data_files` write through `SparkCachedTableCatalog` and until now emitted events with no datasets at all.

#### Problem

Iceberg's rewrite actions (`rewrite_data_files`, `rewrite_position_delete_files`, …) do not write through the table's own catalog. They stage the table in `SparkTableCache` under a random UUID and then read and write through `org.apache.iceberg.spark.SparkCachedTableCatalog`, which Iceberg registers as `spark.sql.catalog.default_cache_iceberg`.

That catalog implements `TableCatalog` directly — it is neither a `SparkCatalog` nor a `SparkSessionCatalog` — so `IcebergHandler.isClass` rejects it, `CatalogUtils` throws `UnsupportedCatalogException`, and `DataSourceV2RelationDatasetExtractor` returns an empty list. The resulting `append_data` event names the table it compacted in the job name but carries **no datasets at all**, so a compaction silently contributes no lineage.

#### Solution

Resolve such datasets from the *relation* instead of the catalog.

A new `IcebergRelationHandler` unwraps `relation.table()` to the underlying Iceberg `Table`, recovers the catalog that originally loaded it from the table's fully qualified name, and delegates back to `IcebergHandler` — so a compaction reports exactly the same dataset identifier, symlinks included, as a regular write to the same table. It falls back to the table location when the owning catalog cannot be recovered, and raises `UnsupportedCatalogException` — the one exception the caller is contracted to handle — when there is no location either, rather than letting `new Path(null)` throw out of the handler.

Relation handlers so far were a context-free static list, which a handler needing the Spark session cannot join. `DatasetBuilderFactory` now contributes them per Spark version the same way it already contributes catalog handlers, and `CatalogUtils` merges the shared and the contributed ones. The context-free `CatalogUtils.getDatasetIdentifierFromRelation(relation)` is kept, deprecated in favour of the context-taking overload, so out-of-tree callers still compile. Handler `hasClasses()` answers are memoized per handler class: the probe costs a thrown `ClassNotFoundException`, cannot change within a JVM, and the list is otherwise rebuilt per relation per event.

Both the identifier and the facets take this fallback on the same condition — that no `CatalogHandler` supports the relation's catalog. Gating on *that* rather than on resolution merely having failed leaves the existing Unity Catalog fallback untouched: it names a table after the catalog it was handed, which is right for a real Unity Catalog proxy and wrong for a cached catalog, whose name is the UUID cache key. So storage, catalog and dataset version resolve against the owning catalog too, and a compaction event carries the same facets as a plain write.

Both the relation resolution and the facet lookup catch linkage errors as well as exceptions, since a relation handler reaches into the underlying table's own API and `InputFieldsCollector` calls `getDatasetIdentifierExtended` outside any `PlanUtils#safeApply` — one unresolvable relation must not take the whole plan's lineage down with it.

### Checklist

### DIFF
--- a/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/DefaultRelationHandlers.java
+++ b/integration/spark/app/src/main/java/io/openlineage/spark/agent/lifecycle/DefaultRelationHandlers.java
@@ -1,0 +1,21 @@
+/*
+/* Copyright 2018-2026 contributors to the OpenLineage project
+/* SPDX-License-Identifier: Apache-2.0
+*/
+
+package io.openlineage.spark.agent.lifecycle;
+
+import com.google.common.collect.ImmutableList;
+import io.openlineage.spark.agent.lifecycle.plan.catalog.RelationHandler;
+import io.openlineage.spark.api.OpenLineageContext;
+import io.openlineage.spark3.agent.lifecycle.plan.catalog.iceberg.IcebergRelationHandler;
+import java.util.List;
+
+/** Relation handlers available to every Spark 3.x / 4.x version. */
+final class DefaultRelationHandlers {
+  private DefaultRelationHandlers() {}
+
+  static List<RelationHandler> list(OpenLineageContext context) {
+    return ImmutableList.of(new IcebergRelationHandler(context));
+  }
+}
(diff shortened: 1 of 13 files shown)


mobuchowski:
As long as it's for maintanence operations - like mentioned `rewrite_data_files`, `rewrite_position_delete_files` - it would also make sense to be able to turn off this feature, as they are not _meaningfully_ changing the dataset.

Might make sense to make that a followup though

mobuchowski (approved):

Issue #3431, opened 2025-01-22
[FEATURE][SPARK][STREAMING] Lineage tracking for Apache Iceberg sink and source tables


LuVlk:
### Motivation

Currently, OpenLineage provides support for lineage tracking with Apache Iceberg tables in batch contexts and certain streaming environments. 
However, there is limited or incomplete support for lineage tracking specifically in streaming scenarios involving Iceberg tables as sources and sinks.

This feature requests aims to enhance OpenLineage’s capabilities to fully support lineage tracking in these scenarios.

### Description

Spark: 3.5.2
OpenLineage: 1.25.0
Iceberg: 1.6.1

When we create read or write data to/from iceberg tables, inputs and outputs for the lineage can not be found.

Spark Environment (Snippet)
```
  .config("spark.extraListeners", "io.openlineage.spark.agent.OpenLineageSparkListener")
  .config('spark.openlineage.namespace', 'mynamespace')
  .config('spark.openlineage.parentJobNamespace', 'parent-namespace')
  .config('spark.openlineage.parentJobName', 'hello-world')
  .config('spark.openlineage.parentRunId', str(uuid4()))
  .config('spark.openlineage.transport.type', 'kafka')
  .config('spark.openlineage.transport.topicName', 'openlineage.events')
  .config('spark.openlineage.transport.properties.bootstrap.servers', 'https://<redacted>:9094')
  .config('spark.openlineage.transport.properties.acks', 'all')
  .config('spark.openlineage.transport.properties.retries', '3')
  .config('spark.openlineage.transport.properties.key.serializer', 'org.apache.kafka.common.serialization.StringSerializer')
  .config('spark.openlineage.transport.properties.value.serializer', 'org.apache.kafka.common.serialization.StringSerializer')
```

Spark Streaming Job code
```
spark
  .read
  .format("kafka")
  .option("kafka.bootstrap.servers", "https://<redacted>:9094")
  .option("subscribe", "test-topic")
  .load()
  .writeTo("iceberg.test_table")
  .create()
```
Please refer to [example_message](https://github.com/user-attachments/files/18502481/example_message.json) in pseudo JSON format without the iceberg table as output.

We noticed that same when reading from iceberg in via spark streaming and writing to another iceberg table in foreachBatch mode. The streaming source iceberg table is missing from the generated messages.

### Related issues

- https://github.com/OpenLineage/OpenLineage/issues/3298
- https://github.com/OpenLineage/OpenLineage/issues/2915
[...]

PR #4908, opened 2026-08-31
spark: emit outputs for relation-backed streaming writes

MSDehghan:

Spark: emit catalog-aware output datasets for relation-backed structured-streaming writes on Spark 3.2+.

### Meaningful description

Structured Streaming V2 writes currently emit output datasets only for the exact Kafka `StreamingWrite` class. Iceberg exposes `SparkWrite$StreamingAppend`, so OpenLineage reports an unsupported implementation and emits no output even though Spark 3.2+ exposes the target relation on `WriteToDataSourceV2`.

This PR adds a Spark 3.2+ output builder that:

- handles non-Kafka `MicroBatchWrite` plans when Spark provides a target relation
- delegates to the existing relation output builder so connector extension outputs remain authoritative and ordinary relations fall back to `DataSourceV2RelationDatasetExtractor`
- preserves catalog, storage, schema, symlink, and version facets without extracting ordinary relations twice
- preserves the existing Kafka handling and the Spark 3.1 unsupported-warning path
- returns no fabricated output when the relation is absent

The builder is registered for Spark 3.2, 3.3, 3.4, 3.5, and 4.x. This is a focused implementation of the Iceberg sink portion discussed in #4401 without connector-specific reflection.

Validation completed:

- extension-routing coverage verifies that all connector-provided outputs are preserved and generic extraction is skipped
- generic-routing coverage verifies that ordinary relations are extracted exactly once
- Spark3 tests on Scala 2.12 and focused coverage on Scala 2.13, including Kafka exclusion and empty relations
- functional Iceberg streaming regression on Spark 3.2.4, 3.3.4, and 3.5.6
- Spark 4.1, Iceberg 1.11, and Polaris REST lab run produced exactly one Iceberg output per matching raw COMPLETE event with catalog, schema, symlink, storage, and version facets
- full pre-commit suite, Spark PMD, and Spotless checks

The DataHub-oriented aggregate assertions in the Polaris lab cover lifecycle coalescing and downstream naming outside this upstream extraction change.

Related: #3431

### Checklist

### DIFF
--- a/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/WriteToDataSourceV2DatasetBuilder.java
+++ b/integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/WriteToDataSourceV2DatasetBuilder.java
@@ -8,9 +8,7 @@ package io.openlineage.spark3.agent.lifecycle.plan;
 import io.openlineage.client.OpenLineage;
 import io.openlineage.spark.agent.util.ScalaConversionUtils;
 import io.openlineage.spark.api.AbstractQueryPlanOutputDatasetBuilder;
-import io.openlineage.spark.api.DatasetFactory;
 import io.openlineage.spark.api.OpenLineageContext;
-import io.openlineage.spark3.agent.utils.DataSourceV2RelationDatasetExtractor;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
@@ -37,12 +35,8 @@ public class WriteToDataSourceV2DatasetBuilder
   private static final String KAFKA_STREAMING_WRITE_CLASS_NAME =
       "org.apache.spark.sql.kafka010.KafkaStreamingWrite";
 
-  private final DatasetFactory<OpenLineage.OutputDataset> factory;
-
-  public WriteToDataSourceV2DatasetBuilder(
-      OpenLineageContext context, DatasetFactory<OpenLineage.OutputDataset> factory) {
+  public WriteToDataSourceV2DatasetBuilder(OpenLineageContext context) {
     super(context, false);
-    this.factory = factory;
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/WriteToDataSourceV2DatasetBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+      return Collections.emptyList();
+    }
+
+    DataSourceV2Relation target = relation.get();
+    // Run query-plan visitors first so connector extensions can attach their side effects, such as
+    // Iceberg metrics reporter injection. The relation visitor's datasets are deliberately not
+    // combined with the command result, which keeps a single output.
+    delegate(target, event);
+    return DataSourceV2RelationDatasetExtractor.extract(
+        factory, context, target, includeDatasetVersion(event));
```
mobuchowski: When a non-Kafka V2 streaming sink's table is handled by an OpenLineageExtensionProvider, delegate(target, event) already invokes DataSourceV2RelationOutputDatasetBuilder, which produces the extension-specific dataset. Discarding that result and unconditionally using the generic extractor can emit the wrong name, namespace, or facets—or no output—and also extracts ordinary relations twice; preserve the delegated output instead of replacing it.

mobuchowski (changes requested):

integration/spark/spark3/src/main/java/io/openlineage/spark3/agent/lifecycle/plan/WriteToDataSourceV2DatasetBuilder.java
MSDehghan: Thank you. I think it's fixed now

mobuchowski (approved):

PR #4919, opened 2026-09-04
client: prevent event loss on HTTP redirects

### LINKED ISSUE #4918: [BUG] HTTP redirects can silently discard lineage event bodies

- Java client
- Python client (`http` and `async_http` transports)
- Go client

### OpenLineage version

- Current `main`: `47b78f69cb11e280947e67b3a37c2bd80203bcb3`
- Latest release: `1.53.0`

The affected transport code is unchanged between these revisions.

- Java 17.0.20.1, Apache HttpClient 5.6.4
- Python 3.13.13, Requests 2.34.2, httpx2 2.12.0
- Go 1.26.0, `github.com/hashicorp/go-retryablehttp` 0.7.8
- macOS/Darwin arm64

### Environment configuration

The reproductions use each client's default HTTP endpoint (`api/v1/lineage`), no authentication, no compression, and a local loopback HTTP server. No sensitive configuration is involved.

### Deployment details

This applies when an ingress, reverse proxy, load balancer, or canonical-route handler returns an HTTP 301 or 302 for the configured lineage endpoint and the redirect target returns a successful status.

### Problem details

The Java, synchronous Python, asynchronous Python, and Go HTTP transports can report successful lineage delivery even though the redirected endpoint never receives the event.

All four clients automatically follow a `302 Location: /accepted`. Their HTTP libraries rewrite the event POST as a GET, so the event body is discarded. If `/accepted` returns 200, the transports treat the final response as success. In particular, the asynchronous Python transport records `success: 1, failed: 0`.

Observed request sequence in every client:

```text
POST /api/v1/lineage body=<complete lineage event>
GET /accepted body=
```

Observed client results:

| Client | Result after the redirected GET returns 200 |
| --- | --- |
| Java | `transport_result=success` |
| Python `http` | final status `200`, no exception |
| Python `async_http` | `success: 1, failed: 0` |
| Go | `transport_result=success` |

The relevant behavior is currently configured here:

- Java builds Apache HttpClient with its default redirect strategy and `throwOnHttpError` only rejects 400-599 despite its `non-2xx` comment.
- Python `http` uses Requests' redirect-following default and checks only the final response.
- Python `async_http` explicitly sets `follow_redirects=True` and marks the final 2xx successful.
- Go uses the standard `http.Client` redirect policy and checks only the final response.

Successful emission must mean that a POST carrying the complete lineage event reached the accepted endpoint.

For redirect responses, the clients should use one consistent policy:

- reject 301/302/303 for event POSTs, because those statuses may change the method and discard the body;
- either follow 307/308, which preserve method and body, or return an explicit delivery error;
- never treat a final non-2xx response as successful delivery.

### How to reproduce

1. Start a local HTTP server with two routes:
   - `POST /api/v1/lineage` records the request and returns `302 Location: /accepted`;
   - `/accepted` records the request and returns 200 for any method.
2. Configure any affected OpenLineage HTTP transport with the server's base URL.
3. Emit a minimal RunEvent.
4. Observe that the client reports success while the server records a bodyless `GET /accepted`.
5. Change the redirect status to 307 as a control. The server then records `POST /accepted` with the complete event body.

The 302 case was repeated twice for Java, synchronous Python, asynchronous Python, and Go with identical results. The 307 control preserved the POST and body in all four paths.

Focused existing suites pass on the affected revision:

```text
client/java: ./gradlew :test --tests io.openlineage.client.transports.HttpTransportTest
             13 tests passed

client/python: python -m pytest tests/test_http.py tests/test_async_http.py -q
               78 tests passed

client/go: go test ./pkg/transport
           passed
```

### Additional details

I searched issues and pull requests across all states for combinations of `301`, `302`, `redirect`, `HttpTransport`, `api/v1/lineage`, `lineage event`, and `bodyless GET`; I found no report or fix for this behavior.

#2486 established that Java should raise on non-2xx responses, but the current predicate still accepts 3xx. #4799 currently changes Java HTTP retry configuration but does not alter redirect or response-success semantics.

mattfaltyn:

Prevent Java, Python, and Go HTTP transports from silently discarding lineage event bodies on method-changing redirects and fail emission on terminal non-2xx responses.

### Meaningful description

HTTP clients commonly rewrite a POST to a bodyless GET when following a `301`, `302`, or `303` response. The OpenLineage Java, Python, and Go transports followed those redirects and accepted a successful response from the redirect target, so an event could be reported as emitted even though no lineage payload reached that target.

This change applies one consistent method-preservation policy across the three clients, while retaining each HTTP library's native redirect limit:

- reject `301`, `302`, and `303` before following them;
- continue following `307` and `308`, which preserve the POST method and body; and
- surface a terminal non-2xx response as an emission failure.

Redirect hop limits remain library-native: Java allows 50, Python Requests allows 30, and Go allows 10.

Java enforces the policy on each lineage POST and handles safe redirects itself, including when callers inject a default Apache client through the public constructor or builder. Python uses request-scoped response hooks so caller-owned sessions and unrelated requests retain their existing behavior. Go installs the policy on both retryablehttp's internal client and the returned standard client, while retaining the standard ten-redirect limit.

No dependencies or configuration surface are added. Disabling all redirects was considered, but that would unnecessarily break valid `307` and `308` deployments. Making the behavior configurable was also avoided because silently dropping an event body is never a safe default.

Closes #4918

### Validation

- `cd client/java && DOCKER_HOST=unix:///Users/mattfaltyn/.colima/default/docker.sock TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE=/var/run/docker.sock ./gradlew --no-daemon build check`
- `cd client/python && uv run pytest tests/ --ignore=tests/test_http_integration.py -q` (698 passed)
- `cd client/python && uv run pytest tests/test_http_integration.py -q` (34 passed)
- `cd client/python && uv run mypy --explicit-package-bases src`
- `cd client/go && go test ./...`
- `cd client/go && go build ./...`
- `cd client/go && go vet ./...`
- `cd client/go && go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --config .golangci.yml ./...`
- `prek run --files <changed files>` (all applicable hooks passed)

### Checklist

## Summary by CodeRabbit

* **Behavior Changes**
  * HTTP transports reject 301, 302, and 303 redirects to prevent event data loss.
  * 307 and 308 redirects preserve the POST method and event body.
  * Redirect chains are limited, and circular or cross-origin redirects are rejected where applicable.
  * Non-2xx responses consistently fail event delivery.

* **Documentation**
  * Updated Java, Python, and Go transport guidance to describe redirect handling and proxy requirements.

### DIFF
--- a/client/go/pkg/transport/http_test.go
+++ b/client/go/pkg/transport/http_test.go
@@ -119,6 +119,56 @@ func TestHTTPTransport_Emit_ServerError(t *testing.T) {
 	}
 }
 
+func TestHTTPTransport_Emit_RedirectPolicy(t *testing.T) {
+	tests := []struct {
+		status       int
+		wantErr      bool
+		wantRequests int
+	}{
+		{http.StatusMovedPermanently, true, 1},
+		{http.StatusFound, true, 1},
+		{http.StatusSeeOther, true, 1},
+		{http.StatusTemporaryRedirect, false, 2},
+		{http.StatusPermanentRedirect, false, 2},
+	}
+
+	for _, tt := range tests {
+		t.Run(http.StatusText(tt.status), func(t *testing.T) {
+			var methods []string
+			var bodies [][]byte
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				body, _ := io.ReadAll(r.Body)
+				methods = append(methods, r.Method)
+				bodies = append(bodies, body)
+				if r.URL.Path == "/api/v1/lineage" {
+					w.Header().Set("Location", "/accepted")
+					w.WriteHeader(tt.status)
+					return
+				}
+				w.WriteHeader(http.StatusOK)
+			}))
+			defer srv.Close()
+
+			tr := newHTTPTransport(t, HTTPConfig{URL: srv.URL})
+			_, err := tr.Emit(context.Background(), map[string]string{"event": "lineage"})
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("Emit() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if len(methods) != tt.wantRequests {
+				t.Fatalf("requests = %d, want %d", len(methods), tt.wantRequests)
+			}
+			if !tt.wantErr {
+				if methods[1] != http.MethodPost {
+					t.Errorf("redirected method = %q, want POST", methods[1])
+				}
+				if !bytes.Equal(bodies[0], bodies[1]) {
+					t.Error("redirected body does not match original body")
+				}
+			}
+		})
+	}
+}
+
 // TestHTTPTransport_Emit_GzipCompression verifies that when CompressionGzip is
 // configured the request body is gzip-compressed and the Content-Encoding header
 // is set to "gzip", with the payload still decodable after decompression.
--- a/client/go/pkg/transport/http_test.go
+++ b/client/go/pkg/transport/http_test.go
@@ -169,6 +169,25 @@ func TestHTTPTransport_Emit_RedirectPolicy(t *testing.T) {
 	}
 }
 
+func TestHTTPTransport_Emit_StopsAfterTenRedirects(t *testing.T) {
+	requests := 0
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		requests++
+		w.Header().Set("Location", "/api/v1/lineage")
+		w.WriteHeader(http.StatusTemporaryRedirect)
+	}))
+	defer srv.Close()
+
+	tr := newHTTPTransport(t, HTTPConfig{URL: srv.URL})
+	_, err := tr.Emit(context.Background(), map[string]string{"event": "lineage"})
+	if err == nil || !strings.Contains(err.Error(), "stopped after 10 redirects") {
+		t.Fatalf("Emit() error = %v, want ten-redirect limit error", err)
+	}
+	if requests != 10 {
+		t.Fatalf("requests = %d, want 10", requests)
+	}
+}
+
 // TestHTTPTransport_Emit_GzipCompression verifies that when CompressionGzip is
 // configured the request body is gzip-compressed and the Content-Encoding header
 // is set to "gzip", with the payload still decodable after decompression.
--- a/client/go/pkg/transport/transport.go
+++ b/client/go/pkg/transport/transport.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"time"
 
@@ -64,6 +65,14 @@ func NewWithContext(ctx context.Context, config *Config) (Transport, error) {
 	case TransportTypeHTTP:
 		retryClient := retryablehttp.NewClient()
 		retryClient.Logger = nil // suppress default debug logging
+		checkRedirect := func(req *http.Request, _ []*http.Request) error {
+			switch req.Response.StatusCode {
+			case http.StatusMovedPermanently, http.StatusFound, http.StatusSeeOther:
+				return http.ErrUseLastResponse
+			}
+			return nil
+		}
+		retryClient.HTTPClient.CheckRedirect = checkRedirect
 
 		timeout := defaultTimeout
 		if config.HTTP.TimeoutInMillis > 0 {
@@ -72,6 +81,7 @@ func NewWithContext(ctx context.Context, config *Config) (Transport, error) {
 
 		httpClient := retryClient.StandardClient()
 		httpClient.Timeout = timeout
+		httpClient.CheckRedirect = checkRedirect
 
 		u, err := url.Parse(config.HTTP.URL)
 		if err != nil {
--- a/client/go/pkg/transport/transport.go
+++ b/client/go/pkg/transport/transport.go
@@ -65,7 +65,10 @@ func NewWithContext(ctx context.Context, config *Config) (Transport, error) {
 	case TransportTypeHTTP:
 		retryClient := retryablehttp.NewClient()
 		retryClient.Logger = nil // suppress default debug logging
-		checkRedirect := func(req *http.Request, _ []*http.Request) error {
+		checkRedirect := func(req *http.Request, via []*http.Request) error {
+			if len(via) >= 10 {
+				return errors.New("stopped after 10 redirects")
+			}
 			switch req.Response.StatusCode {
 			case http.StatusMovedPermanently, http.StatusFound, http.StatusSeeOther:
 				return http.ErrUseLastResponse
--- a/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
+++ b/client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
@@ -11,6 +11,7 @@ import static org.apache.hc.core5.http.HttpHeaders.ACCEPT;
 import static org.apache.hc.core5.http.HttpHeaders.AUTHORIZATION;
 import static org.apache.hc.core5.http.HttpHeaders.CONTENT_ENCODING;
 import static org.apache.hc.core5.http.HttpHeaders.CONTENT_TYPE;
+import static org.apache.hc.core5.http.HttpHeaders.LOCATION;
 
 import io.openlineage.client.OpenLineage;
 import io.openlineage.client.OpenLineageClientException;
@@ -26,7 +27,10 @@ import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLContext;
 import lombok.NonNull;
@@ -35,11 +39,11 @@ import lombok.experimental.Delegate;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hc.client5.http.SystemDefaultDnsResolver;
+import org.apache.hc.client5.http.classic.methods.HttpPost;
 import org.apache.hc.client5.http.config.ConnectionConfig;
 import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.entity.GzipCompressingEntity;
 import org.apache.hc.client5.http.impl.DefaultSchemePortResolver;
-import org.apache.hc.client5.http.impl.DefaultRedirectStrategy;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.client5.http.impl.io.DefaultHttpClientConnectionOperator;
@@ -48,22 +52,17 @@ import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManagerBuilder;
 import org.apache.hc.client5.http.io.DetachedSocketFactory;
 import org.apache.hc.client5.http.io.HttpClientConnectionOperator;
-import org.apache.hc.client5.http.protocol.RedirectStrategy;
 import org.apache.hc.client5.http.ssl.DefaultClientTlsStrategy;
 import org.apache.hc.client5.http.ssl.TlsSocketStrategy;
+import org.apache.hc.core5.http.ClassicHttpRequest;
 import org.apache.hc.core5.http.ClassicHttpResponse;
 import org.apache.hc.core5.http.HttpEntity;
-import org.apache.hc.core5.http.HttpRequest;
-import org.apache.hc.core5.http.HttpResponse;
 import org.apache.hc.core5.http.ParseException;
 import org.apache.hc.core5.http.config.Registry;
 import org.apache.hc.core5.http.config.RegistryBuilder;
-import org.apache.hc.core5.http.ProtocolException;
 import org.apache.hc.core5.http.io.SocketConfig;
 import org.apache.hc.core5.http.io.entity.EntityUtils;
 import org.apache.hc.core5.http.io.entity.StringEntity;
-import org.apache.hc.core5.http.io.support.ClassicRequestBuilder;
-import org.apache.hc.core5.http.protocol.HttpContext;
 import org.apache.hc.core5.net.URIBuilder;
 import org.apache.hc.core5.pool.PoolConcurrencyPolicy;
 import org.apache.hc.core5.pool.PoolReusePolicy;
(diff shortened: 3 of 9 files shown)


client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -112,6 +128,7 @@ private static CloseableHttpClient withTimeout(HttpConfig httpConfig) {
     return HttpClientBuilder.create()
         .setDefaultRequestConfig(requestConfig)
         .setConnectionManager(connectionManagerBuilder.build())
+        .setRedirectStrategy(METHOD_PRESERVING_REDIRECT_STRATEGY)
```
mobuchowski: Could we also enforce this policy for injected clients? `HttpTransport(CloseableHttpClient, HttpConfig)` and `Builder.http(...)` bypass `withTimeout`, so this strategy is never installed. With Apache’s default redirect behavior, a POST receiving 301/302/303 can still be followed as a bodyless GET, and a final 2xx would make `emit` succeed even though the event was lost. Please cover this public construction path too, ideally with a regression test using an injected default client.

client/python/src/openlineage/client/transport/http.py
```diff
@@ -426,6 +437,8 @@ def _auth_headers(self, token_provider: TokenProvider) -> dict:  # type: ignore[
         return {}
 
     def _prepare_session(self, session: Session) -> None:
+        if _raise_on_method_changing_redirect not in session.hooks["response"]:
+            session.hooks["response"].append(_raise_on_method_changing_redirect)
```
mobuchowski: Could we scope this hook to the lineage emission request? `HttpConfig.session` may be caller-owned or shared, and `_prepare_session` currently mutates its global response hooks. Unrelated requests through that session—including GETs—will therefore also reject 301/302/303 responses. Passing the guard for `session.post(...)` only, or otherwise filtering it to this POST, would avoid changing behavior outside the transport.

client/go/pkg/transport/transport.go
```diff
@@ -64,6 +65,14 @@ func NewWithContext(ctx context.Context, config *Config) (Transport, error) {
 	case TransportTypeHTTP:
 		retryClient := retryablehttp.NewClient()
 		retryClient.Logger = nil // suppress default debug logging
+		checkRedirect := func(req *http.Request, _ []*http.Request) error {
```
mobuchowski: Could we retain `net/http`’s default redirect-count guard here? Providing `CheckRedirect` replaces the default ten-redirect limit, and because `via` is ignored, a circular or long 307/308 chain can keep replaying the POST body until the overall timeout. Please stop once `len(via) >= 10` while continuing to allow 307/308, ideally with a redirect-loop regression test.

mattfaltyn:
Thanks for the thoughtful review. These were excellent catches, and I implemented all three recommendations in `7970264d5`.

> Could we also enforce this policy for injected clients?

Done. Java now enforces redirect handling on each lineage POST, independent of the injected Apache client's defaults. A regression test with an injected default client verifies that `302` is rejected and `307` preserves the POST method and event body.

> Could we scope this hook to the lineage emission request?

Done. Python now passes the guard only on the lineage POST. Shared session hooks remain unchanged, caller hooks are preserved, and the redirect guard runs first. The regression test covers both the request-local guard and the unchanged shared session.

> Could we retain `net/http`'s default redirect-count guard here?

Done. Go now stops after ten redirects while continuing to allow `307` and `308`. The new redirect-loop regression verifies the ten-request limit and terminal error.

All local validation is green: Java `build check`, Python 698 unit tests plus 34 HTTP integration tests and mypy, Go test/build/vet/golangci-lint, and the applicable repository hooks. Thanks again for the precise feedback!

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -326,30 +337,59 @@ public void emit(@NonNull OpenLineage.JobEvent jobEvent) {
...
-            throwOnHttpError(response);
-            return null;
-          });
+      URI requestUri = uri;
+      Set<URI> visited = new HashSet<>();
+      visited.add(requestUri);
+      int redirectCount = 0;
+      while (true) {
+        HttpPost request = new HttpPost(requestUri);
+        request.setConfig(requestConfig);
```
kacpermuda, line 352: `HttpPost` implements `Configurable`, and in HC5 a per-request `RequestConfig` **replaces** the client's default instead of merging. The old `ClassicRequestBuilder` output was not `Configurable`, so an injected client's `setDefaultRequestConfig(...)` used to apply and now silently doesn't — timeouts, proxy, cookie spec, expect-continue, content-compression are all dropped for lineage POSTs.

Verified with an injected client at `setResponseTimeout(200ms)` against a handler sleeping 1500ms: before this PR `emit` failed after ~319ms, after it succeeds after ~1653ms. That's the same public path the earlier review comment asked you to cover. Could we either let callers supply the `RequestConfig`, or keep builder-level `disableRedirectHandling()` for the internally-built clients and document the override for injected ones?

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -326,30 +337,59 @@ public void emit(@NonNull OpenLineage.JobEvent jobEvent) {
...
+              }
+              return null;
+            });
+        String redirectLocation = redirect.get();
+        if (redirectLocation == null) {
+          return;
+        }
+        redirectCount++;
+        if (redirectCount > MAX_REDIRECTS) {
+          throw new OpenLineageClientException("Maximum redirects (50) exceeded");
```
kacpermuda: The message hardcodes `50` instead of `MAX_REDIRECTS`, and `> MAX_REDIRECTS` permits 51 hops. Also, since `visited` catches any real loop first, this counter only fires on 50 *distinct* URLs — a short comment on why both guards exist would help.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -326,30 +337,59 @@ public void emit(@NonNull OpenLineage.JobEvent jobEvent) {
...
-            return null;
-          });
+      URI requestUri = uri;
+      Set<URI> visited = new HashSet<>();
+      visited.add(requestUri);
+      int redirectCount = 0;
+      while (true) {
+        HttpPost request = new HttpPost(requestUri);
+        request.setConfig(requestConfig);
+        setHeaders(request);
```
kacpermuda, line 353: `setHeaders` re-applies `Authorization` on every hop with no host check. `net/http` and `requests` both strip auth across a host change, and Apache's `RedirectExec` is no longer in the path, so Java is now the outlier. Consider refusing a scheme/host/port change, or dropping auth headers across one.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
```diff
@@ -369,7 +409,7 @@ private void throwOnHttpError(@NonNull ClassicHttpResponse response)
     HttpEntity entity = response.getEntity();
     String body = EntityUtils.toString(entity, UTF_8);
     EntityUtils.consume(entity);
-    if (code >= 400 && code < 600) { // non-2xx
+    if (code < 200 || code >= 300) {
```
kacpermuda, line 444: Widening to all non-2xx is a behavior change beyond the redirect fix — a terminal 3xx or 1xx now fails emission where it used to succeed. Looks intentional, but it isn't reflected in the one-line changelog summary (same widening on the Python side).

client/python/src/openlineage/client/transport/http.py
```diff
@@ -31,6 +31,13 @@
 log = logging.getLogger(__name__)
 
 
+def _raise_on_method_changing_redirect(response: Response, **_: Any) -> None:
+    if response.status_code in (301, 302, 303):
+        response.close()
```
kacpermuda: `response.close()` here throws the body away. Response hooks run before `Session.send` reads `r.content`, so `err.response.text` comes back `''`. Measured against a 302 carrying a body: with `close()` -> `''`, without it -> the body survives, and the connection is still released once content is read.

The async hook does `await response.aread()` first, so the two transports are inconsistent about this.

client/python/src/openlineage/client/transport/http.py
```diff
@@ -386,15 +393,28 @@ def emit(self, event: Event) -> Response:
...
             resp = self.session.post(
                 url=urljoin(self.url, self.endpoint),
                 data=body,
                 headers=headers,
                 timeout=self.timeout,
                 verify=self.verify,
+                hooks={"response": response_hooks},
             )
             resp.close()
+            if isinstance(resp.status_code, int) and not 200 <= resp.status_code < 300:
```
kacpermuda: Two things here:

- `raise_for_status()` is called twice, and for 4xx/5xx the `raise requests.HTTPError` between them is unreachable.
- `isinstance(resp.status_code, int)` looks like it's only present because several existing tests return bare `MagicMock()` responses whose `status_code` isn't order-comparable — which also means a real non-int would silently skip the check.

Could we set `status_code=200` in those mocks and simplify to `if not 200 <= resp.status_code < 300:`?

client/go/pkg/transport/transport.go
```diff
@@ -64,6 +65,17 @@ func NewWithContext(ctx context.Context, config *Config) (Transport, error) {
 	case TransportTypeHTTP:
 		retryClient := retryablehttp.NewClient()
 		retryClient.Logger = nil // suppress default debug logging
+		checkRedirect := func(req *http.Request, via []*http.Request) error {
+			if len(via) >= 10 {
+				return errors.New("stopped after 10 redirects")
```
kacpermuda: This string is load-bearing: retryablehttp suppresses retries by matching `stopped after \d+ redirects\z` against the error. Reword it and a redirect loop quietly becomes 5 retried redirect loops (the test's `requests == 10` assertion is what currently pins this). Worth a comment saying so, and maybe a shared const with the `len(via) >= 10` check above.

client/go/pkg/transport/http_test.go
```diff
@@ -119,6 +119,75 @@ func TestHTTPTransport_Emit_ServerError(t *testing.T) {
...
+		{http.StatusPermanentRedirect, false, 2},
+	}
+
+	for _, tt := range tests {
+		t.Run(http.StatusText(tt.status), func(t *testing.T) {
+			var methods []string
+			var bodies [][]byte
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				body, _ := io.ReadAll(r.Body)
+				methods = append(methods, r.Method)
```
kacpermuda, line 173: `methods`/`bodies` are written from the handler goroutine and read from the test goroutine without the mutex that `testServer` above uses — same for `requests++` in the next test. Passes `-race` today, but it's latently flaky and inconsistent with the file's own convention. The new tests are also missing the `// TestXxx verifies …` doc comments the rest of the file has.

kacpermuda (commented):
Confirmed all three earlier comments are addressed, thanks. Some LLM review, that I think is still valid:

One new regression on the injected-client path, plus smaller notes inline. Three things without a good inline anchor:

- "One consistent redirect policy across the three clients" holds for the 301/302/303 vs 307/308 decision, but not for the hop limit: Java 50, requests 30, Go 10. Each matches its library default so this is fine — just not quite what the description says.
- The Python tests call `_raise_on_method_changing_redirect` directly, and `test_http_transport_does_not_mutate_shared_session_hooks` patches `session.post`, so nothing pins requests' hook-dispatch ordering or `merge_hooks` precedence — the part most likely to break on a requests upgrade. `test_http_integration.py` already has a fastapi/uvicorn harness that could cover it.
- Worth a line in the transport docs: a deployment behind a 301/302-returning proxy that previously appeared to work (while dropping every event) now hard-fails emission, so operators need to move to 307/308.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
mattfaltyn: > Could we either let callers supply the `RequestConfig`, or keep builder-level `disableRedirectHandling()` for the internally-built clients and document the override for injected ones?

Thanks, fixed in `f763b818b`. Java now copies the injected client's complete `RequestConfig` and changes only redirect handling, so caller timeouts, proxy, cookie, compression, expect-continue, and related settings remain intact. A real response-timeout regression test covers this path.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
mattfaltyn: > The message hardcodes `50` instead of `MAX_REDIRECTS`, and `> MAX_REDIRECTS` permits 51 hops.

Thanks, fixed in `f763b818b`. The comparison now uses `>=`, the message uses `MAX_REDIRECTS`, and a comment explains that the visited set catches cycles while the counter bounds distinct redirect chains.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
mattfaltyn: > Consider refusing a scheme/host/port change, or dropping auth headers across one.

Thanks, fixed in `f763b818b`. Java now refuses any scheme, host, or effective-port change before replaying the POST. A regression test sends an Authorization header and verifies that the cross-origin target receives no request.

client/java/src/main/java/io/openlineage/client/transports/HttpTransport.java
mattfaltyn: > Widening to all non-2xx is a behavior change beyond the redirect fix.

Thanks, clarified. The one-line changelog summary now explicitly says terminal non-2xx responses fail emission.

client/python/src/openlineage/client/transport/http.py
mattfaltyn: > `response.close()` here throws the body away.

Thanks, fixed in `f763b818b`. The synchronous hook now consumes the response body before raising, preserving `err.response.text`. Unit and real HTTP integration coverage both verify the redirect body.

client/python/src/openlineage/client/transport/http.py
mattfaltyn: > Could we set `status_code=200` in those mocks and simplify to `if not 200 <= resp.status_code < 300:`?

Thanks, fixed in `f763b818b` and `9e7054590`. The non-int bypass and duplicate `raise_for_status()` are gone, every affected success mock now uses 200, and the failure fixture uses 500.

client/go/pkg/transport/transport.go
mattfaltyn: > This string is load-bearing: retryablehttp suppresses retries by matching `stopped after \d+ redirects\z` against the error.

Thanks, documented in `f763b818b`. The limit and message share `maxRedirects`, the comment records retryablehttp's dependency on the error shape, and the regression still asserts exactly ten requests.

client/go/pkg/transport/http_test.go
mattfaltyn: > `methods`/`bodies` are written from the handler goroutine and read from the test goroutine without the mutex that `testServer` above uses.

Thanks, fixed in `f763b818b`. Both new tests now use mutex-protected writes and snapshots, and both have the repository's standard doc comments. The focused race run passes.

mattfaltyn:
> "One consistent redirect policy across the three clients" holds for the 301/302/303 vs 307/308 decision, but not for the hop limit.

Thanks, clarified in the PR description. The shared behavior is now described as the method-preservation policy, with Java's 50, Python Requests' 30, and Go's 10 redirect limits stated explicitly.

> The Python tests call `_raise_on_method_changing_redirect` directly, and `test_http_transport_does_not_mutate_shared_session_hooks` patches `session.post`, so nothing pins requests' hook-dispatch ordering or `merge_hooks` precedence.

Added Docker-backed FastAPI integration cases for 301, 302, 303, 307, and 308. They verify guard ordering, caller-hook execution, preserved response bodies, method-preserving delivery, and unchanged shared session hooks.

> Worth a line in the transport docs: a deployment behind a 301/302-returning proxy that previously appeared to work now hard-fails emission.

Added operator guidance to the Java, Python, and Go transport docs: proxies must use 307 or 308 to preserve POST bodies, and terminal non-2xx responses fail emission.

Thanks for the thorough follow-up!

PR #4932, opened 2026-09-07
spark: don't skip AQE for Delta for non-internal queries

mobuchowski:
When the Delta extension is configured, AdaptivePlanEventFilter currently rejects every query whose physical plan uses Adaptive Query Execution. This drops all lineage for legitimate top-level aggregations, joins, and other shuffle queries even when they are not internal Delta work.

Spark 3.4 and newer identify nested SQL executions with rootExecutionId. Use that relationship to reject an execution only when it is a child of a root command, preserving the user-visible root execution while continuing to suppress its internal duplicates.

The listener resolves the root QueryExecution at SQL START and retains the filtering decision through SQL END. Spark 3.1 through 3.3 do not expose root execution IDs, so they keep the existing adaptive-plan fallback.

### Checklist


integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkDeltaIntegrationTest.java
```diff
@@ -388,46 +388,54 @@ void testNoDuplicateEventsForDelta() {
...
-              String lastRequestBody = requests[requests.length - 1].getBody().toString();
+  @Test
+  @EnabledIfSystemProperty(named = "spark.version", matches = "^(3\\.[4-9].*|[4-9].*)$")
+  void testTopLevelAdaptiveAggregationEmitsTerminalPair() {
+    clearTables("aqe_read_source");
 
-              return lastRequestBody.contains("/tmp/delta/t4")
-                  && lastRequestBody.contains("create_table_as_select")
-                  && lastRequestBody.contains("COMPLETE")
-                  && requests.length == 2;
```
pawel-big-lebowski, line 429: Where is this assertion now? Wasn't this main idea behind testNoDuplicateEventsForDelta?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/ExecutionContext.java
```diff
@@ -50,6 +50,8 @@ default Optional<Integer> getActiveJobId() {
 
   default void setActiveJobId(Integer activeJobId) {}
 
+  default void setCommandChildExecution(Optional<Boolean> commandChildExecution) {}
```
pawel-big-lebowski: is this the deliberate decision to limit Optional resolving code when method is called?

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DeltaEventFilter.java
```diff
@@ -50,17 +50,18 @@ public DeltaEventFilter(OpenLineageContext context) {
...
-    return isFilterRoot()
-        || isLocalRelationOnly()
+    // These broad plan shapes also occur in user queries; restrict them when Spark knows the root.
+    boolean commandChildOrLegacy = context.getCommandChildExecution().orElse(true);
+    return isOnJobStartOrEnd(event)
+        || (commandChildOrLegacy
+            && (isFilterRoot() || isLocalRelationOnly() || isSerializeFromObject()))
         || isLogicalRDDWithInternalDataColumns()
         || isStagedDeltaTable(event)
-        || isDeltaLogProjection()
```
pawel-big-lebowski, line 61: Isn't this changing behaviour for pre 3.4 Spark? I mean, this can be fine but PR description mentions the behaviour won't change.

pawel-big-lebowski (commented):
Would love to see first this CI succeeding with full-tests label added.

integration/spark/shared/src/main/java/io/openlineage/spark/agent/filters/DeltaEventFilter.java
mobuchowski: Spark before 3.4 does not have rootExecutionId, so it does not change it

integration/spark/app/src/test/java/io/openlineage/spark/agent/SparkDeltaIntegrationTest.java
mobuchowski: It's in `assertSingleTerminalPair`

integration/spark/shared/src/main/java/io/openlineage/spark/agent/lifecycle/ExecutionContext.java
mobuchowski: made it more explicit

pawel-big-lebowski (approved):

PR #4935, opened 2026-09-09
client: add OAuth2 client credentials token provider

rahul-madaan:
Add `oauth2_client_credentials` auth type for the HTTP transport in the Python and Java clients, with the token-endpoint logic of the JWT token provider extracted into a shared base.

### Meaningful description
OpenLineage backends protected by an OAuth 2.0 authorization server issue short-lived access tokens to a client ID and client secret. The `api_key` auth cannot refresh such a token and the `jwt` provider cannot request one, since it exchanges an API key.

This adds `OAuth2ClientCredentialsTokenProvider` (`auth.type: oauth2_client_credentials`) to both clients. The token endpoint handling, caching, refresh buffer and response parsing of `JwtTokenProvider` are extracted into `TokenEndpointTokenProvider`, so both providers share it and only differ in the token request they send.

Options: `clientId`, `clientSecret`, `tokenEndpoint` (required), `scope`, `clientAuthMethod` (`client_secret_basic` by default, or `client_secret_post`), plus the inherited `tokenFields`, `expiresInField` and `tokenRefreshBuffer`. Options are accepted in camelCase and snake_case, so `OPENLINEAGE__TRANSPORT__AUTH__CLIENT_ID` style environment variables work. For `client_secret_basic` the credentials are form-urlencoded before being encoded, as [RFC 6749, section 2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) requires, and both clients now produce the same bytes.

#### Backwards compatibility of the `JwtTokenProvider` refactor

`JwtTokenProvider` keeps its public API, option lookup order, defaults, validation order, exception types and message strings, and the token request it sends is unchanged. Its existing tests are untouched and pass in both languages. I verified the Python provider against the version on `main` and found identical instance attributes across camelCase, snake_case and custom-option configs, identical `KeyError` and `RuntimeError` messages, an identical token request, and refresh triggering on exactly the same second. On the Java side the full public API is preserved, with some members now inherited rather than declared, and the base resolves a logger per concrete class so `JwtTokenProvider` keeps its own log category.

Three deliberate changes, all of which also apply to `jwt`:

- Token fetching is serialized, so concurrent emitters share one token request instead of racing.
- A token endpoint that returns neither an expiry field nor a JWT `exp` claim leaves the token uncacheable and costs a request per event. That was only visible at debug level and is now a warning.
- `JwtTokenProvider.toString()` now nests the inherited fields under `super=`. Redaction is unchanged: `apiKey`, `clientSecret` and the cached token are all excluded.

#### Notes

- Naming: `TokenEndpointTokenProvider` follows the `<source>TokenProvider` pattern of `ApiKeyTokenProvider` and `JwtTokenProvider`. Happy to rename if you prefer something else.
- The Java transport docs listed `apiKey` as an auth type name where the resolver registers `api_key`; corrected in passing on a line this PR already touches.
- Alternative considered: implementing this in the Airflow OpenLineage provider (apache/airflow#72711). Moved here on maintainer feedback so that every integration and every Airflow version can use it; the Airflow PR will be reduced to resolving the credentials from an Airflow connection.
- Tested against a Keycloak-protected backend from both clients: `client_secret_basic` and `client_secret_post`, `composite` and `async_http` transports, environment variable configuration, cached token reuse, rejection of a wrong client secret, and a soak run confirming the token was refreshed before expiry while the expired token was rejected.

### Checklist

https://claude.ai/code/session_01QBfMWG5AG33JMqGGQgAV6v


rahul-madaan:
Heads-up on CI so nobody has to chase it: the two failing `test-integration-spark-java` jobs fail on `WriteToDataSourceV2DatasetBuilderTest.preservesAllExtensionProvidedOutputs` and `extractsOrdinaryRelationExactlyOnceThroughDelegation`, which were added in #4908. They fail the same way on `main` itself at this PR's base commit `9ac19298c` (CircleCI builds 594904 and 594913), so they are unrelated to this change and a rebase will not clear them. Everything this PR touches is green, including `build-client-java`, `build-client-python`, `unit-tests-client-python` and the CPython 3.10-3.14 matrix.

Happy to open a `proposal` issue for the new auth type if you would like one; I skipped it because #4313 added the `jwt` provider the same way without a linked issue.

jakub-moravec:
Hi, thanks for adding the support! I went through the PR and the logic and structure (java) make sense to me. 

A generic question to the community is what about the Go client - @mobuchowski is it OK to leave it out, or should we try to keep the supported mechanisms in sync for all clients?

rahul-madaan:
@jakub-moravec I have added same support for go client as well. https://github.com/OpenLineage/OpenLineage/pull/4940
let's take it up separately.

website/docs/client/python/configuration.md
```diff
@@ -238,10 +238,11 @@ The HTTP transport provides synchronous, blocking event emission. This is the de
 - `timeout` - float specifying timeout (in seconds) value used while connecting to server. Optional, default: `5`.
 - `verify` - boolean specifying whether the client should verify TLS certificates from the backend. Optional, default: `true`.
 - `auth` - dictionary specifying authentication options. Optional, by default no authorization is used. If set, requires the `type` property.
-  - `type` - string specifying value for one of the out-of-the-box available authentication methods (`api_key` or `jwt`), or the fully qualified class name of your TokenProvider. Required if `auth` is provided.
+  - `type` - string specifying value for one of the out-of-the-box available authentication methods (`api_key`, `jwt` or `oauth2_client_credentials`), or the fully qualified class name of your TokenProvider. Required if `auth` is provided.
```
kacpermuda, line 241: Maybe just oauth2? How many other options do we expect here?

kacpermuda (commented):
Python part looks good, one comment. Will take a look again on Monday and merge if no other objections will appear.

website/docs/client/python/configuration.md
rahul-madaan: Happy to go with `oauth2` — it reads better and is more consistent with how `api_key`  and `jwt` are named here. Let me know.

kacpermuda (approved):

mobuchowski (approved):

PR #4939, opened 2026-09-10
go: accept any 2xx response in the HTTP transport

### LINKED ISSUE #4938: Go client rejects HTTP 202 responses from the lineage endpoint
### Problem

The Go client's HTTP transport treats any response other than `200 OK` or `201 Created` as a failure, so a backend that answers with `202 Accepted` makes every `Emit` return an error even though the event was accepted.

`client/go/pkg/transport/http.go`:

```go
if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
    respBody, _ := io.ReadAll(resp.Body)
    return nil, fmt.Errorf("server responded with status %v: %s", resp.StatusCode, respBody)
}
```

The Python client accepts it. `client/python/src/openlineage/client/transport/http.py` calls `response.raise_for_status()`, which raises only for 4xx and 5xx. The Java client behaves the same way: `HttpTransport.throwOnHttpError` throws only when `code >= 400 && code < 600`.

So `202 Accepted`, which is a natural response for an ingestion endpoint that queues events, works with the Python and Java clients but not the Go one.

### Observed

Emitting to a backend that returns `202` gives:

```
server responded with status 202: Event received
```

The event was accepted and the caller still sees an error.

### Suggested fix

Treat any 2xx as success, so the Go client agrees with the other two.

### Contributing

Happy to raise a draft PR for this.

rahul-madaan:
closes: #4938

Go client: treat any 2xx response as success in the HTTP transport, so `202 Accepted` no longer reports a failure.

### Meaningful description
The Go HTTP transport accepted only `200 OK` and `201 Created`, so a consumer that queues events and answers `202 Accepted` made every `Emit` return an error even though the event had been accepted:

```
server responded with status 202: Event received
```

The Python client raises only for 4xx and 5xx (`response.raise_for_status()`), and the Java client does the same (`HttpTransport.throwOnHttpError` throws when `code >= 400 && code < 600`). This change brings the Go client in line by treating any 2xx as success.

Two tests are added: `202 Accepted` succeeds, and a 4xx is still reported as an error. Verified against a real consumer that answers `202`.

Marked draft only because I raised it unprompted alongside #4938; happy to mark it ready whenever you would like it reviewed.

### Checklist

https://claude.ai/code/session_01QBfMWG5AG33JMqGGQgAV6v

## Summary by CodeRabbit

- **Bug Fixes**
  - HTTP requests now succeed for all 2xx response statuses, including `202 Accepted`.
  - Non-2xx responses, such as `401 Unauthorized`, continue to return an error with status and response details.

### DIFF
--- a/client/go/pkg/transport/http.go
+++ b/client/go/pkg/transport/http.go
@@ -157,7 +157,7 @@ func (h *httpTransport) Emit(ctx context.Context, event any) (map[string]string,
 		_ = resp.Body.Close()
 	}()
 
-	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusMultipleChoices {
 		respBody, _ := io.ReadAll(resp.Body)
 		return nil, fmt.Errorf("server responded with status %v: %s", resp.StatusCode, respBody)
 	}
--- a/client/go/pkg/transport/http_test.go
+++ b/client/go/pkg/transport/http_test.go
@@ -105,6 +105,34 @@ func TestHTTPTransport_Emit_CreatedStatusAccepted(t *testing.T) {
 	}
 }
 
+// TestHTTPTransport_Emit_AcceptedStatusAccepted verifies that a 202 Accepted
+// response is treated as success, as consumers that queue events answer with it.
+func TestHTTPTransport_Emit_AcceptedStatusAccepted(t *testing.T) {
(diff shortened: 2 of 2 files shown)


rahul-madaan:
@mobuchowski requesting your review here to make behaviour of python and go client similar.

mobuchowski (approved):

mobuchowski:
Thanks @rahul-madaan

PR #4941, opened 2026-09-11
sql: expand wildcards over derived tables, and respect wildcard modifiers

### LINKED ISSUE #3809: [BUG] SQL parser produces no column lineage for `SELECT * FROM (SELECT ...)`

SQL parser

### OpenLineage version

`openlineage-sql==1.34.0`

_No response_

### Environment configuration

_No response_

### Deployment details

_No response_

### Problem details

I'm currently working on custom integration which uses `openlineage-sql` to extract lineage from Oracle queries. And some of these queries are generated, leading to queries like:
```sql
SELECT tbl.* FROM (SELECT col1, col2, col2 FROM table) tbl
```

`openlineage-sql` is able to extract table names from a query, but it doesn't produce column lineage for this case, because it skips `SELECT *`.

_No response_

### How to reproduce

```python
import openlineage_sql

r = openlineage_sql.parse([
  """
INSERT INTO table1 SELECT a, b, c FROM table2
"""], dialect="oracle"
)

assert len(r.column_lineage) > 0

r = openlineage_sql.parse([
  """
INSERT INTO table1 SELECT * FROM (SELECT a, b, c FROM table2)
"""], dialect="oracle"
)

assert len(r.column_lineage) > 0  # raises AssertionError

r = openlineage_sql.parse([
  """
INSERT INTO table1 SELECT tbl.* FROM (SELECT a, b, c FROM table2) tbl
"""], dialect="oracle"
)

assert len(r.column_lineage) > 0  # raises AssertionError
```

### Additional details

_No response_

- [ ] I am interested in doing this work

lhadhazy:
closes: #3809

Expand wildcards over derived tables, and stop discarding the modifiers attached to a wildcard.

### Meaningful description

**This PR carries two commits, and the first is not mine.** `sql: add cte column registry for wildcard expansion` is @simonjobs' work from his fork, rebased onto current `main`. He offered it for anyone to continue and it had never been opened here, so it is included rather than reimplemented — the CTE case it solves is what the derived-table case builds on, and separating them would mean either rewriting his code as mine or submitting something that does not compile. His authorship is preserved in the commit.

#### The problem

The parser produces no column lineage for the form this issue reports:

```sql
INSERT INTO t1 SELECT * FROM (SELECT a, b, c FROM t2)
```

which is identical in meaning to `INSERT INTO t1 SELECT a, b, c FROM t2`, and that already works. The projection is known inside the query, so there is nothing to guess at.

A second problem turned up while testing the CTE branch. The match arm bound `SelectItem::Wildcard(_)`, discarding `WildcardAdditionalOptions`, so every modifier attached to a wildcard was ignored:

| SQL | Reported | Correct |
|---|---|---|
| `SELECT * EXCLUDE (secret)` | `id`, `secret` | `secret` is not in the output |
| `SELECT * EXCEPT (secret)` | `id`, `secret` | same, BigQuery spelling |
| `SELECT * RENAME (x AS y)` | `id`, `x` | the output column is `y` |

That is a **wrong edge rather than a missing one**, which seems worse: a missing facet is visibly absent, whereas these look like ordinary lineage. It matters for anything treating lineage as evidence rather than as a browsing aid.

#### The solution

A derived table registers its output columns when it is visited, keyed on its alias, or on the subquery's own text when it has none. The FROM clause is visited before the projection, so the entry is in place by the time a wildcard reaches it. This reuses the registry the first commit introduces rather than adding a second mechanism.

`WildcardFilter` reduces `WildcardAdditionalOptions` to the exclusions and renames expansion needs. EXCLUDE and EXCEPT collapse together, being the Snowflake and BigQuery spellings of one idea.

`expand_wildcard_for_cte_table` becomes `expand_wildcard_for_known_table`, since it now serves CTEs and aliased derived tables both.

#### The boundary is unchanged

A wildcard over a physical table still yields nothing. The parser has no schema there, so guessing would be worse than silence — @tnazarew's point when this issue was opened, and the reason the fix is scoped to projections that are knowable from the query text. There is a test pinning it, alongside the equivalent one for the CTE case.

#### One thing I changed in the first commit

I restored the `_ => {}` arm on the `SelectItem` match. Enumerating every variant turned a `sqlparser` bump into a compile error in this file, and that is not hypothetical: rebasing onto current `main` broke on `SelectItem::ExprWithAliases`, which did not exist when the branch was written. The same rebase needed two other fixes — `ExcludeSelectItem` now carries `ObjectName` rather than `Ident`, and a newer clippy flags the `if let` inside the `QualifiedWildcard` arm.

#### Alternatives considered

Implementing derived-table expansion separately from the CTE registry, which would have duplicated the lookup for no benefit. And keying unaliased derived tables on a counter rather than on the subquery text, which is cheaper but breaks if the visit order ever changes.

#### Tests

Seven added in `tests_wildcard.rs`, six of which fail without this change. **163 pass, none regressed**, including the eight CTE tests from the commit this builds on. Verified with the same three commands CI runs, from the workspace root: `cargo test --no-default-features`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo fmt -- --check`.

### Checklist

---

#### A note on the sign-off, since it is unusual

The first commit is authored by @simonjobs and **signed off by me rather than by him**. He never opened a PR here, so it carried no sign-off and the DCO check failed on it.

I signed it under DCO clause (b): the work is covered by this project's own Apache-2.0 licence, and it is submitted here with modifications under the same licence. The author field is untouched — GitHub still shows him as the author of that commit, and the commit message records what changed during the rebase.

If you would rather have his own sign-off on it, I am happy to wait for him, or to squash the two commits into one authored by me with a `Co-authored-by` trailer. Whichever you prefer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQL column lineage for wildcard selections from CTEs and derived tables.
  - Added support for qualified wildcards, including aliased and nested queries.
  - Correctly applies positional column aliases and preserves projection order.
  - Honors `EXCLUDE`/`EXCEPT` and `RENAME` modifiers with appropriate identifier matching.
  - Preserves source mappings across chained CTEs, joins, and mixed wildcard projections.

- **Tests**
  - Added coverage for wildcard lineage, CTEs, derived tables, nested queries, column aliases, and dialect-specific modifiers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased](https://github.com/OpenLineage/OpenLineage/compare/1.53.0...HEAD)
 
+### Fixed
+
+* **SQL: Expand wildcards over derived tables, and respect wildcard modifiers** [`#4941`](https://github.com/OpenLineage/OpenLineage/pull/4941) [@lhadhazy](https://github.com/lhadhazy) with [@simonjobs](https://github.com/simonjobs)
+  *Emits column lineage for `SELECT * FROM (SELECT ...)`, and stops reporting columns that `EXCLUDE`, `EXCEPT` or `RENAME` removed or renamed from the output.*
+
 ## [1.53.0](https://github.com/OpenLineage/OpenLineage/compare/1.52.0...1.53.0)
 
 ### Added
--- a/integration/sql/impl/src/context/mod.rs
+++ b/integration/sql/impl/src/context/mod.rs
@@ -81,6 +81,9 @@ pub struct Context<'a> {
     dialect: &'a dyn CanonicalDialect,
     // Used to generate unique names for unaliased columns created from compound expressions
     column_id: u32,
+    // Registry of known column names for CTE tables, keyed by qualified_name().
+    // Populated during With::visit(), consumed during wildcard expansion in Select::visit().
+    cte_column_registry: HashMap<String, Vec<String>>,
 }
 
 impl<'a> Context<'a> {
@@ -94,6 +97,7 @@ impl<'a> Context<'a> {
             default_database: None,
             dialect: &SnowflakeDialect,
             column_id: 0,
+            cte_column_registry: HashMap::new(),
         }
     }
 
@@ -110,6 +114,7 @@ impl<'a> Context<'a> {
             default_database,
             dialect,
             column_id: 0,
+            cte_column_registry: HashMap::new(),
         }
     }
 
@@ -578,6 +583,25 @@ impl<'a> Context<'a> {
             for (alias, t) in old.aliases.tables() {
                 frame.aliases.add_table_alias(t.clone(), alias.clone());
             }
+        }
+    }
+
+    // --- CTE Column Registry ---
+
+    pub fn register_cte_columns(&mut self, table_qualified_name: String, columns: Vec<String>) {
+        self.cte_column_registry
+            .insert(table_qualified_name, columns);
+    }
+
+    pub fn cte_columns(&self, table_qualified_name: &str) -> Option<&Vec<String>> {
+        self.cte_column_registry.get(table_qualified_name)
+    }
+
+    pub fn resolve_table_qualified_name(&self, table: &DbTableMeta) -> String {
+        if let Some(frame) = self.frames.last() {
+            frame.aliases.resolve_table(table).qualified_name()
+        } else {
+            table.qualified_name()
         }
     }
 
--- a/integration/sql/impl/src/visitor.rs
+++ b/integration/sql/impl/src/visitor.rs
@@ -6,12 +6,12 @@ use crate::lineage::*;
 
 use anyhow::{anyhow, Result};
 use sqlparser::ast::{
-    AccessExpr, AlterTableOperation, CopyIntoSnowflakeKind, CreateTableLikeKind, Expr, FromTable,
-    Function, FunctionArg, FunctionArgExpr, FunctionArguments, Ident, Join, JoinConstraint,
-    JoinOperator, ObjectName, ObjectNamePart, Query, RenameTableNameKind, Select, SelectItem,
-    SelectItemQualifiedWildcardKind, SetExpr, Statement, Subscript, Table, TableFactor,
-    TableFunctionArgs, TableObject, UpdateTableFromKind, Use, Value, ValueWithSpan, WindowSpec,
-    WindowType, With,
+    AccessExpr, AlterTableOperation, CopyIntoSnowflakeKind, CreateTableLikeKind, ExcludeSelectItem,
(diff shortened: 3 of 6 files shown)


lhadhazy:
CI note: the two red `test-integration-spark-java` jobs are not from this change. They fail on `WriteToDataSourceV2DatasetBuilderTest.preservesAllExtensionProvidedOutputs` and `extractsOrdinaryRelationExactlyOnceThroughDelegation`, added by commit `aab43df6` on 2026-09-07. `main` was green on `1f389d9e` immediately before that commit and has been red on both jobs since, and this branch is based on `9ac19298`, so a rebase will not clear them — the same two jobs are red on several other open PRs.

Everything this change touches is green, including the Rust SQL jobs and the CPython matrix. Happy to rebase once `main` is fixed.

lhadhazy:
Rebased onto `e248b98e`, and the two `test-integration-spark-java` jobs that were red when I opened this are green — they pass now on every open PR that runs them, and on #4936's merge run from 2026-09-14. Nothing appears to have fixed them deliberately: no commit has touched `WriteToDataSourceV2DatasetBuilderTest` since `aab43df6` introduced the failures, which reads as environment-dependent rather than a code defect.

The automated review then found four real defects in this change, and the fourth commit fixes them. Each has a test that fails without its fix.

**Column order was arbitrary.** The registry took its names from `column_ancestry`, a `HashMap`. Emitted lineage is sorted by descendant so nothing showed, but a positional column alias has to bind to the first projected column. A frame now records the order its columns are first seen.

**The registry outlived its query.** It sat on the `Context`, so a CTE named `d` in a nested query replaced an outer `d` for the rest of the statement. It now lives on the frame and is read innermost-outward, so an inner declaration shadows rather than overwrites.

**Explicit column aliases were ignored.** `WITH d(x) AS (SELECT a FROM t)` publishes `x`; both registration paths read the inner name. Aliases now bind positionally and rename the ancestry keys with them.

**Modifier matching was byte equality.** `EXCLUDE`, `EXCEPT` and `RENAME` exist in Snowflake, BigQuery and DuckDB, where an unquoted identifier resolves without regard to case, so `EXCLUDE (SECRET)` left `secret` in the output. Unquoted identifiers now match case-insensitively, quoted ones exactly.

Happy to split the modifier fix into its own PR if that is easier to review, and happy to make any changes you want.

lhadhazy:
@kacpermuda — you merged the last two changes to this crate, so rather than let this keep sitting: it is rebased onto current main and green.

It was red for its first nine days on two `test-integration-spark-java` jobs that had nothing to do with this change, which is probably why it slipped past triage; they have been green since main recovered, and the PR has been rebased twice since.

The automated review found seven issues across three rounds and all of them are fixed, each with a test that fails without its fix — the substantive ones were positional column aliases binding to the wrong columns, the registry outliving the query that declared it, and the wildcard modifiers this change adds handling for. `integration/sql` is at 182 tests.

No rush at all, and happy to split the modifier half into its own PR if that is easier to review.

Issue #4951, opened 2026-09-13
[FEATURE] Spark: Support Iceberg metrics reporting for cached-table operations


MSDehghan:
## Motivation

It would be useful to collect Iceberg metrics for maintenance operations such as `rewrite_data_files`, including scan planning time and the final commit's duration, retry count, and file changes.

These operations can use `SparkCachedTableCatalog`, which loads tables from Iceberg's Spark table cache. OpenLineage's reporter injector only knows how to unwrap `SparkCatalog` and `SparkSessionCatalog`, so it can't register a reporter through this catalog.

Skipping the cached catalog fixes the injector exception, but leaves this path unsupported. Some scan reports are already available directly from scan objects, and an existing reporter on the original catalog may capture other reports. It would help to establish what's still missing in a real rewrite and add support for those reports.

## Description

Could we collect reports from the underlying table or operation when Spark is using `SparkCachedTableCatalog`? The existing `icebergScanReport` and `icebergCommitReport` facets should be enough; I don't think this needs a new facet.

The main question is where to attach the reporter and when the report becomes available. The cached catalog doesn't expose an underlying Iceberg catalog through `icebergCatalog()`, so it won't fit the current injection code directly. It may be possible to use the table, scan, or original catalog instead.

Commit timing also needs care. An internal append may only stage rewritten files, with the enclosing maintenance action committing the new snapshot later. The final commit report needs to belong to that operation, rather than whichever append happened to touch the same table.

A useful test would run `rewrite_data_files` on a table with several small files, confirm that files were actually rewritten, and check that the newly collected reports reach the right lineage events. It should also cover repeated operations so reports don't get duplicated or carried into a later run.

The existing scan-report path and any user-configured reporter should keep working. The new collection should respect `spark.openlineage.vendors.iceberg.metricsReporterDisabled`, and unavailable metrics should not interrupt normal lineage collection.

If only some reports can be captured reliably, supporting those first and documenting the remaining limitations seems reasonable.

## Related issues

This follows #4950. The exception can be fixed by safely skipping the catalog; this request is about adding metrics support for it.

[...]

PR #4953, opened 2026-09-13
spark: document Iceberg maintenance report ownership

MSDehghan:
Document Iceberg maintenance report ownership and verify it in the real compaction integration test.

### Meaningful description
A real `rewrite_data_files` replay produces a planning `ScanReport` and a final `replace` `CommitReport`, but the maintenance CALL has no datasets on which to attach them. The cached-catalog append reads preplanned tasks and stages files; assigning the CALL's reports to that append would misattribute the metrics.

This PR follows the investigation/design path for #4951. It does **not** implement cached-catalog metrics collection and does **not** close the feature issue. The proposal records the baseline, traces Iceberg's reporter ownership and commit timing, and describes the earlier collection hook and execution correlation needed for an implementation. The dataset metrics documentation states the current limitation.

The existing compaction test now observes the native table reporter during the CALL, forwards callbacks to the original reporter, and restores the reporter afterward. It verifies the reports' enclosing SQL execution and before/after snapshots, actual file rewriting, the real table's output identity, and the absence of a commit report on the staging append. It does not require the CALL to remain without datasets, so future support can extend this coverage.

Related to #4951. The injector exception fix is separate in #4952; this PR is based directly on `main` and contains no commits from that PR. The uninstrumented baseline documented in the proposal was collected with #4952 applied.

Validation on Java 17:
- Spark 3.5.6 / Scala 2.13 / Iceberg 1.6.0: the compaction ownership test and ordinary scan/commit report integration tests passed together.
- Spark 3.2.4 / Scala 2.12 / Iceberg 1.4.3: the compaction ownership test passed.
- `:app:pmdTest`, Spotless, `uvx prek run --all-files`, and `git diff --check` passed.

The full Spark/JDK matrix was not run locally. No production reporter hook, new facet, or dependency minimum is introduced.

### Checklist


MSDehghan:
closed

PR #4956, opened 2026-09-13
client/java: do not count GC time accumulated before the first circuit breaker check

### LINKED ISSUE #4954: [BUG] client/java: JavaRuntimeCircuitBreaker counts GC time accumulated before its first check, tripping the breaker spuriously

Java client

### OpenLineage version

main branch (1.54.0-SNAPSHOT)

Java 17, `io.openlineage:openlineage-java`

### Environment configuration

```yaml
circuitBreaker:
  type: javaRuntime
  memoryThreshold: 20
  gcCpuThreshold: 10
```

### Problem details

`JavaRuntimeCircuitBreaker.currentState()` keeps a per-thread baseline of the old generation GC
collection time in `lastTotalGCTimeNS` and derives the GC CPU percentage from the difference
between checks. The early-return branch stores the wrong quantity in that baseline:

```java
long currentTimeInNanoseconds = System.nanoTime();
long gcCpuTime = getGCCpuTimeNS() - lastTotalGCTimeNS.get();
long elapsedTime = currentTimeInNanoseconds - lastTimestampInNanoseconds.get();
double gcCpuTimePercentage = (gcCpuTime / (double) elapsedTime) * 100;
if (elapsedTime <= 0) {
  lastTimestampInNanoseconds.set(currentTimeInNanoseconds);
  lastTotalGCTimeNS.set(gcCpuTime);      // <- a delta, not the absolute collection time
  return new CircuitBreakerState(false);
}
```

Everywhere else `lastTotalGCTimeNS` holds the absolute collection time - the normal path sets
`lastTotalGCTimeNS.get() + gcCpuTime`, which is exactly the value returned by `getGCCpuTimeNS()`.

That branch is not an edge case, it is taken on the first call on every thread.
`lastTimestampInNanoseconds` is a lazily initialized `ThreadLocal` whose `initialValue()` calls
`System.nanoTime()`, and it is read *after* `currentTimeInNanoseconds` has already been captured,
so on the first call `elapsedTime` is always negative. The existing tests all work around this
with a discarded `currentState()` call commented `// first run to collect elapsed time`.

On that first call `gcCpuTime` is the difference between two reads of the same counter taken
microseconds apart, i.e. approximately zero. So after the breaker is primed the baseline is ~0
instead of the collection time reported by the old generation GC bean.

The second check then computes

```
gcCpuTime = <total GC CPU time since JVM startup> - ~0
```

and divides it by the interval between the two checks. On a JVM that has been running for a
while this is a value in the thousands of percent, so `gcCpuTimePercentage >= gcCPUThreshold`
is satisfied by garbage collection that happened before OpenLineage was even initialized. As
soon as free memory is below `memoryThreshold` the breaker trips and lineage collection is
skipped, and the reason string reports the same inflated number:

```
Circuit breaker tripped at memory 19.99%  GC CPU time 412700.00% (freeMemoryThreshold 20%, gcCPUThreshold 10%)
```

A related smaller problem in the same block: `gcCpuTimePercentage` is computed before the
`elapsedTime <= 0` guard, so on that path it divides by zero or by a negative interval.

The baseline recorded on the priming call should be the absolute collection time, the same value
the normal path records. Then a second check taken when no garbage collection has happened
reports 0% GC CPU time and the breaker stays open.

### How to reproduce

Mock a JVM that accumulated GC CPU time before the breaker was created and does not collect
anything between the two checks:

```java
GarbageCollectorMXBean gcBean = mock(GarbageCollectorMXBean.class);
when(gcBean.getName()).thenReturn("oldGcBean");
when(gcBean.getCollectionCount()).thenReturn(2L);
when(gcBean.getCollectionTime())
    .thenReturn(TimeUnit.MILLISECONDS.convert(100, TimeUnit.SECONDS));
when(RuntimeUtils.getGarbageCollectorMXBeans()).thenReturn(Collections.singletonList(gcBean));

// free memory below memoryThreshold, so only the GC CPU time decides the state
when(RuntimeUtils.totalMemory()).thenReturn(26000L * MEGABYTE + MEGABYTE);
when(RuntimeUtils.freeMemory()).thenReturn(2000L * MEGABYTE - MEGABYTE);
when(RuntimeUtils.maxMemory()).thenReturn(30000L * MEGABYTE);

JavaRuntimeCircuitBreaker circuitBreaker = new JavaRuntimeCircuitBreaker(config);
circuitBreaker.currentState().isClosed();               // priming call, false
circuitBreaker.currentState().isClosed();               // expected false, actually true
```

The second call reports the full 100 seconds of GC CPU time as if it had happened during the
few microseconds between the two checks.

### Additional details

Reproducible on every JVM: the priming call always takes the `elapsedTime <= 0` branch, so the
breaker is always evaluated against a zero baseline on its second check. The practical impact
depends on how much GC the process did before OpenLineage started and on whether free memory is
below the configured threshold.

Daksha1611:
closes: #4954

**client/java**: fix `JavaRuntimeCircuitBreaker` counting garbage collection that happened before its first check.

### Meaningful description

`JavaRuntimeCircuitBreaker` keeps a per-thread baseline of the old generation GC collection time
in `lastTotalGCTimeNS` and derives the GC CPU percentage from the difference between two checks.
The `elapsedTime <= 0` branch stored a delta there instead of the absolute collection time that
every other path stores.

That branch runs on the first call on every thread: `lastTimestampInNanoseconds` is a lazily
initialized `ThreadLocal` whose `initialValue()` calls `System.nanoTime()`, and it is read after
`currentTimeInNanoseconds` has already been captured, so `elapsedTime` is always negative there.
The existing tests already work around this with a discarded `currentState()` call commented
`// first run to collect elapsed time`.

Since the delta between two reads of the same counter taken microseconds apart is approximately
zero, the breaker was left with a ~0 baseline. The next check then measured the whole GC CPU time
accumulated since JVM startup against the interval between the two checks, producing percentages
in the thousands and tripping the breaker on garbage collection that predates OpenLineage.

The fix reads `getGCCpuTimeNS()` once into a local and records that same absolute value on both
paths. The normal path previously wrote `lastTotalGCTimeNS.get() + gcCpuTime`, which is the same
value, so this is not a behaviour change there. Computing `gcCpuTimePercentage` was also moved
below the `elapsedTime <= 0` guard, since on that path it divided by zero or by a negative
interval.

The added test primes the breaker against a GC bean that already reports 100 seconds of
collection time and does not collect anything between the two checks. The second check has to
report 0% GC CPU time; before this change it reported the full 100 seconds as if it had happened
between the two calls, and the breaker tripped.

### Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved circuit-breaker handling of garbage-collection timing during initial checks.
  - Prevented pre-existing garbage-collection activity from incorrectly contributing to utilization calculations.
  - Ensured subsequent checks use the latest garbage-collection baseline for more accurate state evaluation.
  - Improved reliability of circuit-breaker state evaluation when timing intervals are unavailable or non-positive.
  - Ensured garbage-collection utilization is calculated only from activity occurring between valid checks.


Daksha1611:
The failing Spark jobs here are unrelated to this change.

Both jobs fail on the same two tests:

```
WriteToDataSourceV2DatasetBuilderTest > preservesAllExtensionProvidedOutputs()
WriteToDataSourceV2DatasetBuilderTest > extractsOrdinaryRelationExactlyOnceThroughDelegation()
```

They already fail on main at 9ac19298c, the commit this branch is based on. On that commit every
build that actually executed the `spark3` test task failed on these two tests (595525, 595625).
The two green builds on the same commit (595532, 595646) never ran the task at all - it was
`UP-TO-DATE` from the Gradle build cache - so they are not evidence that the tests pass.

This PR does not touch the Spark integration, and `build-client-java` passes.

mobuchowski:
@Daksha1611 can you rebase the PR?

Daksha1611:
@mobuchowski rebased onto main, done. The Spark failure I mentioned above is fixed there now, so the build is green.

PR #4957, opened 2026-09-13
sql: make DbTableMeta hashCode consistent with equals

### LINKED ISSUE #4955: [BUG] sql: DbTableMeta violates the equals/hashCode contract, breaking ColumnMeta lookups in hash-based collections

SQL parser, Spark integration

### OpenLineage version

main branch (1.54.0-SNAPSHOT), `io.openlineage:openlineage-sql-java`

Java 17

### Environment configuration

_No response_

### Deployment details

_No response_

### Problem details

`DbTableMeta.equals` compares the qualified name, which folds away a null database or schema:

```java
public String qualifiedName() {
  return String.format(
      "%s%s%s", database != null ? database + "." : "", schema != null ? schema + "." : "", name);
}

@Override
public boolean equals(Object o) {
  ...
  DbTableMeta other = (DbTableMeta) o;
  return qualifiedName().equals(other.qualifiedName());
}
```

`hashCode` is derived from the three fields separately instead:

```java
@Override
public int hashCode() {
  return new HashCodeBuilder().append(database).append(schema).append(name).toHashCode();
}
```

So two instances can be equal and still hash differently, which breaks the
`equals`/`hashCode` contract:

```java
new DbTableMeta(null, "db", "table").equals(new DbTableMeta("db", null, "table"))  // true
new DbTableMeta(null, "db", "table").hashCode()
    == new DbTableMeta("db", null, "table").hashCode()                             // false
```

Both shapes come out of the parser. `DbTableMeta::new_with_namespace_and_schema` fills `schema`
from the second-to-last identifier but falls back to `default_database` for `database`:

```rust
database: reversed.get(2).map(...).or(default_database),
schema:   reversed.get(1).map(...).or(default_schema),
```

so `FROM db.table` parsed with no defaults yields `(None, Some("db"), "table")`, while
`FROM table` parsed with `default_database = Some("db")` yields `(Some("db"), None, "table")`.
Those two describe the same table, compare equal, and hash differently.

This matters because `DbTableMeta` is reachable from `ColumnMeta`, whose `hashCode` delegates to
it through `Optional`, and `ColumnMeta` is used as a hash key:

- `ColumnLevelLineageBuilder.externalExpressionMappings` is a `Map<ColumnMeta, ExprId>`
- `SqlCollector` and `QueryRelationColumnLineageCollector` collect `ColumnMeta` into `Set`s

Equal keys landing in different buckets means a lookup for a column that is present returns
null, and a set that should hold one entry holds two, so column-level lineage for that column is
dropped or duplicated.

`hashCode` should be derived from the same value `equals` compares, so that equal tables always
hash equally.

### How to reproduce

```java
DbTableMeta fromQualifiedName = new DbTableMeta(null, "db", "table");
DbTableMeta fromDefaultDatabase = new DbTableMeta("db", null, "table");

fromQualifiedName.equals(fromDefaultDatabase);   // true
Set<DbTableMeta> tables = new HashSet<>();
tables.add(fromQualifiedName);
tables.add(fromDefaultDatabase);
tables.size();                                   // 2, expected 1

Map<ColumnMeta, String> lineage = new HashMap<>();
lineage.put(new ColumnMeta(fromQualifiedName, "column"), "expression");
lineage.get(new ColumnMeta(fromDefaultDatabase, "column"));   // null, expected "expression"
```

### Additional details

No native library is needed to reproduce this, the classes are plain Java.

Daksha1611:
closes: #4955

**sql**: make `DbTableMeta.hashCode` consistent with `equals` so equal tables resolve to the same hash bucket.

### Meaningful description

`DbTableMeta.equals` compares `qualifiedName()`, which folds away a null database or schema, while
`hashCode` was built from `database`, `schema` and `name` separately. Two instances describing the
same table could therefore be equal and still hash differently, breaking the `equals`/`hashCode`
contract:

```java
new DbTableMeta(null, "db", "table").equals(new DbTableMeta("db", null, "table"))  // true
// hash codes differ
```

Both shapes are produced by the parser. `new_with_namespace_and_schema` fills `schema` from the
second-to-last identifier but falls back to `default_database` for `database`, so `FROM db.table`
parsed without defaults and `FROM table` parsed against a default database describe the same table
through different fields.

This is reachable from `ColumnMeta`, which delegates its hash to `DbTableMeta` through `Optional`
and is used as a hash key in `ColumnLevelLineageBuilder.externalExpressionMappings` and in the sets
built by `SqlCollector` and `QueryRelationColumnLineageCollector`. Equal keys in different buckets
means a lookup for a column that is present returns null and a set that should hold one entry holds
two, so column-level lineage for that column is dropped or duplicated.

`hashCode` is now derived from `qualifiedName()`, the same value `equals` compares. The added tests
are plain Java and do not need the native library.

### Checklist


Daksha1611:
The failing Spark jobs here are unrelated to this change.

Both jobs fail on the same two tests:

```
WriteToDataSourceV2DatasetBuilderTest > preservesAllExtensionProvidedOutputs()
WriteToDataSourceV2DatasetBuilderTest > extractsOrdinaryRelationExactlyOnceThroughDelegation()
```

They already fail on main at 9ac19298c, the commit this branch is based on. On that commit every
build that actually executed the `spark3` test task failed on these two tests (595525, 595625).
The two green builds on the same commit (595532, 595646) never ran the task at all - it was
`UP-TO-DATE` from the Gradle build cache - so they are not evidence that the tests pass.

This PR does not touch the Spark integration, and `build-client-java` passes.

PR #4972, opened 2026-09-19
client: add NATS transport for Python and Java

### LINKED ISSUE #4971: [FEATURE] NATS (JetStream) transport for the Python and Java clients
### Motivation

OpenLineage clients can emit events over HTTP, Kafka, Kinesis and several cloud-specific transports, but not over NATS. NATS is widely deployed as the messaging layer in edge, IoT and hybrid setups, and in platforms that have standardised on it instead of Kafka. Teams in that position currently have to run a side-channel (an HTTP relay, or a Kafka cluster just for lineage) to get events from their producers to a lineage backend.

A NATS transport would let producers publish lineage events to a NATS subject. With JetStream, those events are stored durably and acknowledged, so a backend can consume them on its own schedule. This is the same decoupling the Kafka and Kinesis transports give today, and it matters most for consumers that run as periodic batch jobs rather than always-on services. I'm working on the consumer side for OpenMetadata, which already reads OpenLineage events from Kafka and Kinesis.

There is recent precedent in OpenTelemetry: a NATS exporter/receiver for the Collector has been accepted for donation to collector-contrib (open-telemetry/opentelemetry-collector-contrib#50735, originally requested in #39540). The implementation is maintained by Synadia (the company behind NATS): https://github.com/synadia-io/opentelemetry-collector-nats. I'd like the OpenLineage transport's configuration to follow the same conventions where they apply.

### Description

Add a `nats` transport to the Python client and the Java client.

**Behaviour**

- Serialise each event to JSON (the same payload as the Kafka transport) and publish it to a configured subject.
- With `jetstream.enabled: true` (the recommended setting), publish through JetStream and wait for the stream's acknowledgement, up to `publishTimeout`. A failed or timed-out publish raises in the same way the Kafka transport does. With JetStream disabled, publish over core NATS (fire-and-forget), for setups where losing events is acceptable.
- Optionally set the `Nats-Msg-Id` header to `<runId>:<eventType>:<eventTime>`, so JetStream's duplicate window drops retried publishes.
- The transport does not create or configure streams. Retention (`max_age`, `max_bytes`, retention policy, replicas) is left to whoever operates the NATS server, and the docs will include an example `nats stream add` command.

**Configuration (Python / `openlineage.yml`)**

```yaml
transport:
  type: nats
  url: nats://nats.example.com:4222      # comma-separated list allowed
  subject: openlineage.events
  jetstream:
    enabled: true
    publishTimeout: 5                    # seconds
  msgIdHeader: true
  auth:
    type: userPassword                   # userPassword | token | nkey | creds
    username: openlineage
    password: ${NATS_PASSWORD}
  tls:
    caFile: /etc/ssl/nats/ca.pem
    certFile: /etc/ssl/nats/client.pem
    keyFile: /etc/ssl/nats/client-key.pem
```

The same settings would be available through the usual `OPENLINEAGE__TRANSPORT__*` environment variables and, for Spark, `spark.openlineage.transport.*`.

**Implementation**

- Python: `openlineage/client/transport/nats.py`, with `nats-py` behind an optional `openlineage-python[nats]` extra, as `confluent-kafka` is for Kafka. `nats-py` is asyncio-only, so the transport would run its client on a background event loop behind the synchronous `emit()`, following the approach in `async_http.py`.
- Java: a separate `client/java/transports-nats` module using `jnats`, registered through `ServiceLoader` like `transports-kinesis`, so the core client gains no new dependency.
- Tests against a `nats -js` container (Testcontainers), covering JetStream ack, publish timeout, auth, TLS and the dedupe header.
- Docs pages for both clients, including the stream setup and retention guidance.

### Related issues

I couldn't find an existing issue or PR about NATS in this repository.

- open-telemetry/opentelemetry-collector-contrib#39540: request for a NATS receiver/exporter
- open-telemetry/opentelemetry-collector-contrib#50735: accepted component donation (synadia-io/opentelemetry-collector-nats)

oluies:
closes: #4971

Client/Python and Client/Java: Add a NATS transport that publishes events to a NATS subject, through JetStream by default.

### Meaningful description

OpenLineage clients can't currently publish events to NATS. Teams that use NATS as their messaging layer have to run an HTTP relay, or a Kafka cluster just for lineage. This PR adds a `nats` transport to both clients.

**Behaviour (same in both clients)**

- Events are serialized to JSON and published to `subject`.
- With `jetstream: true` (the default), `emit` waits for the stream's acknowledgement. It raises if no stream captures the subject, the connection fails, or the ack doesn't arrive within `publishTimeout`. With `jetstream: false`, events go over core NATS and only reach subscribers connected at that moment.
- `Nats-Msg-Id` is built from a SHA-256 digest of the serialized event, so JetStream's duplicate window drops a retried publish of the same event, while any two different events keep different ids:
  - `{runId}:{eventType}:{digest}` for RunEvent (START, RUNNING, COMPLETE and FAIL of one run stay distinct even at the same `eventTime`)
  - `job:{digest}` and `dataset:{digest}` for JobEvent and DatasetEvent. Names are left out because they can contain characters that NATS headers reject.
- `messageTtl` optionally sets a per-message TTL in whole seconds (NATS Server 2.11+, stream created with per-message TTL allowed).
- Auth: user/password, token, NKey seed file or `.creds` file (at most one). The docs note that user/password and token should be used over TLS.
- The transport never creates or configures streams. Retention is left to whoever runs the server, and the docs include an example `nats stream add` command with `max-age` guidance.

**Python** (`openlineage/client/transport/nats.py`, `openlineage-python[nats]` → `nats-py[nkeys]>=2.13.0`)

- `nats-py` is asyncio-only, so the client runs on a background event loop behind the synchronous `emit()`. A timed-out emit is cancelled, so it can never publish after `emit` has raised.
- No background reconnect loop: a lost connection is replaced by the next emit, bounded by `connectTimeout`. Concurrent first emits share one connection. State is reset in forked children (Airflow workers fork).
- TLS uses PEM files. Config keys work in camelCase and in the snake_case form they arrive in from `OPENLINEAGE__TRANSPORT__*` environment variables; values are coerced to strings, since env var values arrive JSON-decoded.

**Java** (new `client/java/transports-nats` module, registered through `ServiceLoader` like `transports-kinesis`)

- jnats 2.26.3 (Java 8 bytecode). The shadow jar relocates jnats and BouncyCastle; documented `io.nats.client.*` keys in `properties` are mapped onto the relocated names, and explicit settings take precedence over them.
- TLS uses keystore/truststore and verifies the server host name. jnats itself only checks the certificate chain, so the transport wraps the `SSLContext` to add endpoint identification.
- jnats threads are daemon threads owned by the transport, so an unclosed transport does not keep the JVM alive.
- Every failure, including oversized payloads (NATS `max_payload`, 1 MB by default), is reported as `OpenLineageClientException`. After a failed connection attempt, emits fail fast for `connectTimeout` instead of each blocking on a new attempt.

**Tests**

Both suites run against real NATS servers: `NATS_URL` when set, else a local `nats-server` binary, else the `nats:2` image (Testcontainers on the Java side). Tests that need auth or TLS start their own servers with `nats-server` and generate throwaway credentials and certificates (`openssl`, operator/account/user JWTs); they skip when the tools are missing.

Covered: JetStream persistence and ack, de-duplication and the message id scheme, per-message TTL, missing stream, core NATS, server down and restart, composite transport, NKey / creds / user-password auth, TLS with trusted and unknown CA, wrong host name and mTLS, non-ASCII and multi-line names, oversized events, env var and YAML config, a real fork (Python), import without the extra (Python), and a JVM that must exit without `close()` (Java). `./gradlew :transports-nats:shadowJarTest` repeats creds auth and `properties` against the shadow jar.

`dev/nats-offline/run.sh` runs both suites in containers without internet access, next to a `nats:2 -js` container, and fails if the internet is reachable (Apple `container` by default, `RUNTIME=docker` also works).

Local results: Python client suite 755 passed (41 NATS); Java `:transports-nats:build` passed (33 tests including the shadow jar, PMD, spotless, javadoc); offline harness passed for both.

**Review**

After the first push I ran an adversarial multi-reviewer pass; the fixes are in the second and third commits. The main ones: Python emits that timed out could still publish later, Python NKey/creds auth needed the `nkeys` extra, Java threads kept the JVM alive, and Java TLS did not verify the host name. CodeRabbit's five inline comments (message id collisions, connection race, timeout validation, null `properties` merge) are addressed by the same commits.

Two pre-existing things surfaced along the way, outside this PR: in every Java transport module `jar` and `shadowJar` write the same file (this module orders them so the shaded jar wins), and the Java client's `generateCode` fetches each schema's `$id` from openlineage.io, so the build cannot run fully offline from scratch.

**Configuration shape:** the issue sketched nested `auth:` and `tls:` blocks. The implementation uses flat keys (`user`, `credsFile`, `tlsCaFile`, …), because they map directly onto environment variables and Spark conf keys, as the Kafka transport's keys do.

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Added NATS transport support for Java and Python clients.
  - Supports JetStream and core NATS publishing, configurable subjects, authentication, TLS, timeouts, deduplication, and message expiration.
  - Added stable, digest-based message IDs, automatic transport discovery, NKey and credential-file authentication, and connection recovery.
  - Added configuration precedence and validation for credentials, payloads, and message TTLs.

- **Documentation**
  - Added setup, configuration, security, stream requirements, and usage examples for Java and Python.

- **Tests**
  - Added comprehensive integration and configuration coverage, including persistence, reconnection, authentication, TLS, and core NATS delivery.


oluies:
The consumer side is now built and running against this transport, which exercised it end to end.

**OpenMetadata**: [`feat/openlineage-nats-broker`](https://github.com/oluies/OpenMetadata/tree/feat/openlineage-nats-broker) adds NATS as a third broker option to OpenMetadata's OpenLineage connector (beside Kafka and Kinesis), consuming a JetStream stream through a durable pull consumer so a scheduled ingestion run resumes where the previous one stopped. It includes an integration test that publishes events to a stream and consumes them with the connector.

**A DuckDB producer**: [duckdbpreso/10-lineage](https://github.com/oluies/duckdbpreso/tree/main/10-lineage) publishes a dbt-duckdb run's lineage with this transport instead of POSTing it to OpenMetadata, which is the case the transport was written for: the producer only needs the broker, and the consumer can be down for as long as the stream's `max-age` allows.

Running that turned up one gap, fixed in 20b74d9: a producer that already holds OpenLineage JSON emits it as a dict, and those events got no `Nats-Msg-Id` at all, so republishing a batch stored it twice. The id now falls back to the payload digest, and the same batch published twice leaves one copy of each event.

oluies:
The consumer side is now a draft PR rather than just a branch: **open-metadata/OpenMetadata#33665** (issue open-metadata/OpenMetadata#33664). It adds NATS JetStream as a third broker option for OpenMetadata's OpenLineage connector, consuming a stream through a durable pull consumer, and its manual testing is the DuckDB producer publishing through this transport.

oluies:
open-metadata/OpenMetadata#33665 is out of draft and ready for review. Since the last update it also fixes two consumer-side problems found in review:

- Acknowledgements are now sent from an event loop on its own thread, so they still reach the server while the connector is busy downstream. Before this fix, a slow run could consume every event and acknowledge none of them, and the next run replayed the whole stream.
- The connection schema now rejects a zero poll timeout, because with it the consumer never ended a run.

Neither fix changes the producer side. The consumer still relies on this transport's payload-derived `Nats-Msg-Id` to deduplicate events within the stream's duplicate window.

